PR #73843, opened 2024-06-04
Use EnableCodeStyleSeverity instead of AnalysisLevel to control new diagnostic severity behavior

sharwell:
* Use built-in features for marking unnecessary code
* Verify missing code style severity falls back to old behavior
* Use `EnableCodeStyleSeverity` instead of `AnalysisLevel` to control new diagnostic severity behavior

Follow-up to #72238

### DIFF
--- a/src/Analyzers/CSharp/Tests/UseObjectInitializer/UseObjectInitializerTests.cs
+++ b/src/Analyzers/CSharp/Tests/UseObjectInitializer/UseObjectInitializerTests.cs
@@ -1259,15 +1259,12 @@ public partial class UseObjectInitializerTests
     }
 
     [Theory]
-    [InlineData(8.0)]
-    [InlineData(9.0)]
+    [CombinatorialData]
     [WorkItem("https://github.com/dotnet/roslyn/issues/72094")]
-    public async Task TestWithConflictingSeverityConfigurationEntries(double analysisLevel)
+    public async Task TestWithConflictingSeverityConfigurationEntries(bool enabled)
     {
-        var expectFix = analysisLevel >= 9.0;
-
         string testCode, fixedCode;
-        if (expectFix)
+        if (enabled)
         {
             testCode =
                 """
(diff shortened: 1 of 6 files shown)


arkalyanms (approved):

cremor:
If I understand this PR correctly then these documentation pages need to be updated for the change, correct?
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview#enable-on-build
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/language-rules#option-format

sharwell:
Yes, those would need to be updated. We'll also need to update the SDK code style build targets.

src/Analyzers/CSharp/Tests/UseObjectInitializer/UseObjectInitializerTests.cs
```diff
@@ -20,17 +20,6 @@ namespace Microsoft.CodeAnalysis.Editor.CSharp.UnitTests.UseObjectInitializer;
 [Trait(Traits.Feature, Traits.Features.CodeActionsUseObjectInitializer)]
 public partial class UseObjectInitializerTests
 {
-    private static async Task TestInRegularAndScriptAsync(string testCode, string fixedCode, OutputKind outputKind = OutputKind.DynamicallyLinkedLibrary)
```
jaredpar, line 23: Why did this change produce so much churn in the object init tests?

sharwell: This analyzer was using an old strategy for code fading that was interfering with some new testing we needed for the feature change. I converted the analyzer to the new style as a separate commit at the beginning of this PR.

chsienki (approved):

333fred (approved):

jjonescz (approved):

sharwell:
Rebased after reviews completed to apply the fixup commit.

Issue #73691, opened 2024-05-24
Compile Error After Upgrading - SDK 8.0.300 - 'Class1' does not contain a definition for 'DisposeAsync'


bitbonk:
### Describe the bug
After upgrading our solution from 8.0.205 to 8.0.300 we now get compiler errors like the one below without making any changes to the code:

`Error CS1929 : 'Class1' does not contain a definition for 'DisposeAsync' and the best extension method overload 'EnumerableExtensions.DisposeAsync(IEnumerable<object>)' requires a receiver of type 'System.Collections.Generic.IEnumerable<object>'`

### To Reproduce
1. Create a project with the following code:
    ```csharp
    
    namespace CompilerErrorRepro.ClassLibrary
    {
        internal class Class1 : IAsyncDisposable
        {
            ValueTask IAsyncDisposable.DisposeAsync()
            {
                return ValueTask.CompletedTask;
            }

            public async Task MethodWithCompilerError()
            {
                await using var sut = Create();
            }

            private Class1 Create()
            {
                return new Class1();
            }
        }
    }

    namespace System
    {
        internal static class EnumerableExtensions
        {
            public static async ValueTask DisposeAsync(this IEnumerable<object> objects)
            {
                foreach (var disposable in objects.OfType<IAsyncDisposable>()) await disposable.DisposeAsync();
            }
        }
    }
    ```

2. Compile using SDK 8.0.300 with this `global.json`

   ```json
   {
     "sdk": {
       "version": "8.0.300", "rollForward": "disable"
     }
   }
   ```
3. See the error mentioned above
   `Error CS1929 : 'Class1' does not contain a definition for 'DisposeAsync' and the best extension method overload 'EnumerableExtensions.DisposeAsync(IEnumerable<object>)' requires a receiver of type 'System.Collections.Generic.IEnumerable<object>'`

4. Compile using the same code but with SDK  SDK 8.0.205 
   ```json
   {
     "sdk": {
       "version": "8.0.205", "rollForward": "disable"
     }
   }
   ```
5. No compiler error

### Further technical details
- using latest Rider as the IDE
- output of `dotnet --info`

```text
.NET SDK:
 Version:           8.0.300
 Commit:            326f6e68b2
 Workload version:  8.0.300-manifests.0d1e46b7
 MSBuild version:   17.10.4+10fbfbf2e

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win-x64
[...]

Issue #72819, opened 2024-03-30
Likely unexpected breaking change introduced recently for `await using`


AlekseyTs:
The tests below reflects the current behavior:
```
        [Fact]
        public void Test()
        {
            var src = @"
using System;
using System.Threading.Tasks;

interface IMyAsyncDisposable1
{
    ValueTask DisposeAsync();
}

interface IMyAsyncDisposable2
{
    ValueTask DisposeAsync();
}

struct S2 : IMyAsyncDisposable1, IMyAsyncDisposable2, IAsyncDisposable
{
    ValueTask IMyAsyncDisposable1.DisposeAsync() => throw null;
    ValueTask IMyAsyncDisposable2.DisposeAsync() => throw null;

    public ValueTask DisposeAsync()
    {
        System.Console.Write('D');
        return ValueTask.CompletedTask;
    }
}

class C
{
    static async Task Main()
    {
        await Test<S2>();
    }

    static async Task Test<T>() where T : IMyAsyncDisposable1, IMyAsyncDisposable2, IAsyncDisposable, new()
    {
        await using (new T())
        {
            System.Console.Write(123);
        }
    }
}
";
            var comp = CreateCompilation(src, targetFramework: TargetFramework.Net80, options: TestOptions.ReleaseExe);

            comp.VerifyDiagnostics(
                // (36,22): error CS0121: The call is ambiguous between the following methods or properties: 'IMyAsyncDisposable1.DisposeAsync()' and 'IMyAsyncDisposable2.DisposeAsync()'
                //         await using (new T())
                Diagnostic(ErrorCode.ERR_AmbigCall, "new T()").WithArguments("IMyAsyncDisposable1.DisposeAsync()", "IMyAsyncDisposable2.DisposeAsync()").WithLocation(36, 22)
                );
        }
```

The code used to compile successfully and was printing "123D". This is likely a regression from https://github.com/dotnet/roslyn/pull/72598.

Note, similar scenario in `await foreach` works the same way the `await using` used to work (i.e. it falls back to using `IAsyncDisposable` implementation).
```
using System;
using System.Threading;
using System.Threading.Tasks;

interface ICustomEnumerator
{
    public int Current {get;}

    public ValueTask<bool> MoveNextAsync();
}

interface IGetEnumerator<TEnumerator> where TEnumerator : ICustomEnumerator
{
    TEnumerator GetAsyncEnumerator(CancellationToken token = default);
}

struct S1 : IGetEnumerator<S2>
{
[...]

PR #72598, opened 2024-03-19
Prefer pattern-based over interface-based disposal in await using

jcouv:
Closes https://github.com/dotnet/roslyn/issues/72573

Issue #30257, opened 2018-10-02
Confirm whether obsolete DisposeAsync should be reported


jcouv:
```C#
        [Fact]
        public void TestWithObsoleteDisposeAsync()
        {
            string source = @"
class C : System.IAsyncDisposable
{
    public static async System.Threading.Tasks.Task Main()
    {
        using await (var x = new C())
        {
        }
    }
    [System.Obsolete]
    public async System.Threading.Tasks.ValueTask DisposeAsync()
    {
        await System.Threading.Tasks.Task.Delay(10);
    }
}
";
            var comp = CreateCompilationWithTasksExtensions(source + s_interfaces, options: TestOptions.DebugExe);
            comp.VerifyDiagnostics();
            // PROTOTYPE(async-streams) Confirm whether this behavior is ok (currently matching behavior of obsolete Dispose in non-async using)
        }

```

333fred:
Note that this is the same behavior as `Dispose`, which definitely seems like bug. It would be a breaking change to fix it though...

333fred:
Unfortunately, it doesn't matter. People have existing code, and if they make no changes to their code or libraries then upgrading the compiler version should not break their code.

I've gone and verified that this is the same behavior as the native compiler, which reinforces this.

333fred:
It does seem like something that we should enforce with the warning waves stuff we're looking at, or in strict mode.

333fred:
So, the only reasoning for this behavior that I can think of is, if the compile-time type is does not implement `IDisposable`, we still insert a runtime check to see if it needs to be disposed. If the runtime type has an error obsolete `Dispose` method, it will still be called with no indications of an issue. I think we're likely stuck with this behavior.

BillWagner:
@333fred @jcouv I think any resolution of this would change the compiler behavior, not the standard. I don't think I should move this issue to the dotnet/csharpstandard repo per dotnet/csharplang#173. Would you agree?

333fred:
I don't believe this is a spec bug, no.

jcouv:
@BillWagner I agree this is a compiler bug (not a spec bug).

Here are the relevant LDM notes I could find to support that expectation: https://github.com/dotnet/csharplang/blob/3c8559f186d4c5df5d1299b0eaa4e139ae130ab6/meetings/2019/LDM-2019-01-09.md#obsolete

PR #73850, opened 2024-06-04
Delay reporting of pattern-based using diagnostics

jcouv:
In the reported scenario, we try to bind pattern-based first, but incorrectly report the diagnostics from that binding despite successfully binding via interface as fallback.

Regression from https://github.com/dotnet/roslyn/pull/72598
Fixes https://github.com/dotnet/roslyn/issues/73691
Fixes https://github.com/dotnet/roslyn/issues/72819

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
+++ b/src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
@@ -190,14 +190,15 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                 // Pattern-based binding
                 // If this is a ref struct, or we're in a valid asynchronous using, try binding via pattern.
+                BindingDiagnosticBag? patternDiagnostics = null;
                 if (type is object && (type.IsRefLikeType || hasAwait))
                 {
                     BoundExpression? receiver = fromExpression
                                                ? expressionOpt
                                                : new BoundLocal(syntax, declarationsOpt[0].LocalSymbol, null, type) { WasCompilerGenerated = true };
 
-                    BindingDiagnosticBag patternDiagnostics = originalBinder.Compilation.IsFeatureEnabled(MessageID.IDS_FeatureDisposalPattern)
-                                                       ? diagnostics
+                    patternDiagnostics = originalBinder.Compilation.IsFeatureEnabled(MessageID.IDS_FeatureDisposalPattern)
+                                                       ? BindingDiagnosticBag.GetInstance(diagnostics)
                                                        : BindingDiagnosticBag.Discarded;
                     MethodSymbol disposeMethod = originalBinder.TryFindDisposePatternMethod(receiver, syntax, hasAwait, patternDiagnostics, out bool expanded);
                     if (disposeMethod is object)
@@ -228,6 +229,8 @@ namespace Microsoft.CodeAnalysis.CSharp
                         {
                             awaitableType = disposeMethod.ReturnType;
                         }
+
+                        diagnostics.AddRange(patternDiagnostics);
                         return true;
                     }
                 }
@@ -245,7 +248,13 @@ namespace Microsoft.CodeAnalysis.CSharp
                         awaitableType = originalBinder.Compilation.GetWellKnownType(WellKnownType.System_Threading_Tasks_ValueTask);
                     }
 
-                    return !ReportUseSite(disposableInterface, diagnostics, hasAwait ? awaitKeyword : usingKeyword);
+                    var wasSuccess = !ReportUseSite(disposableInterface, diagnostics, hasAwait ? awaitKeyword : usingKeyword);
+                    if (!wasSuccess)
+                    {
+                        diagnostics.AddRange(patternDiagnostics);
+                    }
+
+                    return wasSuccess;
                 }
 
                 if (type is null || !type.IsErrorType())
@@ -261,6 +270,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                     Error(diagnostics, errorCode, syntax, declarationTypeOpt ?? expressionOpt!.Display);
                 }
 
+                diagnostics.AddRange(patternDiagnostics);
                 return false;
             }
 
--- a/src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
+++ b/src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
@@ -190,19 +190,19 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                 // Pattern-based binding
                 // If this is a ref struct, or we're in a valid asynchronous using, try binding via pattern.
-                BindingDiagnosticBag? patternDiagnostics = null;
                 if (type is object && (type.IsRefLikeType || hasAwait))
                 {
                     BoundExpression? receiver = fromExpression
                                                ? expressionOpt
                                                : new BoundLocal(syntax, declarationsOpt[0].LocalSymbol, null, type) { WasCompilerGenerated = true };
 
-                    patternDiagnostics = originalBinder.Compilation.IsFeatureEnabled(MessageID.IDS_FeatureDisposalPattern)
+                    BindingDiagnosticBag patternDiagnostics = originalBinder.Compilation.IsFeatureEnabled(MessageID.IDS_FeatureDisposalPattern)
                                                        ? BindingDiagnosticBag.GetInstance(diagnostics)
                                                        : BindingDiagnosticBag.Discarded;
                     MethodSymbol disposeMethod = originalBinder.TryFindDisposePatternMethod(receiver, syntax, hasAwait, patternDiagnostics, out bool expanded);
                     if (disposeMethod is object)
                     {
+                        diagnostics.AddRange(patternDiagnostics);
                         MessageID.IDS_FeatureDisposalPattern.CheckFeatureAvailability(diagnostics, originalBinder.Compilation, syntax.Location);
 
                         var argumentsBuilder = ArrayBuilder<BoundExpression>.GetInstance(disposeMethod.ParameterCount);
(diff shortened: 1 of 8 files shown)


src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
```diff
@@ -245,7 +248,13 @@ bool bindDisposable(bool fromExpression, out MethodArgumentInfo? patternDisposeI
                         awaitableType = originalBinder.Compilation.GetWellKnownType(WellKnownType.System_Threading_Tasks_ValueTask);
                     }
 
-                    return !ReportUseSite(disposableInterface, diagnostics, hasAwait ? awaitKeyword : usingKeyword);
+                    var wasSuccess = !ReportUseSite(disposableInterface, diagnostics, hasAwait ? awaitKeyword : usingKeyword);
+                    if (!wasSuccess)
+                    {
+                        diagnostics.AddRange(patternDiagnostics);
```
AlekseyTs, line 254: >diagnostics.AddRange(patternDiagnostics); [](http://example.com/codeflow?start=24&length=41)

Are we testing this code path?

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
```diff
@@ -13557,13 +13557,10 @@ await using (new T())
 }
 ";
             var comp = CreateCompilation(src, targetFramework: s_targetFrameworkSupportingByRefLikeGenerics, options: TestOptions.ReleaseExe);
-
-            // https://github.com/dotnet/roslyn/issues/72819: The failure is likely unexpected, but is not specific to `allow ref struct` scenario.
-            comp.VerifyDiagnostics(
-                // (36,22): error CS0121: The call is ambiguous between the following methods or properties: 'IMyAsyncDisposable1.DisposeAsync()' and 'IMyAsyncDisposable2.DisposeAsync()'
+            comp.VerifyEmitDiagnostics(
```
AlekseyTs, line 13560: >comp.VerifyEmitDiagnostics( [](http://example.com/codeflow?start=12&length=27)

I think we should convert this test case to a success case, execute the scenario and observe expected disposal behavior. See, for example, how ```SystemActivatorCreateInstance_03``` takes care of ```CreateInstance``` problem.

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAwaitUsingTests.cs
```diff
@@ -3519,5 +3519,147 @@ public ValueTask DisposeAsync()
...
+        throw null;
+    }
+}
+""";
+            var comp = CreateCompilationWithTasksExtensions([source, IAsyncDisposableDefinition]);
+            comp.VerifyEmitDiagnostics(
+                // (5,1): error CS8410: 'Class1': type used in an asynchronous using statement must be implicitly convertible to 'System.IAsyncDisposable' or implement a suitable 'DisposeAsync' method.
+                // await using var x = new Class1();
+                Diagnostic(ErrorCode.ERR_NoConvToIAsyncDisp, "await using var x = new Class1();").WithArguments("Class1").WithLocation(5, 1),
+                // (5,1): error CS1929: 'Class1' does not contain a definition for 'DisposeAsync' and the best extension method overload 'EnumerableExtensions.DisposeAsync(IEnumerable<object>)' requires a receiver of type 'System.Collections.Generic.IEnumerable<object>'
```
AlekseyTs, line 3579: >// (5,1): error CS1929: 'Class1' does not contain a definition for 'DisposeAsync' and the best extension method overload 'EnumerableExtensions.DisposeAsync(IEnumerable<object>)' requires a receiver of type 'System.Collections.Generic.IEnumerable<object>' [](http://example.com/codeflow?start=16&length=254)

This doesn't really look like a very useful error. The error above feels informative above.

src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
```diff
@@ -190,14 +190,15 @@ bool bindDisposable(bool fromExpression, out MethodArgumentInfo? patternDisposeI
 
                 // Pattern-based binding
                 // If this is a ref struct, or we're in a valid asynchronous using, try binding via pattern.
+                BindingDiagnosticBag? patternDiagnostics = null;
```
AlekseyTs, line 193: >BindingDiagnosticBag? patternDiagnostics = null; [](http://example.com/codeflow?start=16&length=48)

I find this change unnecessary complicated and it feels like we can simply follow what ```GetDisposalInfoForEnumerator``` is doing in terms of diagnostics. Specifically, it looks like it reports ```patternDiagnostics``` only when a pattern method is found. Otherwise, we either find an interface implementation, or report an error about that if we don't.

AlekseyTs:

src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
jcouv: Yes, `PatternBasedFails_WithInterfaceImplementation_UseSite`. But the diagnostic from the pattern-based binding is the same, so it's hard to tell by looking at the test...

src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
jcouv: > Specifically, it looks like it reports patternDiagnostics only when a pattern method is found.

That's not what I see in `GetDisposalInfoForEnumerator` (we always discard there): `TryFindDisposePatternMethod(receiver, syntax, isAsync, BindingDiagnosticBag.Discarded, out bool expanded)`

The options so far:
1. always discard pattern-based diagnostics
2. report pattern-based diagnostics when we commit to pattern method
3. report pattern-based diagnostics when we commit to pattern method or when interface fallback fails

I don't see what's wrong with option 3.

If we go with option 2 as you're describing above then we lose the accessibility diagnostic in this scenario, which seems a useful diagnostic:
```
        public void TestPatternBasedDisposal_InstanceMethod_Inaccessible()
        {
            string source = @"
public class D
{
    public static async System.Threading.Tasks.Task<int> Main()
    {
        await using (var x = new C())
        {
        }

        return 1;
    }
}
public class C
{
    private System.Threading.Tasks.ValueTask DisposeAsync()
        => throw null;
}
";
            var comp = CreateCompilationWithTasksExtensions(new[] { source, IAsyncDisposableDefinition });
            comp.VerifyDiagnostics(
                // (6,22): error CS0122: 'C.DisposeAsync()' is inaccessible due to its protection level
                //         await using (var x = new C())
                Diagnostic(ErrorCode.ERR_BadAccess, "var x = new C()").WithArguments("C.DisposeAsync()").WithLocation(6, 22),
                // (6,22): error CS8410: 'C': type used in an async using statement must be implicitly convertible to 'System.IAsyncDisposable' or implement a suitable 'DisposeAsync' method.
                //         await using (var x = new C())
                Diagnostic(ErrorCode.ERR_NoConvToIAsyncDisp, "var x = new C()").WithArguments("C").WithLocation(6, 22)
                );
        }
```

If we go with option 1 as `GetDisposalInfoForEnumerator` does, then we furthermore lose the obsolete diagnostic in this scenario, which also seems a good diagnostic:
```
        public void TestWithObsoleteDisposeAsync_02()
        {
            string source = @"
class C : System.IAsyncDisposable
{
    public static async System.Threading.Tasks.Task Main()
    {
        await using (var x = new C())
        {
        }
    }
    [System.Obsolete]
    public async System.Threading.Tasks.ValueTask DisposeAsync()
    {
        await System.Threading.Tasks.Task.Yield();
    }
}
";
            var comp = CreateCompilationWithTasksExtensions(new[] { source, IAsyncDisposableDefinition }, options: TestOptions.DebugExe);
            comp.VerifyDiagnostics(
                // 0.cs(6,22): warning CS0612: 'C.DisposeAsync()' is obsolete
                //         await using (var x = new C())
                Diagnostic(ErrorCode.WRN_DeprecatedSymbol, "var x = new C()").WithArguments("C.DisposeAsync()").WithLocation(6, 22)
                );
        }
```

jcouv: From offline discussion, we'll go with option 2.
There's already a tracking issue for `GetDisposalInfoForEnumerator` dropping obsolete diagnostics: [https://github.com/dotnet/roslyn/issues/30257](https://github.com/dotnet/roslyn/issues/30257)

jcouv:
Sorry, looks like CodeFlow broke on second commit :-/

AlekseyTs:
@jcouv It looks like there are legitimate test failures.

src/Compilers/CSharp/Portable/Binder/UsingStatementBinder.cs
```diff
@@ -197,11 +197,12 @@ bool bindDisposable(bool fromExpression, out MethodArgumentInfo? patternDisposeI
                                                : new BoundLocal(syntax, declarationsOpt[0].LocalSymbol, null, type) { WasCompilerGenerated = true };
 
                     BindingDiagnosticBag patternDiagnostics = originalBinder.Compilation.IsFeatureEnabled(MessageID.IDS_FeatureDisposalPattern)
-                                                       ? diagnostics
+                                                       ? BindingDiagnosticBag.GetInstance(diagnostics)
```
AlekseyTs, line 200: It doesn't look like we put this bag back into the pool.

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
```diff
@@ -13555,15 +13554,22 @@ await using (new T())
...
+namespace System
+{
+    public class Activator
+    {
+         public static T CreateInstance<T>() where T : allows ref struct => default;
+    }
+}
 ";
             var comp = CreateCompilation(src, targetFramework: s_targetFrameworkSupportingByRefLikeGenerics, options: TestOptions.ReleaseExe);
+            comp.VerifyEmitDiagnostics();
```
AlekseyTs, line 13567: This looks redundant since we verify diagnostics on CompileAndVerify below

AlekseyTs:
Consider including tests for scenarios from #72819. The adjusted test case is ref struct specific.

AlekseyTs:

jcouv:

chsienki (approved):

AlekseyTs:
@jcouv It looks like there is a legitimate test failure

AlekseyTs (approved):
LGTM (commit 6)

PR #73851, opened 2024-06-04
Use a serialization compatible marker

davidwengier:
FatalError tries to tag exceptions so that they don't get reported twice, but it uses a plain object which fails serialization/deserialization if that exception ever crosses a json rpc boundary that uses System.Text.Json. This PR changes it to something serializable.

### DIFF
--- a/src/Compilers/Core/Portable/InternalUtilities/FatalError.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/FatalError.cs
@@ -225,6 +225,9 @@ namespace Microsoft.CodeAnalysis.ErrorReporting
 
 #endif
 
+        // We use a Guid for the marker because it is used as a key in an exceptions Data dictionary, so we must make sure
+        // it's serializable if the exception crosses an RPC boundary. In particular System.Text.Json doesn't like plain
+        // object dictionary keys.
         private static readonly object s_reportedMarker = Guid.NewGuid();
 
         // Do not allow this method to be inlined.  That way when we have a dump we can see this frame in the stack and
--- a/src/Compilers/Core/Portable/InternalUtilities/FatalError.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/FatalError.cs
@@ -225,7 +225,7 @@ namespace Microsoft.CodeAnalysis.ErrorReporting
 
 #endif
 
-        private static readonly Guid s_reportedMarker = Guid.NewGuid();
+        private static readonly object s_reportedMarker = Guid.NewGuid();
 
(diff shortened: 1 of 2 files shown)


dibarbet (approved):

jasonmalinowski (approved):

dibarbet:
(disabling automerge, think it needs 2 compiler reviews)
@dotnet/roslyn-compiler for 2 reviews

davidwengier:
Whoops, thank you, didn't notice this was in the compiler

src/Compilers/Core/Portable/InternalUtilities/FatalError.cs
```diff
@@ -225,7 +225,7 @@ public static bool ReportAndCatchUnlessCanceled(Exception exception, Cancellatio
 
 #endif
 
-        private static readonly object s_reportedMarker = new();
+        private static readonly Guid s_reportedMarker = Guid.NewGuid();
```
sharwell, line 228: Might as well box this so we don't repeatedly allocate the box.

```suggestion
        private static readonly object s_reportedMarker = Guid.NewGuid();
```

sharwell: Note that if you make this change, you'll also need to change the `==` operators to use `Equals` instead. They'll be just as fast for the identity cases, but when it crosses an RPC boundary it will switch to use value equality. The other option is to check it for null instead.

davidwengier: Thanks. The only two places this is checked is already doing a null check, so should be good.

src/Compilers/Core/Portable/InternalUtilities/FatalError.cs
```diff
@@ -225,7 +225,7 @@ public static bool ReportAndCatchUnlessCanceled(Exception exception, Cancellatio
 
 #endif
 
-        private static readonly object s_reportedMarker = new();
+        private static readonly object s_reportedMarker = Guid.NewGuid();
```
sharwell, line 228: 💡 Consider having a comment or test covering the serialization requirement. It seems like it would be trivial to regress.

sharwell (approved):

dibarbet (approved):

davidwengier:
Ping @dotnet/roslyn-compiler for reviews please. Very small change to shared code

cston (approved):

CyrusNajmabadi (approved):

jjonescz (approved):

PR #73861, opened 2024-06-05
Ignore workspaces that do not have IEditAndContinueWorkspaceService

tmat:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2048920

Workspaces that do not include Features layer wouldn't have the service (not sure what those are though).


tmat:
@dibarbet @jasonmalinowski Any idea what workspace doesn't have features layer?

jasonmalinowski:
@tmat: oh strange; the only workspaces I can imagine would be somebody who created their own workspaces with the Default MEF services...but then how those ended up getting in a place where ENC is trying to operate doesn't make sense to me. If you want to request cabs on this one that might be wise; maybe this is really a setup issue where assemblies didn't get loaded?

ToddGrun:
Approving after chatting with Tomas. He's requested cabs, but we'd like to avoid this from getting hit in new builds.

ToddGrun (approved):
:shipit:

tmat:
Will follow up on the memory dump, once we get one.

PR #73867, opened 2024-06-06
Fix serialization of capabilities in onautoinsert

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/7184


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/LanguageServer/LanguageServerHost.cs
```diff
@@ -29,7 +29,7 @@ internal sealed class LanguageServerHost
 
     public LanguageServerHost(Stream inputStream, Stream outputStream, ExportProvider exportProvider, ILogger logger, AbstractTypeRefResolver typeRefResolver)
     {
-        var messageFormatter = CreateJsonMessageFormatter();
+        var messageFormatter = RoslynLanguageServer.CreateJsonMessageFormatter();
```
dibarbet, line 32: this was the bug - the vscode serializer wasn't adding all of the vs extension converters.  I deleted the vscode one so we always have consistent serialization.

PR #73902, opened 2024-06-08
Throw when there is a syntax version mismatch in codelens

dibarbet:
Noticed in integration tests that sometimes we would get silent failures.  We should be throwing here with a specific exception that will tell the client to re-request the codelens items.

This also matches how inlay hints behaves.

### DIFF
--- a/src/LanguageServer/Protocol/Handler/CodeLens/CodeLensResolveHandler.cs
+++ b/src/LanguageServer/Protocol/Handler/CodeLens/CodeLensResolveHandler.cs
@@ -9,6 +9,7 @@ using Roslyn.Utilities;
 using Microsoft.CodeAnalysis.Shared.Extensions;
 using LSP = Roslyn.LanguageServer.Protocol;
 using System.Text.Json;
+using StreamJsonRpc;
 
 namespace Microsoft.CodeAnalysis.LanguageServer.Handler.CodeLens;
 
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Protocol/Handler/CodeLens/CodeLensResolveHandler.cs
```diff
@@ -48,11 +49,13 @@ public async Task<LSP.CodeLens> HandleRequestAsync(LSP.CodeLens request, Request
             ]
         };
 
-        // If the request is for an older version of the document, return a request with '- references'
+        // If the request is for an older version of the document, throw an exception so the client knows to re-query us.
         if (resolveData.SyntaxVersion != currentDocumentSyntaxVersion.ToString())
         {
-            context.TraceInformation($"Requested syntax version {resolveData.SyntaxVersion} does not match current version {currentDocumentSyntaxVersion}");
-            return request;
+            throw new LocalRpcException($"Resolve version {resolveData.SyntaxVersion} does not match current version {currentDocumentSyntaxVersion}")
```
sharwell, line 55: Placeholders should be wrapped in single quotes in exception messages

```suggestion
            throw new LocalRpcException($"Resolve version '{resolveData.SyntaxVersion}' does not match current version '{currentDocumentSyntaxVersion}'")
```

CyrusNajmabadi (approved):

Issue #72149, opened 2024-02-16
Add more incremental docs


333fred:
The incremental generator cookbook needs some more examples on:

* [ ] Testing
* [ ] Auto interface implementation examples

lsoft:
I would suggest to add subitem `- [ ] Testing incrementability` to avoid issues like https://github.com/dotnet/roslyn/issues/71629

333fred:
@jaredpar these things still need documentation.

PR #73906, opened 2024-06-08
Doc Cookbook Auto Interface Implementation

bdbonazz:
Added documentation of IncrementalGenerators Cookbook about Auto interface implementation

#72149

Working solution: https://github.com/bdbonazz/AutoImplementGenerator


bdbonazz:
@dotnet-policy-service agree

jaredpar:
@333fred, @CyrusNajmabadi PTAL

333fred (changes requested):
Unfortunately, this type of generator is not something that we want to put in the cookbook or recommend that people do. Going through all interfaces implemented by a type, particularly the way that ends up being implemented here (using a `Compilation` as an incremental parameter) is not actually incremental in any fashion and will kill IDE performance. As of yet, we have no ideas for how to implement this type of generator efficiently enough to suggest that people should write a generator in this fashion.

A restructuring of the generator that attributes the types that get their interfaces auto-implemented, rather than the interfaces themselves, would be acceptable to include. But the current structure is something that we explicitly recommend _against_ doing in this very document.

bdbonazz:
@333fred
I agree with you about performance issues, and I agree with @sharwell about using "ForAttributeWithMetadataName "
Do you think a solution of this type would be acceptable?

```csharp
[AutoImplement(nameof(IUserInterface), nameof(IUserInterface2))]
public partial class UserClass : IUserInterface, IUserInterface2
{
    public string UserProp { get; set; }
}
```

Moving the attribute to the class would let me use "ForAttributeWithMetadataName" to get only the decorated classes.
I would no longer cycle through all the interfaces of a class, but only the interfaces passed as an argument to the attribute.
Bonus point: the developer could decide when a class should auto-implement an interface and when to implement the interface manually

If at least one of you gives me the ok, I will start implementing this solution

333fred:
Yes, that's exactly the type of structure I meant meant when I said a change could be acceptable to include.

bdbonazz:
@333fred 
Perfect
Just one last thing:
You rightly had doubts about "using a Compilation as an incremental parameter".
Still I think I need Compilation in order to get INamedTypeSymbol interfaceSymbol from the interface name (and I need interfaceSymbol in order to have interfaceUsings, interfaceProperties, etc.)
I don't know if there is a more efficient way to do this.
Would you prefer I get the interface infos in the RegisterSourceOutput method like I'm doing now? Or should I get that infos in the transform argument of the ForAttributeWithMetadataName method?

333fred:
That info is already available in the context of your transform callback in `FAWMN`. See https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.cookbook.md#augment-user-code.

bdbonazz:
@333fred  You are right, now it should be more performant

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
 
 ### Auto interface implementation
 
-TODO: https://github.com/dotnet/roslyn/issues/72149
+**User scenario:** As a generator author I want to be able to implement the properties of the interfaces decorated with a specific attribute automatically for a user.
+
+**Solution:** Require the user to decorate the interface that they want to self-implement with the `[AutoImplement]` Attribute; Require the user to make the classes that implement the decorated interface be a `partial class`.
+Provide that attribute in a `RegisterPostInitializationOutput` step. Register for callbacks on the classes with
+`CreateSyntaxProvider` using the predicate `node is ClassDeclarationSyntax classDeclarationSyntax && classDeclarationSyntax.BaseList is not null && classDeclarationSyntax.BaseList.Types.Count > 0` to collect the classes that possibly implements at least one interface, And use tuples (or create an equatable model) to pass along that information.
```
333fred, line 643: Looks like this description still needs to be updated.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+    private const string AutoImplementAttributeClassName = $"{AutoImplementAttributeName}Attribute";
+    private const string FullyQualifiedMetadataName = $"{AutoImplementAttributeNameSpace}.{AutoImplementAttributeClassName}";
+
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        context.RegisterPostInitializationOutput(ctx =>
+        {
+            //Generate the AutoImplement Attribute
+            const string autoImplementAttributeDeclarationCode = $$"""
+using System;
```
333fred, line 684: It's good practice to include the autogenerated comment as well, even when using the `.g.cs` extension, just to make sure it's obvious.

```suggestion
// <auto-generated/>
using System;
```

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+namespace {{AutoImplementAttributeNameSpace}};
+
+[AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
```
333fred, line 687: Structs too? Fine if you want to say that's too complex for the workbook sample, but in that case, it would be good to call out the limitation in a comment explicitly.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+            return expression.Substring(NameOfString.Length, expression.Length - NameOfString.Length - 1);
+
+        //Trimming ""
+        string ret = expression.Trim('"');
+        return ret;
+    }
+    private static INamedTypeSymbol? GetInterfaceSymbol(Compilation compilation, string[] nameSpaces, string implementedInterfaceName)
+    {
+        // Try Get without NameSpace if interface was already Fully Qualified
+        INamedTypeSymbol? interfaceSymbol = compilation.GetTypeByMetadataName(implementedInterfaceName);
```
333fred, line 822: You don't need to do this. Change the `AutoImplement` attribute ctor to be `Type[]` instead of `string[]`, and then just leverage the attribute arguments which will already be types. It's both more robust doesn't need the lookup mechanism you're using here, which will likely fail.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+
+        IncrementalValuesProvider<Model> provider = context.SyntaxProvider.ForAttributeWithMetadataName(
+            fullyQualifiedMetadataName: FullyQualifiedMetadataName,
+            predicate: static (node, cancellationToken_) => node is ClassDeclarationSyntax,
+            transform: static (ctx, cancellationToken) =>
+            {
+                ClassDeclarationSyntax classDeclarationSyntax = (ClassDeclarationSyntax)ctx.TargetNode;
+                
+                string className = classDeclarationSyntax.Identifier.ValueText;
+                string classNameSpace = GetClassNameSpace(classDeclarationSyntax.Parent);
```
333fred, line 708: Don't do either of these things. The context has the type symbol for the attributed class; get this info directly from the symbol, not from syntax. This code will not work robustly; it won't handle nested types, types in the global namespace, or syntax that deviates from the standard single-namespace form.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+        return root.DescendantNodes()
+                .OfType<UsingDirectiveSyntax>()
+                .Select(usingDirective =>
+                {
+                    //get string "System" from string "using System;"
+                    string usingName = usingDirective.ToString().Split(' ').Last().TrimEnd(';');
+                    return usingName;
+                })
+                .ToArray();
+    }
```
333fred, line 787: I would not try to copy usings over. Instead, fully qualify everything. There's no telling what will conflict in the rest of the user's file with your own code.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+        string[] interfacesNames = GetInterfacesNames(classDeclarationSyntax.AttributeLists);
+
+        List<InterfaceModel> ret = [];
+        foreach (string interfaceName in interfacesNames)
+        {
+            INamedTypeSymbol? interfaceSymbol = GetInterfaceSymbol(compilation, usings, interfaceName);
+            if (interfaceSymbol is null)
+                continue;
+
+            //Get interface usings, in order to be sure to have the needed usings for the properties that will be generated
```
333fred, line 742: I would not try to copy usings over. Instead, fully qualify everything. There's no telling what will conflict in the rest of the user's file with your own code. It will also not handle interfaces defined in another assembly.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+                .GetMembers()
+                .OfType<IPropertySymbol>()
+                .ToArray();
+
+            InterfacePropertyModel[] propertyModels = interfaceProperties
+                .Select(interfaceProperty =>
+                {
+                    /*Using "interfaceProperty.Type" instead of "interfaceProperty.Type.Name" in order to not have error in specific cases.
+                    Es. the type "int" has Name "Int32", but writing "Int32" casue compilation error if we are not using "System" namespace*/
+                    string type = interfaceProperty.Type.ToString();
```
333fred, line 764: Use an appropriate `ToDisplayString()` instead. The same holds true for everywhere else you are converting type symbols to strings. You likely want `SymbolDisplayFormat.FullyQualifiedFormat`.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+            interfaceSymbol = compilation.GetTypeByMetadataName($"{nameSpace}.{implementedInterfaceName}");
+            if (interfaceSymbol is not null)
+                return interfaceSymbol;
+        }
+
+        return null;
+    }
+
+    private record Model(string ClassName, string ClassNameSpace, InterfaceModel[] Interfaces);
+    private record InterfaceModel(string Name, string NameSpace, string[] Usings, InterfacePropertyModel[] Properties);
```
333fred, line 837: Use an equatable list implementation, rather than arrays.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+                    """);
+                foreach (InterfacePropertyModel property in interfaceModel.Properties)
+                {
+                    //Check if property has a setter
+                    string setter = property.HasSetter
+                        ? "set; "
+                        : string.Empty;
+
+                    sourceBuilder.AppendLine($$"""
+                            public {{property.Type}} {{property.Name}} { get; {{setter}}}
```
333fred, line 866: Since this is only auto-implementing properties as `get; set;`, it would probably be good to narrow the name of `AutoImplementAttribute` to convey that up front for people reading the sample.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,248 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+                InterfaceModel[] interfaces = GetInterfaceModels(ctx.SemanticModel.Compilation, classDeclarationSyntax);
+
+                return new Model(
+                    className,
+                    classNameSpace,
+                    interfaces
+                    );
+            }).Where(m => m is not null);
+
+        context.RegisterSourceOutput(provider.Collect(), Execute);
```
333fred, line 718: Don't collect here; just pass in `provider` directly, and remove the outer `foreach` from `Execute`. I'd also rename `Execute` to be something more descriptive, rather than letting it be a relic of `ISourceGenerator`.

333fred (commented):
Done review pass. Large step in the right direction, but there's still plenty of room for improvement.

docs/features/incremental-generators.cookbook.md
bdbonazz: I'm really grateful for all your suggestions
By 'equatable list implementation' do you mean a custom implementation like 'public class EquatableList<T> : List<T>, IEquatable<EquatableList<T>>' or would you prefer another solution?

333fred: That would work. Here's what the cookbook says further up:

> Be careful of collection types in your models. Most built-in collection types in .NET do not do value equality by default. Arrays, ImmutableArray<T> and List<T>, for example, use reference equality, not value equality. We suggest that most generator authors use a wrapper type around arrays to augment them with value-based equality.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,188 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
 
 ### Auto interface implementation
 
-TODO: https://github.com/dotnet/roslyn/issues/72149
+**User scenario:** As a generator author I want to be able to implement the properties of interfaces passed as arguments to a decorator of a class automatically for a user
+
+**Solution:** Require the user to decorate the class with the `[AutoImplement]` Attribute and pass as arguments the types of the interfaces they want to self-implement themselves; The classes that implement the attribute have to be `partial class`.
+Provide that attribute in a `RegisterPostInitializationOutput` step. Register for callbacks on the classes with
+`ForAttributeWithMetadataName` using the fullyQualifiedMetadataName `FullyQualifiedAttributeName`, and use tuples (or create an equatable model) to pass along that information.
+The Attribute could work for Structs too, the example was kept simple on purpose for the workbook sample
```
333fred, line 644:
```suggestion
The attribute could work for structs too, the example was kept simple on purpose for the workbook sample.
```

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,188 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+    int InterfaceProperty { get; set; }
+}
+
+public interface IUserInterface2
+{
+    float InterfacePropertyOnlyGetter { get; }
+}
+```
+
+```csharp
```
333fred, line 660: Let's keep these snippets together so it's clear that they're both user code.

```suggestion

```

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,188 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+""";
+            ctx.AddSource($"{AttributeClassName}.g.cs", autoImplementAttributeDeclarationCode);
+        });
+
+        IncrementalValuesProvider<ClassModel> provider = context.SyntaxProvider.ForAttributeWithMetadataName(
+            fullyQualifiedMetadataName: FullyQualifiedAttributeName,
+            predicate: static (node, cancellationToken_) => node is ClassDeclarationSyntax,
+            transform: static (ctx, cancellationToken) =>
+            {
+                ISymbol classSymbol = ctx.SemanticModel.GetDeclaredSymbol(ctx.TargetNode, cancellationToken)!;
```
333fred, line 706:
```suggestion
                ISymbol classSymbol = ctx.TargetSymbol;
```

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,188 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+                context.AddSource(generatedFileName, sourceBuilder.ToString());
+            }
+        });
+    }
+
+    private static EquatableList<InterfaceModel> GetInterfaceModels(ISymbol classSymbol)
+    {
+        EquatableList<InterfaceModel> ret = [];
+
+        AttributeData attribute = classSymbol.GetAttributes().First(a => a.AttributeClass!.Name == AttributeClassName);
```
333fred, line 748: Just get this from the context. The `Attributes` property will have all the matched attributes; since `AllowMultiple` is false, you can just access the `[0]`th index and not care about any duplicates with errors.

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,188 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+        INamedTypeSymbol[] interfaceSymbols = attribute.ConstructorArguments[0].Values.Select(x => (x.Value as INamedTypeSymbol)!).ToArray();
```
333fred, line 749: You can't just assume that everything bound correctly and is a non-null type. The user could have an error.

```suggestion
        // A separate analyzer should error on using a non-interface type in this list.
        INamedTypeSymbol[] interfaceSymbols = attribute.ConstructorArguments[0].Values.Select(x => x.Type).Where(t => t is { TypeKind: TypeKind.Interface }).ToArray()!;
```

bdbonazz: You are right about "You can't just assume that everything bound correctly and is a non-null type. The user could have an error.", but it seems your solution is not working.
1) Every element of "_**attribute.ConstructorArguments[0].Values**_" in this case has _**Type.TypeKind == TypeKind.Class**_
2) "_**attribute.ConstructorArguments[0].Values.Select(x => x.Type).Where(t => t is { TypeKind: TypeKind.Interface }).ToArray()!**_" is a "_**ITypeSymbol[]**_", not a "_**INamedTypeSymbol[]**_". So the subsequent foreach is not working, cause it finds properties whose type is assembly.

I'll work on this

333fred: 1. Well, then you're not passing an interface.
2. All interfaces are `INamedTypeSymbol`s, you can just add a `Cast<INamedTypeSymbol>()` and it'll work fine.

bdbonazz: Sorry, probably I'm not understanding something
Screenshot 2024-06-21 213852
As you can see, even if I write "_**[AutoImplementProperties(typeof(IUserInterface), typeof(IUserInterface2))]**_" the TypeKind is "_**Class**_"

And both "_**tempo**_" and "**_tempo2_**" arrays have 61 elements (you can see some of them in the screenshot) instead of the 1 I expected, as every Interface in my sample only have 1 property

I'm keeping updated this repo in order to test every update I'm committing
https://github.com/bdbonazz/AutoImplementGenerator

333fred: Ah, I think I see. The select should select the `Value`, not `Type`, and check for `INamedTypeSymbol` there. `Type` in this case is the actual type of the argument, which is `System.Type`, not the type bound by the `typeof`. `Value` has that.

bdbonazz: Perfect
I added the check on TypeKind.Interface

docs/features/incremental-generators.cookbook.md
```diff
@@ -741,33 +738,37 @@ sealed class {{AttributeClassName}} : Attribute
...
-            EquatableList<string> properties = new();
+        if (attribute.ConstructorArguments.Length == 0)
+            return ret;
 
-            foreach (IPropertySymbol interfaceProperty in interfaceSymbol
-                .GetMembers()
-                .OfType<IPropertySymbol>())
+        foreach(TypedConstant constructorArgumentValue in attribute.ConstructorArguments[0].Values)
+        {
+            if (constructorArgumentValue.Value is INamedTypeSymbol interfaceSymbol && interfaceSymbol.TypeKind == TypeKind.Interface)
```
333fred, line 750:
```suggestion
            if (constructorArgumentValue.Value is INamedTypeSymbol { TypeKind: TypeKind.Interface } interfaceSymbol)
```

docs/features/incremental-generators.cookbook.md
```diff
@@ -636,7 +636,189 @@ TODO: https://github.com/dotnet/roslyn/issues/72149
...
+```csharp
+public interface IUserInterface
+{
+    int InterfaceProperty { get; set; }
+}
+
+public interface IUserInterface2
+{
+    float InterfacePropertyOnlyGetter { get; }
+}
```
333fred, line 657:
```suggestion
}

```

333fred (approved):
@CyrusNajmabadi @chsienki for reviews as well please.

chsienki (approved):

333fred:
Thanks @bdbonazz!

bdbonazz:
> Thanks @bdbonazz!
@333fred 
You are welcome
Thank you too, I learned a lot from your comments

PR #73919, opened 2024-06-10
Improve EnC log messages

tmat:
Log project name and path. The name includes TFM, which allows us to disambiguate between different flavors.


tmat:
@davidwengier ptal

davidwengier (approved):

PR #73942, opened 2024-06-11
Expose Roslyn LSP types to XAML and add readme

dibarbet:
We're exposing Roslyn LSP types to Razor and XAML to allow them to consume them when they run as a separate language inside the same Roslyn language server instance.

This will allow us to drastically simplify our server implementation as we do not have to handle different request handlers for the same LSP request using different types, and allows them to call into Roslyn helpers with concrete LSP types (instead of doing complicated wrapping).


src/LanguageServer/Protocol/Protocol/LanguageServer.Protocol.csproj
```diff
@@ -1,59 +0,0 @@
-﻿<Project>
```
dibarbet, line 1: this was an unused project file sitting in the repo (probably from when these types got moved from internal)

mgoertz-msft (approved):

src/LanguageServer/Protocol/Protocol/ServerCapabilities.cs
```diff
@@ -57,6 +57,13 @@ internal class ServerCapabilities
             set;
         }
 
+        [JsonIgnore]
+        public bool? HoverProvider2
```
davidwengier, line 61: Did you mean to commit this?

dibarbet: oops

davidwengier (approved):

PR #73947, opened 2024-06-11
Parse `field` as a contextual keyword

cston:
Spec change: https://github.com/dotnet/csharplang/pull/8239, https://github.com/dotnet/csharplang/pull/8283


src/Compilers/CSharp/Portable/Parser/SyntaxFactoryContext.cs
```diff
@@ -34,5 +34,15 @@ internal class SyntaxFactoryContext
...
+
+        /// <summary>
+        /// If an accessor kind changes, "field" and "value" within the accessor may need to be reinterpreted.
+        /// </summary>
+        internal bool IsInFieldKeywordContext;
+
+        /// <summary>
+        /// If an accessor kind changes, "field" and "value" within the accessor may need to be reinterpreted.
+        /// </summary>
+        internal bool IsInValueKeywordContext;
```
CyrusNajmabadi, line 46: you might want to mention how this differs from the above.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -3835,23 +3835,28 @@ private SyntaxToken TryEatCheckedOrHandleUnchecked(ref SyntaxToken operatorKeywo
...
 
-            // Check for expression body
-            if (this.CurrentToken.Kind == SyntaxKind.EqualsGreaterThanToken)
-            {
-                expressionBody = this.ParseArrowExpressionClause();
-            }
-            // Check if we have an initializer
-            else if (this.CurrentToken.Kind == SyntaxKind.EqualsToken)
+            // PROTOTYPE: SyntaxKind.GetAccessorDeclaration does seem correct for an initializer.
+            // PROTOTYPE: Confirm that field should be a keyword, and bind to the backing field, in a property initializer: object P { get; } = field;
```
CyrusNajmabadi, line 3845: this is interesting case for sure.  i actually have no idea what should happen here.  Property intitializers can't actually reference `this` or fields.  But if we do *not* have this bind to the `field` keyword, this could then bind to somethgin higher... and i have no idea if we want that :)

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -3876,7 +3881,32 @@ private SyntaxToken TryEatCheckedOrHandleUnchecked(ref SyntaxToken operatorKeywo
                 semicolon);
         }
 
-        private AccessorListSyntax ParseAccessorList(bool isEvent)
+        private readonly struct FieldAndValueKeywordContext : IDisposable
+        {
+            private readonly LanguageParser _parser;
+            private readonly (bool, bool) _isInFieldAndValueKeywordContext;
```
CyrusNajmabadi, line 3887: maybe `_originalXXX` to indicate it was the starting value.

CyrusNajmabadi: nit: separate fields just seems simpler :)

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4191,6 +4221,9 @@ private AccessorDeclarationSyntax ParseAccessorDeclaration(bool isEvent)
                 accessorName = ConvertToKeyword(accessorName);
             }
 
+            // PROTOTYPE: How should ‘field’ and ‘value’ be interpreted in the attributes, modifiers? Those parts have already been parsed.
```
CyrusNajmabadi, line 4224: we already decided this in ldm, but the answer, afair, is: they are not in scope there.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4266,6 +4299,20 @@ private static SyntaxKind GetAccessorKind(SyntaxToken accessorName)
             };
         }
 
+        private static (bool, bool) GetFieldAndValueKeywordContext(AccessorDeclaringKind declaringKind, SyntaxKind accessorKind)
+        {
+            return accessorKind switch
+            {
+                SyntaxKind.GetAccessorDeclaration => (declaringKind == AccessorDeclaringKind.Property, false),
```
CyrusNajmabadi, line 4306: named args for the literals.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -5108,7 +5155,7 @@ private static bool CanReuseVariableDeclarator(CSharp.Syntax.VariableDeclaratorS
             // declaration.  However, in the specific case of variable declarators, Dev10
             // specifically treats it as a variable name, even if it could be interpreted as a
             // keyword.
-            var name = this.ParseIdentifierToken();
+            var name = this.ParseIdentifierToken(reportErrorRatherThanConvertingFieldOrValue: true);
```
CyrusNajmabadi, line 5158: for somethign as core as ParseIdentifierToken, it feels *very* strange to need such a specialized arg.  esp.  since we haven't needed that sort of arg for other contextual tokens?

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -5439,7 +5486,7 @@ private EnumMemberDeclarationSyntax ParseEnumMemberDeclaration()
                 equalsValue = _syntaxFactory.EqualsValueClause(
                     this.EatToken(SyntaxKind.EqualsToken),
                     this.CurrentToken.Kind is SyntaxKind.CommaToken or SyntaxKind.CloseBraceToken
-                        ? this.ParseIdentifierName(ErrorCode.ERR_ConstantExpected)
+                        ? this.ParseIdentifierName(null, ErrorCode.ERR_ConstantExpected)
```
CyrusNajmabadi, line 5489: named arg for literal.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -5548,12 +5603,24 @@ private SyntaxToken ParseIdentifierToken(ErrorCode code = ErrorCode.ERR_Identifi
...
+                bool isFieldOrValueInKeywordContext = IsCurrentTokenFieldOrValueInKeywordContext();
                 SyntaxToken identifierToken = this.EatToken();
 
                 if (this.IsInAsync && identifierToken.ContextualKind == SyntaxKind.AwaitKeyword)
                 {
                     identifierToken = this.AddError(identifierToken, ErrorCode.ERR_BadAwaitAsIdentifier);
                 }
+                else if (isFieldOrValueInKeywordContext) // PROTOTYPE: Are we testing this with "value"?
+                {
+                    if (reportErrorRatherThanConvertingFieldOrValue.GetValueOrDefault())
```
CyrusNajmabadi, line 5615: not gettnig why we need this `bool?`.  seems out of sorts with how we genrally do contexts and parsing of identifiers/contextual-keywords.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -8562,7 +8641,7 @@ private bool IsPossibleStatement(bool acceptAccessibilityMods)
                     return true;
 
                 case SyntaxKind.IdentifierToken:
-                    return IsTrueIdentifier();
+                    return IsTrueIdentifier() || IsCurrentTokenFieldOrValueInKeywordContext();
```
CyrusNajmabadi, line 8644: this logic is what IsTrueIdentifier is supposed to already check.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -11624,11 +11718,11 @@ private ArgumentSyntax ParseArgumentExpression(bool isIndexer)
 
             if (isIndexer && this.CurrentToken.Kind is SyntaxKind.CommaToken or SyntaxKind.CloseBracketToken)
             {
-                expression = this.ParseIdentifierName(ErrorCode.ERR_ValueExpected);
+                expression = this.ParseIdentifierName(null, ErrorCode.ERR_ValueExpected);
```
CyrusNajmabadi, line 11721: named parametres.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -13722,17 +13834,26 @@ private new struct ResetPoint
             internal readonly TerminatorState TerminatorState;
             internal readonly bool IsInAsync;
             internal readonly bool IsInQuery;
+            // PROTOTYPE: Why do we need this? We don't need something similar for 'partial'
```
CyrusNajmabadi, line 13837: we need this because partial can look locally to figure out what it means.  it is not affected by outer contexts.  that's not the case for async/queries/field/value.  field/value shuld be treated very closely to async/await.  An outer context pushes us into a context where those tokens now become keywords.

fortunately field/value is also simpler because we never change the value as we go deeper into the accessor.

src/Compilers/CSharp/Portable/Syntax/Syntax.xml
```diff
@@ -910,6 +910,36 @@
...
+      <PropertyComment>
+        <summary>SyntaxToken representing the value keyword.</summary>
+      </PropertyComment>
+    </Field>
+    <TypeComment>
+      <summary>Class which represents the syntax node for a value expression.</summary>
+    </TypeComment>
+    <FactoryComment>
+      <summary>Creates a ValueExpressionSyntax node.</summary>
+    </FactoryComment>
```
CyrusNajmabadi, line 941: Alternatively, we could use LiteralExpressionSyntax, just like we use that for boolean expressions with `true` and `false` keyword.  but i'm not sure how compiler would feel about fallout of these being thought of as literals, so i think it's totally ok to not do that.

src/Compilers/Core/Portable/Syntax/GreenNode.cs
```diff
@@ -286,6 +286,9 @@ internal enum NodeFlags : ushort
             ContainsStructuredTrivia = 1 << 9,
 
             InheritMask = IsNotMissing | ContainsAnnotations | ContainsAttributes | ContainsDiagnostics | ContainsDirectives | ContainsSkippedText | ContainsStructuredTrivia,
+
+            FactoryContextIsInFieldKeywordContext = 1 << 10,
+            FactoryContextIsInValueKeywordContext = 1 << 11,
```
CyrusNajmabadi, line 291: i would move these up next to the other FactoryContext values.

CyrusNajmabadi: note: this also makes us out of bits again.

src/Compilers/Core/Portable/Syntax/GreenNode.cs
```diff
@@ -336,6 +339,22 @@ internal bool ParsedInIterator
...
 
+        internal bool ParsedInFieldKeywordContext
+        {
+            get
+            {
+                return (this.Flags & NodeFlags.FactoryContextIsInFieldKeywordContext) != 0;
+            }
+        }
+
+        internal bool ParsedInValueKeywordContext
```
CyrusNajmabadi, line 350: where is the code for this checked.   there should be some point that is checking this, and thus ensuring that if we do an incremental parse, where a statement used to be in a property, and it moves to be in a method, that the meaning of `field/value` go back to being an identifier (and vice versa).

note: this is not a strange scenario to be in.  people convert properties to methods and back.  So incremental parsing def needs to be good here.

CyrusNajmabadi:
done with initial pass.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
cston: Removed this case, assuming we will decide `field` is not a keyword (and should not bind to the backing field) in the property initializer.

src/Compilers/CSharp/Portable/Syntax/Syntax.xml
cston: Will leave as is.

src/Compilers/Core/Portable/Syntax/GreenNode.cs
cston: Is this property necessary? To switch the context between a property accessor and a method would require changing the containing member definition kind which I suspect invalidates the body of property, accessor, or method.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -11163,6 +11201,15 @@ private ExpressionSyntax ParseTermWithoutPostfix(Precedence precedence)
                             {
                                 return ParseDeclarationExpression(ParseTypeMode.Normal, isScoped: false);
                             }
+                            else if (IsCurrentTokenFieldInKeywordContext() && PeekToken(1).Kind != SyntaxKind.ColonColonToken)
+                            {
```
cston, line 11205: >{ [](http://example.com/codeflow?start=28&length=1)

Should report `ErrorCode.INF_IdentifierConflictWithContextualKeyword` here rather than in the binder, and should be reported regardless of whether the `field` token is treated as an identifier or a keyword.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -1744,10 +1782,16 @@ internal void ReportFieldContextualKeywordConflictIfAny(SyntaxNode syntax, Synta
...
-                diagnostics.Add(ErrorCode.INF_IdentifierConflictWithContextualKeyword, syntax, name, requiredVersion.ToDisplayString());
+                ReportFieldContextualKeywordConflict(syntax, identifier, diagnostics);
             }
         }
+
+        private static void ReportFieldContextualKeywordConflict(SyntaxNode syntax, SyntaxToken identifier, BindingDiagnosticBag diagnostics)
+        {
+            string name = identifier.Text;
+            var requiredVersion = MessageID.IDS_FeatureFieldKeyword.RequiredVersion();
+            diagnostics.Add(ErrorCode.INF_IdentifierConflictWithContextualKeyword, syntax, name, requiredVersion.ToDisplayString());
```
cston, line 1793: >ErrorCode.INF_IdentifierConflictWithContextualKeyword [](http://example.com/codeflow?start=28&length=53)

Should be reported in `LanguageParser.ParseTermWithoutPostfix` rather than here.

333fred (commented):
I've looked through most of the tests, but not `FieldKeywordTests.cs` yet. Will get to that later.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -1433,6 +1435,42 @@ private BoundExpression BindSizeOf(SizeOfExpressionSyntax node, BindingDiagnosti
...
+                    field = backingField;
+                    break;
+                case MethodSymbol { AssociatedSymbol: SourcePropertySymbol property }:
+                    field = property.BackingField;
+                    break;
+                default:
+                    {
+                        Debug.Assert((this.Flags & BinderFlags.InContextualAttributeBinder) != 0);
+                        var contextualAttributeBinder = TryGetContextualAttributeBinder(this);
+                        if (contextualAttributeBinder is { AttributeTarget: MethodSymbol { AssociatedSymbol: SourcePropertySymbol property } })
```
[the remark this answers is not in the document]
cston: I'd like to leave this as is since `PropertySymbol.BackingField` is not annotated currently so it seems appropriate to check `field is null` for the two code paths that rely on that property.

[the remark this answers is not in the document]
333fred: I'm not sure what you mean? I'll put a separate comment with the suggested change.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -1433,6 +1435,42 @@ private BoundExpression BindSizeOf(SizeOfExpressionSyntax node, BindingDiagnosti
...
+                            field = property.BackingField;
+                        }
+                        break;
+                    }
+            }
+
+            if (field is null)
+            {
+                throw ExceptionUtilities.UnexpectedValue(ContainingMember());
+            }
```
333fred, line 1468:
```suggestion
            SynthesizedBackingFieldSymbolBase field;

            ReportFieldContextualKeywordConflict(node, node.Token, diagnostics);

            switch (ContainingMember())
            {
                case SynthesizedBackingFieldSymbolBase backingField:
                    field = backingField;
                    break;
                case MethodSymbol { AssociatedSymbol: SourcePropertySymbol property }:
                    field = property.BackingField;
                    break;
                default:
                    {
                        Debug.Assert((this.Flags & BinderFlags.InContextualAttributeBinder) != 0);
                        var contextualAttributeBinder = TryGetContextualAttributeBinder(this);
                        if (contextualAttributeBinder is { AttributeTarget: MethodSymbol { AssociatedSymbol: SourcePropertySymbol property } })
                        {
                            field = property.BackingField;
                        }
                        else
                        {
                            throw ExceptionUtilities.UnexpectedValue(ContainingMember());
                        }
                        break;
                    }
            }
```

cston: Yes, that's what I understood your original suggestion to be. But `PropertySymbol.BackingField` is not annotated as returning a non-`null` value so I'd still want to check `field is { }` after the `switch`. (I could check `{ AssociatedSymbol: SourcePropertySymbol { BackingField: { } field }` in each of the two `switch` cases, but that seems redundant.)

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        [Fact]
+        public void FieldReference_01()
+        {
+            string source = """
+                class C
+                {
+                    static C _other = new();
+                    object P => _other.field;
```
333fred, line 163: `C` doesn't have any backing fields here; consider adding a setter that does use `field` to ensure that it's actually generated.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -0,0 +1,344 @@
...
+
+        [Fact]
+        public void FieldReference_03()
+        {
+            string source = """
+                class C
+                {
+                    int P
+                    {
+                        set { _ = this is { field: 0 }; }
```
333fred, line 204: Same comment about not having any backing fields.

src/Compilers/CSharp/Test/Syntax/Parsing/FieldKeywordParsingTests.cs
```diff
@@ -0,0 +1,1853 @@
...
+                                    N(SyntaxKind.IntKeyword);
+                                }
+                                N(SyntaxKind.IdentifierToken, "i");
+                            }
+                            N(SyntaxKind.CloseBracketToken);
+                        }
+                        N(SyntaxKind.ArrowExpressionClause);
+                        {
+                            N(SyntaxKind.EqualsGreaterThanToken);
+                            IdentifierNameOrFieldExpression(languageVersion, escapeIdentifier: false, expectedParsedAsToken);
```
333fred, line 366: I thought we said that `field` wasn't a keyword in indexers?

src/Compilers/CSharp/Test/Syntax/Parsing/FieldKeywordParsingTests.cs
```diff
@@ -0,0 +1,1853 @@
...
+                            {
+                                N(useRemove ? SyntaxKind.RemoveKeyword : SyntaxKind.AddKeyword);
+                                N(SyntaxKind.Block);
+                                {
+                                    N(SyntaxKind.OpenBraceToken);
+                                    N(SyntaxKind.ExpressionStatement);
+                                    {
+                                        N(SyntaxKind.SimpleAssignmentExpression);
+                                        {
+                                            IdentifierNameOrFieldExpression(languageVersion, escapeIdentifier: false, expectedParsedAsToken);
```
333fred, line 630: I thought `field` wasn't a contextual keyword in event accessors?

src/Compilers/CSharp/Test/Syntax/Parsing/FieldKeywordParsingTests.cs
333fred, line 1: Consider adding some tests of nested contexts, like lambdas and local functions.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
cston: `SourcePropertySymbolBase.BackingField` is declared in a nullable-disabled context, and if it was annotated, it would be nullable since the value will be `null` for properties other than auto properties.

333fred: Ah, ok, I was misunderstanding your comment: I was reading as "definitely doesn't return null", and you meant "is null-oblivious". Thanks.

src/Compilers/CSharp/Test/Syntax/Parsing/FieldKeywordParsingTests.cs
cston: `field` is not a keyword in indexers, and `expectedParsedAsToken` is `false` here. I've updated the tests where `expectedParsedAsToken` is always `false` to check for `N(SyntaxKind.IdentifierToken, "field");` explicitly instead.

333fred (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
```diff
@@ -254,17 +260,27 @@ public override OneOrMany<SyntaxList<AttributeListSyntax>> GetAttributeDeclarati
...
                     }
 
-                    if (accessor.Body != null || accessor.ExpressionBody != null)
+                    var body = (SyntaxNode?)accessor.Body ?? accessor.ExpressionBody;
+                    if (body != null)
                     {
                         accessorsHaveImplementation = true;
                     }
+
+                    usesFieldKeyword = usesFieldKeyword || containsFieldKeyword(accessor);
```
RikkiGibson, line 269: The other usage of this local function passes the `body` as an argument, should we do the same here?

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -160,13 +158,15 @@ private enum Flags : byte
...
+            if (usesFieldKeyword || (isAutoProperty && hasGetAccessor) || hasInitializer)
             {
                 Debug.Assert(!IsIndexer);
                 string fieldName = GeneratedNames.MakeBackingFieldName(_name);
                 BackingField = new SynthesizedBackingFieldSymbol(this,
                                                                       fieldName,
-                                                                      isReadOnly: (hasGetAccessor && !hasSetAccessor) || isInitOnly,
+                                                                      // Synthesized backing field for 'field' should not be marked 'initonly'
+                                                                      // since the field might be modified in the get accessor.
+                                                                      isReadOnly: !usesFieldKeyword && ((hasGetAccessor && !hasSetAccessor) || isInitOnly),
```
RikkiGibson, line 169: I'm not sure I follow this logic.. what if we are in a `readonly` property or a property on a readonly struct? But it seems like this would give a non-`readonly` field in that case.

I would expect to be able to use the `field` keyword in such cases, but it would need to refer to a `readonly` field. You would be able to write to the `field` in an `init` accessor, or a property initializer, or by assigning to the property in the constructor.

src/Compilers/CSharp/Portable/Syntax/InternalSyntax/CSharpSyntaxNode.cs
```diff
@@ -230,21 +230,11 @@ protected void SetFactoryContext(SyntaxFactoryContext context)
             {
                 SetFlags(NodeFlags.FactoryContextIsInQuery);
             }
-        }
-
-        internal static NodeFlags SetFactoryContext(NodeFlags flags, SyntaxFactoryContext context)
```
RikkiGibson, line 235: Was this just dead code?

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
```diff
@@ -254,17 +260,27 @@ public override OneOrMany<SyntaxList<AttributeListSyntax>> GetAttributeDeclarati
...
-                accessorsHaveImplementation = GetArrowExpression(syntax) is object;
+                var body = GetArrowExpression(syntax);
+                accessorsHaveImplementation = body is object;
+                usesFieldKeyword = body is { } && containsFieldKeyword(body);
                 Debug.Assert(accessorsHaveImplementation); // it's not clear how this even parsed as a property if it has no accessor list and no arrow expression.
             }
+
+            static bool containsFieldKeyword(SyntaxNode syntax)
+            {
+                return syntax.DescendantTokens().Any(static t => t.Kind() == SyntaxKind.FieldKeyword);
```
RikkiGibson, line 282: Does this traversal avoid creating the red tree?

RikkiGibson (commented):
Looks good but haven't gotten through tests yet.

src/Compilers/CSharp/Test/Syntax/Parsing/FieldKeywordParsingTests.cs
```diff
@@ -0,0 +1,2006 @@
...
+                            {
+                                N(SyntaxKind.SetKeyword);
+                                N(SyntaxKind.SemicolonToken);
+                            }
+                            N(SyntaxKind.CloseBraceToken);
+                        }
+                        N(SyntaxKind.EqualsValueClause);
+                        {
+                            N(SyntaxKind.EqualsToken);
+                            IdentifierNameOrFieldExpression(languageVersion, escapeIdentifier: false, expectedParsedAsToken);
```
RikkiGibson, line 98: It feels like it would be more straightforward to just use the generated baseline when we expect it to be the same in all the cases of the theory.

src/Compilers/CSharp/Test/Syntax/Syntax/FieldAndValueKeywordTests.cs
```diff
@@ -561,6 +582,7 @@ class C
             [CombinatorialValues(LanguageVersion.CSharp12, LanguageVersion.Preview)] LanguageVersion languageVersion)
         {
             string source = """
+                #pragma warning disable 168
```
RikkiGibson, line 585: Please include something like the message of the diagnostic which is being disabled or its name in ErrorCode to hint at what is going on here.

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
cston: I was assuming the comment *is* the tracking. (It might be useful to move the diagnostic to the parser, but it doesn't seem necessary, so I didn't think it required an issue.)

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
cston: The reason for passing `accessor` rather than `body` is to include attributes from the accessor.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
cston: No, but haven't we already created the red tree for the property simply by referencing the root `SyntaxNode`?

CyrusNajmabadi: no.  walking down the red tree causes the red allocations.  ideally this walk is done on the green tree to avoid actual allocs.

src/Compilers/CSharp/Portable/Syntax/InternalSyntax/CSharpSyntaxNode.cs
cston: Yes.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -3876,7 +3879,32 @@ private SyntaxToken TryEatCheckedOrHandleUnchecked(ref SyntaxToken operatorKeywo
                 semicolon);
         }
 
-        private AccessorListSyntax ParseAccessorList(bool isEvent)
+        private readonly struct FieldKeywordContext : IDisposable
```
CyrusNajmabadi, line 3882: could make a ref-struct, so this doesn't get put on the heap accidentally.

333fred (approved):

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
RikkiGibson: It's fine to not have an issue, but in that case I would expect a prototype comment, is there one present here that I missed?

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -0,0 +1,355 @@
...
+                        get { return field; }
+                        set { _ = this is { field: 0 }; }
+                    }
+                }
+                """;
+            var comp = CreateCompilation(source);
+            comp.VerifyEmitDiagnostics(
+                // (5,22): info CS9258: 'field' is a contextual keyword in property accessors starting in language version preview. Use '@field' instead.
+                //         get { return field; }
+                Diagnostic(ErrorCode.INF_IdentifierConflictWithContextualKeyword, "field").WithArguments("field", "preview").WithLocation(5, 22),
```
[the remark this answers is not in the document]
cston: I believe we want to report this info-level diagnostic always, even with language version preview, to allow users to discover where there may be a breaking change. But perhaps we should only report the diagnostic when `field` would otherwise bind to a member other than the synthesized field. cc @jaredpar

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -0,0 +1,355 @@
...
+            var model = comp.GetSemanticModel(tree);
+            var attributeArguments = tree.GetRoot().DescendantNodes().OfType<AttributeArgumentSyntax>().Select(arg => arg.Expression).ToArray();
+
+            var argument = attributeArguments[0];
+            Assert.IsType<IdentifierNameSyntax>(argument);
+            Assert.Null(model.GetSymbolInfo(argument).Symbol);
+        }
+
+        [Fact]
+        public void Attribute_02()
```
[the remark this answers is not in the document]
cston: I've added an item to the test plan to verify attributes can be attached to the synthesized field using `[field: ...]`, and I'll handle that in a subsequent PR.

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
cston: Should the synthesized field of a readonly property be readonly? Or should the field be mutable, regardless? Perhaps this should be an open question for language design?
```
readonly class C
{
    object P
    {
        get
        {
            field ??= ""; // ok?
            return field;
        }
    }
}
```

RikkiGibson: The field should be writable except when the accessor is `readonly` in the "readonly modifier" sense. i.e. the get-only case you have shown here should be writable.

cston: Added a PROTOTYPE comment for now.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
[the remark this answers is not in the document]
cston: Added a PROTOTYPE comment to `Binder.ReportFieldContextualKeywordConflict`.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
cston: If we refine the diagnostic reporting to only report when `field` would not bind to a distinct symbol, reporting will need to stay in the `Binder`. Given that, I'll leave the code as is, with no PROTOTYPE comment for now.

PR #73950, opened 2024-06-11
Add Microsoft.IdentityModel.JsonWebTokens to InternalTools.props and add restore source

akhera99:


JoeRobich (approved):
Saw that restore internal tools passed in the test build.

PR #73966, opened 2024-06-12
Add generated semantic search API list

tmat:
Allows to easily review the exact set of APIs available to semantic search.


tmat:
@jasonmalinowski ptal

src/Tools/SemanticSearch/BuildTask/GenerateFilteredReferenceAssembliesTask.cs
```diff
@@ -49,8 +50,7 @@ public sealed class GenerateFilteredReferenceAssembliesTask : Task
         \s*
         (?<Inclusion>[+|-]?)
         ((?<Kinds>[A-Za-z]+):)?
-        (?<MetadataName>[^#]*)
```
tmat, line 52: Can't use `#` for comment in the metadata name since `#ctor` is document id for constructor.

PR #73976, opened 2024-06-13
Fix error in compile loop

jaredpar:
This is a silly error where I was tracknig the wrong value in the loop condition. Hadn't noticed because the bad code works as long as `maxParallel` is greater than 1.


jaredpar:
@dotnet/roslyn-compiler PTAL. Silly error.

jjonescz (approved):

AlekseyTs (approved):

Issue #66722, opened 2023-02-06
Test plan for "roles" features


jcouv:
Spec: https://github.com/dotnet/csharplang/blob/main/proposals/extensions.md
Championed issue: https://github.com/dotnet/csharplang/issues/8697

Confirm metadata format for forward compatibility requirements before shipping. Consider marking assemblies with an attribute.

## TODO list
- remove logic for roles (`explicit` modifier, `IsExplicitExtension`, type erasure, treat extensions as static classes)
- add API guidelines to spec
- interop story with F# and VB (instance methods blocked for now)
- switch codegen to `static class`
- signature conflicts
- static scenarios
  - spec: block or allow extension resolution in `usings`
  - emit type references, satisfying constraints
  - static usings
  - problem with extension members in usings
  - spec: checking for implicit conversion when looking for compatible implicit extensions?
  - semantic model (`GetMemberGroup`, which plays in SigHelp), speculative semantic model, remaining Lookup API
  - fix downlevel support (update tests, maybe synthesize missing attribute)
  - hiding (warnings)
  - use-site diagnostics
  - dealing with substitution/unification ambiguities (`MergeEquivalentTypes`)
  - nullability
  - Public APIs (`IsExtension`)
  - XML docs
  - spec: Attributes on extension types (just use AttributeTarget.Struct?)
  - reserve `extension` keyword
  - events
  - spec: warn on slight type mismatches in extension resolution
  - Expression trees (see what we do for extension methods)
  - test ref analysis (expecting no impact)
  - pointer access, pointer element access, 
  - compilation events
  - interceptors
  - EnC
  - if we don't ship instance scenarios, we need to block them
- instance scenarios
  - use `modreq` instead of `modopt` to prevent accidental consumption of generated static members
  - erasure, access on instances of receiver type (`e.UnderlyingMember`)
  - type parameter receiver
  - merging extension methods and extension members within scopes instead of priority order
  - remaining lookup APIs
  - resolve extensions on type parameter values, adjust receiver accordingly
  - pattern-based lookups (`using`, implicit indexer access, ...)
  - patterns
  - spec: operators
  - spec: maybe add a tie breaker or preference in overload resolution 
  - indexer access on extension type
  - IOperation, CFG (should be minimal, if no new bound nodes from initial binding)
[...]

PR #73994, opened 2024-06-13
Emit extension marker method as public

jcouv:
Since we need the marker to roundtrip through reference assemblies, it should not be `private`.
Since the name is already unspeakable, I chose `public` accessibility.
Added a PROTOTYPE marker to record that the synthesized method is still not properly emitted in reference assemblies.

Corresponding spec update: https://github.com/dotnet/csharplang/pull/8121

### DIFF
--- a/src/Compilers/Test/Core/Metadata/MetadataReaderUtils.cs
+++ b/src/Compilers/Test/Core/Metadata/MetadataReaderUtils.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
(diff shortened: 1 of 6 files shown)


jjonescz (approved):

src/Compilers/Test/Core/Metadata/MetadataReaderUtils.cs
```diff
@@ -472,11 +473,26 @@ internal static void VerifyPEMetadata(string pePath, string[] types, string[] me
...
                     attributes);
             }
+
+            static StringHandle getAttributeTypeName(MetadataReader metadataReader, EntityHandle constructorHandle)
+            {
+                // See MetadataWriter.GetCustomAttributeTypeCodedIndex
+                if (constructorHandle.Kind == HandleKind.MemberReference)
+                {
+                    var typeRef = metadataReader.GetMemberReference((MemberReferenceHandle)constructorHandle).Parent;
+                    return metadataReader.GetTypeReference((TypeReferenceHandle)typeRef).Name;
```
AlekseyTs, line 487: >(TypeReferenceHandle)typeRef [](http://example.com/codeflow?start=59&length=28)

Strictly speaking this cast isn't always valid, I think. From spec for MemberRef metadata table: "Class (an index into the MethodDef, ModuleRef,TypeDef, TypeRef, or TypeSpec
tables; more precisely, a MemberRefParent (§II.24.2.6) coded index)." I am not going to block on this because the change is for a test helper, it is an improvement and it is sufficient for the test scenarios we currently have.

AlekseyTs (approved):
LGTM (commit 2)

Issue #73445, opened 2024-05-13
Test plan for "First-class Span Types"


jjonescz:
Proposal issue: https://github.com/dotnet/csharplang/issues/7905
Speclet: https://github.com/dotnet/csharplang/blob/main/proposals/first-class-span-types.md
Label: https://github.com/dotnet/roslyn/labels/New%20Feature%20-%20First-class%20Span%20Types
Feature branch: https://github.com/dotnet/roslyn/tree/features/FirstClassSpan

- [ ] conversions
  - [x] based on language version
  - [x] conversions:
    - [x] from `T[]` to `Span<U>`
    - [x] from `T[]` to `ReadOnlySpan<U>`
    - [x] from `Span<T>` to `ReadOnlySpan<U>`
    - [x] from `ReadOnlySpan<T>` to `ReadOnlySpan<U>`
    - [x] from `string` to `ReadOnlySpan<U>`
  - [x] implicit conversions between the above where the conversion from `T` to `U` is:
    - [x] no conversion
    - [x] identity conversion
    - [x] reference conversion
    - [x] non-reference conversion
    - [x] user-defined conversion
  - [x] explicit conversions between the above where the conversion from `T` to `U` is:
    - [x] *see implicit conversions*
  - [x] implicit and explicit conversions from the reverse of the above are disallowed (see `FirstClassSpanTests.*_Opposite*`)
  - [x] span conversions are *standard implicit conversions* (see `FirstClassSpanTests.*_ThroughUserImplicit`)
  - [x] implicit conversions for extension method receiver
    - [x] not considered for extension receiver in method group
  - [x] limited co-variance from `ReadOnlySpan<T>` to `ReadOnlySpan<U>`
  - [x] conversions to `params {ReadOnly}Span<T>` (see `FirstClassSpanTests.*_Params*`)
  - [x] no span conversion with `ref`, `ref readonly`, `in` modifiers
  - [x] no span conversion from multi-dimensional arrays
  - [x] method group parameter and return type conversions (disallowed) (see `FirstClassSpanTests.*_MethodGroup*`)
- [ ] pattern matching ignores span conversions
- [x] nullable analysis of element types
- [x] ref safety of implicit and explicit conversions (see `FirstClassSpanTests.*_RefSafety`)
- [x] null coalesce expressions (see `FirstClassSpanTests.*_Coalesce`)
- [x] conversions in expression trees (see `FirstClassSpanTests.*_ExpressionTree*`)
[...]

PR #74002, opened 2024-06-14
Ignore user-defined Span conversions

jjonescz:
- Ignores span UDCs
- Adds new "explicit span conversion"
- Fixes nullability analysis of span conversions

Corresponding speclet update: https://github.com/dotnet/csharplang/pull/8167

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionKind.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionKind.cs
@@ -70,5 +70,6 @@ namespace Microsoft.CodeAnalysis.CSharp
         InlineArray, // A conversion from an inline array to Span/ReadOnlySpan
 
         ImplicitSpan, // A conversion between array, (ReadOnly)Span, string - part of the "first-class Span types" feature
+        ExplicitSpan, // A conversion from array to (ReadOnly)Span
     }
 }
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionKind.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionKind.cs
@@ -69,7 +69,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
         InlineArray, // A conversion from an inline array to Span/ReadOnlySpan
 
-        ImplicitSpan, // A conversion between array, (ReadOnly)Span, string - part of the "first-class Span types" feature
+        ImplicitSpan, // A conversion from array to (ReadOnly)Span, or from string or (ReadOnly)Span to ReadOnlySpan
         ExplicitSpan, // A conversion from array to (ReadOnly)Span
     }
 }
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
@@ -1909,6 +1909,7 @@ namespace Microsoft.CodeAnalysis.CSharp
         public Conversion ClassifyImplicitExtensionMethodThisArgConversion(BoundExpression sourceExpressionOpt, TypeSymbol sourceType, TypeSymbol destination, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
         {
             Debug.Assert(sourceExpressionOpt is null || Compilation is not null);
+            // PROTOTYPE: Revert `TypeSymbol.Equals(...)` to `(object)sourceExpressionOpt.Type == sourceType` when implicit span is a "conversion from type".
             Debug.Assert(sourceExpressionOpt == null || TypeSymbol.Equals(sourceExpressionOpt.Type, sourceType, TypeCompareKind.ConsiderEverything));
             Debug.Assert((object)destination != null);
 
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
@@ -848,6 +848,11 @@ namespace Microsoft.CodeAnalysis.CSharp
                 return Conversion.ExplicitDynamic;
             }
 
+            if (HasExplicitSpanConversion(source, destination, ref useSiteInfo))
+            {
+                return Conversion.ExplicitSpan;
+            }
+
             return Conversion.NoConversion;
         }
 
@@ -995,6 +1000,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 case ConversionKind.ImplicitTupleLiteral:
                 case ConversionKind.ImplicitNullable:
                 case ConversionKind.ConditionalExpression:
+                case ConversionKind.ImplicitSpan:
                     return true;
 
                 default:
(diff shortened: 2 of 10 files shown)


jjonescz:

jjonescz:
@cston @333fred for reviews, thanks. Switching the conversions to be "from type" can be done in a follow up PR, I believe.

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -2710,6 +2874,40 @@ static class C
...
+        var comp = CreateCompilationWithSpan(source, parseOptions: TestOptions.Regular12);
+        CompileAndVerify(comp, expectedOutput: "11").VerifyDiagnostics();
+
+        // PROTOTYPE: Need to consider "implicit span conversion" in "better conversion target" for this to work.
+
+        var expectedDiagnostics = new[]
+        {
+            // (3,3): error CS0121: The call is ambiguous between the following methods or properties: 'C.M(object[])' and 'C.M(ReadOnlySpan<object>)'
+            // C.M(null);
+            Diagnostic(ErrorCode.ERR_AmbigCall, "M").WithArguments("C.M(object[])", "C.M(System.ReadOnlySpan<object>)").WithLocation(3, 3),
```
jjonescz, line 2901: I'm not sure if this should be an ambiguity - the problem is that `null` to Span UDC is not ignored and we don't have an equivalent standard conversion defined. So UDC is happening here instead of implicit span conversion which would be preferred (and there would be no ambiguity). Should the spec be updated with `null` to Span conversions?

If not then when we make the conversions "from type" ,"better conversion target" will kick in and the array overload will be picked.

jjonescz: That's true. However, consider `M(ReadOnlySpan<int>)` called as `M(null)` will use a user-defined conversion operator defined on ReadOnlySpan (from `int[]`) - so there is a "hole" in that we do not handle this case via standard implicit span conversions. Is that fine?

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
```diff
@@ -1903,7 +1909,7 @@ public Conversion ConvertExtensionMethodThisArg(TypeSymbol parameterType, TypeSy
         public Conversion ClassifyImplicitExtensionMethodThisArgConversion(BoundExpression sourceExpressionOpt, TypeSymbol sourceType, TypeSymbol destination, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
         {
             Debug.Assert(sourceExpressionOpt is null || Compilation is not null);
-            Debug.Assert(sourceExpressionOpt == null || (object)sourceExpressionOpt.Type == sourceType);
+            Debug.Assert(sourceExpressionOpt == null || TypeSymbol.Equals(sourceExpressionOpt.Type, sourceType, TypeCompareKind.ConsiderEverything));
```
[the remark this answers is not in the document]
jjonescz: It's failing when called from nullable walker - it passes equivalent types but not the same instances.

However, it shouldn't be necessary to change this when the expression is "from type".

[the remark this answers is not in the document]
333fred: Then let's leave a prototype comment to put this back.

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
333fred: > so there is a "hole" in that we do not handle this case via standard implicit span conversions. Is that fine?

That seems fine to me. We don't stack implicit standard conversions, and `null` being allowed to convert to `Span` via the UDC is often considered not great.

jjonescz: > and `null` being allowed to convert to `Span` via the UDC is often considered not great.

Just to be clear - that will continue to be allowed.

333fred: Right, but it's ok to me to not handle it through stacked standard implicit conversions.

333fred (approved):

jjonescz:

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionKind.cs
```diff
@@ -70,5 +70,6 @@ internal enum ConversionKind : byte
         InlineArray, // A conversion from an inline array to Span/ReadOnlySpan
 
         ImplicitSpan, // A conversion between array, (ReadOnly)Span, string - part of the "first-class Span types" feature
```
cston, line 72: >// A conversion between array, (ReadOnly)Span, string [](http://example.com/codeflow?start=22&length=53)

Consider making the comment for `ImplicitSpan` more specific, to match the comment for `ExplicitSpan`. Perhaps:
```
// A conversion from array to (ReadOnly)Span, or from string or (ReadOnly)Span to ReadOnlySpan
```

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -497,7 +509,7 @@ class C<T, U>
 
             var op = (IConversionOperation)model.GetOperation(cast)!;
             var conv = op.GetConversion();
-            Assert.Equal(ConversionKind.ImplicitSpan, conv.Kind);
+            Assert.Equal(ConversionKind.ExplicitSpan, conv.Kind);
```
cston, line 512: >ExplicitSpan [](http://example.com/codeflow?start=40&length=12)

Should the test be renamed `..._Explicit_...`?

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -1665,10 +1686,56 @@ class C
...
+            //     ReadOnlySpan<I<object>> M(I<string>[] x) => x;
+            Diagnostic(ErrorCode.ERR_NoImplicitConvCast, "x").WithArguments("I<string>[]", "System.ReadOnlySpan<I<object>>").WithLocation(5, 49)
+        };
+
+        CreateCompilationWithSpan(source, parseOptions: TestOptions.RegularNext).VerifyDiagnostics(expectedDiagnostics);
+        CreateCompilationWithSpan(source).VerifyDiagnostics(expectedDiagnostics);
+    }
+
+    [Fact]
+    public void Conversion_Array_ReadOnlySpan_Interface_Cast()
```
cston, line 1706: >ReadOnlySpan [](http://example.com/codeflow?start=33&length=12)

Consider adding a corresponding pair of tests (corresponding to this test and the one above) for `Span<T>`.

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -2426,6 +2577,19 @@ .maxstack  1
               IL_000b:  ret
             }
             """);
+
+        // Note: although a breaking change, the previous would fail with a runtime exception
+        // (Span's constructor checks that the element types are identical).
```
cston, line 2582: This definition of `Span<T>` does not contain a constructor.

cston (approved):

333fred (approved):

PR #74012, opened 2024-06-14
Enable end-to-end binding and emit for basic scenarios involving instance extension members.

AlekseyTs:
- Translate instance declarations in source to static declarations in metadata
  - A new parameter is added at the beginning.
  - The parameter's name is unspeakable
  - The type of the parameter is the extended type
  -  A modopt(System.Runtime.CompilerServices.ExtensionAttribute) is added to the type. On import the presence of the modopt is checked by verifying fully qualified name of the ExtensionAttribute type.
  - The parameter is a 'ref' parameter, unless it is known to be a reference type
- Rewrite usages of the instance APIs in bound tree to usages of the corresponding static APIs.
- Rewrite bound tree corresponding to a body of an instance extension method to a body corresponding to the static/metadata form of the method. This rewrite is done by `InstanceExtensionMethodBodyRewriter` right after the `LocalRewriter`. All the following rewrite phases and emit phase are performed with the static/metadata form as the owner of the body.
- Translate static declarations in metadata to language representation (i.e. an instance form). This allows binding to them as to instance API during semantic analysis.

### DIFF
--- a/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
+++ b/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
@@ -1433,6 +1433,10 @@ namespace Microsoft.CodeAnalysis.CSharp
                     loweredBody = (BoundStatement)extensionRewriter.Visit(loweredBody);
                     method = metadataMethod;
                 }
+                else
+                {
+                    loweredBody = InstanceExtensionMethodReferenceRewriter.Rewrite(method, loweredBody);
+                }
 
                 if (sawAwaitInExceptionHandler)
                 {
--- a/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
+++ b/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
@@ -296,6 +296,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                     var emittedBody = GenerateMethodBody(
                         moduleBeingBuilt,
                         synthesizedEntryPoint,
+                        methodBodySyntaxOpt: null,
                         methodOrdinal,
                         loweredBody,
                         ImmutableArray<EncLambdaInfo>.Empty,
@@ -775,9 +776,13 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                         if (_emitMethodBodies && !diagnosticsThisMethod.HasAnyErrors() && !_globalHasErrors)
                         {
+                            var sourceMethod = method as SourceMemberMethodSymbol;
+                            (BlockSyntax blockBody, ArrowExpressionClauseSyntax expressionBody) = sourceMethod?.Bodies ?? default;
+
                             emittedBody = GenerateMethodBody(
                                 _moduleBeingBuiltOpt,
                                 method,
+                                blockBody ?? expressionBody ?? sourceMethod?.SyntaxNode,
                                 methodOrdinal,
                                 loweredBody,
                                 ImmutableArray<EncLambdaInfo>.Empty,
@@ -1327,9 +1332,14 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                             lambdaRuntimeRudeEditsBuilder.Sort(static (x, y) => x.LambdaId.CompareTo(y.LambdaId));
 
+                            MethodSymbol metadataSymbol = (MethodSymbol)methodSymbol.ContainingType.TryGetCorrespondingStaticMetadataExtensionMember(methodSymbol) ?? methodSymbol;
+
+                            (BlockSyntax blockBody, ArrowExpressionClauseSyntax expressionBody) = sourceMethod?.Bodies ?? default;
+
                             var emittedBody = GenerateMethodBody(
                                 _moduleBeingBuiltOpt,
-                                methodSymbol,
+                                metadataSymbol,
+                                (SyntaxNode)blockBody ?? expressionBody ?? sourceMethod?.SyntaxNode,
                                 methodOrdinal,
                                 boundBody,
                                 lambdaDebugInfoBuilder.ToImmutable(),
@@ -1345,7 +1355,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                                 codeCoverageSpans,
                                 entryPointOpt: null);
 
-                            _moduleBeingBuiltOpt.SetMethodBody(methodSymbol.PartialDefinitionPart ?? methodSymbol, emittedBody);
+                            _moduleBeingBuiltOpt.SetMethodBody(metadataSymbol.PartialDefinitionPart ?? metadataSymbol, emittedBody);
                         }
                     }
 
@@ -1415,6 +1425,13 @@ namespace Microsoft.CodeAnalysis.CSharp
                 if (loweredBody.HasErrors)
                 {
                     return loweredBody;
+                }
+
+                if (method.ContainingType.TryGetCorrespondingStaticMetadataExtensionMember(method) is MethodSymbol metadataMethod)
+                {
+                    var extensionRewriter = new InstanceExtensionMethodBodyRewriter(method, (SourceExtensionMetadataMethodSymbol)metadataMethod);
+                    loweredBody = (BoundStatement)extensionRewriter.Visit(loweredBody);
+                    method = metadataMethod;
                 }
 
                 if (sawAwaitInExceptionHandler)
@@ -1493,6 +1510,7 @@ namespace Microsoft.CodeAnalysis.CSharp
         private static MethodBody GenerateMethodBody(
             PEModuleBuilder moduleBuilder,
             MethodSymbol method,
+            SyntaxNode methodBodySyntaxOpt,
             int methodOrdinal,
             BoundStatement block,
             ImmutableArray<EncLambdaInfo> lambdaDebugInfo,
@@ -1522,7 +1540,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             {
                 StateMachineMoveNextBodyDebugInfo moveNextBodyDebugInfoOpt = null;
 
-                var codeGen = new CodeGen.CodeGenerator(method, block, builder, moduleBuilder, diagnosticsForThisMethod, optimizations, emittingPdb);
+                var codeGen = new CodeGen.CodeGenerator(method, methodBodySyntaxOpt, block, builder, moduleBuilder, diagnosticsForThisMethod, optimizations, emittingPdb);
 
                 if (diagnosticsForThisMethod.HasAnyErrors())
                 {
--- a/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
+++ b/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
@@ -777,13 +777,10 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                         if (_emitMethodBodies && !diagnosticsThisMethod.HasAnyErrors() && !_globalHasErrors)
                         {
-                            var sourceMethod = method as SourceMemberMethodSymbol;
-                            (BlockSyntax blockBody, ArrowExpressionClauseSyntax expressionBody) = sourceMethod?.Bodies ?? default;
-
                             emittedBody = GenerateMethodBody(
                                 _moduleBeingBuiltOpt,
                                 method,
-                                blockBody ?? expressionBody ?? sourceMethod?.SyntaxNode,
+                                TryGetMethodBodySyntax(method as SourceMemberMethodSymbol),
                                 methodOrdinal,
                                 loweredBody,
                                 ImmutableArray<EncLambdaInfo>.Empty,
@@ -832,6 +829,16 @@ namespace Microsoft.CodeAnalysis.CSharp
                 stateMachineStateDebugInfoBuilder.Free();
             }
         }
+
+#nullable enable
+
+        private static SyntaxNode? TryGetMethodBodySyntax(SourceMemberMethodSymbol? sourceMethod)
+        {
+            (BlockSyntax blockBody, ArrowExpressionClauseSyntax expressionBody) = sourceMethod?.Bodies ?? default;
+            return blockBody ?? expressionBody ?? sourceMethod?.SyntaxNode;
+        }
+
+#nullable disable
 
         /// <summary>
         /// In some circumstances (e.g. implicit implementation of an interface method by a non-virtual method in a
@@ -1335,12 +1342,10 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                             MethodSymbol metadataSymbol = (MethodSymbol)methodSymbol.ContainingType.TryGetCorrespondingStaticMetadataExtensionMember(methodSymbol) ?? methodSymbol;
 
-                            (BlockSyntax blockBody, ArrowExpressionClauseSyntax expressionBody) = sourceMethod?.Bodies ?? default;
-
                             var emittedBody = GenerateMethodBody(
                                 _moduleBeingBuiltOpt,
                                 metadataSymbol,
-                                (SyntaxNode)blockBody ?? expressionBody ?? sourceMethod?.SyntaxNode,
+                                TryGetMethodBodySyntax(sourceMethod),
                                 methodOrdinal,
                                 boundBody,
                                 lambdaDebugInfoBuilder.ToImmutable(),
--- a/src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
@@ -940,7 +940,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             {
                 if (m.Kind == SymbolKind.Event)
                 {
-                    yield return (EventSymbol)m;
+                    yield return (EventSymbol)(TryGetCorrespondingStaticMetadataExtensionMember(m) ?? m);
                 }
             }
         }
@@ -1019,7 +1019,8 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             {
                 if (m.Kind == SymbolKind.Method)
                 {
-                    var method = (MethodSymbol)m;
+                    var method = (MethodSymbol)(TryGetCorrespondingStaticMetadataExtensionMember(m) ?? m);
+
                     if (method.ShouldEmit())
                     {
                         yield return method;
@@ -1036,7 +1037,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             {
                 if (m.Kind == SymbolKind.Property)
                 {
-                    yield return (PropertySymbol)m;
+                    yield return (PropertySymbol)(TryGetCorrespondingStaticMetadataExtensionMember(m) ?? m);
                 }
             }
         }
--- a/src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
@@ -1019,6 +1019,10 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             {
                 if (m.Kind == SymbolKind.Method)
                 {
+                    // PROTOTYPE(roles): Make sure this transformation is not going to break handling of partial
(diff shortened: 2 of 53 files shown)


AlekseyTs:
@jcouv, @jjonescz, @dotnet/roslyn-compiler Please review.

src/Compilers/CSharp/Portable/Lowering/InstanceExtensionMethodReferenceRewriter.cs
```diff
@@ -0,0 +1,219 @@
...
+            ImmutableArray<MethodSymbol> originalMethodsOpt,
+            BoundExpression? receiverOpt,
+            ImmutableArray<BoundExpression> arguments,
+            ImmutableArray<RefKind> argumentRefKinds,
+            bool invokedAsExtensionMethod,
+            TypeSymbol type)
+        {
+            var temps = ImmutableArray<LocalSymbol>.Empty;
+
+            if (receiverOpt is not null && method.OriginalDefinition.ContainingSymbol is NamedTypeSymbol declaringTypeDefinition &&
```
jcouv, line 122: Consider inverting the condition and returning early (without update) when possible.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -1417,6 +1427,17 @@ public override object VisitField(FieldSymbol symbol, TypeCompilationState argum
...
 
+                if (method.ContainingType.TryGetCorrespondingStaticMetadataExtensionMember(method) is MethodSymbol metadataMethod)
+                {
+                    var extensionRewriter = new InstanceExtensionMethodBodyRewriter(method, (SourceExtensionMetadataMethodSymbol)metadataMethod);
+                    loweredBody = (BoundStatement)extensionRewriter.Visit(loweredBody);
+                    method = metadataMethod;
+                }
+                else
+                {
+                    loweredBody = InstanceExtensionMethodReferenceRewriter.Rewrite(method, loweredBody);
```
jcouv, line 1438: The body of an instance extension method may reference instance methods from other extension types so shouldn't it also get the `InstanceExtensionMethodReferenceRewriter` treatment?

jcouv (commented):
Done with review pass (iteration 2). Tests not looked at yet (the test file won't load in VSCode so I think it'll have to be done in GitHub or maybe wait a couple of days for CodeFlow fix)

AlekseyTs:
> Tests not looked at yet (the test file won't load in VSCode so I think it'll have to be done in GitHub or maybe wait a couple of days for CodeFlow fix)

The majority of changes in the extensions test file are new tests added at the end. So, it shouldn't be that bad to review on GitHub, I think.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -775,9 +776,13 @@ private void CompileSynthesizedMethods(TypeCompilationState compilationState)
 
                         if (_emitMethodBodies && !diagnosticsThisMethod.HasAnyErrors() && !_globalHasErrors)
                         {
+                            var sourceMethod = method as SourceMemberMethodSymbol;
+                            (BlockSyntax blockBody, ArrowExpressionClauseSyntax expressionBody) = sourceMethod?.Bodies ?? default;
+
                             emittedBody = GenerateMethodBody(
                                 _moduleBeingBuiltOpt,
                                 method,
+                                blockBody ?? expressionBody ?? sourceMethod?.SyntaxNode,
```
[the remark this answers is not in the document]
AlekseyTs: > consider factoring this logic for computing methodBodySyntax here and below (line 1342) to a helper

Extracting a helper.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
AlekseyTs: > The body of an instance extension method may reference instance methods from other extension types so shouldn't it also get the `InstanceExtensionMethodReferenceRewriter` treatment?

It does, `InstanceExtensionMethodBodyRewriter` includes the same logic. This way we save on an extra rewrite pass, but the logic is duplicated to some degree. Unfortunately there is no common base between the rewriters where the logic could belong, and there is no inheritance relationship between the two rewriters. There is, however, an advantage from doing the work during `InstanceExtensionMethodBodyRewriter` apart from saving on an extra pass. We are rewriting `BoundThisReference`s to the synthesized parameter references. Those, have different types. By doing call rewrite at the same time we never end up with `BoundCall` nodes having type mismatch between "this" value and the target synthesized parameter. Quite possibly this (the mismatch) is not a big deal, since extension instances can come from sources other than `this` references. I will add a PROTOTYPE comment to reevaluate this later.

src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
```diff
@@ -1019,7 +1019,8 @@ internal virtual IEnumerable<MethodSymbol> GetMethodsToEmit()
             {
                 if (m.Kind == SymbolKind.Method)
                 {
-                    var method = (MethodSymbol)m;
+                    var method = (MethodSymbol)(TryGetCorrespondingStaticMetadataExtensionMember(m) ?? m);
```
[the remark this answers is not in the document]
AlekseyTs: > The logic below in `ShouldEmit` checks whether a partial method has an implementation. Will that work with a static metadata extension method?

This is a good observation. Adding a prototype comment.

src/Compilers/CSharp/Portable/Lowering/InstanceExtensionMethodBodyRewriter.cs
```diff
@@ -0,0 +1,200 @@
...
+    /// that corresponds to its static/metadata form.
+    /// </summary>
+    internal sealed class InstanceExtensionMethodBodyRewriter : BoundTreeToDifferentEnclosingContextRewriter
+    {
+        private readonly SourceExtensionMetadataMethodSymbol _metadataMethod;
+        private ImmutableDictionary<Symbol, Symbol> _symbolMap;
+        private RewrittenMethodSymbol _rewrittenContainingMethod;
+
+        public InstanceExtensionMethodBodyRewriter(MethodSymbol sourceMethod, SourceExtensionMetadataMethodSymbol metadataMethod)
+        {
```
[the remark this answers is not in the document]
AlekseyTs: > nit: consider asserting the relationship between those two methods. Maybe also asserting that `sourceMethod` is an instance method in an extension type

Adding an assert

src/Compilers/CSharp/Portable/Lowering/InstanceExtensionMethodBodyRewriter.cs
```diff
@@ -0,0 +1,200 @@
...
+                }
+                _symbolMap = builder.ToImmutable();
+            }
+
+            _rewrittenContainingMethod = rewritten;
+
+            return (savedContainer, saveSymbolMap);
+        }
+
+        private (RewrittenMethodSymbol, ImmutableDictionary<Symbol, Symbol>) EnterMethod(MethodSymbol symbol, RewrittenMethodSymbol rewritten)
```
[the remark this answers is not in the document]
AlekseyTs: > nit: consider tightening the `rewritten` parameter to be a `RewrittenLambdaOrLocalFunctionSymbol`

Will do

src/Compilers/CSharp/Portable/Lowering/InstanceExtensionMethodReferenceRewriter.cs
AlekseyTs: > Consider inverting the condition and returning early (without update) when possible.

It is quite possible I misunderstand the suggestion, but we cannot skip update, ever, because the arguments and the receiver might not match what we have in the original BoundCall, they have been rewritten. In this `if` we might do additional adjustments. Given that, I do not see much advantage in inverting the condition.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_ObjectOrCollectionInitializerExpression.cs
```diff
@@ -663,7 +663,9 @@ private BoundExpression EvaluateSideEffects(BoundExpression arg, RefKind refKind
 
 #if DEBUG
             var discardedUseSiteInfo = CompoundUseSiteInfo<AssemblySymbol>.Discarded;
-            Debug.Assert(_compilation.Conversions.ClassifyConversionFromType(rewrittenReceiver.Type, memberSymbol.ContainingType, isChecked: false, ref discardedUseSiteInfo).IsImplicit);
+            Debug.Assert(_compilation.Conversions.ClassifyConversionFromType(rewrittenReceiver.Type, memberSymbol.ContainingType, isChecked: false, ref discardedUseSiteInfo).IsImplicit ||
+                         (memberSymbol.ContainingType.GetExtendedTypeNoUseSiteDiagnostics(null) is { } extended &&
+                          _compilation.Conversions.ClassifyConversionFromType(rewrittenReceiver.Type, extended, isChecked: false, ref discardedUseSiteInfo).IsImplicit));
```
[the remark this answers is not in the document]
AlekseyTs: > Consider leaving a PROTOTYPE comment here. In the future, there should be an implicit conversion to/from extension type, so we shouldn't need the additional case.

I am going to revert this change instead. It looks like all affected tests are properly disabled or excluded from "UsedAssemblies" leg.

src/Compilers/CSharp/Portable/Lowering/SyntheticBoundNodeFactory.cs
```diff
@@ -1690,12 +1695,47 @@ internal BoundExpression Not(BoundExpression expression)
...
+            SyntaxNode? syntaxOpt = null
+#if DEBUG
+            , [CallerLineNumber] int callerLineNumber = 0
+            , [CallerFilePath] string? callerFilePath = null
 #endif
             )
         {
             Debug.Assert(argument.Type is { });
-            MethodSymbol? containingMethod = this.CurrentFunction;
-            Debug.Assert(containingMethod is { });
```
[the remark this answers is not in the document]
AlekseyTs: > Consider keeping the assertion that `containingMethod` is not null

Will do.

src/Compilers/CSharp/Portable/Symbols/Extensions/RewrittenExtensionMethodSymbol.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    internal abstract class RewrittenExtensionMethodSymbol : RewrittenMethodSymbol
+    {
+        protected RewrittenExtensionMethodSymbol(MethodSymbol originalMethod) : base(originalMethod, TypeMap.Empty)
+        {
+            Debug.Assert(originalMethod.ContainingType.GetExtendedTypeNoUseSiteDiagnostics(null) is not null);
+        }
+
+        public sealed override bool IsExtensionMethod => false;
+        public sealed override bool IsVirtual => false;
```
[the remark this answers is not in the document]
AlekseyTs: > Taking this property as an example, would it be worthwhile to assert that the original method was also not virtual?

I prefer to not do that because we do not control what comes from metadata and I do not want to worry how erroneous modifiers are are handled on source. The current implementation reflects how we decided to see extension methods and I don't really care what the original method says.

src/Compilers/CSharp/Portable/Symbols/Extensions/RewrittenLambdaOrLocalFunctionSymbol.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        protected override ImmutableArray<ParameterSymbol> MakeParameters()
+        {
+            var sourceParameters = _originalMethod.Parameters;
+            var parameters = ArrayBuilder<ParameterSymbol>.GetInstance(ParameterCount);
+
+            foreach (var parameter in sourceParameters)
+            {
+                parameters.Add(new RewrittenMethodParameterSymbol(this, parameter));
```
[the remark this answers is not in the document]
AlekseyTs: > This logic seems duplicated below.

Good catch. An artifact of an incomplete refactoring. Will remove.

AlekseyTs:
@jcouv, @jjonescz, @dotnet/roslyn-compiler Please review.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
jcouv: Would it be possible to only use `InstanceExtensionMethodBodyRewriter`, but disable/skip the parts that only apply to extension method bodies?

AlekseyTs: At the moment ```InstanceExtensionMethodBodyRewriter``` is very specialized for the task of rewriting in a different context. Having two different rewriters feel simpler clearer to me.

src/Compilers/CSharp/Portable/Symbols/Extensions/SynthesizedExtensionThisParameterMetadataSymbol.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        // PROTOTYPE(roles): Confirm what name should we use for this parameter
+        //                   Is this name going to hurt or help EE when this parameter is lifted into a closure?
+        public override string Name => GeneratedNames.ThisProxyFieldName();
+
+        public override bool IsDiscard => false;
+
+        public override TypeWithAnnotations TypeWithAnnotations
+            => TypeWithAnnotations.Create(
+                    ((SourceExtensionTypeSymbol)_containingSymbol.ContainingSymbol).GetExtendedTypeNoUseSiteDiagnostics(null),
```
jcouv, line 32: Just want to confirm, from our discussion earlier this week, we'll want to change this to the extension type at some point (instead of the extended type) and we'll let erasure deal with it.

AlekseyTs: > Just want to confirm, from our discussion earlier this week, we'll want to change this to the extension type at some point (instead of the extended type) and we'll let erasure deal with it.

While that approach probably is going to work, at the moment I am not sure what would be an advantage of doing that.

AlekseyTs: Though, I think it will be worth following up once the "erase" functionality is there.

jcouv:
>         var verifier = CompileAndVerify(comp, expectedOutput: IncludeExpectedOutput("2"), verify: ExecutionConditionUtil.IsMonoOrCoreClr ? Verification.Passes : Verification.Skipped).VerifyDiagnostics();

Can this be simplified to `Verification.FailsPeVerify` (here and below)?

---
Refers to: src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs:39921 in 2b11520. [](commit_id = 2b115207af9e705cae61188e3eadd3a5dc99b2bb, deletion_comment = False)

jcouv:
>     // PROTOTYPE(roles): Add flavors of "_Metadata" types for an underlying type a type parameter known/unknown to be a reference type

typo? types -> tests

---
Refers to: src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs:42692 in 2b11520. [](commit_id = 2b115207af9e705cae61188e3eadd3a5dc99b2bb, deletion_comment = False)

AlekseyTs:
>         var verifier = CompileAndVerify(comp, expectedOutput: IncludeExpectedOutput("2"), verify: ExecutionConditionUtil.IsMonoOrCoreClr ? Verification.Passes : Verification.Skipped).VerifyDiagnostics();

> Can this be simplified to ```Verification.FailsPeVerify``` (here and below)?

I do not know. It definitely doesn't have the same meaning for me and I wouldn't think they are equivalent if I saw the two values side by side. To summarize, I would prefer to keep the code as is. 

---
In reply to: [2181290271](https://github.com/dotnet/roslyn/pull/74012#issuecomment-2181290271) [](http://example.com/codeflow?ancestors=2181290271)

---
Refers to: src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs:39921 in 2b11520. [](commit_id = 2b115207af9e705cae61188e3eadd3a5dc99b2bb, deletion_comment = False)

jcouv:
> 		.addon void E::add_E2(valuetype C modopt([System.Runtime]System.Runtime.CompilerServices.ExtensionAttribute)&, class [System.Runtime]System.Action)

You'd mentioned the CLS-compliance issue for this signature for event accessors. How do you want to capture that?

---
Refers to: src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs:42669 in 2b11520. [](commit_id = 2b115207af9e705cae61188e3eadd3a5dc99b2bb, deletion_comment = False)

jcouv (commented):
LGTM Thanks (iteration 6)

AlekseyTs:
@jjonescz, @dotnet/roslyn-compiler For the second review.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PENamedTypeSymbol.cs
```diff
@@ -2336,25 +2379,61 @@ private IEnumerable<PENamedTypeSymbol> CreateNestedTypes()
...
+                        (PEMethodSymbol setMethod, PEExtensionInstanceMethodSymbol extensionSet) = GetAccessorMethod(module, methodHandleToSymbol, _handle, methods.Setter);
 
                         if (((object)getMethod != null) || ((object)setMethod != null))
                         {
-                            members.Add(PEPropertySymbol.Create(moduleSymbol, this, propertyDef, getMethod, setMethod));
+                            PEPropertySymbol property = PEPropertySymbol.Create(moduleSymbol, this, propertyDef, getMethod, extensionGet, setMethod, extensionSet);
+
+                            // If this check causes a perf problem, we could optimize it.
+                            if (extendedType is not null && property.IsStatic && property.ParameterCount > 0 &&
+                                IsSynthesizedExtensionThisParameter(extendedType, property.Parameters[0]))
```
jjonescz, line 2417: Should we also check `extensionGet is not null || extensionSet is not null` like we do for events?

jjonescz (approved):

AlekseyTs:
> 		.addon void E::add_E2(valuetype C modopt([System.Runtime]System.Runtime.CompilerServices.ExtensionAttribute)&, class [System.Runtime]System.Action)

> You'd mentioned the CLS-compliance issue for this signature for event accessors. How do you want to capture that?

I am planning to add some tests demonstrating that events cannot be consumed by VB. And also would like to check previous version of C# compiler.

---
In reply to: [2181329185](https://github.com/dotnet/roslyn/pull/74012#issuecomment-2181329185) [](http://example.com/codeflow?ancestors=2181329185)

---
Refers to: src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs:42669 in 2b11520. [](commit_id = 2b115207af9e705cae61188e3eadd3a5dc99b2bb, deletion_comment = False)

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PENamedTypeSymbol.cs
AlekseyTs: > Should we also check extensionGet is not null || extensionSet is not null like we do for events?

Unlike an event, a property has its own signature, which by itself defines if we need to transform it.

src/Compilers/CSharp/Portable/Lowering/InstanceExtensionMethodBodyRewriter.cs
```diff
@@ -0,0 +1,213 @@
...
+                    rewrittenReceiver,
+                    this.VisitList(node.Arguments),
+                    node.ArgumentRefKindsOpt,
+                    node.InvokedAsExtensionMethod,
+                    this.VisitType(node.Type));
+            }
+        }
+
+        // PROTOTYPE(roles): Here we are pretty much duplicating what InstanceExtensionMethodReferenceRewriter would do.
+        //                   We need to reevaluate whether we are getting enough advantage from this duplication.
```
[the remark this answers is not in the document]
AlekseyTs: > Is this comment still relevant? Looks like this method delegates to InstanceExtensionMethodReferenceRewriter.

Yes it is relevant. The comment is about the fact that two rewriters  are doing the same thing, not about code duplication.

jcouv (approved):

PR #72755, opened 2024-03-27
Warn for `yield return` in `lock`

jjonescz:
Resolves https://github.com/dotnet/roslyn/issues/72443.
Test plan: https://github.com/dotnet/roslyn/issues/72662
Speclet: https://github.com/dotnet/csharplang/blob/main/proposals/ref-unsafe-in-iterators-async.md

Issue #72662, opened 2024-03-22
Test plan for "ref and unsafe in iterators and async" feature


jjonescz:
Speclet: https://github.com/dotnet/csharplang/blob/main/proposals/ref-unsafe-in-iterators-async.md
Feature branch: https://github.com/dotnet/roslyn/tree/features/RefInAsync
Label: https://github.com/dotnet/roslyn/labels/New%20Feature%20-%20Ref%2FUnsafe%20in%20Iterators%2FAsync

- [ ] Review feature specification at LDM
- [x] Breaking change around `unsafe` context within iterators is documented 
- [x] language version
  - [x] `ref` / `ref struct` local in 
    - [x] `async` method
    - [x] iterator method
  - [x] `unsafe` block is allowed in C# 13 in
    - [x]  `async` method (this is a preserved legacy behavior)
    - [x] iterator method
  - [x] `unsafe` modifier is allowed on an iterator in C# 13, still an error (a language version error) in previous versions
    - [x] regular method
    - [x] property get
    - [x] local function
    - [x] operator 
  - [x] effect on `unsafe` context within iterator's body (the breaking change) 
    - [x] regular method
    - [x] property get
    - [x] local function
    - [x] operator 
- [ ] use `ref` / `ref struct` local
  - [x] within / across `await`
  - [x] within / across `yield return;` in
    - [x] regular iterators
    - [x] `async` iterators
  - [x] across `yield break;` in
    - [x] regular iterators
    - [x] `async` iterators
  - [x] within / across `async foreach` block
  - [x] `yield` in nested block
  - [x] within `async` lambda / local function
  - [x] within iterator local function
  - [x] within property / indexer `get` iterator
  - [ ] error in `async` / iterator even in `unsafe` context
  - [x] await in elements of stackalloc, the result span should be usable afterwards 
  - [x] ```async Task<int[]> M() => ... ReadOnlySpan<int> x = await M();``` is expected to work
  - [x] ```Buffer2<int> b = [await a, await b]; // inline array type (ref struct)```
  - [x] ```using (refStruct = ...) { await ...; }```
  - [x] ```foreach (var _ in refStruct) { await ...; }```
  - [x] Interpolation handlers with awaits in the middle handler is a ref struct
    - [x] no awaits in async method should work
    - [x] Has await should be blocked (turns out it works but doesn't use the interpolation handler)
- [x] `unsafe` context
  - [x] unannotated / `unsafe` containing type
[...]

Issue #72443, opened 2024-03-07
Async iterators permit yield return inside of lock blocks


stephentoub:
**Version Used**: 
https://github.com/dotnet/roslyn/commit/d8d7886808a68c720a246b09028f870bc62c30fe

**Steps to Reproduce**:
[SharpLab](https://sharplab.io/#v2:EYLgtghglgdgPgAQAwAIEEYAsBuAsAKGTXQFY98CEBOFGAUwHcUBhACgEoA6ABQCcB7AMZ0AziIBqEADYBXOgEERATxiCO5SgGY0AJhYEA3gRQm02hAA40ANhR8hoidLmKVa9sdNH8p32hoAZvy8dBCCABYorLAALihQMXRg8TAoAOJ0MZKyoq6qHB4+fibexcXUNpwAInRSEEqs6OzkZSgAvp4mHRRFJgAOvFAAbhCJaFYAknmCAKIwMmB0vBDAUnQAPLEAfOmZ2XIi0xydKKVlFQjWnACaULUAJuonvlJCANZRMeFQIoWtZ60TEFeFFYvEUABeFBIbDg9YodAw+IAamRf0BJWeGOI6DQAHZ4i1sd1WiTTN02kA)
```C#
await new C().ProcessValueAsync();

public class C
{
    public async Task ProcessValueAsync()
    {
        await foreach (int item in GetValuesAsync())
        {
            await Task.Delay(1);
        }
    }

    private async IAsyncEnumerable<int> GetValuesAsync()
    {
        await Task.Yield();
        lock (this)
        {
            for (int i = 0; i < 10; i++)
            {
                yield return i;
            }
        }
    }
}
```

**Expected Behavior**:
Fails to compile, with an error about not yielding while holding a lock.

**Actual Behavior**:
Compiles without error, but blows up at runtime with an exception:
```
Unhandled exception. System.Threading.SynchronizationLockException: Object synchronization method was called from an unsynchronized block of code.
   at System.Threading.Monitor.Exit(Object obj)
   at C.GetValuesAsync()+MoveNext()
```

Since normal synchronous `IEnumerable<T>` iterators were introduced, `yield return` has unfortunately been allowed while holding a `lock`. This is problematic because it means the caller iterating through the iterator will also be holding the lock, e.g. this:
```C#
object o = new object();
Console.WriteLine($"Before: {Monitor.IsEntered(o)}");
using (IEnumerator<int> e = GetValues(o).GetEnumerator())
{
[...]

PR #74024, opened 2024-06-17
Remove warning for `yield return` in `lock`

jjonescz:
See https://github.com/dotnet/roslyn/issues/72443#issuecomment-2147915772.
Reverts https://github.com/dotnet/roslyn/pull/72755 (except tests).
Speclet update: https://github.com/dotnet/csharplang/pull/8220
Test plan: https://github.com/dotnet/roslyn/issues/72662


src/Compilers/CSharp/Test/Semantic/Semantics/IteratorTests.cs
```diff
@@ -210,18 +197,6 @@ static IEnumerable<int> GetValues(object obj)
 
             CompileAndVerify(source, options: TestOptions.ReleaseExe.WithWarningLevel(8),
                 expectedOutput: expectedOutput).VerifyDiagnostics();
-
-            var expectedDiagnostics = new[]
```
AlekseyTs, line 214: >var expectedDiagnostics = new[] [](http://example.com/codeflow?start=12&length=31)

Should the `WithWarningLevel(8)` be removed above?

AlekseyTs:

jjonescz:
@cston @dotnet/roslyn-compiler for a second review, thanks

cston (approved):

Issue #74033, opened 2024-06-17
`GeneratorDriver.GetRunResult()` should not throw


chsienki:
If you call `.GetRunResult()` on a driver that hasn't run, but has generators in it, we null-ref.

```csharp
        [Fact]
        public void RunResults_Are_Empty_Before_Generation_With_Generators()
        {
            var generator = new SingleFileTestGenerator("public class D {}", "source.cs");

            GeneratorDriver driver = CSharpGeneratorDriver.Create([generator], parseOptions: TestOptions.Regular);
            var results = driver.GetRunResult();

            Assert.Empty(results.GeneratedTrees);
            Assert.Empty(results.Diagnostics);
            Assert.Empty(results.Results);
        }
```

See [internal link](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2080052?src=WorkItemMention&src-action=artifact_link)

Peter-Juhasz:
Is this the same as this one I encountered in VS 17.11.0 P2?

```
StreamJsonRpc.RemoteInvocationException: Object reference not set to an instance of an object.
   at StreamJsonRpc.JsonRpc.<InvokeCoreAsync>d__154`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at Microsoft.CodeAnalysis.Remote.BrokeredServiceConnection`1.<TryInvokeAsync>d__18`1.MoveNext()
RPC server exception:
System.NullReferenceException: Object reference not set to an instance of an object.
      at Microsoft.CodeAnalysis.GeneratorDriver.<GetRunResult>g__getGeneratorSources|15_1(GeneratorState generatorState)
      at Microsoft.CodeAnalysis.GeneratorDriver.<>c.<GetRunResult>b__15_0(ISourceGenerator generator, GeneratorState generatorState)
      at Microsoft.CodeAnalysis.ImmutableArrayExtensions.ZipAsArray[T1,T2,TResult](ImmutableArray`1 self, ImmutableArray`1 other, Func`3 map)
      at Microsoft.CodeAnalysis.GeneratorDriver.GetRunResult()
[...]

PR #74034, opened 2024-06-17
Ensure an empty run result doesn't throw when generators are present.

chsienki:
Fixes https://github.com/dotnet/roslyn/issues/74033

### DIFF
--- a/src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
+++ b/src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
@@ -39,7 +39,7 @@ namespace Microsoft.CodeAnalysis
         internal GeneratorDriver(ParseOptions parseOptions, ImmutableArray<ISourceGenerator> generators, AnalyzerConfigOptionsProvider optionsProvider, ImmutableArray<AdditionalText> additionalTexts, GeneratorDriverOptions driverOptions)
         {
             var incrementalGenerators = GetIncrementalGenerators(generators, SourceExtension);
-            _state = new GeneratorDriverState(parseOptions, optionsProvider, generators, incrementalGenerators, additionalTexts, generators.SelectAsArray(g => GeneratorState.Empty), DriverStateTable.Empty, SyntaxStore.Empty, driverOptions, runtime: TimeSpan.Zero, parseOptionsChanged: true);
+            _state = new GeneratorDriverState(parseOptions, optionsProvider, generators, incrementalGenerators, additionalTexts, generators.SelectAsArray(g => GeneratorState.Uninitialized), DriverStateTable.Empty, SyntaxStore.Empty, driverOptions, runtime: TimeSpan.Zero, parseOptionsChanged: true);
         }
 
         public GeneratorDriver RunGenerators(Compilation compilation, CancellationToken cancellationToken = default)
@@ -91,7 +91,7 @@ namespace Microsoft.CodeAnalysis
                 }
                 else
                 {
-                    states.Add(GeneratorState.Empty);
+                    states.Add(GeneratorState.Uninitialized);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


chsienki:
@dotnet/roslyn-compiler for reviews please :)

src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
```diff
@@ -39,7 +39,7 @@ internal GeneratorDriver(GeneratorDriverState state)
         internal GeneratorDriver(ParseOptions parseOptions, ImmutableArray<ISourceGenerator> generators, AnalyzerConfigOptionsProvider optionsProvider, ImmutableArray<AdditionalText> additionalTexts, GeneratorDriverOptions driverOptions)
         {
             var incrementalGenerators = GetIncrementalGenerators(generators, SourceExtension);
-            _state = new GeneratorDriverState(parseOptions, optionsProvider, generators, incrementalGenerators, additionalTexts, ImmutableArray.Create(new GeneratorState[generators.Length]), DriverStateTable.Empty, SyntaxStore.Empty, driverOptions, runtime: TimeSpan.Zero, parseOptionsChanged: true);
+            _state = new GeneratorDriverState(parseOptions, optionsProvider, generators, incrementalGenerators, additionalTexts, generators.SelectAsArray(g => GeneratorState.Empty), DriverStateTable.Empty, SyntaxStore.Empty, driverOptions, runtime: TimeSpan.Zero, parseOptionsChanged: true);
```
RikkiGibson, line 42: Could you please explain how this change prevents an NRE which was occurring in the original code?

sharwell: I'm guessing it's in `getGeneratorSources` access to `generatorState.PostInitTrees.Length`. The value for `PostInitTrees` is a default immutable array for `default(GeneratorState)`, as opposed to an empty immutable array for `GeneratorState.Empty`.

chsienki: When calling `.GetRunResult()` we call into a local function `getGeneratorSources` that creates an array to handle the generated sources of the given generator

https://github.com/dotnet/roslyn/blob/b8320e1f710db367d33288ed5fee0fb5ee07dd88/src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs#L174

That calls into the collections of the individual states to get their sizes. Prior to the change, we created `default` `GeneratorState`s, so the `PostInitTrees` and `GeneratedTrees` were uninitialized immutable arrays which throw when attempting to get their length.

Instead we now store `GeneratorState.Empty` instances, which in turn have _empty_ result arrays rather than uninitialized ones. 

We had a test that checked you could call `GetRunResult` before calling `RunGenerators` but that test didn't add any generators, so the `generatorStates` array was empty. Adding a generator ensures that array has content, and we see the issue in the added test.

333fred:
So, the change itself looks fine to me, but there are a lot of failures in CI here. Is something not merging correctly?

chsienki:
Test failures seem weird and unrelated, kicked off a re-run and will investigate further if we still see it.

333fred:
> Test failures seem weird and unrelated, kicked off a re-run and will investigate further if we still see it.

Seems like some investigation may be required.

jaredpar:
> src\Compilers\Core\Portable\CommandLine\AnalyzerConfig.cs(30,38): error CS8795: (NETCORE_ENGINEERING_TELEMETRY=Build) Partial method 'AnalyzerConfig.GetSectionMatcherRegex()' must have an implementation part because it has accessibility modifiers.

All the bootstrap legs broke with this message. Guessing something fairly fundamental about generators broke with this change.

sharwell:
Yeah, `GeneratorState.Initialized` is false for `default(GeneratorState)` but true for `GeneratorState.Empty`.

chsienki:
Turns out a bunch of generator tests were failing; they were just hiding amongst the mass of other issues that I didn't see them.

@dotnet/roslyn-compiler for reviews please :)

src/Compilers/Core/Portable/SourceGeneration/GeneratorState.cs
```diff
@@ -25,6 +28,11 @@ namespace Microsoft.CodeAnalysis
                                                                          exception: null,
                                                                          elapsedTime: TimeSpan.Zero);
 
+        /// <summary>
+        /// A generator state for a generator that has not yet been initialized
+        /// </summary>
+        public static readonly GeneratorState Uninitialized = Empty with { Initialized = false };
```
333fred, line 34: Honestly, not to nitpick even more, but this makes me think that the right answer is just to be resilient to `default(ImmutableArray)`, not to introduce a separate uninitialized state that is separate from `default`.

chsienki:
@dotnet/roslyn-compiler for review please :)

jaredpar:
@333fred, @RikkiGibson PTAL

333fred (approved):

jaredpar (approved):

RikkiGibson (approved):

PR #74036, opened 2024-06-17
Add LanguageServer.Protocol.Test.Utilities.csproj

tmat:
As with other Test.Utilities projects, this is a project to define test helpers/mocks for that layer.

### DIFF
--- a/Roslyn.sln
+++ b/Roslyn.sln
@@ -569,6 +569,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ExternalAccess", "ExternalA
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ExternalAccess", "ExternalAccess", "{806F0C6F-3640-4C92-8D55-6767B1535467}"
 EndProject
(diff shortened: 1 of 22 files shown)


tmat:
@dibarbet PTAL

BTW, seems like the LSP layer is being tested only with the editor, which means we only test on Windows :(

dibarbet:
> @dibarbet PTAL
> 
> BTW, seems like the LSP layer is being tested only with the editor, which means we only test on Windows :(

Yup.  There shouldn't be anything fundamental about the LSP tests that require it.  The last time I looked the main issue was that the TestWorkspace was not compatible with non-windows platforms (some WPF concept in it).

dibarbet:
> > @dibarbet PTAL
> > BTW, seems like the LSP layer is being tested only with the editor, which means we only test on Windows :(
> 
> Yup. There shouldn't be anything fundamental about the LSP tests that require it. The last time I looked the main issue was that the TestWorkspace was not compatible with non-windows platforms (some WPF concept in it).

Although there were possibly some tests that required the editorfeatures composition (for testing custom VS LSP stuff).  But those tests are expected to only run on windows.

Issue #66734, opened 2023-02-07
CSharpMoveStaticMembersRefactoringProvider


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/CSharpMoveStaticMembersRefactoringProvid/10258818)._

---
[severity:Other]
Using the Intellisense refactoring option "Move static members to another type" on a static method inside a static (helper class).
The refactoring dialog opens and I choose an existing static class.  The method I choose has one private dependency so I choose that as well.

After clicking OK the refactoring provider crashes. The "Show Stack Trace" button provides the error below.

System.InvalidOperationException : Sequence contains no elements
   at System.Linq.Enumerable.First[TSource](IEnumerable\`1 source)
   at System.Linq.ImmutableArrayExtensions.First[T](ImmutableArray\`1 immutableArray,Func`2 predicate)
   at async Microsoft.CodeAnalysis.MoveStaticMembers.MoveStaticMembersWithDialogCodeAction.RefactorAndMoveAsync()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.MoveStaticMembers.MoveStaticMembersWithDialogCodeAction.ComputeOperationsAsync()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeActionWithOptions.GetOperationsAsync()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedAction.InvokeWorkerAsync()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Extensions.IExtensionManagerExtensions.PerformActionAsync()

Example code:

```
    public static class WorkflowTypes
    {
        public const string FirstType = "firstType";
    }

public static class WorkflowValidations
    {
        private static readonly System.Collections.Generic.List validWorkflowTypes = new System.Collections.Generic.List()
        {
            "firstType"
        };

//  Moving this method and above dependency into WorkflowTypes static class 
        public static bool IsValidWorkflowType(this string workflowType) => validWorkflowTypes.Contains(workflowType);
    }
```

---
#### Feedback Bot on 1/23/2023, 09:40 PM: 

(private comment, text removed)

---
[...]

PR #74037, opened 2024-06-17
Fix move static members when destination is in the same file

dibarbet:
Resolves https://github.com/dotnet/roslyn/issues/66734


LondonRain:
Thanks a lot for taking a look at this! ♥

Will this also work for moving between nested classes in the same file?

dibarbet:
> Thanks a lot for taking a look at this! ♥
> 
> Will this also work for moving between nested classes in the same file?

Looks like it kind of works - there seem to be some issues with the formatting after it gets applied.  Filed https://github.com/dotnet/roslyn/issues/74054 to track.

Cosifne (approved):

PR #74044, opened 2024-06-18
Avoid span conversion in delegate extension receiver

jjonescz:
> An extension method `Cᵢ.Mₑ` is ***eligible*** if:
> 
> - `Cᵢ` is a non-generic, non-nested class
> - The name of `Mₑ` is *identifier*
> - `Mₑ` is accessible and applicable when applied to the arguments as a static method as shown above
> - An implicit identity, reference ~~or boxing~~ **, boxing, or span** conversion exists from *expr* to the type of the first parameter of `Mₑ`.
>   👉 **Span conversion is not considered when overload resolution is performed for a method group conversion.**

Speclet updates: https://github.com/dotnet/csharplang/pull/8221, https://github.com/dotnet/csharplang/pull/8287


jjonescz:
@cston @333fred for reviews, thanks.

cston:
It looks like there is still an open question in the speclet about whether to take the breaking change or not. Was that discussed in LDM?

jjonescz:
> It looks like there is still an open question in the speclet about whether to take the breaking change or not. Was that discussed in LDM?

Not yet, I'm optimistically assuming we won't take a breaking change here. But if we will, we can always revert this. Or we can postpone this PR until after LDM if you prefer.

cston:
> Or we can postpone this PR until after LDM if you prefer.

Is this change blocking other work? If not, I think my preference is to postpone this PR until after an LDM decision.

jjonescz:
@cston @333fred this is ready for reviews again, thanks

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
```diff
@@ -1891,16 +1891,16 @@ private static bool HasIdentityConversionToAny(NamedTypeSymbol type, ArrayBuilde
...
             Debug.Assert((object)thisType != null);
-            var conversion = this.ClassifyImplicitExtensionMethodThisArgConversion(sourceExpressionOpt: null, thisType, parameterType, ref useSiteInfo);
+            var conversion = this.ClassifyImplicitExtensionMethodThisArgConversion(sourceExpressionOpt: null, thisType, parameterType, ref useSiteInfo, isMethodGroupConversion);
             return IsValidExtensionMethodThisArgConversion(conversion) ? conversion : Conversion.NoConversion;
         }
 
         // Spec 7.6.5.2: "An extension method ... is eligible if ... [an] implicit identity, reference,
         // or boxing conversion exists from expr to the type of the first parameter"
-        public Conversion ClassifyImplicitExtensionMethodThisArgConversion(BoundExpression sourceExpressionOpt, TypeSymbol sourceType, TypeSymbol destination, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+        public Conversion ClassifyImplicitExtensionMethodThisArgConversion(BoundExpression sourceExpressionOpt, TypeSymbol sourceType, TypeSymbol destination, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo, bool isMethodGroupConversion)
```
333fred, line 1903: Should we update the comment above this method?

333fred (approved):
LGTM (commit 2)

jjonescz:

cston (approved):

Issue #73923, opened 2024-06-10
Keep  ref struct interfaces behind preview flag


jaredpar:
Language design is leaning towards keeping `ref struct` interfaces for .NET 9. There aren't a lot of real world cases and that is leading to concerns on whether we've battle tested the design enough. Leaving in preview would give us more time to evaluate against real world cases and ensure the design is correct.

This issue is a place holder to ensure we don't attach it to the new language version when it's defined in the 17.12 milestone.

hamarb123:
I have some code like this in some scenarios that require very low overhead to be useful:
```csharp
interface IAction
{
    void Run(/*extra parameters here for some variants of it*/);
}
```
And I was hoping to update it to allow ref struct IActions so that I didn't have to do the pointer trick in these cases. (Note that I pass the generic `TAction` by reference to allow state changes to it when it runs)

I will post more use cases when I have a chance to go through more of the code where I ran into similar issues as I find or think of them.

AlekseyTs:
@hamarb123 Note, the plan is to keep ability for ref structs to implement interface behind a preview flag, the `allows ref struct` constraint on type parameters is not affected and will be fully supported.

jaredpar:
Related https://github.com/dotnet/runtime/issues/103270

tannergooding:
While `ref structs implementing interfaces` isn't something terribly important to me for .NET 9/C# 13, there is the slightly related feature of being able to mark interface members with `[UnscopedRef]` that is important and makes a difference in how the .NET Libraries team will ship `ITensor<T>`

We want to expose `ROSpan<nint> Lengths { get; }` which for structs (regular -or- ref struct) requires it be annotated as `[UnscopedRef]` so that you can return an `InlineArray` or equivalent to allow allocation free tracking of the multidimensional lengths buffer.

PR #74062, opened 2024-06-19
Use separate feature IDs for ref struct interfaces and allows ref struct constraint.

AlekseyTs:
Related to #73923.


jcouv (approved):
LGTM Thanks (iteration 1)

333fred (approved):

PR #74068, opened 2024-06-20
Add some missing LSP protocol definitions

dibarbet:
Used by xaml

### DIFF
--- a/src/LanguageServer/Protocol/Protocol/Internal/VSInternalDocumentOnAutoInsertRegistrationOptions.cs
+++ b/src/LanguageServer/Protocol/Protocol/Internal/VSInternalDocumentOnAutoInsertRegistrationOptions.cs
@@ -1,0 +1,34 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Text.Json.Serialization;
+
+namespace Roslyn.LanguageServer.Protocol;
+
+/// <summary>
+/// Class representing the registration options for on auto insert support.
+/// </summary>
+internal class VSInternalDocumentOnAutoInsertRegistrationOptions : VSInternalDocumentOnAutoInsertOptions, ITextDocumentRegistrationOptions, IStaticRegistrationOptions
+{
+    /// <summary>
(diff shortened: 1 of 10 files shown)


src/LanguageServer/Protocol/Protocol/Internal/VSInternalDocumentOnAutoInsertRegistrationOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Text.Json.Serialization;
+
+namespace Roslyn.LanguageServer.Protocol;
+
+/// <summary>
+/// Class representing the registration options for on auto insert support.
+/// </summary>
+internal class VSInternalDocumentOnAutoInsertRegistrationOptions : VSInternalDocumentOnAutoInsertOptions, ITextDocumentRegistrationOptions, IStaticRegistrationOptions
```
mgoertz-msft, line 12: I noticed that `VSInternalDocumentOnAutoInsertOptions` doesn't use `JsonPropertyName` but still `DataMember`. We need to update all of those as well so they serialize correctly.

dibarbet: @mgoertz-msft hmm, it looks correct to me?
https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/Protocol/Protocol/Internal/VSInternalDocumentOnAutoInsertOptions.cs

mgoertz-msft: Hah, I see what happened. I made sure I was looking at main but I ended up in your fork :)

https://github.com/dibarbet/roslyn/blob/main/src/Features/LanguageServer/Protocol/Protocol/Internal/VSInternalDocumentOnAutoInsertOptions.cs

mgoertz-msft: That's what I get for using the go to definition feature in a PR :)

mgoertz-msft (approved):

JoeRobich (approved):

PR #74069, opened 2024-06-20
[LSP] Support "implementation" folding range kinds

davidwengier:
Fixes https://github.com/dotnet/razor/issues/9492 along with https://github.com/dotnet/razor/pull/10509

### DIFF
--- a/src/LanguageServer/Protocol/Handler/FoldingRanges/FoldingRangesHandler.cs
+++ b/src/LanguageServer/Protocol/Handler/FoldingRanges/FoldingRangesHandler.cs
@@ -110,6 +110,7 @@ namespace Microsoft.CodeAnalysis.LanguageServer.Handler
                     BlockTypes.Comment => FoldingRangeKind.Comment,
                     BlockTypes.Imports => FoldingRangeKind.Imports,
                     BlockTypes.PreprocessorRegion => FoldingRangeKind.Region,
+                    BlockTypes.Member => VSFoldingRangeKind.Implementation,
                     _ => null,
                 };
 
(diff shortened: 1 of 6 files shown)


src/LanguageServer/Protocol/Handler/FoldingRanges/FoldingRangesHandler.cs
```diff
@@ -110,6 +110,7 @@ private static FoldingRange[] GetFoldingRanges(BlockStructure blockStructure, So
                     BlockTypes.Comment => FoldingRangeKind.Comment,
                     BlockTypes.Imports => FoldingRangeKind.Imports,
                     BlockTypes.PreprocessorRegion => FoldingRangeKind.Region,
+                    BlockTypes.Member => VSFoldingRangeKind.Implementation,
```
davidwengier, line 113: In my blind optimism that this will one day be added to the spec, and because "kind" is just a string, I didn't gate this on VS extended capabilities, but let me know if you don't agree @dibarbet

dibarbet: Can you check and make sure the VSCode behavior doesn't change if we give them null vs. a kind they don't know about?  Checking capabilities would be safer since that guarantees they won't ever barf on invalid kinds

davidwengier: VS Code looks perfectly normal. There is even a note in the spec:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange
> When [the foldingRangeKind] property exists the client also guarantees that it will handle values outside its set gracefully and falls back to a default value when unknown.

davidwengier: But it's up to you, I won't be offended if you want me to make this more strict. It's your server :)

dibarbet: ok, if its spec'd like that then fine with me.

dibarbet (approved):

PR #74084, opened 2024-06-20
Create new AsyncServiceProvider instead of using AsyncServiceProvider.GlobalProvider

sharwell:
`AsyncServiceProvider.GlobalProvider` is no longer an instance of `AsyncServiceProvider`, so the original code is failing.

### DIFF
--- a/src/VisualStudio/IntegrationTest/New.IntegrationTests/InProcess/EditorInProcess.cs
+++ b/src/VisualStudio/IntegrationTest/New.IntegrationTests/InProcess/EditorInProcess.cs
@@ -50,6 +50,7 @@ using Roslyn.VisualStudio.IntegrationTests.InProcess;
 using Roslyn.VisualStudio.NewIntegrationTests.InProcess;
 using WindowsInput.Native;
 using Xunit;
+using COMAsyncServiceProvider = Microsoft.VisualStudio.Shell.Interop.COMAsyncServiceProvider;
 using IComponentModel = Microsoft.VisualStudio.ComponentModelHost.IComponentModel;
 using IObjectWithSite = Microsoft.VisualStudio.OLE.Interop.IObjectWithSite;
 using IOleServiceProvider = Microsoft.VisualStudio.OLE.Interop.IServiceProvider;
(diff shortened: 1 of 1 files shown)


src/VisualStudio/IntegrationTest/New.IntegrationTests/InProcess/EditorInProcess.cs
```diff
@@ -455,7 +456,7 @@ public async Task<ImmutableArray<string>> GetF1KeywordsAsync(CancellationToken c
         ErrorHandler.ThrowOnFailure(vsView.GetBuffer(out var textLines));
         ErrorHandler.ThrowOnFailure(textLines.GetLanguageServiceID(out var languageServiceGuid));
 
-        var languageService = await ((AsyncServiceProvider)AsyncServiceProvider.GlobalProvider).QueryServiceAsync(languageServiceGuid).WithCancellation(cancellationToken);
+        var languageService = await new AsyncServiceProvider((COMAsyncServiceProvider.IAsyncServiceProvider)AsyncServiceProvider.GlobalProvider).QueryServiceAsync(languageServiceGuid).WithCancellation(cancellationToken);
```
AArnott, line 459: As this is a fix for a downcast that wasn't documented to succeed, maybe your fix should avoid doing the same thing again. I believe the only supported way to get the COM IAsyncServiceProvider is by querying for it.

sharwell: Querying for it is easy enough. Do you know which service interface matches it?

AArnott: There isn't a different service interface for it. So it would look like this:

```cs
AsyncServiceProvider.GlobalProvider.GetService<COMAsyncServiceProvider.IAsyncServiceProvider>()
```

You'll have to test it though. This is all based on my recollection.

sharwell: ➡️ I'll submit it as a separate follow-up PR since the current one already passed integration tests.

genlu (approved):

src/VisualStudio/IntegrationTest/New.IntegrationTests/InProcess/EditorInProcess.cs
davkean: Yeah that will "work" as the managed version implements the native interface, but the right change is to query it as suggested by Andrew.

sharwell: ➡️ Submitted #74100

PR #74088, opened 2024-06-20
On the fly docs - add telemetry regarding documentation comments

akhera99:
Adds some telemetry to see if documentation comments impact the number of users clicking "tell me more" link.

### DIFF
--- a/src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs
+++ b/src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs
@@ -212,11 +212,6 @@ internal sealed partial class OnTheFlyDocsView : UserControl, INotifyPropertyCha
         OnTheFlyDocsLogger.LogOnTheFlyDocsResultsRequested();
         if (_onTheFlyDocsElement.HasComments)
         {
-            Logger.Log(FunctionId.CodeAnalysisService_CalculateDiagnosticsAsync, KeyValueLogMessage.Create(m =>
-            {
-                m["SymbolHeaderText"] = _onTheFlyDocsElement.SymbolSignature;
-            }, LogLevel.Information));
-
             OnTheFlyDocsLogger.LogOnTheFlyDocsResultsRequestedWithDocComments();
         }
 
--- a/src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs
+++ b/src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs
@@ -208,6 +208,17 @@ internal sealed partial class OnTheFlyDocsView : UserControl, INotifyPropertyCha
             m["SymbolHeaderText"] = _onTheFlyDocsElement.SymbolSignature;
         }, LogLevel.Information));
 
+        OnTheFlyDocsLogger.LogOnTheFlyDocsResultsRequested();
+        if (_onTheFlyDocsElement.HasComments)
+        {
+            Logger.Log(FunctionId.CodeAnalysisService_CalculateDiagnosticsAsync, KeyValueLogMessage.Create(m =>
+            {
+                m["SymbolHeaderText"] = editorFeaturesOnTheFlyDocsElement.OnTheFlyDocsElement.SymbolSignature;
+            }, LogLevel.Information));
+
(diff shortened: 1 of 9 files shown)


src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsViewFactory.cs
```diff
@@ -60,6 +60,8 @@ public OnTheFlyDocsViewFactory(IViewElementFactoryService factoryService, IAsync
             throw new InvalidOperationException("QuickInfoSession is null");
         }
 
+        OnTheFlyDocsLogger.LogShowedOnTheFlyDocsLink();
```
genlu, line 63: Could we log whether a symbol has doc-comment here? That would provide us the numbers regardless of whether user clicked the link or not, right?

src/Features/Core/Portable/QuickInfo/OnTheFlyDocsElement.cs
```diff
@@ -12,9 +12,13 @@ namespace Microsoft.CodeAnalysis.QuickInfo;
 /// <param name="symbolSignature">formatted string representation of a symbol/></param>
 /// <param name="declarationCode">the symbol's declaration code</param>
 /// <param name="language">the language of the symbol</param>
-internal sealed class OnTheFlyDocsElement(string symbolSignature, ImmutableArray<string> declarationCode, string language)
+/// <param name="hasComments">whether the symbol has comments</param>
+internal sealed class OnTheFlyDocsElement(string symbolSignature, ImmutableArray<string> declarationCode, string language, bool hasComments)
```
genlu, line 16: Maybe also differentiate source symbols and metadata symbols?

src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs
```diff
@@ -206,8 +206,20 @@ public void RequestResults()
...
+            m["HasDocumentationComments"] = _onTheFlyDocsElement.HasComments;
         }, LogLevel.Information));
 
+        OnTheFlyDocsLogger.LogOnTheFlyDocsResultsRequested();
+        if (_onTheFlyDocsElement.HasComments)
+        {
+            Logger.Log(FunctionId.CodeAnalysisService_CalculateDiagnosticsAsync, KeyValueLogMessage.Create(m =>
+            {
+                m["SymbolHeaderText"] = _onTheFlyDocsElement.SymbolSignature;
+            }, LogLevel.Information));
```
genlu, line 218: what's this log message for?

akhera99: Oh I meant to change that

src/Features/CSharp/Portable/QuickInfo/CSharpSemanticQuickInfoProvider.cs
```diff
@@ -181,6 +183,30 @@ protected override NullableFlowState GetNullabilityAnalysis(SemanticModel semant
             return sourceText.GetSubText(new Text.TextSpan(span.Start, Math.Min(maxLength, span.Length))).ToString();
         }).ToImmutableArray();
 
-        return new OnTheFlyDocsElement(symbol.ToDisplayString(), symbolStrings, symbol.Language);
+        return new OnTheFlyDocsElement(symbol.ToDisplayString(), symbolStrings, symbol.Language, hasDocumentationComment);
+
+        bool HasDocumentationComment(ImmutableArray<SyntaxReference> references)
```
genlu, line 188: when creating a tooltip in QuickInfoUtilities.CreateQuickInfoItemAsync, it looks like we know whether if contains  doc-comment, is it possible log there instead so we don't have to check the syntax?

src/Features/CSharp/Portable/QuickInfo/CSharpSemanticQuickInfoProvider.cs
```diff
@@ -168,13 +168,20 @@ protected override NullableFlowState GetNullabilityAnalysis(SemanticModel semant
             return null;
         }
 
+        if (symbol.MetadataToken != 0)
```
genlu, line 171: so this does what we want? Great!

genlu (approved):

PR #74089, opened 2024-06-20
Fix error recovery for various nested conversion scenarios

RikkiGibson:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2088611

- We had various places where we visited the underlying conversions of a conversion, but did not call MarkUnderlyingConversionsChecked. This caused assert failures in various nested conversion scenarios (exercised in new tests). It seems like this method is just a sanity check to ensure we actually traversed the whole conversion "tree" and should be called whenever we are applying the nested conversions to the corresponding subexpressions. See pre-existing call sites of the method for comparison.
- We were not calling BindToNaturalType for the operand of a user-defined conversion which was not valid e.g. due to ambiguity. This results in unexpected unconverted target-typed expressions in the bound tree in later phases.
- Increased uniformity of NullableWalker assertions for switch expressions versus conditional operators. This fixed assertion failures in some of the new tests.
- NullableWalker would report nullability warnings for user-defined conversion arguments, when the conversion was not valid in initial binding anyway. Note that a user-defined conversion may *exist* but still not be *valid*, when it is ambiguous which user-defined conversion method should be used. Entering this code path in such cases was causing an assertion to fail and is not needed.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
@@ -683,7 +683,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                 }
 
                 var elementConversions = conversion.UnderlyingConversions;
-                conversion.MarkUnderlyingConversionsChecked();
 
                 Debug.Assert(elementType is { });
                 Debug.Assert(elements.Length == elementConversions.Length);
@@ -709,6 +708,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                             diagnostics);
                     builder.Add(convertedElement!);
                 }
+                conversion.MarkUnderlyingConversionsChecked();
             }
 
             return new BoundCollectionExpression(
@@ -1645,8 +1645,8 @@ namespace Microsoft.CodeAnalysis.CSharp
         {
             bool targetTyped = conversionIfTargetTyped is { };
             Debug.Assert(targetTyped || destination.IsErrorType() || destination.Equals(source.Type, TypeCompareKind.ConsiderEverything));
-            ImmutableArray<Conversion> underlyingConversions = conversionIfTargetTyped.GetValueOrDefault().UnderlyingConversions;
-            conversionIfTargetTyped.GetValueOrDefault().MarkUnderlyingConversionsChecked();
+            var conversion = conversionIfTargetTyped.GetValueOrDefault();
+            ImmutableArray<Conversion> underlyingConversions = conversion.UnderlyingConversions;
             var condition = source.Condition;
             hasErrors |= source.HasErrors || destination.IsErrorType();
 
@@ -1658,6 +1658,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 targetTyped
                 ? CreateConversion(source.Alternative.Syntax, source.Alternative, underlyingConversions[1], isCast: false, conversionGroupOpt: null, destination, diagnostics)
                 : GenerateConversionForAssignment(destination, source.Alternative, diagnostics);
+            conversion.MarkUnderlyingConversionsChecked();
             var constantValue = FoldConditionalOperator(condition, trueExpr, falseExpr);
             hasErrors |= constantValue?.IsBad == true;
             if (targetTyped && !destination.IsErrorType() && !Compilation.IsFeatureEnabled(MessageID.IDS_FeatureTargetTypedConditional))
@@ -1684,7 +1685,6 @@ namespace Microsoft.CodeAnalysis.CSharp
             Conversion conversion = conversionIfTargetTyped ?? Conversion.Identity;
             Debug.Assert(targetTyped || destination.IsErrorType() || destination.Equals(source.Type, TypeCompareKind.ConsiderEverything));
             ImmutableArray<Conversion> underlyingConversions = conversion.UnderlyingConversions;
-            conversion.MarkUnderlyingConversionsChecked();
             var builder = ArrayBuilder<BoundSwitchExpressionArm>.GetInstance(source.SwitchArms.Length);
             for (int i = 0, n = source.SwitchArms.Length; i < n; i++)
             {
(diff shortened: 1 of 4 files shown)


RikkiGibson:
@dotnet/roslyn-compiler for review

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -683,6 +683,7 @@ private static void CheckInlineArrayTypeIsSupported(SyntaxNode syntax, TypeSymbo
                 }
 
                 var elementConversions = conversion.UnderlyingConversions;
+                conversion.MarkUnderlyingConversionsChecked();
```
AlekseyTs, line 686: >conversion.MarkUnderlyingConversionsChecked(); [](http://example.com/codeflow?start=16&length=46)

Consider moving this after the work of checking the underlying conversions is done. I assume that is after the loop below.

AlekseyTs: BTW, it is not clear to me what is this operation going to change. It look like we are mutating a local value of a struct, which is not returned or stored anywhere.

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -1645,6 +1646,7 @@ internal bool TryGetCollectionIterationType(SyntaxNode syntax, TypeSymbol collec
             bool targetTyped = conversionIfTargetTyped is { };
             Debug.Assert(targetTyped || destination.IsErrorType() || destination.Equals(source.Type, TypeCompareKind.ConsiderEverything));
             ImmutableArray<Conversion> underlyingConversions = conversionIfTargetTyped.GetValueOrDefault().UnderlyingConversions;
+            conversionIfTargetTyped.GetValueOrDefault().MarkUnderlyingConversionsChecked();
```
AlekseyTs, line 1649: >conversionIfTargetTyped.GetValueOrDefault().MarkUnderlyingConversionsChecked(); [](http://example.com/codeflow?start=12&length=79)

It is not clear to me how mutating a local struct value going to help.

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
AlekseyTs: > it is not clear to me what is this operation going to change

I guess this operation is mutating a class instance referenced by the struct, therefore, the mutation is preserved even when the struct value is not preserved.
Is this change backed by a test?

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
AlekseyTs, line 1649: >conversionIfTargetTyped.GetValueOrDefault() [](http://example.com/codeflow?start=12&length=43)

Consider extracting a local and reusing.

AlekseyTs, line 1649: >MarkUnderlyingConversionsChecked [](http://example.com/codeflow?start=56&length=32)

Consider doing this after the conversions are actually checked.

AlekseyTs: Is this change backed by a test?

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -1682,6 +1684,7 @@ private BoundExpression ConvertSwitchExpression(BoundUnconvertedSwitchExpression
             Conversion conversion = conversionIfTargetTyped ?? Conversion.Identity;
             Debug.Assert(targetTyped || destination.IsErrorType() || destination.Equals(source.Type, TypeCompareKind.ConsiderEverything));
             ImmutableArray<Conversion> underlyingConversions = conversion.UnderlyingConversions;
+            conversion.MarkUnderlyingConversionsChecked();
```
AlekseyTs, line 1687: >MarkUnderlyingConversionsChecked [](http://example.com/codeflow?start=23&length=32)

Consider doing this after the conversions are actually checked.

AlekseyTs:

RikkiGibson:
It looks like some test runs are timing out. I would be surprised if just the changes in c39c9183ecae009da86cd1c2387658e2020d2ebc were causing this. Following up with infraswat.

src/Compilers/CSharp/Test/Semantic/Semantics/UserDefinedConversionTests.cs
```diff
@@ -1932,5 +1934,361 @@ .locals init (S? V_0)
...
+                    public static implicit operator C((string, string) pair)
+                    {
+                        Console.WriteLine("string");
+                        return new C();
+                    }
+                }
+                """;
+
+            var comp = CreateCompilation(source);
+            comp.VerifyEmitDiagnostics();
```
RikkiGibson, line 2020: I meant to verify output of this one.

AlekseyTs (approved):
LGTM (commit 3)

RikkiGibson:

RikkiGibson:

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -1723,7 +1727,7 @@ private BoundExpression ConvertSwitchExpression(BoundUnconvertedSwitchExpression
 
                 return new BoundConversion(
                     syntax,
-                    source,
+                    BindToNaturalType(source, diagnostics),
```
333fred, line 1730: Should we be calling `BindToTypeForErrorRecovery` instead? It seems like the main difference is around error reporting; any errors from `BindToNaturalType` will be put into `diagnostics`, while `BindToTypeForErrorRecovery` will discard them. Do we already have a diagnostic at this point?

RikkiGibson: We will definitely have *some* diagnostic about this expression, because of the preceding `if (!hasErrors) GenerateImplicitConversionError(...)`.

re: the difference in practice. If we used BindToTypeForErrorRecovery here, then test `Repro_VsFeedback_2088611` wouldn't report `ERR_ImplicitObjectCreationNoTargetType`. It would only report `ERR_ConversionNotTupleCompatible`. I feel like in this particular case, it's maybe better to have a squiggle on the new() because it happens to be related to the ambiguity. But, ideally, we would be able to report some kind of user-defined operator overload resolution error here, I think.

333fred (approved):

PR #74091, opened 2024-06-20
Handle unspecified anonymous type property in VisualBasicOperationFactory

cston:
Handle unspecified anonymous type property in `VisualBasicOperationFactory`.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2063559

### DIFF
--- a/src/Compilers/VisualBasic/Portable/Operations/VisualBasicOperationFactory.vb
+++ b/src/Compilers/VisualBasic/Portable/Operations/VisualBasicOperationFactory.vb
@@ -1602,12 +1602,12 @@ Namespace Microsoft.CodeAnalysis.Operations
             Return New AnonymousObjectCreationOperation(initializers, _semanticModel, syntax, type, isImplicit)
         End Function
 
-        Private Function CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess As BoundAnonymousTypePropertyAccess) As IOperation
+        Private Function CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess As BoundAnonymousTypePropertyAccess) As IPropertyReferenceOperation
             Dim [property] As IPropertySymbol = DirectCast(boundAnonymousTypePropertyAccess.ExpressionSymbol, IPropertySymbol)
             Return CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess, [property])
         End Function
 
-        Private Function CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess As BoundAnonymousTypePropertyAccess, [property] As IPropertySymbol) As IOperation
+        Private Function CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess As BoundAnonymousTypePropertyAccess, [property] As IPropertySymbol) As IPropertyReferenceOperation
             Dim instance As IOperation = CreateAnonymousTypePropertyAccessImplicitReceiverOperation([property], boundAnonymousTypePropertyAccess.Syntax.FirstAncestorOrSelf(Of AnonymousObjectCreationExpressionSyntax))
             Dim arguments = ImmutableArray(Of IArgumentOperation).Empty
             Dim syntax As SyntaxNode = boundAnonymousTypePropertyAccess.Syntax
--- a/src/Compilers/VisualBasic/Portable/Operations/VisualBasicOperationFactory.vb
+++ b/src/Compilers/VisualBasic/Portable/Operations/VisualBasicOperationFactory.vb
@@ -1602,8 +1602,12 @@ Namespace Microsoft.CodeAnalysis.Operations
             Return New AnonymousObjectCreationOperation(initializers, _semanticModel, syntax, type, isImplicit)
         End Function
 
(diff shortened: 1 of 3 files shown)


333fred (approved):

src/Compilers/VisualBasic/Test/IOperation/IOperation/IOperationTests_IAnonymousObjectCreationOperation.vb
```diff
@@ -1821,5 +1821,189 @@ Block[B3] - Exit
...
+        Dim item = New C() With {.}
+                                  ~
+BC30203: Identifier expected.
+        Dim item = New C() With {.}
+                                  ~
+BC30984: '=' expected (object initializer).
+        Dim item = New C() With {.}
+                                  ~
+]]>.Value
+            Dim expectedFlowGraph = <![CDATA[
```
AlekseyTs, line 1847: >expectedFlowGraph [](http://example.com/codeflow?start=16&length=17)

Please assert original IOperation tree for all added tests

src/Compilers/VisualBasic/Portable/Operations/VisualBasicOperationFactory.vb
```diff
@@ -1602,8 +1602,12 @@ Namespace Microsoft.CodeAnalysis.Operations
             Return New AnonymousObjectCreationOperation(initializers, _semanticModel, syntax, type, isImplicit)
         End Function
 
-        Private Function CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess As BoundAnonymousTypePropertyAccess) As IPropertyReferenceOperation
+        Private Function CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess As BoundAnonymousTypePropertyAccess) As IOperation
             Dim [property] As IPropertySymbol = DirectCast(boundAnonymousTypePropertyAccess.ExpressionSymbol, IPropertySymbol)
+            Return CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess, [property])
```
AlekseyTs, line 1607: >Return CreateBoundAnonymousTypePropertyAccessOperation(boundAnonymousTypePropertyAccess, [property]) [](http://example.com/codeflow?start=12&length=100)

Alternatively, consider creating ```InvalidOperation``` when ```[property] Is Nothing``` and reverting changes in ```GetAnonymousTypeCreationInitializers```.

AlekseyTs:

src/Compilers/VisualBasic/Portable/Operations/VisualBasicOperationFactory.vb
cston: I tried that initially. However, `TestOperationVisitor.VisitAnonymousObjectCreation()` expects the initializer target to be an `IPropertyReferenceOperation`, and I assumed there might be consumers other than test code that also expect the same.

AlekseyTs: I think we should simply change the assertion in ```TestOperationVisitor.VisitAnonymousObjectCreation```, what it is doing is mostly driven by implementation. When we are fixing bugs, we are changing shape of ```IOperation``` tree even for success scenarios, not to mention error scenarios. The binder says, it doesn't know what the property is (that is what the bound node returns) and I think we should stick with that rather than trying to re-guess what that property might be. Otherwise, we probably should fix what the bound node returns, so that ```SemanticModel``` and ```IOperation``` tree produce consistent result.

src/Compilers/VisualBasic/Test/IOperation/IOperation/IOperationTests_IAnonymousObjectCreationOperation.vb
```diff
@@ -1821,5 +1821,286 @@ Block[B3] - Exit
...
+          IVariableDeclaratorOperation (Symbol: item As <anonymous type: $0 As ?>) (OperationKind.VariableDeclarator, Type: null) (Syntax: 'item')
+            Initializer:
+              null
+      Initializer:
+        IVariableInitializerOperation (OperationKind.VariableInitializer, Type: null, IsInvalid) (Syntax: '= New With {.}')
+          IAnonymousObjectCreationOperation (OperationKind.AnonymousObjectCreation, Type: <anonymous type: $0 As ?>, IsInvalid) (Syntax: 'New With {.}')
+            Initializers(1):
+                ISimpleAssignmentOperation (OperationKind.SimpleAssignment, Type: ?, IsInvalid) (Syntax: '.')
+                  Left:
+                    IPropertyReferenceOperation: Property <anonymous type: $0 As ?>.$0 As ? (OperationKind.PropertyReference, Type: ?, IsInvalid) (Syntax: '')
```
AlekseyTs, line 1954: >Property <anonymous type: $0 As ?>.$0 As ? [](http://example.com/codeflow?start=49&length=42)

I assume this is not consistent with symbol information returned by ```SemanticModel``` for the node. Let's confirm that and assert what ```SemanticModel``` API returns for the syntax node.

AlekseyTs:

src/Compilers/VisualBasic/Portable/Operations/VisualBasicOperationFactory.vb
cston: Thanks. Based on our offline discussion, I've left this as is, and added a check of the SemanticModel to the tests.

AlekseyTs (approved):
LGTM (commit 4)

PR #74095, opened 2024-06-21
Add LSP notification for starting project initialization

JoeRobich:
This will allow us to show a work indicator in the C# extension when loading a solution or projects.

See C# ext PR: https://github.com/dotnet/vscode-csharp/pull/7254

### DIFF
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/LanguageServerProjectSystem.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/LanguageServerProjectSystem.cs
@@ -98,6 +98,8 @@ internal sealed class LanguageServerProjectSystem
     {
         using (await _gate.DisposableWaitAsync())
         {
+            await ProjectInitializationHandler.SendProjectInitializationStartedNotificationAsync();
+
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/LanguageServerProjectSystem.cs
```diff
@@ -98,6 +98,8 @@ public async Task OpenSolutionAsync(string solutionFilePath)
     {
         using (await _gate.DisposableWaitAsync())
         {
+            await ProjectInitializationHandler.SendProjectInitializationStartedNotificationAsync();
```
dibarbet, line 101: Fine with a notification for now.  But definitely in the future I would like us to move to a generic progress request that is triggered when the project load/reload is triggered - right now the project could be reloading and we won't show a busy indicator.

dibarbet, line 101: Note - this won't get triggered in devkit - is that fine?

JoeRobich:
Closing as unnecessary. Likely will revisit something similar for project reloading.

PR #74102, opened 2024-06-21
Enable some type test scenarios involving ref like type parameters

AlekseyTs:


src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
```diff
@@ -23551,23 +23551,19 @@ public void IsOperator_07()
 class Helper1<T, U>
     where T : allows ref struct
 {
-    static void Test1(U h1)
+    public static void Test1(U h1)
     {
         if (h1 is T)
```
jjonescz, line 23556: Does the spec needs updating? https://github.com/dotnet/csharpstandard/blob/ebf205860866ff7adff0c9876dec8873deadf64c/standard/expressions.md#1212121-the-is-type-operator

> The result of the operation `E is T` [...] is a Boolean value indicating whether `E` is non-null and can successfully be converted to type `T` by a reference conversion, a boxing conversion, an unboxing conversion, a wrapping conversion, or an unwrapping conversion.

This is not one of the listed conversions, right?

AlekseyTs: > This is not one of the listed conversions, right?

It is it not? The check is performed at runtime.

AlekseyTs:
@cston, @jjonescz, @dotnet/roslyn-compiler Please review

AlekseyTs:

AlekseyTs:

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
```diff
@@ -23551,23 +23551,19 @@ public void IsOperator_07()
 class Helper1<T, U>
     where T : allows ref struct
```
cston, line 23552: Are we testing the following where `U` has `allows ref struct`?
```csharp
class Helper1<T, U>
{
    public static void Test1(U h1)
        where U : allows ref struct
    {
        if (h1 is T)
        ...
    }
}

Helper1<I1, S>.Test1(new S());
...
```

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
```diff
@@ -24157,18 +24321,104 @@ interface I1
         [Fact]
         public void IsPattern_07()
         {
-            var src = @"
+            var src1 = @"
 class Helper1<T, U>
```
cston, line 24325: Are we testing the following where `U` has `allows ref struct`?
```csharp
class Helper1<T, U>
    where U : allows ref struct
{
    public static void Test1(U h1)
    {
        if (h1 is T t)
        ...
    }
}

Helper1<I1, S>.Test1(new S());
...
```

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
AlekseyTs: The previous test? The scenario that you suggest (T  has no constraint) is not affected by the change. Doesn't take the affected code path:
```
if (targetType is TypeParameterSymbol { AllowsRefLikeType: true })
```

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
AlekseyTs: Same response

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
cston: > The previous test?

It's not clear to me why `is` check in the previous test is a compile-time error but the `is` check in this test is a runtime check.

> The scenario that you suggest (T  has no constraint) is not affected by the change.

That's true, thanks.

AlekseyTs: > It's not clear to me why `is` check in the previous test is a compile-time error but the `is` check in this test is a runtime check.

The runtime check for this test is safe because we know that the value being checked is never a ref struct. That is not true for the previous test

AlekseyTs: We can box the value being checked in this test

cston: > The scenario that you suggest (T has no constraint) is not affected by the change.

Consider testing if it's not already tested, even though it is not affected by this change.

AlekseyTs: I think it is covered by IsOperator_05

PR #74106, opened 2024-06-21
Make sure task completes on exception path

sharwell:


src/Tools/Source/RunTests/ProcessRunner.cs
```diff
@@ -109,15 +109,31 @@ public static void OpenFile(string file)
...
                         process,
-                        process.ExitCode,
+                        exitCode,
                         new ReadOnlyCollection<string>(outputLines),
                         new ReadOnlyCollection<string>(errorLines));
                     tcs.TrySetResult(result);
+
+                    static async ValueTask<int> GetExitCodeAsync(Process process)
+                    {
+                        await process.WaitForExitAsync();
```
sharwell, line 134: 📝 This pull request fixes the case where `WaitForExit` _throws_, but does not do anything about a case where it never returns at all. Currently it appears more likely that problems are caused by throwing.

jaredpar (approved):

PR #74110, opened 2024-06-21
Handle early attribute required members cycle

333fred:
Fixes https://github.com/dotnet/roslyn/issues/74108. Because this cycle is only in early attribute binding, I've opted for a very simple approach of giving a generic error; it  would be possible to make this error more specific, but it doesn't seem worth the effort to do so.


src/Compilers/CSharp/Portable/Errors/LazyRequiredMembersDiagnosticInfo.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return new LazyRequiredMembersDiagnosticsInfo(this, severity);
+    }
+
+    protected override DiagnosticInfo? ResolveInfo()
+    {
+        var diagnosticsBag = BindingDiagnosticBag.GetInstance(withDiagnostics: true, withDependencies: false);
+        Binder.CheckRequiredMembersInObjectInitializer(_constructor, _initializers, _syntax, diagnosticsBag);
+
+        var diagnostic = diagnosticsBag.HasAnyErrors() ? new CSDiagnosticInfo(ErrorCode.ERR_RequiredMembersAttributeErrors) : null;
```
333fred, line 40: Note: it is not currently possible to hit the true codepath here; for all of our attributes, if there are any potential errors (including unresolved diagnostics, such as this one), we don't report errors from early binding, and instead calculate them again later. I believe that it will be possible to hit this codepath in the overload resolution priority feature, however, so I'm leaving it in and will add an explicit test for that.

src/Compilers/CSharp/Portable/Errors/ErrorCode.cs
```diff
@@ -2121,7 +2121,7 @@ internal enum ErrorCode
         WRN_DuplicateAnalyzerReference = 9067,
         ERR_FileTypeNonUniquePath = 9068,
         ERR_FilePathCannotBeConvertedToUtf8 = 9069,
-        //ERR_ReadOnlyNotSuppAsParamModDidYouMeanIn = 9070,
+        ERR_RequiredMembersAttributeErrors = 9070,
```
jjonescz, line 2124: I think we shouldn't reuse codes from a released version of the compiler. Consider using one of the error codes marked as `// available` below

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Binder/Binder_Attributes.cs
```diff
@@ -263,7 +263,14 @@ private BoundAttribute BindAttributeCore(AttributeSyntax node, NamedTypeSymbol a
 
             if (attributeConstructor is not null)
             {
-                CheckRequiredMembersInObjectInitializer(attributeConstructor, ImmutableArray<BoundExpression>.CastUp(boundNamedArguments), node, diagnostics);
+                if (IsEarlyAttributeBinder)
+                {
+                    diagnostics.Add(new LazyRequiredMembersDiagnosticsInfo(attributeConstructor, ImmutableArray<BoundExpression>.CastUp(boundNamedArguments), node), node.GetLocation());
```
AlekseyTs, line 268: >diagnostics.Add(new LazyRequiredMembersDiagnosticsInfo(attributeConstructor, ImmutableArray<BoundExpression>.CastUp(boundNamedArguments), node), node.GetLocation()); [](http://example.com/codeflow?start=20&length=165)

A comment on ```EarlyWellKnownAttributeBinder``` says: "It only binds those attribute argument syntax which can produce valid attribute arguments, but doesn't report any diagnostics. Subsequent binding phase will rebind such erroneous attributes and generate appropriate diagnostics."

Either the comment is not accurate, or it feels like we don't need to have any code in this branch. Especially given that the interesting part of  ```LazyRequiredMembersDiagnosticsInfo```'s implementation is unreachable.

AlekseyTs: Looking deeper, it looks like what we are doing here will cause ```out bool generatedDiagnostics``` to be true for all ```EarlyWellKnownAttributeBinder.GetAttribute``` calls. That value affects how result of the calls is processed in a fundamental way. It looks like by doing this, we are having an unconditional effect on all those callers without even knowing that we have anything to report.

AlekseyTs: At the moment I see the following alternatives:
1. Do what we do for ```IsAttributeConditionallyOmitted``` helper. I.e. completely skip required members analysis during early attribute binding and then explicitly do it for the early bound attributes later, when we bind remaining attributes in a regular mode.
2. Move the check to ```LoadAndValidateAttributes``` to the section that does some post processing of attributes, including nullable analysis.

I am personally leaning towards the second option. In my opinion, the required members check isn't supposed to invalidate an attribute application. Yes, language wants to report an additional diagnostics for an attribute in some cases, but the attribute application should still be considered good if overload resolution succeeded and other runtime requirements are satisfied.

AlekseyTs:

333fred:
@AlekseyTs addressed feedback.

RikkiGibson (approved):

AlekseyTs (approved):
LGTM (commit 4)

PR #74116, opened 2024-06-23
Do not store temp of ref conditional operator by value

### LINKED ISSUE #74115: Bug: Conditional Operator stores ref result in a non-ref-temporary (Only when SpillSequenceSpiller is used - e.g. for async/await)
**Version Used**: [847d588](https://github.com/dotnet/roslyn/commit/847d58843a139405d633a55c82b232582c996848)

**Steps to Reproduce**:

Compile and execute the following [code](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4AEBMBGAWACgsAGAAizwFYBuYgfXIF5yAxAew4DoA5VMMACmyAEKoA5hICeACgCU3AEpCAzqgA2CeiTwBOWQCIAtKbOnD8nU1ace/QSN4cE4qXKvFiWAMyUc7FwA3sTkYWG+lHgAbJQAHJTRADywCAB85A7CYpIyCqHh5CFEhYVUFKoIyCzkhhA4hjql4ankAB41eE3N5K3SNTg6Bc3IQgBmvTAI5NXMwz2FoxNYeuQAogBuEBoAgqrSMADGspXIcOR48r2qtRB4huQA/DPj7fMLIC/Lq5vbewfHU7nHBXKA3OoNJ5fcjSd49T5LKLKdAIKBgITcADCHDAAAcoBoRABlEQbKCHNTcACqMFUEDGGP4Gg0KjGKSmaQU3WacPCyAA1PzuSNBcLSgKhbywq02rKasgxWU8NwqAZZW1PCUFlQVfpZBrFeEdarZNIrOQAPQW1QACw4mgAJuRhORqOQYBwAO4AQillAA7O1ZcKAL7zeaRKixABq21QQiwyVSGSyThcbjy8nexW1eAqVRqEMaUplnUNk2m/VYgy8WpGr1asz94tev12+yOJyq50uyjUmmmYNu9yhiLazbCCNbW3bAK7Z3IIL76i011ukOeiNhdYW5Cny2VShRaIx2LxBOJpPJlJpdIZfE0LPG7PSXKlfol5Zmoo/P53pRlOVWAVP1jT1dVNV3MC1Ug3NdQMM0vywQMYCET19VlWCwjDHco3IWMNHjRMknKDI23+TtynIVRzlaKAsx3HMemg2QABJDCCVQQyKKAQ0sctkPwuME2ibg2GQHEVBXBATiw8gcJwoA):

```csharp
using System;
using System.Threading.Tasks;
_ = Foo.NumberBuggy().Result;
Console.WriteLine("--------");
_ = Foo.NumberNotBuggy();

public class Foo{
     public static async Task<int> NumberBuggy()
     {
        string str = "a2";
        int x = 1;
        int y = 2;

        ref int r =
              ref await EvalAsync(str, 1) is "a1" ? ref x
            : ref await EvalAsync(str, 2) is "a2" ? ref y
            : ref System.Runtime.CompilerServices.Unsafe.NullRef<int>();
        
         r++;
         r++;
         r++;
         int xxx = r;
         System.Console.WriteLine(xxx);
         System.Console.WriteLine(x);
         System.Console.WriteLine(y); //should be 5 now!
         return xxx;
    }
    
    public static ValueTask<int> NumberNotBuggy()
     {
        string str = "a2";
        int x = 1;
        int y = 2;

        ref int r =
              ref EvalAsync(str, 1).Result is "a1" ? ref x
            : ref EvalAsync(str, 2).Result is "a2" ? ref y
            : ref System.Runtime.CompilerServices.Unsafe.NullRef<int>();
        
         r++;
         r++;
         r++;
         int xxx = r;
         System.Console.WriteLine(xxx);
         System.Console.WriteLine(x);
         System.Console.WriteLine(y);
         return new(xxx);
    }
    static ValueTask<string> EvalAsync(string s, int i)
    {
        System.Console.WriteLine($"{s} {i}");
        return ValueTask.FromResult(s);
    }
}
```

**Expected Behavior**:
Same output for both methods (they are not really async)

**Actual Behavior**:
The result of the conditional expression is stored in a non-ref temp local which causes that the later increment operations do not change the original value - just the temp.

**Error cause**
The error is caused in [SpillSequenceSpiller](https://github.com/dotnet/roslyn/blob/847d58843a139405d633a55c82b232582c996848/src/Compilers/CSharp/Portable/Lowering/SpillSequenceSpiller.cs#L1135
)

The code should be:
```csharp
var tmp = _F.SynthesizedLocal(node.Type, kind: SynthesizedLocalKind.Spill, syntax: _F.Syntax,
   refKind: node.IsRef ? RefKind.Ref : RefKind.None);
```

But how can we determine if the result should be ref-readonly or not?

bernd5:
Fixes #74115


src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -42,6 +42,26 @@ public static RefKind GetRefKind(this BoundExpression node)
                 case BoundKind.ImplicitIndexerAccess:
                     return ((BoundImplicitIndexerAccess)node).IndexerOrSliceAccess.GetRefKind();
 
+                case BoundKind.ConditionalOperator:
```
AlekseyTs, line 45: >case BoundKind.ConditionalOperator: [](http://example.com/codeflow?start=16&length=35)

This case probably doesn't belong here, ```ConditionalOperator``` doesn't represent  a symbol reference.

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -42,6 +42,26 @@ public static RefKind GetRefKind(this BoundExpression node)
...
+                    {
+                        var cond = (BoundConditionalOperator)node;
+                        if (!cond.IsRef)
+                        {
+                            return RefKind.None;
+                        }
+                        var conseqRefKind = cond.Consequence.GetRefKind();
+                        if (conseqRefKind == RefKind.RefReadOnly)
```
AlekseyTs, line 53: >conseqRefKind == RefKind.RefReadOnly [](http://example.com/codeflow?start=28&length=36)

With parameters, do we need to specially handle other RefKind values?

AlekseyTs:
It looks like there are no tests.

AlekseyTs:

bernd5:
@AlekseyTs do you have an idea for a test?

jcouv:
The scenarios in the issue being fixed should be included as tests, for starters.
Additional tests depend on the nature of the product change, but it looks like that's not yet finalized, so it's too early to tell.

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -14,8 +14,10 @@ namespace Microsoft.CodeAnalysis.CSharp
     internal static partial class BoundExpressionExtensions
     {
         /// <summary>
-        /// Returns the RefKind if the expression represents a symbol
-        /// that has a RefKind, or RefKind.None otherwise.
+        /// Returns the RefKind of an expression
+        /// if the expression represents a symbol, the RefKind of the symbol is returned
+        /// if the expression is a "by-ref" expression like conditional operator or InlineArrayAccess the RefKind is calculated and returned
```
AlekseyTs, line 19: >if the expression is a "by-ref" expression like conditional operator or InlineArrayAccess the RefKind is calculated and returned [](http://example.com/codeflow?start=12&length=128)

I am somewhat uncomfortable with this change. This helper is used in several places. It hasn't been demonstrated that any of the existing call sites are actually going to benefit from this change. Moreover, there is a risk that some of the existing call sites could be negatively affected by this change. And that is the risk that we probably don't want to take that late in the release cycle, especially without a proven benefit. I suggest to either demonstrating a benefit for an existing call site, or take a more conservative approach by specially handling ```BoundConditionalOperator``` only in the place where we intend to change the behavior.

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -42,6 +44,45 @@ public static RefKind GetRefKind(this BoundExpression node)
...
                     return ((BoundImplicitIndexerAccess)node).IndexerOrSliceAccess.GetRefKind();
 
+                case BoundKind.ConditionalOperator:
+                    {
+                        var conditionalOperator = (BoundConditionalOperator)node;
+                        if (!conditionalOperator.IsRef)
+                        {
+                            return RefKind.None;
+                        }
+                        {
```
AlekseyTs, line 54: >{ [](http://example.com/codeflow?start=24&length=1)

The extra blocks looks strange, I do not think they help readability of the code.

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -42,6 +44,45 @@ public static RefKind GetRefKind(this BoundExpression node)
...
+                                case RefKind.Ref:
+                                case RefKind.Out:
+                                    return null;
+                                case RefKind.RefReadOnlyParameter:
+                                case RefKind.RefReadOnly:
+                                    return RefKind.RefReadOnly;
+                                case RefKind.None:
+                                    return null; //could be e.g. a throw expression
+                                default:
+                                    return RefKind.RefReadOnly;
```
AlekseyTs, line 81: >return RefKind.RefReadOnly; [](http://example.com/codeflow?start=36&length=27)

It is not obvious why is this a good default

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -42,6 +44,45 @@ public static RefKind GetRefKind(this BoundExpression node)
...
 
+                case BoundKind.ConditionalOperator:
+                    {
+                        var conditionalOperator = (BoundConditionalOperator)node;
+                        if (!conditionalOperator.IsRef)
+                        {
+                            return RefKind.None;
+                        }
+                        {
+                            if (tryGetReadOnlyRefKind(conditionalOperator.Consequence) is { } r)
```
AlekseyTs, line 55: >tryGetReadOnlyRefKind [](http://example.com/codeflow?start=32&length=21)

The intent should be clearly captured in comment explaining why is this necessary. Also, the need to do this should be backed by tests, their behavior should observably change if the logic around "readonly" is "broken" or removed.

src/Compilers/CSharp/Portable/Lowering/SpillSequenceSpiller.cs
```diff
@@ -1132,7 +1132,8 @@ public override BoundNode VisitConditionalOperator(BoundConditionalOperator node
             }
             else
             {
-                var tmp = _F.SynthesizedLocal(node.Type, kind: SynthesizedLocalKind.Spill, syntax: _F.Syntax);
+                var tmp = _F.SynthesizedLocal(node.Type, kind: SynthesizedLocalKind.Spill, syntax: _F.Syntax,
+                    refKind: node.GetRefKind());
```
AlekseyTs, line 1136: >refKind: node.GetRefKind() [](http://example.com/codeflow?start=20&length=26)

I assume there are scenarios when this temp cannot be preserved across await operations that follow ```BoundConditionalOperator``` in the same expression. We should have a test for a scenario like that.

AlekseyTs:

bernd5:
@AlekseyTs can you write a test or help me to do so for the linked issue.

I explain what I would want to test:

compile and execute the following code:
```csharp
string str = "a2";
int x = 1;
int y = 2;

ref int r =
      ref str is "whatever" ? ref x
    : ref str is { Length: >= 2 and <= 10 or 22 } ? ref y
    : ref System.Runtime.CompilerServices.Unsafe.NullRef<int>();

r++;
r++;
r++;
int xxx = r;
System.Console.WriteLine(xxx); //5
System.Console.WriteLine(x); //1
System.Console.WriteLine(y); //expected 5 - but we get 2
```

It should print:
```
5
1
5
```

currently it prints:
```
5
1
2
```

And it already works if no "Or Pattern" is used (without ` or 22 `)

The reason is because the alternative
```csharp
ref str is { Length: >= 2 and <= 10 or 22 } ? ref y
    : ref System.Runtime.CompilerServices.Unsafe.NullRef<int>()
```
needs spilling and stores its result in a temporary by value.

jcouv:
For writing a test, take a look at `TestRefConditionalDifferentTypes3`.
The broad strokes are `CreateCompilation`, `CompileAndVerify` (verifies execution among other things), `VerifyDiagnostics` and `VerifyIL`.

bernd5:
I added the following test method:

```csharp
        [Fact]
        public void TestRefConditionalWithSpilling()
        {
            var source = """
                class C
                {
                    public static void Main()
                    {
                        string str = "a2";
                        int x = 1;
                        int y = 2;
                        int z = 777;

                        ref int r =
                              ref str is "whatever" ? ref x
                            : ref str is { Length: >= 2 and <= 10 or 22 } ? ref y
                            : ref z;

                        r++;
                        r++;
                        r++;
                        int xxx = r;
                        System.Console.WriteLine(xxx); //5
                        System.Console.WriteLine(x); //1
                        System.Console.WriteLine(y); //expected 5 - but we get 2
                    }
                }
                """;

            var comp = CompileAndVerify(source,
                expectedOutput: """
                5
                1
                5
                """);
            comp.VerifyDiagnostics();
        }
```

But if I try to execute it, I get:
```
System.NullReferenceException : Object reference not set to an instance of an object.

C.Main() in :line 10
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

Do you know why?

bernd5:
@AlekseyTs can you review again?

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
jcouv: @bernd5 Was this feedback addressed?

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -78,6 +78,62 @@ public static RefKind GetRefKind(this BoundExpression node)
...
+                        if (!conditionalOperator.IsRef)
+                        {
+                            return RefKind.None;
+                        }
+                        //the ref kind of the entire expression is ref readonly if the ref kind of the consequence
+                        //or the alternative is ref readonly otherwise it just by ref (not readonly)
+                        //
+                        //without this check we loose the readonly specifier
+                        //See Binder.CheckValueKind
+                        if (tryGetReadOnlyRefKind(conditionalOperator.Consequence) is { } readOnlyRefKindOfConsequence)
```
AlekseyTs, line 103: Do we have a test that is going to fail when this `if` statement is removed?

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -78,6 +78,62 @@ public static RefKind GetRefKind(this BoundExpression node)
...
+                        {
+                            return readOnlyRefKindOfConsequence;
+                        }
+                        if (tryGetReadOnlyRefKind(conditionalOperator.Alternative) is { } readOnlyRefKindOfAlternative)
```
AlekseyTs, line 107: Do we have a test that is going to fail when this `if` statement is removed?

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -78,6 +78,62 @@ public static RefKind GetRefKind(this BoundExpression node)
...
+                            {
+                                case RefKind.Ref:
+                                case RefKind.Out:
+                                    return null;
+                                case RefKind.RefReadOnlyParameter:
+                                case RefKind.RefReadOnly: //same as RefKind.In
+                                case RefKindExtensions.StrictIn:
+                                    return RefKind.RefReadOnly;
+                                case RefKind.None:
+                                    return null; //could be e.g. a throw expression
```
AlekseyTs, line 125: Can a `throw` expression be used in a `ref` conditional operator?

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -78,6 +78,62 @@ public static RefKind GetRefKind(this BoundExpression node)
...
+                                case var unhandled:
+                                    Debug.Assert(false, $"unhandled refKind: {unhandled}");
+                                    return RefKind.RefReadOnly;
```
AlekseyTs, line 128: I do not think we should have a `default` case like this. Consider using a switch expression without "default" instead. If it is not exhaustive because a named enum value is not handled, the switch expression is going to produce a warning like
> warning CS8509: The switch expression does not handle all possible values of its input type (it is not exhaustive). For example, the pattern 'RefKind.In' is not covered.

Otherwise, it is going to produce a warning like following:
> warning CS8524: The switch expression does not handle some values of its input type (it is not exhaustive) involving an unnamed enum value. For example, the pattern '(RefKind)6' is not covered.

We should suppress CS8524 with `pragma`

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -78,6 +78,62 @@ public static RefKind GetRefKind(this BoundExpression node)
...
+                        static RefKind? tryGetReadOnlyRefKind(BoundExpression expression)
+                        {
+                            switch (expression.GetRefKindEx())
+                            {
+                                case RefKind.Ref:
+                                case RefKind.Out:
+                                    return null;
+                                case RefKind.RefReadOnlyParameter:
+                                case RefKind.RefReadOnly: //same as RefKind.In
+                                case RefKindExtensions.StrictIn:
```
AlekseyTs, line 122: In some cases ```RefKindExtensions.StrictIn``` is actually handled differently by comparison to RefKind.RefReadOnly. I think we should return ```RefKindExtensions.StrictIn``` for it and, in case of an ambiguity between ```RefKindExtensions.StrictIn``` and ```RefKind.RefReadOnly``` across branches, we should prefer ```RefKindExtensions.StrictIn```

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
```diff
@@ -78,6 +78,62 @@ public static RefKind GetRefKind(this BoundExpression node)
             }
         }
 
+        /// <summary>
+        /// Returns the RefKind of an expression
+        /// if the expression represents a symbol, the RefKind of the symbol is returned
+        /// if the expression is a "by-ref" expression like conditional operator or InlineArrayAccess the RefKind is calculated and returned
+        /// otherwise this function returns RefKind.None 
+        /// </summary>
+        public static RefKind GetRefKindEx(this BoundExpression node)
```
AlekseyTs, line 87: I suggest to move this helper to SpillSequenceSpiller.cs right after ```VisitConditionalOperator```, it doesn't have to be an extension method.

src/Compilers/CSharp/Portable/BoundTree/BoundExpressionExtensions.cs
AlekseyTs: Probably always using ```RefKindExtensions.StrictIn``` instead of ```RefKind.RefReadOnly``` is a better alternative.

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenRefConditionalOperatorTests.cs
```diff
@@ -1015,6 +1015,241 @@ .maxstack  1
 ");
         }
 
+        [Fact]
+        public void TestRefConditionalWithSpilling()
```
AlekseyTs, line 1019: Please add a ```WorkItem``` attribute.

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenRefConditionalOperatorTests.cs
```diff
@@ -1015,6 +1015,241 @@ .maxstack  1
...
+                string str = "a2";
+                int x = 1;
+                int y = 2;
+
+                ref readonly int roy = ref x;
+
+                ref readonly int r =
+                      ref str is "whatever" ? ref roy
+                    : ref str is { Length: >= 2 and <= 10 or 22 } ? ref y
+                    : ref System.Runtime.CompilerServices.Unsafe.NullRef<int>();
```
AlekseyTs, line 1149: It looks like this test is failing in CI because ```Unsafe``` type isn't available in ```TargetFramework.NetLatest``` when the test is executed on Desktop framework. Consider using ```TargetFramework.NetCoreApp``` instead and conditionally supplying expected output:
```
                expectedOutput: !ExecutionConditionUtil.IsMonoOrCoreClr ? null : """
                5
                1
                5
                """);
```

src/Compilers/CSharp/Portable/Lowering/SpillSequenceSpiller.cs
```diff
@@ -1132,13 +1132,16 @@ public override BoundNode VisitConditionalOperator(BoundConditionalOperator node
             }
             else
             {
-                var tmp = _F.SynthesizedLocal(node.Type, kind: SynthesizedLocalKind.Spill, syntax: _F.Syntax);
+                var isRef = node.IsRef;
+                var tmp = _F.SynthesizedLocal(node.Type, kind: SynthesizedLocalKind.Spill, syntax: _F.Syntax,
+                    refKind: isRef ? node.GetRefKindEx() : RefKind.None
```
AlekseyTs, line 1137: It looks like this change introduces a compiler crash for a scenario that worked fine before:
```
        [Fact]
        public void Test1()
        {
            var src1 = @"
using System.Threading.Tasks;

class C
{
    int F;

    static async Task Test(bool b)
    {
        Test(b ? ref (await GetC()).F : ref (await GetC()).F, await GetC());
    }

    static void Test(int x, C y) {}

    static async Task<C> GetC()
    {
        await Task.Yield();
        return new C();
    }
}
";
            var comp1 = CreateCompilation(src1, options: TestOptions.ReleaseDll);
            comp1.VerifyEmitDiagnostics();
        }
```

This scenario also crashes compiler:
```
        [Fact]
        public void Test2()
        {
            var src1 = @"
using System.Threading.Tasks;

class C
{
    int F;

    static async Task Test(bool b)
    {
        Test(ref b ? ref (await GetC()).F : ref (await GetC()).F, await GetC());
    }

    static void Test(ref int x, C y) {}

    static async Task<C> GetC()
    {
        await Task.Yield();
        return new C();
    }
}
";
            var comp1 = CreateCompilation(src1, options: TestOptions.ReleaseDll);
            comp1.VerifyEmitDiagnostics();
        }
```

AlekseyTs: I suggest trying an alternative rewrite here, a rewrite that doesn't introduce ref locals. Instead of creating a local for result, creating a local for condition might work:
```
<condition side-effects>
bool conditionValue = condition;
if (conditionValue)
{
     <consequence side-effects>
} 
else
{
     <alternative side-effects>
}   

return conditionValue ? [ref] consequence : [ref] alternative;
```

AlekseyTs:

Issue #74086, opened 2024-06-20
Run a single generator at a time


chsienki:
## Background and Motivation

Today the generator driver only has the ability to run all generators it contains. As part of the razor cohosting work we would like the ability to only update a single generator (the razor generator in this case).

While we could remove the existing generators, run the driver, and then re-add them, this throws away any saved state associated with the removed generators meaning subsequent runs would not be incremental. 

## Proposed API

```diff
namespace Microsoft.CodeAnalysis
{
    public abstract class GeneratorDriver
    {
+         public GeneratorDriver RunGenerator(ISourceGenerator generator, Compilation compilation, CancellationToken cancellationToken = default)  
    }
}
```

## Usage Examples

``` C#
var compilation = GetCompilation();
var generator1 = new MySourceGenerator();
var generator2 = new MySourceGenerator();
var driver = new CSharpGeneratorDriver([generator1, generator2]);

// run all generators
driver = driver.RunGenerators(compilation);

// run only a single generator
driver = driver.RunGenerator(generator1, compilation);

var runResult = driver.GetRunResult();

var gen1Result = runResult.Results[0]; // result from second run
var gen2Result = runResult.Results[1]; // result from initial 'all generators' run


```

## Alternative Designs

```c#
// choose generator to run by index in the run result
// Seems too tied to impl
public GeneratorDriver RunGenerator(int generatorIndex, Compilation compilation, CancellationToken cancellationToken = default) 

// choose generator to run by type
// May run multiple generators, need to consider constraint on TGenerator to handle both ISourceGenerator and IIncrementalGenerator
public GeneratorDriver RunGenerator<TGenerator>(Compilation compilation, CancellationToken cancellationToken = default) 

// Add a param to RunGenerators that specifies a set of generators to run. 
// Generally, don't see a need to run a subset greater than a single generator.
// Passing null to say 'all' generators seems weird.
public GeneratorDriver RunGenerators(Compilation compilation, CancellationToken cancellationToken = default, IEnumerable<ISourceGenerator> specificGenerators = null) 

// Add a param to RunGenerators that has a filter
// Less weird that the filter being null equals all generators
public GeneratorDriver RunGenerators(Compilation compilation, CancellationToken cancellationToken = default, Func<ISourceGenerator, bool> generatorFilter = null) 

```

333fred:
### API Review

* We think a multiple-generator option is the better approach given balanced mode may want to run multiple generators in the future
[...]

Issue #74039, opened 2024-06-18
.AsIncrementalGenerator extension method


chsienki:
## Background and Motivation

Today, we have an extension method `IIncrementalGenerator.AsSourceGenerator()` that allows the user to wrap an `IIncrementalGenerator` as an `ISourceGenerator` so that it can be passed to other generator APIs that only accept an `ISourceGenerator`. 

The `ISourceGenerator` returned from this call is opaque, and generally not useful for anything other than passing to the driver. 

We also have an API `ISourceGenerator GeneratorRunResult.Generator { get; }` that returns the instance of the generator that was passed to the driver. If a user passes an `IIncrementalGenerator` via `.AsSourceGenerator()` this will be the opaque wrapped type, which is of little use.

We have another extension method `ISourceGenerator.GetGeneratorType()` that returns the _type_ of the generator in the opaque wrapper, but there is no way of getting the actual instance. 

## Proposed API

```diff
namespace Microsoft.CodeAnalysis
{
    public static class GeneratorExtensions
    {
+        public static IIncrementalGenerator AsIncrementalGenerator(ISourceGenerator generator)

+        public static Type GetGeneratorType(IIncrementalGenerator)
    }
}
```

This would allow a user to effectively 'unwrap' the opaque `ISourceGenerator` and get back the original instance of the `IIncrementalGenerator` that was wrapped. 

If a user calls `AsIncrementalGenerator` on a non-wrapped `ISourceGenerator` they will get back an opaque type that can be used anywhere the compiler expects and `IIncrementalGenerator`. 

We extend the `GetGeneratorType` extension method so that you can similarly get at the underlying type without needing to know if it's wrapped or not.

## Usage Examples

``` C#
var generator = new MyIncrementalGenerator();
var driver = new CSharpGeneratorDriver(generator.AsSourceGenerator());
var result = driver.RunGenerators().GetRunResult();
var opaqueWrapper = result.Results[0].Generator;

var underlyingType = opaqueWrapper.GetGeneratorType();
Assert.Equal(typeof(MyIncrementalGenerator), underlyingType);

var unwrapped = opaqueWrapper.AsIncrementalGenerator();
Assert.IsSame(generator, unwrapped);
```

These effectively give the user the inverse of the current `AsSourceGenerator` API. 

[...]

PR #74127, opened 2024-06-24
New generator apis

chsienki:
Adds APIs to convert `ISourceGenerators` to `IIncrementalGenerators` and allows the generator driver to only update a sub-set of generators via a new filter parameter.

Fixes https://github.com/dotnet/roslyn/issues/74039
Fixes https://github.com/dotnet/roslyn/issues/74086


src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
```diff
@@ -42,16 +42,26 @@ internal GeneratorDriver(ParseOptions parseOptions, ImmutableArray<ISourceGenera
...
+        public GeneratorDriver RunGenerators(Compilation compilation, CancellationToken cancellationToken) => RunGenerators(compilation, generatorFilter: null, cancellationToken);
+
+        /// <summary>
+        /// Run generators and produce an updated <see cref="GeneratorDriver"/> containing the results.
+        /// </summary>
+        /// <param name="compilation">The compilation to run generators against</param>
+        /// <param name="generatorFilter">An optional filter that specifies which generators to run. If <c>null</c> all generators will run.</param>
+        /// <param name="cancellationToken">Used to cancel an in progress operation.</param>
+        /// <returns>An updated driver that contains the results of the generators running.</returns>
+        public GeneratorDriver RunGenerators(Compilation compilation, Func<GeneratorFilterContext, bool>? generatorFilter = null, CancellationToken cancellationToken = default)
```
chsienki, line 55: I couldn't figure out a way to make `generatorFilter` non-optional (as discussed in the API review) without breaking our back compat guidelines.

Another option would be to just make this a differently named method, like `RunGeneratorsWithFilter`

chsienki:
@dotnet/roslyn-compiler for review please.

333fred (approved):

jaredpar:
@jjonescz PTAL

src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
chsienki: Hmm, I thought it was complaining about it, but it's not. Fixed.

chsienki: Our analyzer requires that we don't have multiple overloads with optional parameters: https://github.com/dotnet/roslyn/blob/main/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md

That means that the signatures when we make `generatorFilter` required are:
```
RunGenerators(Compilation, CancellationToken)
RunGenerators(Compilation, Func<> filter, CancellationToken = default)
```

Anyone who was calling `RunGenerators(Compilation)` will now be source broken, because we no longer have an overload with only a single required parameter.

Note: It *also* says that we shouldn't add new optional parameters in the middle of a signature for source-compat. Imagine if you had `RunGenerator(compilation, default)` that might bind to a different overload now. So we're in a bit of tricky situation.

We can either: make the filter optional, and bend the 'don't put optional in the middle rule', or we can make it non-optional, and introduce a third overload of `RunGenerators` that *only* takes a `Compilation` parameter. 

Tagging @333fred for any API shape thoughts on this one?

src/Compilers/CSharp/Test/Semantic/SourceGeneration/GeneratorDriverTests.cs
```diff
@@ -4199,6 +4199,160 @@ public static void Interceptor(this Program program, int param)
...
+            driver = driver.RunGenerators(compilation, ctx => true);
+            Assert.True(stepRan);
+
+            // re-run everything
+            stepRan = false;
+            driver = driver.RunGenerators(compilation);
+
+            // step didn't run because the generator was already up to date
+            Assert.False(stepRan);
+        }
```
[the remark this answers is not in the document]
chsienki: Hmmm, this is tricky. It seems logical to say that we shouldn't even init a generator if it's filtered out, which would skip the post init stage. 

However, we also have a step there where we re-parse the post init docs if the parse options change. Right now that's a single flag as it's assumed all the generators run. If we skip a generator on a parse options change then it'll not get them reparsed and cause errors the next time we run. 

We would need to track per-generator if they have had their parse options updated or not, so that we can re-parse them as needed. The alternative is to just always do the init + postinit steps for a generator even if its requested to be skipped.

Let me play around a little and see how hard the per-generator flag would be.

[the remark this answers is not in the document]
chsienki: So, the per generator flag is actually trivial, as we can just compare the produced trees with the latest parse options but turns out we don't need it. 

I've updated it so that the generator doesn't init, and doesn't take part in the compilation at all if it's filtered out + added tests to cover those cases.

src/Compilers/Core/Portable/SourceGeneration/GeneratorState.cs
```diff
@@ -124,5 +125,7 @@ public GeneratorState WithError(Exception exception, Diagnostic error, TimeSpan
         internal ImmutableDictionary<string, ImmutableArray<IncrementalGeneratorRunStep>> OutputSteps { get; }
 
         internal ImmutableArray<(string Key, string Value)> HostOutputs { get; }
+
+        internal bool RequiresPostInitReparse(ParseOptions parseOptions) => PostInitTrees.Any(t => t.Tree.Options != parseOptions);
```
jjonescz, line 129: Consider using `static` lambda if possible:

```suggestion
        internal bool RequiresPostInitReparse(ParseOptions parseOptions) => PostInitTrees.Any(static (t, parseOptions) => t.Tree.Options != parseOptions, parseOptions);
```

src/Compilers/Core/Portable/SourceGeneration/GeneratorDriverState.cs
```diff
@@ -20,8 +20,7 @@ namespace Microsoft.CodeAnalysis
                                       DriverStateTable stateTable,
                                       SyntaxStore syntaxStore,
                                       GeneratorDriverOptions driverOptions,
-                                      TimeSpan runtime,
-                                      bool parseOptionsChanged)
+                                      TimeSpan runtime)
```
jjonescz, line 23: Can we remove the `internal readonly bool ParseOptionsChanged;` field as well?

src/Compilers/Core/Portable/SourceGeneration/GeneratorAdaptor.cs
```diff
@@ -15,6 +15,11 @@ namespace Microsoft.CodeAnalysis
     /// </summary>
     internal sealed class SourceGeneratorAdaptor : IIncrementalGenerator
     {
+        /// <summary>
+        /// A dummy extension that is used to indicate this adaptor was created outside of the driver.
+        /// </summary>
+        public const string DummySourceExtension = ".dummy";
+
         private readonly string _sourceExtension;
```
jaredpar, line 23: Is this a value that is controllable by customers? If so have reservations about using a magic string. Had trouble tracking this through all the APIs to see where it originates.

src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
jaredpar: What is the problem with the signature in the current form? Yes it's optional but it just gets filled in as essentially "run everything". That behavior is compatible with the existing overload so that all seems to track  for me. What am I missing?

src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
```diff
@@ -195,7 +205,7 @@ public GeneratorDriverTimingInfo GetTimingInfo()
             return new GeneratorDriverTimingInfo(_state.RunTime, generatorTimings);
         }
 
-        internal GeneratorDriverState RunGeneratorsCore(Compilation compilation, DiagnosticBag? diagnosticsBag, CancellationToken cancellationToken = default)
+        internal GeneratorDriverState RunGeneratorsCore(Compilation compilation, DiagnosticBag? diagnosticsBag, Func<GeneratorFilterContext, bool>? generatorFilter = null, CancellationToken cancellationToken = default)
```
jaredpar, line 208: Rather than having the filter be optional here why not normalize it in the constructor to be non-null? Essentially somethnig like 

```csharp
generatorFilter ??= static _ => true;
```

Believe that should be amortized to a single alloc, if not could cache as a `static readonly` on the type

src/Compilers/Core/Portable/SourceGeneration/GeneratorAdaptor.cs
chsienki: No, it's only created internally. We track the file extension because we actively parse the tree, so we need to pass it in from the outer driver when we wrap an ISourceGenerator. 

The generators on this path are never actually used directly, so we can just set it to anything. I chose a magic string so we can assert later on that it *isn't* used and dings us if we ever do allow it, so we know to fix it. 

At some point we should probably refactor these so the string isn't needed at all, but this just seemed simpler for now.

src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
chsienki: Was purely to avoid the allocation, but yeah we can amortize it and it's probably simpler.

chsienki: Hmm, actually I think I'm going to leave this as-is. We don't take it in the ctor, it's passed on every call to run generators, we have to re-assign the empty one every run anyway, so it doesn't really make anything simpler.

src/Compilers/Core/Portable/SourceGeneration/GeneratorDriver.cs
chsienki: *shrug* I'm just following the guidelines here https://github.com/dotnet/roslyn/blob/main/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md where it says do not put optional params in the middle of signatures. 

I don't see in this case why it would be an issue, as even in source `RunGenerators(compilation, default)` should still bind to the cancellation token version, but I assume the guidelines exist for a reason and don't want to break them :)

chsienki:
Helps if I actually push my changes 🤦

jjonescz (approved):

PR #74132, opened 2024-06-24
Move PEPropertySymbol to use the same UncommonFields pattern that PEMethodSymbol uses

333fred:
This will be useful for OverloadResolutionPriority as well, as most properties won't have this attribute.

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
@@ -914,7 +914,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols.Metadata.PE
 
                 if (result.DiagnosticInfo is not null || !result.SecondaryDependencies.IsNullOrEmpty())
                 {
-                    AccessUncommonFields()._lazyCachedUseSiteInfo.InterlockedInitialize(PrimaryDependency, result);
+                    AccessUncommonFields()._lazyCachedUseSiteInfo.InterlockedInitializeFromSentinel(PrimaryDependency, result);
                 }
 
                 _flags.SetUseSiteDiagnosticPopulated();
@@ -930,7 +930,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols.Metadata.PE
                 var result = uncommonFields._lazyCachedUseSiteInfo;
                 if (!result.IsInitialized)
                 {
-                    uncommonFields._lazyCachedUseSiteInfo.InterlockedInitialize(primaryDependency, new UseSiteInfo<AssemblySymbol>(primaryDependency));
+                    uncommonFields._lazyCachedUseSiteInfo.InterlockedInitializeFromSentinel(primaryDependency, new UseSiteInfo<AssemblySymbol>(primaryDependency));
                     result = uncommonFields._lazyCachedUseSiteInfo;
                 }
 
--- a/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
@@ -35,11 +35,9 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols.Metadata.PE
         private readonly TypeWithAnnotations _propertyTypeWithAnnotations;
         private readonly PEMethodSymbol _getMethod;
         private readonly PEMethodSymbol _setMethod;
-        private ImmutableArray<CSharpAttributeData> _lazyCustomAttributes;
-        private Tuple<CultureInfo, string> _lazyDocComment;
-        private CachedUseSiteInfo<AssemblySymbol> _lazyCachedUseSiteInfo = CachedUseSiteInfo<AssemblySymbol>.Uninitialized;
-
-        private ObsoleteAttributeData _lazyObsoleteAttributeData = ObsoleteAttributeData.Uninitialized;
+#nullable enable
+        private UncommonFields? _uncommonFields;
+#nullable disable
 
         // CONSIDER: the parameters could be computed lazily (as in PEMethodSymbol).
         // CONSIDER: if the parameters were computed lazily, ParameterCount could be overridden to fall back on the signature (as in PEMethodSymbol).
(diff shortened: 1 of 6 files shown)


src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
```diff
@@ -814,24 +884,28 @@ internal override bool MustCallMethodsDirectly
 
         public override string GetDocumentationCommentXml(CultureInfo preferredCulture = null, bool expandIncludes = false, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return PEDocumentationCommentUtils.GetDocumentationComment(this, _containingType.ContainingPEModule, preferredCulture, cancellationToken, ref _lazyDocComment);
+            return PEDocumentationCommentUtils.GetDocumentationComment(this, _containingType.ContainingPEModule, preferredCulture, cancellationToken, ref AccessUncommonFields()._lazyDocComment);
```
333fred, line 887: This unconditional access is the same pattern used by `PEMethodSymbol`, so I kept it here.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
```diff
@@ -867,8 +938,31 @@ internal override ObsoleteAttributeData ObsoleteAttributeData
         {
             get
             {
-                ObsoleteAttributeHelpers.InitializeObsoleteDataFromMetadata(ref _lazyObsoleteAttributeData, _handle, (PEModuleSymbol)(this.ContainingModule), ignoreByRefLikeMarker: false, ignoreRequiredMemberMarker: false);
-                return _lazyObsoleteAttributeData;
+                if (!_flags.IsObsoleteAttributePopulated)
```
333fred, line 941: Implementation basically copied wholesale from `PEMethodSymbol`, with adjustments to `GetObsoleteDataFromMetadata` as appropriate.

333fred:
@dotnet/roslyn-compiler @jcouv @cston for review. This will contribute to overload resolution priority, but is an effectively unrelated refactoring so I separated it out.

src/Compilers/Core/Portable/Binding/UseSiteInfo.cs
```diff
@@ -526,7 +526,11 @@ public void InterlockedCompareExchange(TAssemblySymbol? primaryDependency, UseSi
             }
         }
 
-        public UseSiteInfo<TAssemblySymbol> InterlockedInitialize(TAssemblySymbol? primaryDependency, UseSiteInfo<TAssemblySymbol> value)
+        /// <summary>
+        /// Atomically initializes the cache with the given value if it is currently fully default.
+        /// This <i>will not</i> initialize <see cref="CachedUseSiteInfo{TAssemblySymbol}.Uninitialized"/>.
+        /// </summary>
+        public UseSiteInfo<TAssemblySymbol> InterlockedInitializeFromDefault(TAssemblySymbol? primaryDependency, UseSiteInfo<TAssemblySymbol> value)
```
333fred, line 533: Note: the calling pattern in `PEMethodSymbol` is pretty complicated, and explicitly calls out not using the `Sentinel` for race safety. It seemed like it could be simplified to me, but I didn't want to touch it here, so I just renamed these methods to make it clear what they actually do: they both initialize a `CachedUseSiteInfo`, one assuming that `_info` is `null`, and one assuming that `info` is `Sentinel`.

333fred:
@jcouv @cston @dotnet/roslyn-compiler for review.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
```diff
@@ -146,7 +179,8 @@ public bool TryGetHasUnscopedRefAttribute(out bool hasUnscopedRefAttribute)
 
             if (propEx != null || isBad)
             {
-                result._lazyCachedUseSiteInfo.Initialize(new CSDiagnosticInfo(ErrorCode.ERR_BindToBogus, result));
+                result.AccessUncommonFields()._lazyCachedUseSiteInfo.Initialize(new CSDiagnosticInfo(ErrorCode.ERR_BindToBogus, result));
```
cston, line 182: >result.AccessUncommonFields()._lazyCachedUseSiteInfo.Initialize [](http://example.com/codeflow?start=16&length=63)

`PEMethodSymbol` has a dedicated `InitializeUseSiteDiagnostic()` method that is used for all (?) cases where use-site diagnostics are calculated, and that also calls `SetUseSiteDiagnosticPopulated()`. Is there a reason t set the fields directly here rather than borrowing that approach?

cston (approved):

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
333fred: I'm frankly not entirely certain why PEMethodSymbol has as complicated an initialization strategy as it does, but it does also initialize the diagnostic from more properties than we do in PEPropertySymbol. Here, we only initialize from:

1. The creation steps, where the created symbol has not yet been exposed to anything else, so we don't have to worry about races.
2. `GetUseSiteDiagnostic`, which does have to worry about races.

The pattern used here is the pattern used by all the rest of the `UncommonFields` members, which correctly handles the occasional race of calculating that there is no use-site diagnostic/setting `_flags.IsUseSiteDiagnostic` and creating `_uncommonFields`.

333fred:
@jcouv @dotnet/roslyn-compiler for a second review

src/Compilers/Core/Portable/Binding/UseSiteInfo.cs
jcouv: Thanks for the rename and this context. It is surprising indeed that we couldn't always use a sentinel for the uninitialized value.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
```diff
@@ -277,6 +312,33 @@ static bool anyUnexpectedRequiredModifiers(ParamInfo<TypeSymbol>[] propertyParam
             }
         }
 
+        private UncommonFields CreateUncommonFields()
```
jcouv, line 315: nit: this could be a local function in `AccessUncommonFields`

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
```diff
@@ -277,6 +312,33 @@ static bool anyUnexpectedRequiredModifiers(ParamInfo<TypeSymbol>[] propertyParam
...
+        {
+            var retVal = new UncommonFields();
+            if (!_flags.IsObsoleteAttributePopulated)
```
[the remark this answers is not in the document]
jcouv: Never mind, figured it out. We only conditionally do steps 1 and 2 (when there is non-trivial data to store). May be worth a comment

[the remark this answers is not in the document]
333fred: The scenario where there is no obsolete attribute. That will not call `AccessUncommonData()`, since it doesn't need to put anything in it.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEPropertySymbol.cs
```diff
@@ -646,10 +708,31 @@ public override ImmutableArray<CSharpAttributeData> GetAttributes()
...
+            var uncommonFields = _uncommonFields;
+            if (uncommonFields == null)
+            {
+                return ImmutableArray<CSharpAttributeData>.Empty;
+            }
+            else
+            {
+                var result = uncommonFields._lazyCustomAttributes;
+                if (result.IsDefault)
+                {
```
jcouv, line 729: Consider leaving a comment

jcouv (approved):
LGTM Thanks (iteration 3)

PR #74134, opened 2024-06-24
Consume new Razor EA

ryzngard:
The Roslyn side of https://github.com/dotnet/razor/pull/10521


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorWorkspaceListenerInitializer.cs
```diff
@@ -47,9 +43,9 @@ internal void Initialize()
                 return;
             }
 
-            _logger.LogTrace("Initializing the Razor workspace listener");
+            _logger.LogTrace("Initializing the Razor workspace listener with pipe name {0}", pipeName);
             _razorWorkspaceListener = new RazorWorkspaceListener(_loggerFactory);
```
ryzngard, line 47: @davidwengier not for right now, but is there a reason we don't just take a workspace in the constructor?

davidwengier: No idea off the top of my head. Maybe left over from something we used to need to do to avoid a circular MEF dependency?

davidwengier (approved):

Issue #74163, opened 2024-06-26
.NET 9 `Preview.5`, `Preview.6`, `Preview.7` C# compiler crashes on LINQ expression with string interpolation


SergeiPavlov:
**Version Used**: 
.NET 9 SDK 9.0.100-preview.5.24307.3
tried also Preview.6 and today's Preview.7

**Steps to Reproduce**:
Following code:
```
using System.Linq;

class Program
{
    static IQueryable<string> GetStrings() => null;

    public static void Main() =>
        GetStrings().Select(o => $"{o} {o} {o} {o}");     // Crashes with 4 or more `{o}`
}
```
with `.csproj`:
```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <LangVersion>Preview</LangVersion>
    <TargetFramework>net9.0</TargetFramework>
  </PropertyGroup>
</Project>
```
**Diagnostic Id**:
```
C:\Program Files\dotnet\sdk\9.0.100-preview.5.24307.3\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(314,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy [D:\WORK\CS\CS.csproj]
C:\Program Files\dotnet\sdk\9.0.100-preview.5.24307.3\Roslyn\Microsoft.CSharp.Core.targets(85,5): error : Process terminated. System.InvalidOperationException: Unexpected value 'Sequence' of type 'Microsoft.CodeAnalysis.CSharp.BoundKind' [D:\WORK\CS\CS.csproj]
C:\Program Files\dotnet\sdk\9.0.100-preview.5.24307.3\Roslyn\Microsoft.CSharp.Core.targets(85,5): error :    at Microsoft.CodeAnalysis.CSharp.ExpressionLambdaRewriter.VisitExpressionWithoutStackGuard(BoundExpression node) [D:\WORK\CS\CS.csproj]
C:\Program Files\dotnet\sdk\9.0.100-preview.5.24307.3\Roslyn\Microsoft.CSharp.Core.targets(85,5): error :    at Microsoft.CodeAnalysis.CSharp.ExpressionLambdaRewriter.VisitInternal(BoundExpression node) [D:\WORK\CS\CS.csproj]
[...]

PR #74166, opened 2024-06-26
Fix param collection expression binding in expression trees

bernd5:
fixes: #74163

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
@@ -1178,7 +1178,7 @@ outerDefault:
 
             ParameterSymbol final = member.GetParameters().Last();
             if ((final.IsParamsArray && final.Type.IsSZArray()) ||
-                (final.IsParamsCollection && !final.Type.IsSZArray() &&
+                (final.IsParamsCollection && !final.Type.IsSZArray() && !binder.InExpressionTree &&
                  (binder.Compilation.LanguageVersion > LanguageVersion.CSharp12 || member.ContainingModule == binder.Compilation.SourceModule)))
             {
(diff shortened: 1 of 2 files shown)


bernd5:
We should may add a test for the concrete issue, too. I just created one for the root issue.
(the project seems to have no ref assembly to Net9.0)

AlekseyTs (changes requested):
We don't want to bind differently in expression tree context. See https://github.com/dotnet/csharplang/blob/main/proposals/params-collections.md#expression-trees

bernd5:
Oh, then it becomes a breaking change to existing code. But this is wanted???

AlekseyTs:
> Oh, then it becomes a breaking change to existing code. But this is wanted???

Please see https://github.com/dotnet/csharplang/blob/main/meetings/2024/LDM-2024-06-17.md#params-span-breaks for more information.

bernd5:
I see - then we would need to fix the lowering (and use an array).

AlekseyTs:
> I see - then we would need to fix the lowering (and use an array).

Please see https://github.com/dotnet/roslyn/issues/74163#issuecomment-2191548114. First, we need to understand the root cause of the problem. Where and what is going wrong. Only then we can decide what the fix should look like. I suggest you to start with that if you would like to continue working on the issue.

AlekseyTs:
> then we would need to fix the lowering (and use an array).

We probably don't want to change the lowering in this fashion either.

bernd5:
> We probably don't want to change the lowering in this fashion either.

I wouldn't like to change the lowering code for interpolated string, too.
We could simply dissallow such code (breaking change) - but this is also not very well.

AlekseyTs:
> I wouldn't like to change the lowering code for interpolated string, too. We could simply dissallow such code (breaking change) - but this is also not very well.

Let's discuss possible fix strategies in the issue. While proposing a strategy, please make the proposal very specific and detailed. For example, the quote above looks neither specific, nor detailed to me.

PR #74171, opened 2024-06-26
Add WorkspaceKinds flags enum

tmat:
Use WorkspaceKinds flags enum in ExportWorkspaceServiceAttribute. Unlike string arrays combined enum values (e.g. `WorkspaceKinds.SemanticSearch | WorkspaceKinds.Interactive`) can be shared among multiple exports.

### DIFF
--- a/src/Workspaces/Core/Portable/Workspace/Host/Mef/ExportWorkspaceServiceAttribute.cs
+++ b/src/Workspaces/Core/Portable/Workspace/Host/Mef/ExportWorkspaceServiceAttribute.cs
@@ -31,16 +31,21 @@ namespace Microsoft.CodeAnalysis.Host.Mef
         public string Layer { get; } = layer ?? throw new ArgumentNullException(nameof(layer));
 
         /// <summary>
-        /// <see cref="WorkspaceKind"/>s that the service is specified for.
+        /// <see cref="WorkspaceKinds"/> that the service is specified for.
         /// If non-empty the service is only exported for the listed workspace kinds and <see cref="Layer"/> is not applied,
         /// unless <see cref="Layer"/> is <see cref="ServiceLayer.Test"/> in which case the export overrides all other exports.
         /// </summary>
-        internal IReadOnlyList<string> WorkspaceKinds { get; } = [];
+        internal WorkspaceKinds WorkspaceKinds { get; } = WorkspaceKinds.Unknown;
 
-        internal ExportWorkspaceServiceAttribute(Type serviceType, string[] workspaceKinds)
+        internal ExportWorkspaceServiceAttribute(Type serviceType, WorkspaceKinds workspaceKinds)
             : this(serviceType)
         {
             WorkspaceKinds = workspaceKinds;
         }
+
+        //internal ExportWorkspaceServiceAttribute(Type serviceType, WorkspaceKinds workspaceKinds)
+        //    : this(serviceType, workspaceKinds.ToArray())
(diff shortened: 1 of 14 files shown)


jasonmalinowski:
> Unlike string arrays combined enum values (e.g. WorkspaceKinds.SemanticSearch | WorkspaceKinds.Interactive) can be shared among multiple exports.

Not quite I followed the benefit there? We can define metadata attributes in other ways that might let us share things differently?

src/Workspaces/Core/Portable/Workspace/Host/Mef/ExportWorkspaceServiceAttribute.cs
```diff
@@ -31,16 +31,21 @@ public class ExportWorkspaceServiceAttribute(Type serviceType, string layer = Se
         public string Layer { get; } = layer ?? throw new ArgumentNullException(nameof(layer));
 
         /// <summary>
-        /// <see cref="WorkspaceKind"/>s that the service is specified for.
+        /// <see cref="WorkspaceKinds"/> that the service is specified for.
         /// If non-empty the service is only exported for the listed workspace kinds and <see cref="Layer"/> is not applied,
         /// unless <see cref="Layer"/> is <see cref="ServiceLayer.Test"/> in which case the export overrides all other exports.
         /// </summary>
-        internal IReadOnlyList<string> WorkspaceKinds { get; } = [];
+        internal WorkspaceKinds WorkspaceKinds { get; } = WorkspaceKinds.Unknown;
```
jasonmalinowski, line 38: Why not just convert the enum to a list of the string and then leave the rest of the code untouched? I think then it'd mean the metadata gets cached directly that way.

src/Workspaces/SharedUtilitiesAndExtensions/Workspace/Core/Workspace/Mef/ILayeredServiceMetadata.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.CodeAnalysis.Host.Mef;
 
 internal interface ILayeredServiceMetadata
 {
-    public IReadOnlyList<string> WorkspaceKinds { get; }
+    public WorkspaceKinds WorkspaceKinds { get; }
```
jasonmalinowski, line 11: Anybody using us a library can make up their own kinds, won't this break them?

jasonmalinowski: Ah I see this is an internal filter, but we _also_ use the Layer as a workspace kind as well.

tmat: Wo don't use layer as workspace kind. I cleaned that up. We use either layer or string[] of workspace kinds. The latter is now enum.

src/Workspaces/Core/Portable/Workspace/Host/Mef/ExportWorkspaceServiceAttribute.cs
tmat: We could, but this seems cleaner. The complexity of both is similar.

CyrusNajmabadi (approved):
i'm personally ok with this.  but i do want Jason to be ok as well.

PR #74182, opened 2024-06-27
Add tests for consuming instance extension methods from VB and previous version of C# compiler.

AlekseyTs:


AlekseyTs:
@jcouv, @jjonescz Please review

src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs
```diff
@@ -42696,4 +42742,1408 @@ .maxstack 8
...
+    {
+    	.maxstack 8
+    	IL_0000: ldarg.0
+    	IL_0001: call instance void [mscorlib]System.Object::.ctor()
+    	IL_0006: ret
+    }
+}
+""";
+
+#if !PREVIOUSVERSIONOFCOMPILER
```
jjonescz, line 42825: What's `PREVIOUSVERSIONOFCOMPILER`? It's not defined anywhere as far as I can see.

AlekseyTs: > It's not defined anywhere as far as I can see.

That is the intent. The branch that is not taken reflects behavior of the compiler in `main`. Obviously, it is going to fail in `roles`. The tests can be run manually in `main' by copy/pasting them into a file and defining the conditional compilation symbol.

AlekseyTs:
@jcouv, @jjonescz, @dotnet/roslyn-compiler Please review

AlekseyTs:
@jcouv, @jjonescz, @dotnet/roslyn-compiler Please review

src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs
jcouv: Consider leaving a comment on one of those directives to provide context.

jcouv (approved):
LGTM Thanks (iteration 1)

src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs
AlekseyTs: > Consider leaving a comment on one of those directives to provide context.

I will do this in a separate PR

Issue #74131, opened 2024-06-24
Test Plan for Overload Resolution Priority


333fred:
Championed issue: https://github.com/dotnet/csharplang/issues/7706
Speclet: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-13.0/overload-resolution-priority.md

## C# Compiler

- [ ] spec: confirm LangVer behavior
- [x] spec: confirm for local functions,
- [x] effect in invocation
  - [x] higher values win, lower values lose, default value is zero, same values remain ambiguous
  - [x] pruning happens on applicable candidates
  - [x] applies to extensions
  - [x] in source and metadata (changes in overrides in metadata have no effect)
  - [x] LangVer doesn't affect overload resolution
- [x] effect in object creation and attributes
- [x] effect in delegate conversion
- [x] effect on properties
  - [x] indexers
  - [x] indexed properties
- [x] effect in unary and binary operators
- [x] no effect in function type (`var x = A.B;` with multiple `B` candidates)
- [x] attribute application
  - [x] LangVer 
  - [x] error on overrides (even those restating priority) and explicit interface implementations
  - [x] error on accessors
  - [x] cycles
  - [x] applied twice
  - [x] applied on partial method or indexer
  - [x] nothing on local functions, 
  - [x] spec: error on lambdas, destructors, conversion operators, static constructors
- [x] public API? (not planned yet)
- [x] BCL types ([added](https://github.com/dotnet/runtime/pull/102176))
- [x] notify other teams/languages
- [x] expression tree. Expression<> x = (a) => A.B(a); // priority should be honored, silent source break like params? can yield an error identify and document known scenario, notify BCL team (apply with care)
- [x] interpolation handlers
- [x] query syntax

## VB Compiler
- [ ] TBD (we'll plan and implement in .NET 9)

## Productivity
No IDE impact

jaredpar:
@333fred make sure to create an issue label we can  use for this feature and attach it to this issue

333fred:
All remaining tests have been covered. Closing out.

Issue #46309, opened 2020-07-24
Bad formatting for pattern block


333fred:
Given this code:

```cs
        private static bool IsCallingConventionModifier(CustomModifier modifier)
        {
            var modifierType = ((CSharpCustomModifier)modifier).ModifierSymbol;
            return (object)modifierType.ContainingAssembly == modifierType.ContainingAssembly.CorLibrary
                   && modifierType.Name != "CallConv"
                   && modifierType.Arity == 0
                   && modifierType.Name.StartsWith("CallConv", StringComparison.Ordinal)
                   && modifierType.ContainingNamespace is
                      {
                          Name: "CompilerServices",
                          ContainingNamespace:
                          {
                              Name: "Runtime",
                              ContainingNamespace:
                              {
                                  Name: "System",
                                  ContainingNamespace: { IsGlobalNamespace: true }
                              }
                          }
                      };
        }
```

The formatter wants to do this:

```cs
        private static bool IsCallingConventionModifier(CustomModifier modifier)
        {
            var modifierType = ((CSharpCustomModifier)modifier).ModifierSymbol;
            return (object)modifierType.ContainingAssembly == modifierType.ContainingAssembly.CorLibrary
                   && modifierType.Name != "CallConv"
                   && modifierType.Arity == 0
                   && modifierType.Name.StartsWith("CallConv", StringComparison.Ordinal)
                   && modifierType.ContainingNamespace is
                   {
                       Name: "CompilerServices",
                       ContainingNamespace:
                       {
                           Name: "Runtime",
                           ContainingNamespace:
                           {
                               Name: "System",
                               ContainingNamespace: { IsGlobalNamespace: true }
                           }
                       }
                   };
        }
```

The braces are on the same level as the `&&`, it's all part of a single component of one arm of the binary expression.

Followup to https://github.com/dotnet/roslyn/issues/46284

jasonmalinowski:
The concern here was how to balance not interfering with people who want things formatted one way or the other but still to lead new code to the ideal thing. The conclusion from the design meeting is we will update the formatter/indentation system to have a primary indentation that we align to ideally, but if the user has aligned the other way we keep the code formatted as it is. That'll require some updates to the formatting engine (since we don't really do that much today) but it's something we've long needed.

PR #74190, opened 2024-06-27
Overload Resolution Priority

333fred:
This implements the `Overload Resolution Priority` language feature.

Test Plan: https://github.com/dotnet/roslyn/issues/74131
Champion issue: https://github.com/dotnet/csharplang/issues/7706
Speclet: https://github.com/dotnet/csharplang/blob/main/proposals/overload-resolution-priority.md


src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1698,6 +1711,78 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
             return currentBestIndex;
         }
 
+        private static void RemoveLowerPriorityMembers<TMember>(ArrayBuilder<MemberResolutionResult<TMember>> results)
```
333fred, line 1714: This method, and the calls to this method in the current file, are the actually interesting parts of the feature. This implements the rules from the speclet, and I've tried to be perf conscious with a couple of early outs to avoid any allocations in the very common scenarios where we have no adjusted priorities.

src/Compilers/CSharp/Portable/Errors/ErrorCode.cs
```diff
@@ -2346,5 +2346,10 @@ internal enum ErrorCode
         // Note: you will need to do the following after adding warnings:
         //  1) Re-generate compiler code (eng\generate-compiler-code.cmd).
         //  2) Update ErrorFacts.IsBuildOnlyDiagnostic (src/Compilers/CSharp/Portable/Errors/ErrorFacts.cs)
+
+        // PROTOTYPE: Condense
```
333fred, line 2350: I'll address these prototypes after we do the test plan review with the compiler team. They're here for now to make it easier for me to merge in `main`.

.vscode/settings.json
```diff
@@ -35,5 +35,5 @@
     "azure-pipelines.customSchemaFile": ".vscode/dnceng-schema.json",
     "dotnet.defaultSolution": "Roslyn.sln",
     "dotnet.completion.showCompletionItemsFromUnimportedNamespaces": true,
-    "dotnet.testWindow.disableAutoDiscovery": true
+    "dotnet.testWindow.disableAutoDiscovery": false
```
cston, line 38: >false [](http://example.com/codeflow?start=46&length=5)

Was this change intentional?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -610,6 +638,21 @@ private void DecodeWellKnownAttributeAppliedToMethod(ref DecodeWellKnownAttribut
             {
                 DecodeInterceptsLocationAttribute(arguments);
             }
+            else if (attribute.IsTargetAttribute(AttributeDescription.OverloadResolutionPriorityAttribute))
+            {
+                if (this.OverriddenMethod is not null)
```
cston, line 643: >OverriddenMethod [](http://example.com/codeflow?start=25&length=16)

Is the attribute allowed on an explicit implementation?

src/Dependencies/PooledObjects/ArrayBuilder.cs
```diff
@@ -634,6 +634,13 @@ public void AddRange(T[] items, int length)
             _builder.AddRange(items, length);
         }
 
+#if COMPILERCORE
+        public void AddRange(OneOrMany<T> items)
```
cston, line 638: >AddRange [](http://example.com/codeflow?start=20&length=8)

Is `RemoveLowerPriorityMembers()` the only caller of this method? If so, consider removing this method and calling `resultsForType.AddRangeTo(results);` directly in that caller.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1698,6 +1711,78 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
...
+                // Can't prune anything unless there's at least 2 candidates
+                case []:
+                case [_]:
+                // We only look at methods and indexers, so if this isn't one of those scenarios, we don't need to do anything.
+                case [{ Member: not (MethodSymbol or PropertySymbol { IsIndexer: true }) }, _, ..]:
+                    return;
+            }
+
+            // Attempt to avoid any allocations by starting with a quick pass through all results and seeing if any have non-default priority. If so, we'll do the full sort and filter.
+            if (results.All(r => r.LeastOverriddenMember.GetOverloadResolutionPriority() == 0))
```
cston, line 1733: >r.LeastOverriddenMember.GetOverloadResolutionPriority() [](http://example.com/codeflow?start=33&length=55)

Consider extracting a `static` local function for this expression.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1698,6 +1711,78 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
...
+                        resultsByContainingType[containingType] = OneOrMany.Create(result);
+                    }
+                    else if (currentOverloadResolutionPriority == previousOverloadResolutionPriority)
+                    {
+                        resultsByContainingType[containingType] = previousResults.Add(result);
+                    }
+                    else
+                    {
+                        removedMembers = true;
+                        Debug.Assert(previousResults.All(r => r.LeastOverriddenMember.GetOverloadResolutionPriority() > currentOverloadResolutionPriority));
```
cston, line 1762: >> currentOverloadResolutionPriority [](http://example.com/codeflow?start=118&length=35)

Would this assert be stronger with `== previousOverloadResolutionPriorty` instead?

src/Dependencies/PooledObjects/ArrayBuilder.cs
333fred: Yes, but I intentionally added this because the calling order is awkward and non-discoverable otherwise. We need AddRangeTo to make sure we avoid allocation, but no one is going to call it in that direction unless they already know it exists. I'd prefer to keep this method.

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -7974,6 +7974,19 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="IDS_FeatureAllowsRefStructConstraint" xml:space="preserve">
     <value>allows ref struct constraint</value>
   </data>
+  <!-- PROTOTYPE: Condense -->
+  <data name="ERR_CannotApplyOverloadResolutionPriorityToOverride" xml:space="preserve">
+    <value>Cannot put 'OverloadResolutionPriorityAttribute' on an overriding member.</value>
```
cston, line 7979: >Cannot put [](http://example.com/codeflow?start=11&length=10)

Perhaps "use" rather than "put", for more consistency with other attribute errors. For instance, "Cannot use 'OverloadResolutionPriorityAttribute' on ..." or "OverloadResolutionPriorityAttribute' cannot be used on ...".

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -7974,6 +7974,19 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="IDS_FeatureAllowsRefStructConstraint" xml:space="preserve">
     <value>allows ref struct constraint</value>
   </data>
+  <!-- PROTOTYPE: Condense -->
+  <data name="ERR_CannotApplyOverloadResolutionPriorityToOverride" xml:space="preserve">
```
cston, line 7978: >ERR_CannotApplyOverloadResolutionPriorityToOverride [](http://example.com/codeflow?start=14&length=51)

Consider collapsing the 3 error messages into 1 or 2, especially if we also want to report an error when the attribute is applied to an explicit interface implementation.

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
```diff
@@ -332,6 +332,21 @@ public sealed override SymbolKind Kind
...
         }
 
+        internal int OverloadResolutionPriority
+        {
+            get
+            {
+                if (!this.IsIndexer)
+                {
+                    return 0;
+                }
```
cston, line 342: Is this check necessary?

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEMethodSymbol.cs
```diff
@@ -1673,5 +1682,27 @@ internal override bool HasAsyncMethodBuilderAttribute(out TypeSymbol builderArgu
...
+            if (!_packedFlags.IsOverloadResolutionPriorityPopulated)
+            {
+                if (_containingType.ContainingPEModule.Module.TryGetOverloadResolutionPriorityValue(_handle, out int priority))
+                {
+                    Interlocked.CompareExchange(ref AccessUncommonFields()._lazyOverloadResolutionPriority, priority, 0);
+                }
+#if DEBUG
+                else
+                {
+                    // 0 is the default if nothing is present in metadata, and we don't care about preserving the difference between "not present" and "set to the default value".
```
cston, line 1697: >we don't care about preserving the difference between "not present" and "set to the default value" [](http://example.com/codeflow?start=79&length=98)

In that case, should the virtual method be `int GetOverloadResolutionPriority()` instead, so callers don't rely on the distinction between `null` and `0`?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -369,6 +369,34 @@ internal override (CSharpAttributeData?, BoundAttribute?) EarlyDecodeWellKnownAt
...
+                            boundAttribute = null;
+                        }
+                    }
+                    else
+                    {
+                        attributeData = null;
+                        boundAttribute = null;
+                    }
+
+                    return (attributeData, boundAttribute);
```
cston, line 398: Consider following the pattern used above:
```
    if (!hasAnyDiagnostics)
    {
        return (attributeData, boundAttribute);
    }
}

return (null, null);
```

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -1273,6 +1273,28 @@ internal override (CSharpAttributeData?, BoundAttribute?) EarlyDecodeWellKnownAt
...
+                        boundAttribute = null;
+                    }
+                }
+                else
+                {
+                    attributeData = null;
+                    boundAttribute = null;
+                }
+
+                return (attributeData, boundAttribute);
```
cston, line 1296: Consider following the pattern used above:
```
    if (!hasAnyDiagnostics)
    {
        return (attributeData, boundAttribute);
    }
}

return (null, null);
```

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -610,6 +638,21 @@ private void DecodeWellKnownAttributeAppliedToMethod(ref DecodeWellKnownAttribut
...
+                    diagnostics.Add(ErrorCode.ERR_CannotApplyOverloadResolutionPriorityToOverride, arguments.AttributeSyntaxOpt);
+                }
+
+                MessageID.IDS_OverloadResolutionPriority.CheckFeatureAvailability(diagnostics, arguments.AttributeSyntaxOpt);
+
+                if (this is SourcePropertyAccessorSymbol or SourceEventAccessorSymbol)
+                {
+                    // Cannot put 'OverloadResolutionPriorityAttribute' on an accessor method.
+                    diagnostics.Add(ErrorCode.ERR_CannotApplyOverloadResolutionPriorityToAccessor, arguments.AttributeSyntaxOpt);
+                }
```
cston, line 654: >} [](http://example.com/codeflow?start=16&length=1)

Consider moving the `if (this.OverriddenMethod is not null)` case here and using `else if (this.OverriddenMethod is not null)` to avoid two errors for an overridden accessor.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1591 @@
...
+                }
+                else
+                {
+                    Assert.Equal(0, m.OverloadResolutionPriority);
+                }
+            }
+        }).VerifyDiagnostics();
+
+        var comp = CreateCompilation([source, OverloadResolutionPriorityAttributeDefinition]);
+        CompileAndVerify(executable, references: [useMetadataReference ? comp.ToMetadataReference() : comp.EmitToImageReference()], expectedOutput: "1").VerifyDiagnostics();
```
cston, line 69: >useMetadataReference ? comp.ToMetadataReference() : comp.EmitToImageReference() [](http://example.com/codeflow?start=50&length=79)

`AsReference(comp, useMetadataReference)`

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1591 @@
...
+            using System.Runtime.CompilerServices;
+
+            public interface I1 {}
+            public interface I2 {}
+            public interface I3 : I1, I2 {}
+
+            public class Base
+            {
+                [OverloadResolutionPriority(1)]
+                public static void M(I2 x) => throw null;
```
cston, line 232: >public static void M(I2 x) => throw null; [](http://example.com/codeflow?start=16&length=41)

Isn't this method dropped from the candidate set even before we get to overload resolution priority?

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1591 @@
...
+            using System.Runtime.CompilerServices;
+
+            public interface I1 {}
+            public interface I2 {}
+            public interface I3 : I1, I2 {}
+
+            public class Base
+            {
+                [OverloadResolutionPriority(1)]
+                public int this[I2 x]
```
cston, line 271: >public int this[I2 x] [](http://example.com/codeflow?start=16&length=21)

Isn't this indexer dropped from the candidate set even before we look at overload resolution priority?

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1591 @@
...
+                {
+                    throw null;
+                }
+            }
+            """;
+
+        CreateCompilation([source, OverloadResolutionPriorityAttributeDefinition], parseOptions: TestOptions.Regular12).VerifyDiagnostics(
+            // (5,6): error CS8652: The feature 'overload resolution priority' is currently in Preview and *unsupported*. To use Preview features, use the 'preview' language version.
+            //     [OverloadResolutionPriority(1)]
+            Diagnostic(ErrorCode.ERR_FeatureInPreview, "OverloadResolutionPriority(1)").WithArguments("overload resolution priority").WithLocation(5, 6),
```
cston, line 901: >ERR_FeatureInPreview [](http://example.com/codeflow?start=33&length=20)

Is it necessary to report an error where the attribute is applied, rather than at the call-site?

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1591 @@
...
+        var definingComp = CreateCompilation([source, OverloadResolutionPriorityAttributeDefinition], parseOptions: TestOptions.RegularNext).VerifyDiagnostics();
+
+        var consumingSource = """
+            var c = new C("test");
+            c["test"] = 0;
+            _ = c["test"];
+            c.M("test");
+            """;
+
+        // We don't error on consumption, only on definition, so this runs just fine.
```
cston, line 918: >We don't error on consumption [](http://example.com/codeflow?start=11&length=29)

Is this covered in the spec?

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1591 @@
...
+            interface I3 : I1, I2 {}
+
+            static class C
+            {
+                [OverloadResolutionPriority(1)]
+                public static void M(this I1 x) => System.Console.WriteLine(1);
+
+                public static void M(this I2 x) => throw null;
+
+                static void Test()
```
cston, line 999: >Test [](http://example.com/codeflow?start=28&length=4)

Consider renaming to `Main` and checking `expectedOutput`.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -361,6 +361,8 @@ private static bool OverloadResolutionResultIsValid<TMember>(ArrayBuilder<Member
 
             if ((options & Options.DynamicResolution) == 0)
```
cston, line 362: >Options.DynamicResolution [](http://example.com/codeflow?start=27&length=25)

Are we testing the case where `Options.DynamicResolution` is set?

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1620,6 +1623,16 @@ private static void RemoveAllInterfaceMembers<TMember>(ArrayBuilder<MemberResolu
 
             if (!dynamicResolution)
```
cston, line 1624: >dynamicResolution [](http://example.com/codeflow?start=17&length=17)

Are we testing the case where `dynamicResolution` is set?

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1591 @@
...
+                    set => System.Console.Write(4);
+                }
+            }
+            """;
+
+        CompileAndVerify([executable, source, OverloadResolutionPriorityAttributeDefinition], expectedOutput: "4321").VerifyDiagnostics();
+
+        var comp = CreateCompilation([source, OverloadResolutionPriorityAttributeDefinition]);
+        CompileAndVerify(executable, references: [useMetadataReference ? comp.ToMetadataReference() : comp.EmitToImageReference()], expectedOutput: "4321").VerifyDiagnostics();
+    }
```
cston, line 1328: >} [](http://example.com/codeflow?start=4&length=1)

Let's test explicit interface implementations, where the attribute is either the interface method, or the explicit implementation, or both.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
333fred: Yes, that's the point of this test.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
333fred: Yes, that's the point of this test.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
333fred: Yes. We don't report errors at the call site, even when the lang version is lower. See the last line of this test.

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
333fred: I intentionally assert that no non-indexers call `TryGetOverloadResolutionPriority`, to try and avoid anyone calling that in a scenario that shouldn't be looking at priority. I suppose I could not do that, but I would prefer to do this, and not even try to crack or decode attributes in scenarios that shouldn't.

.vscode/settings.json
333fred: Yes, though I can revert if you want to keep it separate.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
jjonescz, line 1: Will this feature be implemented for VB as well?

src/Compilers/CSharp/Portable/CSharpResources.resx
333fred: I'm just going to add a 4th for explicit implementations. Sorry, I like my specific errors.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
333fred: Yes, some amount of the feature will be implemented for VB.

333fred:
@cston addressed or answered your comments, please take another look. @jcouv for review.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -1588,4 +1588,181 @@ public class C
...
+
+    [Theory, CombinatorialData]
+    public void Dynamic(bool useMetadataReference)
+    {
+        var source = @"
+            using System.Runtime.CompilerServices;
+
+            public class C
+            {
+                public C(object o) => throw null;
```
cston, line 1733: Should there be attributes on one constructor and one indexer?
```csharp
[OverloadResolutionPriority(1)]
public C(object o) => throw null;

...

[OverloadResolutionPriority(1)]
public int this[object o] { ... }
```

.vscode/settings.json
cston: > Yes, though I can revert if you want to keep it separate.

I'm not sure what the change affects, but it looks like it's unrelated to overload resolution, so consider reverting.

cston (approved):

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
```diff
@@ -332,6 +332,21 @@ public sealed override SymbolKind Kind
             }
         }
 
+        internal int OverloadResolutionPriority
+        {
+            get
+            {
+                if (!this.IsIndexer)
```
jcouv, line 339: I'm not very familiar with the concept, but I see that we also have "indexed properties" in metadata (see `IsIndexedProperty`) which also have parameters. As far as I can tell, they may not be considered indexers (`IsIndexer`). Should the attribute apply to those?

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1698,6 +1711,80 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
...
+            //     - All members that have a lower *overload_resolution_priority* than the highest found within its declaring type group are removed.
+            // - The reduced groups are then recombined into the final set of applicable candidate function members.
+
+            switch (results)
+            {
+                // Can't prune anything unless there's at least 2 candidates
+                case []:
+                case [_]:
+                // We only look at methods and indexers, so if this isn't one of those scenarios, we don't need to do anything.
+                case [{ Member: not (MethodSymbol or PropertySymbol { IsIndexer: true }) }, _, ..]:
```
jcouv, line 1728: Aside from indexed properties, are there any other symbol that hits this case?
Depending on the decision on indexed properties (see comment elsewhere) this case may not be useful.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -361,6 +361,8 @@ private static bool OverloadResolutionResultIsValid<TMember>(ArrayBuilder<Member
 
             if ((options & Options.DynamicResolution) == 0)
             {
+                RemoveLowerPriorityMembers(results);
```
jcouv, line 364: nit: We're introducing more overload resolution features that are not supported by the dynamic binder. Is there a place we should record those divergences?

jcouv (commented):
Done with review pass (iteration 15). Tests not looked at much yet

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1768 @@
...
+            c.M(arg);
+            _ = c[arg];
+            c[arg] = 1;
+            """;
+
+        CompileAndVerify([executable, source, OverloadResolutionPriorityAttributeDefinition], targetFramework: TargetFramework.Mscorlib45AndCSharp, expectedOutput: "1234").VerifyDiagnostics();
+
+        var comp = CreateCompilation([source, OverloadResolutionPriorityAttributeDefinition], targetFramework: TargetFramework.Mscorlib45AndCSharp);
+        CompileAndVerify(executable, references: new[] { AsReference(comp, useMetadataReference) }, targetFramework: TargetFramework.Mscorlib45AndCSharp, expectedOutput: "1234").VerifyDiagnostics();
+    }
```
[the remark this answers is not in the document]
333fred: 1. I can add tests for binary operators.
2. I can add tests for method groups.
3. `LangVersion`
4. They all go through overload resolution. While I can add tests for some of these, I'm not sure how interesting they actually are, since it's just a call to the `OverloadResolution` type.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
333fred: Not as far as I'm aware.

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -7974,6 +7974,22 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
...
+    <value>Cannot use 'OverloadResolutionPriorityAttribute' on an overriding member.</value>
+  </data>
+  <data name="ERR_CannotApplyOverloadResolutionPriorityToNonIndexer" xml:space="preserve">
+    <value>Cannot use 'OverloadResolutionPriorityAttribute' on a property that is not an indexer.</value>
+  </data>
+  <data name="ERR_CannotApplyOverloadResolutionPriorityToAccessor" xml:space="preserve">
+    <value>Cannot use 'OverloadResolutionPriorityAttribute' on an accessor method.</value>
+  </data>
+  <data name="ERR_CannotApplyOverloadResolutionPriorityToExplicitImplementation" xml:space="preserve">
+    <value>Cannot use 'OverloadResolutionPriorityAttribute' on an explicit interface implementation.</value>
```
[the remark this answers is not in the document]
333fred: It's an ignored location. I can certainly update the spec to more completely list every location that is ignored by the language, but it does fall out of the rules.

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
```diff
@@ -1186,6 +1186,16 @@ internal virtual void AddSynthesizedReturnTypeAttributes(PEModuleBuilder moduleB
...
 
+        /// <summary>
+        /// Gets the resolution priority of this method, 0 if not set.
+        /// </summary>
+        /// <remarks>
+        /// Do not call this method from early attribute binding, cycles will occur.
+        /// </remarks>
+        internal int OverloadResolutionPriority => TryGetOverloadResolutionPriority() ?? 0;
+
+        internal abstract int? TryGetOverloadResolutionPriority();
```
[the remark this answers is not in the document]
333fred: We certainly could. My goal was to only have one location (or perhaps 2, with PE symbols) caring about what the default was if it wasn't present. If we think that's too much indirection though, I can remove it.

src/Compilers/CSharp/Portable/Symbols/Source/SourceDestructorSymbol.cs
```diff
@@ -176,5 +176,10 @@ internal override bool GenerateDebugInfo
         {
             get { return true; }
         }
+
+        internal sealed override int? TryGetOverloadResolutionPriority()
+        {
+            return null;
```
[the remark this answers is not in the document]
333fred: I'll add it. It should indeed be an error, since it's ignored.

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
333fred: Yeah, it probably should, thanks for the reminder.

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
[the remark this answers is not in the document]
jcouv: Okay. If I think of the next time we add a symbol and we need to implement the overload, the `null` approach will be a bit easier (more obvious what the right value to return will be).

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1768 @@
...
+
+        var comp2 = CreateCompilation([source2, OverloadResolutionPriorityAttributeDefinition], references: [comp1_1.ToMetadataReference()], targetFramework: TargetFramework.Standard);
+        comp2.VerifyDiagnostics();
+
+        var source3 = """
+            var c = new C();
+            c.M("test");
+            """;
+
+        var comp3 = CreateCompilation(source3, references: [comp2.ToMetadataReference(), comp1_2.ToMetadataReference()], targetFramework: TargetFramework.Standard);
```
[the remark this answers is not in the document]
333fred: Clarified offline: retargeting tests are very annoying to try and run, since we don't generate binding redirects, and ILVerify gets upset.

src/Compilers/Test/Utilities/CSharp/CSharpTestBase.cs
```diff
@@ -716,6 +716,70 @@ public sealed class CollectionBuilderAttribute : Attribute
...
+                    01 00 00 00
+                )
+
+                // Methods
+                .method public hidebysig specialname rtspecialname 
+                    instance void .ctor (
+                        int32 priority
+                    ) cil managed 
+                {
+                    // Method begins at RVA 0x2069
```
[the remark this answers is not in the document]
333fred: I'll do some cleanup, but I always prefer to put attributes here. It's the only usage today, but there's no guarantee that continue to the be case, and in my experience we're more likely to copy definitions around in the future, not move them.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEMethodSymbol.cs
```diff
@@ -1673,5 +1682,27 @@ internal override bool HasAsyncMethodBuilderAttribute(out TypeSymbol builderArgu
             builderArgument = _containingType.ContainingPEModule.TryDecodeAttributeWithTypeArgument(this.Handle, AttributeDescription.AsyncMethodBuilderAttribute);
             return builderArgument is not null;
         }
+
+        internal override int? TryGetOverloadResolutionPriority()
+        {
+            if (!_packedFlags.IsOverloadResolutionPriorityPopulated)
+            {
+                if (_containingType.ContainingPEModule.Module.TryGetOverloadResolutionPriorityValue(_handle, out int priority))
+                {
```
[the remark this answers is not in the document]
333fred: We can't easily do that for accessors (and I have tests demonstrating the behavior), since we don't track, at the PEMethodSymbol layer, if this method is an accessor for some property or event. I instead opted to just read them, and not care whether the language may not permit them to be declared. We could certainly optimize the override/destructor/non-indexer properties, but it didn't seem like it would be worth the effort to me. If a tree falls in the forest but no one is there to hear it, did it make a sound?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -369,6 +369,28 @@ internal override (CSharpAttributeData?, BoundAttribute?) EarlyDecodeWellKnownAt
                     // diagnostics that might later get thrown away as possible when binding method calls.
                     return (null, null);
                 }
+                else if (CSharpAttributeData.IsTargetEarlyAttribute(arguments.AttributeType, arguments.AttributeSyntax, AttributeDescription.OverloadResolutionPriorityAttribute))
+                {
+                    if (this is SourcePropertyAccessorSymbol or SourceEventAccessorSymbol)
```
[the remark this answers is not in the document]
333fred: Technically speaking, I don't think any of these (except lambdas) are places where the attribute would be _ignored_: it may be rather pointless to put them there, but overload resolution _will_ run on operators, conversions, and local function invocations. Lambdas would be ignored though, so we can error there.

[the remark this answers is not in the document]
jcouv: Local functions and unary operators Yes, they do involve overload resolution, but overloads are not possible so an overload resolution attribute doesn't make sense. I thought we error on positions where the attribute will be ignored.

User-defined operators: I don't see where overload resolution is involved in the [spec](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/conversions.md#1054-user-defined-implicit-conversions) or the code.

333fred:
@jcouv addressed feedback

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1698,6 +1711,85 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
...
+            //     - All members that have a lower *overload_resolution_priority* than the highest found within its declaring type group are removed.
+            // - The reduced groups are then recombined into the final set of applicable candidate function members.
+
+            switch (results)
+            {
+                // Can't prune anything unless there's at least 2 candidates
+                case []:
+                case [_]:
+                // We only look at methods and indexers, so if this isn't one of those scenarios, we don't need to do anything.
+                case [Symbol s, _, ..] when !s.CanHaveOverloadResolutionPriority():
```
jcouv, line 1729: Can this case be hit?

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1698,6 +1711,85 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
...
+                if (result.MemberWithPriority is null)
+                {
+                    // Can happen for things like built-in binary operators
+                    continue;
+                }
+
+                var containingType = result.MemberWithPriority.ContainingType;
+                if (resultsByContainingType.TryGetValue(containingType, out var previousResults))
+                {
+                    var previousOverloadResolutionPriority = previousResults.First().MemberWithPriority.GetOverloadResolutionPriority();
```
jcouv, line 1754: nit: consider "previousPriority" and "currentPriority" to simplify (a bit easier to read)

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,2005 @@
...
+            public interface I3 : I1, I2 {}
+
+            public class C
+            {
+                {{(i1First ? i1Source : i2Source)}}
+                {{(i1First ? i2Source : i1Source)}}
+            }
+            """;
+
+        CompileAndVerify([executable, source, OverloadResolutionPriorityAttributeDefinition], expectedOutput: "1", symbolValidator: module =>
```
jcouv, line 51: Consider verifying `OverloadResolutionPriority` in source too (with `sourceSymbolValidator`). Applies to other tests too

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,2005 @@
...
+                // Methods
+                .method public hidebysig virtual 
+                    instance void M (
+                        object o
+                    ) cil managed 
+                {
+                    .custom instance void [assembly1]System.Runtime.CompilerServices.OverloadResolutionPriorityAttribute::.ctor(int32) = (
+                        01 00 00 00 00 00 00 00
+                    )
+                    // Method begins at RVA 0x2050
```
jcouv, line 493: nit: consider cleaning up IL to make the test a bit more compact. Also applies below

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,2005 @@
...
+            }
+            """;
+
+        CreateCompilation([source, OverloadResolutionPriorityAttributeDefinition]).VerifyDiagnostics(
+            // (5,6): error CS9501: Cannot use 'OverloadResolutionPriorityAttribute' on this member.
+            //     [OverloadResolutionPriority(1)]
+            Diagnostic(ErrorCode.ERR_CannotApplyOverloadResolutionPriorityToMember, "OverloadResolutionPriority(1)").WithLocation(5, 6)
+        );
+    }
+}
```
jcouv, line 2005: Do we have a test for method group conversion?
Consider also adding a test where override (in metadata) has a different priority and is ignored. We can observe the value on PEMethodSymbol too

jcouv (commented):
Done with review pass (iteration 16)

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/IMemberResolutionResultWithPriority.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.CodeAnalysis.CSharp;
+
+internal interface IMemberResolutionResultWithPriority<TMember> where TMember : Symbol
+{
+    TMember? MemberWithPriority { get; }
```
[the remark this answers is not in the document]
333fred: No, we need the member itsey, as we need to do a sort.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1698,6 +1711,85 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
...
+            //     - Candidate function members are ordered by *overload_resolution_priority*.
+            //     - All members that have a lower *overload_resolution_priority* than the highest found within its declaring type group are removed.
+            // - The reduced groups are then recombined into the final set of applicable candidate function members.
+
+            switch (results)
+            {
+                // Can't prune anything unless there's at least 2 candidates
+                case []:
+                case [_]:
+                // We only look at methods and indexers, so if this isn't one of those scenarios, we don't need to do anything.
```
cston, line 1728: >// We only look at methods and indexers, so if this isn't one of those scenarios, we don't need to do anything. [](http://example.com/codeflow?start=16&length=111)

Is this comment still meaningful since we're no longer directly checking for methods and indexers?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -1707,5 +1743,18 @@ internal override System.Reflection.MethodImplAttributes ImplementationAttribute
                 return result;
             }
         }
+
+        internal override int? TryGetOverloadResolutionPriority()
+            => GetEarlyDecodedWellKnownAttributeData()?.OverloadResolutionPriority;
+
+#pragma warning disable IDE0055 // Fix formatting
```
jcouv, line 1750: nit: Could we link to a known issue? Just want to confirm that the IDE team is aware of the problem and agrees that the current formatting analyzer is incorrect.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEMethodSymbol.cs
```diff
@@ -1673,5 +1682,29 @@ internal override bool HasAsyncMethodBuilderAttribute(out TypeSymbol builderArgu
...
+                }
+#endif
+
+                _packedFlags.SetIsOverloadResolutionPriorityPopulated();
+            }
+
+            return _uncommonFields?._lazyOverloadResolutionPriority;
+        }
+
+        internal override bool CanHaveOverloadResolutionPriority => !IsExplicitInterfaceImplementation;
```
cston, line 1708: >!IsExplicitInterfaceImplementation [](http://example.com/codeflow?start=68&length=34)

Are there callers that use this property when the method is an explicit implementation? If not, consider returning `true` always since we're not checking for other `MethodKind` cases.

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
```diff
@@ -1186,6 +1186,18 @@ internal virtual void AddSynthesizedReturnTypeAttributes(PEModuleBuilder moduleB
...
+        /// Gets the resolution priority of this method, 0 if not set.
+        /// </summary>
+        /// <remarks>
+        /// Do not call this method from early attribute binding, cycles will occur.
+        /// </remarks>
+        internal int OverloadResolutionPriority => CanHaveOverloadResolutionPriority ? (TryGetOverloadResolutionPriority() ?? 0) : 0;
+
+        internal abstract int? TryGetOverloadResolutionPriority();
+
+        internal abstract bool CanHaveOverloadResolutionPriority { get; }
```
cston, line 1199: >CanHaveOverloadResolutionPriority [](http://example.com/codeflow?start=31&length=33)

Is `CanHaveOverloadResolutionPriority` only used to optimize callers in cases where we definitely know the method cannot use overload priority? If so, is the property necessary?

It looks like the property is only used in `MethodSymbol.OverloadResolutionPriority` and `OverloadResolution.RemoveLowerPriorityMembers()`. For `MethodSymbol.OverloadResolutionPriority`, could we check `MethodKind` instead? For `OverloadResolution.RemoveLowerPriorityMembers()`, we only call `CanHaveOverloadResolutionPriority` when there are at least two overloads, so there may not be much to optimize.

src/Compilers/CSharp/Portable/Symbols/Synthesized/SynthesizedInstanceMethodSymbol.cs
```diff
@@ -88,5 +88,12 @@ internal sealed override bool HasAsyncMethodBuilderAttribute(out TypeSymbol buil
             builderArgument = null;
             return false;
         }
+
+        internal sealed override int? TryGetOverloadResolutionPriority()
+        {
+            return null;
+        }
+
+        internal sealed override bool CanHaveOverloadResolutionPriority => false;
```
cston, line 97: >CanHaveOverloadResolutionPriority [](http://example.com/codeflow?start=38&length=33)

`OverloadResolution.RemoveLowerPriorityMembers()` ignores overloads when the first one returns `false` from `CanHaveOverloadResolutionPriority`. And `SynthesizedInstanceMethodSymbol` has many implementations. Are we sure that all of these implementations are cases where all other overloads will return `false` from `CanHaveOverloadResolutionPriority`? 

For instance, `SynthesizedRecordCopyCtor` inherits from this type, but other constructors may have `[OverloadResolutionPriority]`.

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
333fred: I think you're right, so I'm greatly simplifying this.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
333fred: Added the first. The second is `Overids_ChangePriorityInMetadata_Methods/Indexers` above.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
jcouv: FYI @CyrusNajmabadi This is a formatting issue we run into regularly. Tracked by https://github.com/dotnet/roslyn/issues/46309 (but it was put in backlog). Should we ask for this to be re-triaged?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -610,6 +632,20 @@ private void DecodeWellKnownAttributeAppliedToMethod(ref DecodeWellKnownAttribut
             {
                 DecodeInterceptsLocationAttribute(arguments);
             }
+            else if (attribute.IsTargetAttribute(AttributeDescription.OverloadResolutionPriorityAttribute))
+            {
+                MessageID.IDS_OverloadResolutionPriority.CheckFeatureAvailability(diagnostics, arguments.AttributeSyntaxOpt);
+
+                if (IsOverride)
```
jcouv, line 639: nit: consider moving this case inside the `if (!CanHaveOverloadResolutionPriority)` case. They are really the same case, just with a refinement of the diagnostic ID. 
Maybe even consider `var errorCode = IsOverride ? ErrorCode.ERR_CannotApplyOverloadResolutionPriorityToOverride : ErrorCode.ERR_CannotApplyOverloadResolutionPriorityToMember;`

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
```diff
@@ -1186,6 +1186,24 @@ internal virtual void AddSynthesizedReturnTypeAttributes(PEModuleBuilder moduleB
...
+        /// Gets the resolution priority of this method, 0 if not set.
+        /// </summary>
+        /// <remarks>
+        /// Do not call this method from early attribute binding, cycles will occur.
+        /// </remarks>
+        internal int OverloadResolutionPriority => CanHaveOverloadResolutionPriority ? (TryGetOverloadResolutionPriority() ?? 0) : 0;
+
+        internal abstract int? TryGetOverloadResolutionPriority();
+
+        internal bool CanHaveOverloadResolutionPriority =>
```
jcouv, line 1199: >internal [](http://example.com/codeflow?start=8&length=8)

nit: accessibility likely can be further tightened to protected

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
jcouv: Never mind, it looks like the issue no longer repro'ed...

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
jcouv: Gentle ping

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
```diff
@@ -1186,6 +1186,18 @@ internal virtual void AddSynthesizedReturnTypeAttributes(PEModuleBuilder moduleB
...
 
+        /// <summary>
+        /// Gets the resolution priority of this method, 0 if not set.
+        /// </summary>
+        /// <remarks>
+        /// Do not call this method from early attribute binding, cycles will occur.
+        /// </remarks>
+        internal int OverloadResolutionPriority => CanHaveOverloadResolutionPriority ? (TryGetOverloadResolutionPriority() ?? 0) : 0;
+
+        internal abstract int? TryGetOverloadResolutionPriority();
```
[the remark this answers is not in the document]
jcouv: Gentle ping

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
333fred: I'm using it in the extension method assertion. I'd remove it, but it actually did catch an issue with how I'd written the body.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/IMemberResolutionResultWithPriority.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.CodeAnalysis.CSharp;
+
+internal interface IMemberResolutionResultWithPriority<TMember> where TMember : Symbol
```
jcouv, line 7: nit: consider comment to explain the purpose of this type

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
```diff
@@ -332,6 +332,23 @@ public sealed override SymbolKind Kind
             }
         }
 
+        internal int OverloadResolutionPriority
+        {
+            get
+            {
+                if (!CanHaveOverloadResolutionPriority)
+                {
+                    return 0;
```
jcouv, line 341: Is this reachable?

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -1388,6 +1404,21 @@ protected override void DecodeWellKnownAttributeImpl(ref DecodeWellKnownAttribut
                     diagnostics.Add(ErrorCode.ERR_UnscopedRefAttributeUnsupportedMemberTarget, arguments.AttributeSyntaxOpt.Location);
                 }
             }
+            else if (attribute.IsTargetAttribute(AttributeDescription.OverloadResolutionPriorityAttribute))
+            {
+                MessageID.IDS_OverloadResolutionPriority.CheckFeatureAvailability(diagnostics, arguments.AttributeSyntaxOpt);
+
+                if (!IsIndexer || IsExplicitInterfaceImplementation)
```
jcouv, line 1411: >IsExplicitInterfaceImplementation [](http://example.com/codeflow?start=34&length=33)

Should the `IsExplicitInterfaceImplementation` check be in `CanHaveOverloadResolutionPriority`?
Then we can have the same basic logic here: `if (!CanHaveOverloadResolutionPriority) { ... report diagnostic with error code 1 or 2 ... }`

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1997 @@
...
+                // Methods
+                .method public hidebysig specialname 
+                    instance int32 get_Item (
+                        object x
+                    ) cil managed 
+                {
+                    .custom instance void System.Runtime.CompilerServices.OverloadResolutionPriorityAttribute::.ctor(int32) = (
+                        01 00 01 00 00 00 00 00
+                    )
+                    // Method begins at RVA 0x2050
```
jcouv, line 1383: very nit: consider trimming other IL in tests to make them more compact. I typically just remove the irrelevant attributes and the the method preambles. Converting the multi-line signatures into single line is nice too.

jcouv (commented):
Done with review pass (iteration 17). Only minor questions and nits

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,1997 @@
...
+            }
+            """;
+
+        CreateCompilation([code, OverloadResolutionPriorityAttributeDefinition]).VerifyDiagnostics(
+            // (5,6): error CS9501: Cannot use 'OverloadResolutionPriorityAttribute' on this member.
+            //     [OverloadResolutionPriority(1)]
+            Diagnostic(ErrorCode.ERR_CannotApplyOverloadResolutionPriorityToMember, "OverloadResolutionPriority(1)").WithLocation(5, 6)
+        );
+    }
+}
```
jcouv, line 1997: Have we tried applying the attribute twice?

jcouv: From test plan review, some other suggestions:
- expression tree. Expression<> x = (a) => A.B(a); // priority should be honored, silent source break like params? can yield an error identify and document known scenario, notify BCL team (apply with care)
- interpolation handlers
- query syntax

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
333fred: Not sure what you mean. This is an internal API that may eventually be surfaced as a public API.

jcouv:
> | [Overload Resolution Priority](https://github.com/dotnet/csharplang/issues/7706) | PR not yet available | [In Progress](https://github.com/dotnet/roslyn/issues/74131) | [333fred](https://github.com/333fred) | [jcouv](https://github.com/jcouv), [cston](https://github.com/cston) | Not yet assigned | [333fred](https://github.com/333fred) |

Could be moved at the bottom of the C# 13 table, since the feature is now attached to new language version. Or would you rather wait until VB is done to make that change?

---
Refers to: docs/Language Feature Status.md:13 in 336236d. [](commit_id = 336236dadac709c85460dcb2b88c4f035d53f4a1, deletion_comment = False)

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
jcouv: I'm just asking to make sure my understanding is correct, whether this branch is reachable.

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,2532 @@
...
+            public class C
+            {
+                [OverloadResolutionPriority(1)]
+                [OverloadResolutionPriority(2)]
+                public void M(object o) => System.Console.Write(1);
+                public void M(string s) => System.Console.Write(2);
+            }
+            """;
+
+        var comp = CreateCompilation([source, OverloadResolutionPriorityAttributeDefinition]);
```
jcouv, line 2114: You probably need to use a custom definition of the attribute with AllowMultiple=true. Otherwise, we hit this early check

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
jcouv: What was the resolution?

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/IMemberResolutionResultWithPriority.cs
jcouv: What was the resolution?

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
jcouv: What was the resolution?

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
jcouv: What was the resolution?

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
jcouv: What was the resolution?

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
333fred: I did exactly what you suggested

jcouv: The second part of my suggestion was to move the `IsOverride` case into the `!CanHaveOverloadResolutionPriority` case. If it makes sense, one form could be:
```
if (!CanHaveOverloadResolutionPriority)
{
  diagnostics.Add(IsOverride ? ErrorCode.ERR_CannotApplyOverloadResolutionPriorityToOverride : ErrorCode.ERR_CannotApplyOverloadResolutionPriorityToMember, arguments.AttributeSyntaxOpt.Location);
}
```

src/Compilers/CSharp/Portable/Symbols/PropertySymbol.cs
jcouv: Resolved offline: this is only reachable by tests calling the API directly

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1699,6 +1712,83 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
...
+            //     - All members that have a lower *overload_resolution_priority* than the highest found within its declaring type group are removed.
+            // - The reduced groups are then recombined into the final set of applicable candidate function members.
+
+            switch (results)
+            {
+                // Can't prune anything unless there's at least 2 candidates
+                case []:
+                case [_]:
+                    return;
+            }
```
cston, line 1730: Consider using `if (results.Length < 2) return;`

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,2536 @@
...
+
+using System;
+using System.Linq;
+using Microsoft.CodeAnalysis.CSharp.Symbols;
+using Microsoft.CodeAnalysis.CSharp.Symbols.Retargeting;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.CSharp.Test.Utilities;
+using Microsoft.CodeAnalysis.CSharp.UnitTests;
+using Microsoft.CodeAnalysis.Test.Utilities;
+using Microsoft.DiaSymReader.PortablePdb;
```
cston, line 13: >Microsoft.DiaSymReader.PortablePdb [](http://example.com/codeflow?start=6&length=34)

Is this used?

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -0,0 +1,2536 @@
...
+              IL_0018:  ldc.i4.1
+              IL_0019:  box        "int"
+              IL_001e:  call       "void Handler.AppendFormatted(object)"
+              IL_0023:  ret
+            }
+            """);
+
+        var comp = CreateCompilation([handler, OverloadResolutionPriorityAttributeDefinition, InterpolatedStringHandlerAttribute]);
+        CompileAndVerify(executable, references: [AsReference(comp, useMetadataReference)], expectedOutput: "12").VerifyDiagnostics();
+    }
```
cston, line 2029: >} [](http://example.com/codeflow?start=4&length=1)

Consider testing in object initializers:
```csharp
class C : IEnumerable<int>
{
    private List<int> _list = new();
    public void Add(int x) { throw null; }
    [OverloadResolutionPriority(1)] public void Add(int x, int y = 0) { _list.Add(x); }
    public IEnumerator<int> GetEnumerator() => _list.GetEnumerator();
    IEnumerator IEnumerable.GetEnumerator() => null;
}

class Program
{
    static void Main()
    {
        C c = new() { 1 };
        foreach (var i in c) Console.WriteLine(i);
        c = [2];
        foreach (var i in c) Console.WriteLine(i);
    }
}
```

cston (dismissed):

jcouv (dismissed):
LGTM Thanks (iteration 24)

jcouv (approved):
LGTM Thanks (iteration 26)

cston (approved):

PR #74224, opened 2024-07-01
Report a diagnostic on missing body in partial property implementation

### LINKED ISSUE #74221: Partial properties needs to enforce that all implementation part accessors have bodies
[SharpLab](https://sharplab.io/#v2:EYLgtghglgdgPgBwgJwC5QgGwAQAEBM2AwtgN4CwAUNngMzZJoY6yrYAKyA9gmdgOYBTVAG5sAZ2FiAvlVmUqjdFjyESFanQYplLGG048+QtgF4AfNgCMYyaOzzpQA==)
```cs
partial class C {
  public partial int Prop { get; set; }
}

partial class C {
  public partial int Prop { get => 1; set; }
}
```

Expected behavior: error is reported on 'set' accessor of implementation part.
Actual behavior: NRE in emit layer

Relates to test plan https://github.com/dotnet/roslyn/issues/73090

RikkiGibson:
Closes #74221

### DIFF
--- a/src/Compilers/CSharp/Test/Symbol/Symbols/PartialPropertiesTests.cs
+++ b/src/Compilers/CSharp/Test/Symbol/Symbols/PartialPropertiesTests.cs
@@ -4976,7 +4976,7 @@ public partial class C
             Assert.Equal("SourceFile(Program.cs[52..53))", defSymbol.Locations.Single().ToString());
             Assert.Equal("SourceFile(Program.cs[97..98))", implSymbol.Locations.Single().ToString());
         }
-        
+
         [Fact]
         public void OnlyOneAccessorHasBodyOnImplementation()
         {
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Test/Symbol/Symbols/PartialPropertiesTests.cs
```diff
@@ -4973,5 +4976,29 @@ partial class C
...
         }
+        
+        [Fact]
+        public void OnlyOneAccessorHasBodyOnImplementation()
+        {
+            var source = """
+                partial class C
+                {
+                    public partial int Prop1 { get; set; }
+                    public partial int Prop1 { get => 1; set; }
```
RikkiGibson, line 4987: Note that we already have coverage for when a partial property has no accessors with bodies. In that case, the language considers it to be a definition part, and we don't want to report errors on the missing bodies.

333fred (approved):

RikkiGibson:
@dotnet/roslyn-compiler for second review of a small fix

jcouv (approved):
LGTM Thanks (iteration 2)

PR #74226, opened 2024-07-01
Dispose various locations using PEReader

ToddGrun:
Attempt to address https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2104770

Multiple hdmps gathered for this bug indicate a leak of native memory in NativeHeapMemoryBlock+DisposableData. Tracking this backwards through code leads to a PEReader leak, but I wasn't able to keep tracking back from the data in the dmps. Looking at the PEReader objects in the dmp, I was able to infer that the PEStreamOptions.PrefetchEntireImage option was set when creating the PEReader.

This led to the most likely culprit, CSharpDecompilationService.PerformDecompilation not disposing the PEFile it owns.

Additionally, I noticed CompilationOutputs.TryCopyPdbToAsync wasn't disposing a PEReader it owned through the DebugInformationReaderProvider it owns from the call to OpenPdb.

### DIFF
--- a/src/Features/Core/Portable/Emit/CompilationOutputs.cs
+++ b/src/Features/Core/Portable/Emit/CompilationOutputs.cs
@@ -165,7 +165,7 @@ internal abstract class CompilationOutputs
 
     internal async ValueTask<bool> TryCopyPdbToAsync(Stream stream, CancellationToken cancellationToken)
     {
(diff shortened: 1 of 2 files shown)


ToddGrun:
@tmat -- ptal

PR #74227, opened 2024-07-01
Escape `record` keywords

SteveDunn:
See issue #74117 where [it was decided ](https://github.com/dotnet/roslyn/issues/74117#issuecomment-2192770534)that records should be escaped, even though they're not keywords, as it is the safer thing to do.

See other PR #74125

### DIFF
--- a/src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
+++ b/src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
@@ -103,10 +103,11 @@ namespace Microsoft.CodeAnalysis.CSharp
                 return new SymbolDisplayPart(kind, symbol, text);
             }
 
-            bool escapeRecord = StringComparer.Ordinal.Equals(text, "record") && symbol?.Kind is SymbolKind.NamedType or SymbolKind.Alias;
+            bool isNamedTypeOrAliasName = symbol?.Kind is SymbolKind.NamedType or SymbolKind.Alias;
+            
             bool isEscapable = IsEscapable(kind);
 
-            text = isEscapable ? EscapeIdentifier(text, escapeRecord) : text;
+            text = isEscapable ? EscapeIdentifier(text, isNamedTypeOrAliasName) : text;
 
             return new SymbolDisplayPart(kind, symbol, text);
         }
@@ -136,12 +137,13 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
         }
 
-        private static string EscapeIdentifier(string identifier, bool escapeRecord)
+        private static string EscapeIdentifier(string identifier, bool isNamedTypeOrAliasName)
         {
             SyntaxKind kind = SyntaxFacts.GetKeywordKind(identifier);
-            if (kind == SyntaxKind.None && escapeRecord)
+            
+            if (kind is SyntaxKind.None && isNamedTypeOrAliasName && StringComparer.Ordinal.Equals(identifier, "record"))
             {
-                return $"@{identifier}";
+                kind = SyntaxKind.RecordKeyword;
             }
 
             return kind == SyntaxKind.None ? identifier : $"@{identifier}";
--- a/src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
+++ b/src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
@@ -103,11 +103,10 @@ namespace Microsoft.CodeAnalysis.CSharp
                 return new SymbolDisplayPart(kind, symbol, text);
             }
 
-            bool isNamedTypeOrAliasName = symbol?.Kind is SymbolKind.NamedType or SymbolKind.Alias;
-
-            bool isEscapable = IsEscapable(kind);
-
-            text = isEscapable ? EscapeIdentifier(text, isNamedTypeOrAliasName) : text;
+            if (IsEscapable(kind))
+            {
+                text = EscapeIdentifier(text, symbol?.Kind is SymbolKind.NamedType or SymbolKind.Alias);
+            }
 
             return new SymbolDisplayPart(kind, symbol, text);
         }
--- a/src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
+++ b/src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
@@ -104,7 +104,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
 
             bool isNamedTypeOrAliasName = symbol?.Kind is SymbolKind.NamedType or SymbolKind.Alias;
-            
+
             bool isEscapable = IsEscapable(kind);
 
             text = isEscapable ? EscapeIdentifier(text, isNamedTypeOrAliasName) : text;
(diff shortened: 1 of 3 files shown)


src/Compilers/CSharp/Portable/Syntax/SyntaxKindFacts.cs
```diff
@@ -991,6 +991,8 @@ public static SyntaxKind GetKeywordKind(string text)
                     return SyntaxKind.UsingKeyword;
                 case "class":
                     return SyntaxKind.ClassKeyword;
+                case "record":
+                    return SyntaxKind.RecordKeyword;
```
333fred, line 995: We can't make this appear as keyword kind. We likely need to special case it in the visitor.

src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
```diff
@@ -127,7 +127,7 @@ private static bool IsEscapable(SymbolDisplayPartKind kind)
         private static string EscapeIdentifier(string identifier)
         {
             var kind = SyntaxFacts.GetKeywordKind(identifier);
-            return kind == SyntaxKind.None
+            return kind == SyntaxKind.None && !StringComparer.Ordinal.Equals(identifier, "record")
```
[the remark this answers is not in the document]
333fred: Almost. I think we likely want this, but we also want to only escape some kinds. Let's add a parallel to IsEscapable specifically for records; we only want to escape type names, as that's all we warn on.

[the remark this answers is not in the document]
SteveDunn: > Almost. I think we likely want this, but we also want to only escape some kinds. Let's add a parallel to IsEscapable specifically for records; we only want to escape type names, as that's all we warn on. We can then pass that flag into `EscapeIdentifier` and adjust this second half of the condition appropriately.

Hi Fred, when you say 'we only want to escape type names': I can't see that in the list of `SymbolDisplayPartKind` enum. What about `TypeParameterName`?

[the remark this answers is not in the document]
333fred: You have the symbol itself, just use that.

src/Compilers/CSharp/Test/Symbol/SymbolDisplay/SymbolDisplayTests.cs
```diff
@@ -1093,6 +1093,81 @@ class @true {
...
+        [Fact]
+        public void TestEscapeRecordKeywordIdentifiers_DoesNotEscapesMethodNames()
+        {
+            var text = @"
+class Foo {
+    Foo record() { return default; } }
+";
+
+            Func<NamespaceSymbol, Symbol> findSymbol = global =>
+                global.GetTypeMembers("Foo", 0).Single().
```
jaredpar, line 1151: Please avoid _Foo_ in the code base.

src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
```diff
@@ -93,8 +93,21 @@ protected override void FreeNotFirstVisitor(AbstractSymbolDisplayVisitor visitor
 
         internal SymbolDisplayPart CreatePart(SymbolDisplayPartKind kind, ISymbol? symbol, string text)
         {
-            text = (text == null) ? "?" :
-                   (_escapeKeywordIdentifiers && IsEscapable(kind)) ? EscapeIdentifier(text) : text;
+            if (text == null)
+            {
+
```
jaredpar, line 98:
```suggestion
```

src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
```diff
@@ -93,8 +93,21 @@ protected override void FreeNotFirstVisitor(AbstractSymbolDisplayVisitor visitor
...
+                return new SymbolDisplayPart(kind, symbol, "?");
+            }
+
+            if (!_escapeKeywordIdentifiers)
+            {
+                return new SymbolDisplayPart(kind, symbol, text);
+            }
+
+            bool escapeRecord = StringComparer.Ordinal.Equals(text, "record") && symbol?.Kind is SymbolKind.NamedType or SymbolKind.Alias;
```
333fred, line 107: It feels like the actual text comparison is better in `EscapeIdentifier`. I would simply pass along a flag covering whether the text is a namedtype or alias name, and let `EscapeIdentifier` handle actually looking at the text.

333fred (approved):

src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
```diff
@@ -93,8 +93,21 @@ protected override void FreeNotFirstVisitor(AbstractSymbolDisplayVisitor visitor
...
+            {
+                return new SymbolDisplayPart(kind, symbol, "?");
+            }
+
+            if (!_escapeKeywordIdentifiers)
+            {
+                return new SymbolDisplayPart(kind, symbol, text);
+            }
+
+            bool isNamedTypeOrAliasName = symbol?.Kind is SymbolKind.NamedType or SymbolKind.Alias;
```
cston, line 106: >SymbolKind.Alias [](http://example.com/codeflow?start=82&length=16)

Are we testing the alias case? (See `SymbolDisplayTests.TestAlias1` for an example of an alias test.)

src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
```diff
@@ -93,8 +93,21 @@ protected override void FreeNotFirstVisitor(AbstractSymbolDisplayVisitor visitor
...
+
+            bool isEscapable = IsEscapable(kind);
+
+            text = isEscapable ? EscapeIdentifier(text, isNamedTypeOrAliasName) : text;
```
cston, line 110: Consider rearranging to avoid calculating `isNamedTypeOrAliasName` or reassigning `text` unnecessarily. Perhaps:
```csharp
if (_escapeKeywordIdentifiers && IsEscapable(kind))
{
    text = EscapeIdentifer(text, symbol?.Kind is SymbolKind.NamedType or SymbolKind.Alias);
}
```

SteveDunn: Actually, looking at the previous line, we return if `!escapeKeywordIdentifer` - so we can omit that check here.

src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
SteveDunn: Thanks,  I have added a test for an alias. If `Goo` the new `Foo`? :)

src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.cs
cston: Yes, we're checking `!escapeKeywordIdentifiers` earlier but the two `if` statements could be combined to avoid two `return new SymbolDisplayPart(kind, symbol, text);`.

333fred (approved):

PR #70913, opened 2023-11-21
Pass project info to the Razor extension in VS Code

davidwengier:
Firstly, let me start by saying: This was not the code I expected to work on today 😁

The idea is to bring https://github.com/dotnet/razor/pull/9601 to VS Code. Today, the Razor language server monitors the entire workspace folder for `project.razor.vscode.bin` files, and gets ready to answer questions about those projects. At best this is wasteful, at worst it causes issues because it will find old `obj` folders and use them, say for example when people change from .net 7 to .net 8.

Sadly this was not as easy as I hoped, as the project loading in VS Code, with and without DevKit, was not up to the task for various reasons. I haven't changed any public API, but if these changes go too far then feel free to let me know, and I can just close the PR and leave things how they are.

Some comments below, some minor commits in commit messages. It's mostly plumbing.

Client side: https://github.com/dotnet/vscode-csharp/pull/6688

PR #74229, opened 2024-07-01
Set intermediate path sooner

ryzngard:
Pulled from https://github.com/dotnet/roslyn/pull/70913 (thanks @davidwengier!)

Razor uses the intermediate path for project key. Now that we're removing the file system from being involved in communicating project data, we need to make sure we're communicating the right data. 

This fixes issues found in https://github.com/dotnet/vscode-csharp/pull/7273 and will unblock that insertion


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/WorkspaceProject.cs
```diff
@@ -170,8 +170,7 @@ public async Task SetBuildSystemPropertiesAsync(IReadOnlyDictionary<string, stri
             }
         }
 
-        // Workaround for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1830960
-        _project.CompilationOutputAssemblyFilePath = _project.OutputFilePath;
+        _project.CompilationOutputAssemblyFilePath = _project.CompilationOutputAssemblyFilePath;
```
ryzngard, line 173: Bug has been resolved so can use correct information from devkit

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProjectFactory.cs
```diff
@@ -107,7 +108,7 @@ public async Task<ProjectSystemProject> CreateAndAddToWorkspaceAsync(string proj
                 name: projectSystemName,
                 assemblyName,
                 language,
-                compilationOutputFilePaths: default, // will be updated when command line is set
+                compilationOutputInfo: new(creationInfo.CompilationOutputAssemblyFilePath), // will be updated when command line is set
```
ryzngard, line 111: I'm not sure about the comment after, but anecdotally this can be set here. If it gets updated later I believe that's fine?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/WorkspaceProject.cs
davidwengier: This line doesn't do anything.

Also this method seems to be missing the case where it actually sets the `CompilationOutputAssemblyFilePath` info it gets from the underlying project system: https://github.com/dotnet/roslyn/pull/70913/files#diff-af31a9956c1c1a4a99e36eb99ec37faa9b547e01645b4b9a36c4abc24e0f1476R149

davidwengier: Also missing this bug fix: https://github.com/dotnet/roslyn/pull/70913/files#r1400060227

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProjectFactory.cs
davidwengier: I read the comment as "its okay to be null as teh value is filled in later", but since this fills in the value now, I'd remove the comment.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/WorkspaceProject.cs
dibarbet: this is reassigning to itself, do you need to get the value from somewhere else (is it supposed to come from the properties parameter)?

ryzngard: Oh woops, yup. Need to have `case "IntermediateAssembly": _project.CompilationOutputAssemblyFilePath = GetFullyQualifiedPath(valueOrNull); break;`

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProjectFactory.cs
ryzngard, line 111:
```suggestion
                compilationOutputInfo: new(creationInfo.CompilationOutputAssemblyFilePath),
```

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/WorkspaceProject.cs
```diff
@@ -160,6 +160,7 @@ public async Task SetBuildSystemPropertiesAsync(IReadOnlyDictionary<string, stri
             switch (name)
             {
                 case "AssemblyName": _project.AssemblyName = value; break;
+                case "IntermediateAssembly": _project.CompilationOutputAssemblyFilePath = GetFullyQualifiedPath(valueOrNull); break;
```
davidwengier, line 163: `GetFullyQualifiedPath` has a bug, where it won't work for a relative path, because it does `Path.Combine(_project.FilePath, value)`, but `_project.FilePath` is the full path to the actual project file (ie, `.csproj` file). When `value` is an absolute path, then that will be returned and all will be good, but the intermediate output path is a relative path, so we'd end up with `CompilationOutputAssemblyFilePath` being something like `C:\Goo\Bar.csproj\obj\Debug\net8.0` which is not correct.

ryzngard: Ah I see, that was your comment about why it wasn't completely broken. The paths are coming in as absolute path so `Path.Combine` is just using the second parameter. I think what I did will fix the issue

davidwengier (approved):

PR #74232, opened 2024-07-02
Switch implicit span to conversion from type

jjonescz:
Corresponding speclet update: https://github.com/dotnet/csharplang/pull/8241


src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
```diff
@@ -3931,8 +3934,8 @@ private static bool IsIntegerTypeSupportingPointerConversions(TypeSymbol type)
 #nullable enable
         private bool HasImplicitSpanConversion(TypeSymbol? source, TypeSymbol destination, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
         {
-            // PROTOTYPE: Is it fine that this conversion does not exists when Compilation is null?
-            if (Compilation?.IsFeatureEnabled(MessageID.IDS_FeatureFirstClassSpan) != true)
+            // Note: when Compilation is null, we assume latest LangVersion.
+            if (Compilation?.IsFeatureEnabled(MessageID.IDS_FeatureFirstClassSpan) == false)
```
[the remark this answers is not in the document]
jjonescz: So far I've been able to hit only one scenario with Compilation being null - from `IMethodSymbol.ReduceExtensionMethod` public API - see e.g., test `Conversion_Array_Span_ExtensionMethodReceiver_Implicit_Reduced_01`. And the documentation says we should consider the latest language version in that case (it talks about constraints but it seems consistent to do for other features as well):

```cs
/// <param name="compilation">The compilation in which constraints should be checked.
/// Should not be null, but if it is null we treat constraints as we would in the latest
/// language version.</param>
```

Maybe other scenarios (not just reduced extension methods) should behave similarly? But there might be no more scenarios, I have investigated callsites for a while and haven't found any that I could hit.

jjonescz:

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -486,18 +487,61 @@ void checkConstraintLanguageVersionAndRuntimeSupportForConversion(SyntaxNode syn
...
+                            syntax,
+                            destination.OriginalDefinition,
+                            WellKnownMemberNames.ImplicitConversionName);
+                    }
+                }
+            }
+        }
+
+        internal static MethodSymbol? TryFindImplicitOperatorFromArray(TypeSymbol type)
+        {
```
cston, line 514: >{ [](http://example.com/codeflow?start=8&length=1)

Consider adding `Debug.Assert(type.IsSpan() || type.IsReadOnlySpan());`

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -486,18 +487,61 @@ void checkConstraintLanguageVersionAndRuntimeSupportForConversion(SyntaxNode syn
...
+                {
+                    ParameterCount: 1,
+                    Arity: 0,
+                    IsStatic: true,
+                    DeclaredAccessibility: Accessibility.Public,
+                    Parameters: [{ Type: ArrayTypeSymbol { IsSZArray: true, ElementType: TypeParameterSymbol } }]
+                } method ? method : null);
+        }
+
+        private static T? TryFindSingleMember<T>(TypeSymbol type, string name, Func<Symbol, T?> predicate) where T : class
```
cston, line 526: ><T> [](http://example.com/codeflow?start=45&length=3)

Is the type parameter necessary, or could this be simplified to `TryFindSingleMethod` with `MethodSymbol` directly?

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
```diff
@@ -4003,18 +4005,17 @@ private bool IgnoreUserDefinedSpanConversions(TypeSymbol? source, TypeSymbol? ta
                 return false;
             }
 
-            // PROTOTYPE: Is it fine that this check is not performed when Compilation is null?
-            return Compilation?.IsFeatureEnabled(MessageID.IDS_FeatureFirstClassSpan) == true &&
-                (ignoreUserDefinedSpanConversionsInOneDirection(Compilation, source, target) ||
-                ignoreUserDefinedSpanConversionsInOneDirection(Compilation, target, source));
+            // Note: when Compilation is null, we assume latest LangVersion.
+            return Compilation?.IsFeatureEnabled(MessageID.IDS_FeatureFirstClassSpan) != false &&
```
cston, line 4009: This check is repeated several times, with the same comment. Consider creating a private `IsFirstClassSpansEnabled` property so this is captured in one location.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_Conversion.cs
```diff
@@ -619,28 +619,17 @@ private static bool IsFloatingPointExpressionOfUnknownPrecision(BoundExpression
...
                         }
                         else
                         {
-                            MethodSymbol method = symbol.AsMember(spanType);
+                            MethodSymbol method = methodDefinition.AsMember(spanType);
 
                             rewrittenOperand = _factory.Convert(method.ParameterTypesWithAnnotations[0].Type, rewrittenOperand);
 
-                            if (member == WellKnownMember.System_ReadOnlySpan_T__op_Implicit_Array)
+                            if (_compilation.IsReadOnlySpanType(spanType))
```
cston, line 632: >_compilation.IsReadOnlySpanType(spanType) [](http://example.com/codeflow?start=32&length=41)

This uses `WellKnownMember.System_ReadOnlySpan_T` rather than checking by fully-qualified type name. Is that correct?

cston:
>         CreateCompilationWithSpan(source, parseOptions: TestOptions.Regular12).VerifyDiagnostics(

[This comment is not specific to this PR, so it can be addressed separately.]

`CreateCompilationWithSpan()` is always using the local type definitions for `Span<T>` and `ReadOnlySpan<T>`. On .NET core, we should compile against the BCL implementation instead. Consider using `CreateCompilationWithSpanAndMemoryExtensions()` which covers both cases.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:58 in 4161bd1. [](commit_id = 4161bd1062d6388fc30354d0da6b38d3fa443bf7, deletion_comment = False)

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_Conversion.cs
jjonescz: That was my intent - I even have a test for it: `Conversion_Array_Span_Implicit_ConstantData_NotWellKnownSpan`. The emit of constant data for ReadOnlySpan (i.e., emit of `BoundReadOnlySpanFromArray`) expects the well-known span type and I don't think changing that is worth the effort.

jjonescz:
>         CreateCompilationWithSpan(source, parseOptions: TestOptions.Regular12).VerifyDiagnostics(

Sounds good, thanks! Let me do that after I merge main into the feature branch so I get [https://github.com/dotnet/roslyn/pull/74281](https://github.com/dotnet/roslyn/pull/74281) in.

---
In reply to: [2215489010](https://github.com/dotnet/roslyn/pull/74232#issuecomment-2215489010) [](http://example.com/codeflow?ancestors=2215489010)

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:58 in 4161bd1. [](commit_id = 4161bd1062d6388fc30354d0da6b38d3fa443bf7, deletion_comment = False)

cston (approved):

jjonescz:

333fred (approved):

PR #74250, opened 2024-07-03
Reduce allocations in SymbolDeclaredCompilationEvent

ToddGrun:
The SymbolDeclaredCompilationEvent ctor is showing up as 0.5% of allocations in vbcscompiler.exe in a customer profile I'm looking at. All these costs are attributable to the use of Lazy in this class. Instead, just use a nullable field, and accept the fact that during extremely high contention, it is possible for an ImmutableArray to be created multiple times.

image

### DIFF
--- a/src/Compilers/Core/Portable/DiagnosticAnalyzer/SymbolDeclaredCompilationEvent.cs
+++ b/src/Compilers/Core/Portable/DiagnosticAnalyzer/SymbolDeclaredCompilationEvent.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System;
 using System.Collections.Immutable;
 using Microsoft.CodeAnalysis.Symbols;
 
@@ -15,7 +14,7 @@ namespace Microsoft.CodeAnalysis.Diagnostics
     /// </summary>
     internal sealed class SymbolDeclaredCompilationEvent : CompilationEvent
     {
-        private readonly Lazy<ImmutableArray<SyntaxReference>> _lazyCachedDeclaringReferences;
+        private ImmutableArray<SyntaxReference>? _lazyCachedDeclaringReferences;
 
         public SymbolDeclaredCompilationEvent(
             Compilation compilation,
@@ -25,7 +24,6 @@ namespace Microsoft.CodeAnalysis.Diagnostics
         {
             SymbolInternal = symbolInternal;
             SemanticModelWithCachedBoundNodes = semanticModelWithCachedBoundNodes;
-            _lazyCachedDeclaringReferences = new Lazy<ImmutableArray<SyntaxReference>>(() => Symbol.DeclaringSyntaxReferences);
         }
 
         public ISymbol Symbol => SymbolInternal.GetISymbol();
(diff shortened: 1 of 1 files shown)


jaredpar (dismissed):

src/Compilers/Core/Portable/DiagnosticAnalyzer/SymbolDeclaredCompilationEvent.cs
```diff
@@ -35,7 +33,18 @@ internal sealed class SymbolDeclaredCompilationEvent : CompilationEvent
...
 
         // PERF: We avoid allocations in re-computing syntax references for declared symbol during event processing by caching them directly on this member.
-        public ImmutableArray<SyntaxReference> DeclaringSyntaxReferences => _lazyCachedDeclaringReferences.Value;
+        public ImmutableArray<SyntaxReference> DeclaringSyntaxReferences
+        {
+            get
+            {
+                if (!_lazyCachedDeclaringReferences.HasValue)
+                {
+                    _lazyCachedDeclaringReferences = Symbol.DeclaringSyntaxReferences;
```
AlekseyTs, line 42: >_lazyCachedDeclaringReferences = Symbol.DeclaringSyntaxReferences; [](http://example.com/codeflow?start=20&length=66)

This assignment is not atomic. Why we are using a nullable type at all. Is using ```default(ImmutableArray<SyntaxReference>)``` as an initial value and using interlocked initialize going to work?

AlekseyTs:

src/Compilers/Core/Portable/DiagnosticAnalyzer/SymbolDeclaredCompilationEvent.cs
sharwell: @ToddGrun Specifically, you're going to want to use `InterlockedOperations.Initialize` here with a static lambda callback.

sharwell: I believe the full form would be:

```cs
return InterlockedOperations.Initialize(
  ref _lazyCachedDeclaringReferences,
  static symbol => symbol.DeclaringSyntaxReferences,
  Symbol);
```

sharwell (changes requested):

src/Compilers/Core/Portable/DiagnosticAnalyzer/SymbolDeclaredCompilationEvent.cs
ToddGrun: OK, I've gone ahead with your and Sam's suggestion. I did consider this originally, but wasn't sure whether the Symbol.DeclaringSyntaxReferences could return a default immutable array. But, if that's not a concern, then this is definitely better.

ToddGrun:
@sharwell -- Mostly matched your suggested change (I didn't pass in the Symbol directly, as there's no need to call into that property in the already initialized case)

src/Compilers/Core/Portable/DiagnosticAnalyzer/SymbolDeclaredCompilationEvent.cs
333fred: Even if it could, we would still need to use some kind of sentinel array to compare with. `Nullable<T>` cannot be used in this scenario, period; you could have two threads race and observe a sheared `Nullable<ImmutableArray>` value in thread 2, such that the write to `Nullable.hasValue` is true, but `Nullable.value` has not been written yet.

333fred (approved):

src/Compilers/Core/Portable/DiagnosticAnalyzer/SymbolDeclaredCompilationEvent.cs
ToddGrun: As I wrote in the original description, I didn't consider multiple assignments during periods of high contention as being a deal breaker. However, I'm more than glad to have made the change such that the value won't be calculated twice as that seems to be the desired consensus.

ToddGrun: Fred stopped by and explained what I was missing. Not the first time a torn-write has slipped past me.

sharwell (approved):

AlekseyTs (approved):
LGTM (commit 2)

PR #74251, opened 2024-07-03
Do not include collection expression conversion to nullable value type in ClassifyStandardImplicitConversion()

### LINKED ISSUE #74185: CS7038: Failed to emit module if new collection syntax is used with an implicit ImmutableArray operator.
**Version Used**: 
.Net 8.0.302, also occurs with 8.0.301, I haven't tested other versions.

**Steps to Reproduce**:

[sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEAMACdBGAOgGEB7AGzIFMBjAFyhJgGcBuAWACgtc8AWdrjnwEAMrACOA7sNIUa9RkwIBJMGACutCMCpShhACJQIAcxgkm9aktIJKAQRgQyATyZRWnTugDMuAEzYRGQQTEx4nADenNixuH6kMFqwlABOADwkwABW8gB8QamUELSURqbmllDUABQAlDFx0RxxrbgA7NgA2jCUAO7YWbl09QC6Aq0AvpzTHN5+6IGJyb0ZACoFIDw+6RtRjbEADqlQAG4llNhFEAiMrtu7BQD6UKVgnnOfrb5BjCtp9WwW1oAAsPDUAKIwdRgNLaKgECFgQ60FyPeoNFpNA7YWatHE/ZbQVY1fA7DbYV6Ud51bAAXmeVPe9MpbyUaxI9lSqQgLnqAgJfmUUJhcNoJHWBQA4pRaCLYTzxalAQzsC82QQZXLoQqShL+V4sbFhTqxRLtprZfKzcraaqtdbFfq6gKjd10O0tgA5Ei0L3qCjKABmPr9AbINQAREymJG6qNBTwAGyU5FkaqvQaHG2/JLEtKPAD8NVg2FUGi0Ogc3N5RdZ1KYdoKMewTD6r2oIOwzTacXMtGw0Io9IKvQGRJSkpLbLqcBxrSevdiqqHZBxkwEsyAA===)

The error occurs locally with .Net 8.0.302 of the sdk and on sharplab with main (11 Jun 2024), if I choose "Default", the error becomes CS1061 instead of CS7038

**Diagnostic Id**: CS7038: Failed to emit module

**Expected Behavior**: CS1061

**Actual Behavior**: Outputs the cryptic error CS7038: Failed to emit module

The code is based on the [omnisharp container](https://github.com/OmniSharp/csharp-language-server-protocol/blob/a9688416c957caf26645621712333614901bb68f/src/Protocol/Models/Container.cs#L9), that's where I encountered the error. I reduced it to the minimal version where I was still able to get the error, the implicit operator from ImmutableArray seems to be important for this. The original container class has a bunch of other operators, but they don't seem related to the error.

cston:
Do not include collection expression conversions to nullable value types in `ClassifyStandardImplicitConversion()`.

User-defined conversions should not be considered in collection expression conversions. And it looks like the scenarios were not supported previously, even though `ClassifyStandardImplicitConversion()` allowed user-defined conversions *from nullable value types*, because `Binder.ConvertCollectionExpression()` created a `BoundCollectionExpression` with `HasErrors=true` when the conversion was a user-defined conversion rather than a collection expression conversion.

In some cases, such as the repro from #74185 with 17.8, a specific error was reported for the target type: `CS1061: 'Container<object>' does not contain a definition for 'Add'`

In other cases, no diagnostic was reported from binding, but the invalid `BoundCollectionExpression` would cause emit to fail with a general error: `CS7038: Failed to emit module: Unable to determine specific cause of the failure.`

Now, the user-defined conversion is ignored in `ClassifyStandardImplicitConversion()`, and an error should be reported for the target type in all cases.

Fixes #74185

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
@@ -669,7 +669,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             Conversion conversion = ClassifyImplicitBuiltInConversionFromExpression(sourceExpression, source, destination, ref useSiteInfo);
             if (conversion.Exists &&
                 !conversion.IsInterpolatedStringHandler &&
-                !conversion.IsCollectionExpression)
+                !isImplicitCollectionExpressionConversion(conversion))
             {
                 Debug.Assert(IsStandardImplicitConversionFromExpression(conversion.Kind));
                 return conversion;
@@ -681,6 +681,16 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
 
             return Conversion.NoConversion;
+
+            static bool isImplicitCollectionExpressionConversion(Conversion conversion)
+            {
+                return conversion switch
+                {
+                    { Kind: ConversionKind.CollectionExpression } => true,
+                    { Kind: ConversionKind.ImplicitNullable, UnderlyingConversions: [{ Kind: ConversionKind.CollectionExpression }] } => true,
+                    _ => false,
+                };
+            }
         }
 
         private Conversion ClassifyStandardImplicitConversion(TypeSymbol source, TypeSymbol destination, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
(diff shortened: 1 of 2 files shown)


AlekseyTs (approved):
LGTM (commit 1)

333fred (approved):

PR #74258, opened 2024-07-03
Reduce closures allocated during invocation of CapturedSymbolReplacement.Replacement

ToddGrun:
This method previously took in a Func<NamedTypeSymbol, BoundExpression> and most callers allocated a closure to implement the callback. This PR uses the args pattern and adds a parameter to the Replacement method that is passed back into that callback to better allow static lambdas to be used.

This was the top allocating closure allocation in vbcscompiler in a customer profile I'm looking at (0.2% -- 42 MB)

image

### DIFF
--- a/src/Compilers/CSharp/Portable/Lowering/ClosureConversion/ClosureConversion.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/ClosureConversion/ClosureConversion.cs
@@ -624,12 +624,8 @@ namespace Microsoft.CodeAnalysis.CSharp
                 // frame pointers are organized in a linked list.
                 return proxyField.Replacement(
                     syntax,
-                    static (frameType, arg) =>
-                    {
-                        var (syntax, @this) = arg;
-                        return @this.FramePointer(syntax, frameType);
-                    },
-                    (syntax, this));
+                    static (frameType, arg) => arg.self.FramePointer(arg.syntax, frameType),
+                    (syntax, self: this));
             }
 
             var localFrame = (LocalSymbol)framePointer;
@@ -786,11 +782,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                 var left = proxy.Replacement(
                     syntax,
-                    static (frameType1, arg) =>
-                    {
-                        var (syntax, framePointer) = arg;
-                        return new BoundLocal(syntax, framePointer, null, framePointer.Type);
-                    },
+                    static (frameType1, arg) => new BoundLocal(arg.syntax, arg.framePointer, null, arg.framePointer.Type),
                     (syntax, framePointer));
 
                 var assignToProxy = new BoundAssignmentOperator(syntax, left, value, value.Type);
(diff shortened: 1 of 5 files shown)


src/Compilers/CSharp/Portable/Lowering/ClosureConversion/ClosureConversion.cs
```diff
@@ -622,7 +622,14 @@ protected override BoundExpression FramePointer(SyntaxNode syntax, NamedTypeSymb
...
                 // frame pointers are organized in a linked list.
-                return proxyField.Replacement(syntax, frameType => FramePointer(syntax, frameType));
+                return proxyField.Replacement(
+                    syntax,
+                    static (frameType, arg) =>
+                    {
+                        var (syntax, @this) = arg;
+                        return @this.FramePointer(syntax, frameType);
+                    },
+                    (syntax, this));
```
sharwell, line 632: It seems like many of these could be significantly more succinct. Feels like we're "paying more than we need to" for the gains.

```suggestion
                    static (frameType, arg) => arg.self.FramePointer(arg.syntax, frameType),
                    (syntax, self: this));
```

ToddGrun: Sure, I can switch to that form.

src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/CapturedSymbol.cs
```diff
@@ -23,7 +23,7 @@ public CapturedSymbolReplacement(bool isReusable)
         /// Rewrite the replacement expression for the hoisted local so all synthesized field are accessed as members
         /// of the appropriate frame.
         /// </summary>
-        public abstract BoundExpression Replacement(SyntaxNode node, Func<NamedTypeSymbol, BoundExpression> makeFrame);
+        public abstract BoundExpression Replacement<TArg>(SyntaxNode node, Func<NamedTypeSymbol, TArg, BoundExpression> makeFrame, TArg arg);
```
333fred, line 26: Would really prefer if these put `arg` before `makeFrame`.

ToddGrun:
> Not going to block on it, but would prefer if we made the args more developer-friendly.

I'm flexible on this, although it's a bit unfortunate for the compilers and editor folks to have different orderings. How about, I'll go with whatever the other compiler dev indicates is there desire for the ordering?

333fred:
> I'm flexible on this, although it's a bit unfortunate for the compilers and editor folks to have different orderings.

We keep procrastinating on fixing the ordering across the codebase. I'm just going to submit a PR to do it.

sharwell:
@333fred another problems is the lambda being first is the way public APIs are written in the .NET libraries. It might be better to fix the experience for that than to make up a different pattern from everyone else.

AlekseyTs (approved):
LGTM (commit 2)

ToddGrun:
OK, going to merge with this parameter ordering for now.

PR #74267, opened 2024-07-03
Improve performance of SourceText.LineInfo indexer.

ToddGrun:
Saw something new in a customer performance profile, a *ton* of time spent in the LSP DidChangeHandler. It's uncertain, but my guess is the customer did a Find/Replace that ended up doing a huge number of edits within the current file. The way the Roslyn didChange handler currently works, it takes each of those edits and creates a new SourceText. (This is something I'm going to pursue in a separate PR).

Each of those new source texts ends up calling into ChangedText.GetLinesCore, and subsequently the LineInfo indexer, during the didChangeHandler. This is the method that ends up showing in the profile (23.5 seconds of CPU time in the trace, which was 40% of CPU time in the VS process)

The change here is to allow SourceText a way to create TextLine's with a known extent, avoiding the overhead involved with calling the LineInfo indexer.

image

### DIFF
--- a/src/Compilers/Core/Portable/Text/TextLine.cs
+++ b/src/Compilers/Core/Portable/Text/TextLine.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Diagnostics;
 using Roslyn.Utilities;
 
 namespace Microsoft.CodeAnalysis.Text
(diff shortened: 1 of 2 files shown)


src/Compilers/Core/Portable/Text/TextLine.cs
```diff
@@ -81,6 +81,14 @@ public static TextLine FromSpan(SourceText text, TextSpan span)
             }
         }
 
+        // Do not use unless you are certain the span you are passing in is valid!
+        // This was added to allow SourceText's indexer to directly create TextLines
+        // without the performance implications of calling FromSpan.
+        internal static TextLine FromKnownSpan(SourceText text, TextSpan span)
```
333fred, line 87: Can we add an assert to verify the span in debug mode?

ToddGrun: Good idea, will do!

ToddGrun:
@dotnet/roslyn-compiler -- ptal

333fred (approved):
Generally LGTM, but I do also like Jared's suggestion of adding `Unsafe` to convey the danger of misusing the method.

src/Compilers/Core/Portable/Text/TextLine.cs
ToddGrun: Seems like a good idea, changed.

ToddGrun:
@jaredpar  -- any remaining concerns / requests?

ToddGrun:

jjonescz (approved):

PR #74294, opened 2024-07-08
Fix SourceCodeKind setting

tmat:
For documents that have ParseOptions the SourceCodeKind should be set on the options and also on DocumentAttributes, for other documents it should be set on just the attributes.

### DIFF
--- a/src/Workspaces/Core/Portable/Workspace/Solution/DocumentState.cs
+++ b/src/Workspaces/Core/Portable/Workspace/Solution/DocumentState.cs
@@ -29,10 +29,10 @@ internal partial class DocumentState : TextDocumentState
 
     // properties inherited from the containing project:
     public LanguageServices LanguageServices { get; }
-    private readonly ParseOptions? _options;
+    public ParseOptions? ParseOptions { get; }
 
     // null if the document doesn't support syntax trees:
-    private readonly ITreeAndVersionSource? _treeSource;
+    public ITreeAndVersionSource? TreeSource { get; }
 
     private ImmutableArray<byte> _contentHash;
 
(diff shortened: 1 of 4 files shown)


src/Workspaces/Core/Portable/Workspace/Solution/DocumentState.cs
```diff
@@ -29,10 +29,10 @@ internal partial class DocumentState : TextDocumentState
 
     // properties inherited from the containing project:
     public LanguageServices LanguageServices { get; }
-    private readonly ParseOptions? _options;
+    public ParseOptions? ParseOptions { get; }
```
tmat, line 32: Simplify.

tmat:
@dotnet/roslyn-ide ptal

src/Workspaces/Core/Portable/Workspace/Solution/DocumentState.cs
```diff
@@ -397,12 +377,17 @@ private DocumentState SetParseOptions(ParseOptions options, bool onlyPreprocesso
...
             return this;
         }
 
-        return this.SetParseOptions(this.ParseOptions.WithKind(kind), onlyPreprocessorDirectiveChange: false);
+        if (ParseOptions != null)
+        {
+            return UpdateParseOptionsAndSourceCodeKind(ParseOptions.WithKind(kind), onlyPreprocessorDirectiveChange: false);
+        }
+
+        return UpdateAttributes(Attributes.With(sourceCodeKind: kind));
```
ToddGrun, line 390: >return UpdateAttributes(Attributes.With(sourceCodeKind: kind)); [](http://example.com/codeflow?start=8&length=63)

This doesn't need to happen if the ParseOptions are set?

ToddGrun: Oh, I see. UpdateParseOptionsAndSourceCodeKind does it when it creates the new DocumentState.

ToddGrun (approved):
:shipit:

PR #74299, opened 2024-07-09
Disable BuildWithNetFrameworkHostedCompiler

jjonescz:


jjonescz:
Note: DartLab build succeeds which demonstrates this works - previously DartLab builds were failing with `error NU1010: The PackageReference items Microsoft.Net.Compilers.Toolset.Framework do not have corresponding PackageVersion.`. The DartLab CI check is red because some tests are failing, might be flaky integration tests or something the IDE team should look at.

arkalyanms (approved):

AmadeusW:
Just wanted to add that I've been running into the same NU1010 issue in all of my local development environments (not DartLab). Applying this fix allows me to restore and build roslyn.

PR #74305, opened 2024-07-09
Avoid using constants for large string literals

jjonescz:
Follow up on https://github.com/dotnet/roslyn/pull/74281. Fixes:

> CSC : error CS8103: Combined length of user strings used by the program exceeds allowed limit. Try to decrease use of string literals. [D:\a\_work\1\s\src\Compilers\CSharp\Test\Emit2\Microsoft.CodeAnalysis.CSharp.Emit2.UnitTests.csproj::TargetFramework=net9.0]

### DIFF
--- a/src/Compilers/Test/Utilities/CSharp/TestSources.cs
+++ b/src/Compilers/Test/Utilities/CSharp/TestSources.cs
@@ -8,7 +8,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Test.Utilities
 {
     internal static class TestSources
     {
-        internal const string Span = @"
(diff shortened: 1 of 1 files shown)


jaredpar (approved):

AlekseyTs (approved):

AlekseyTs:
I think it would be good to check for other constant fields that our unit-test projects can reference from our shared test libraries.

jjonescz:
> I think it would be good to check for other constant fields that our unit-test projects can reference from our shared test libraries.

https://github.com/dotnet/roslyn/pull/74307

PR #74313, opened 2024-07-09
Extension type erasure

jcouv:
# Erasure
The erasure is in `PEModuleBuilder.Translate`. 

Here are some cases where we need to erase:
- top-level: `E` => `Underlying`
- type arguments: `C<E>` => `C<Underlying>` (or similar cases with arrays and pointer types)

Extensions are erased by default, but in various scenarios we need to keep extensions:
1. When serializing the type for purpose of the extension erasure attribute, no extensions should be erased.
2. Extension types in containing type position should never be erased (`E.Nested`)

One caveat is that CCI translation is lazy. If you translate `E` with erasure, you get `Underlying`. But if you translate `C<E>` with erasure, we'll just return `C<E>`. The nested extension type will be erased at a later step (when `GetGenericArguments` is called). As a result, we rely on context to carry the information of whether further translations should erase or keep extensions. 

Note: because we ran into some limitations with this approach, we're planning to encode the "keepExtensions" flag into CCI adapters instead. This is described in test `TypeReference_Modopt_ExtensionAsTypeArgument` but will be a later PR.

# Synthesizing the erasure attribute
In `Compilation_WellKnownMembers.cs`, we add `SynthesizeExtensionErasureAttribute` which synthesizes an attribute with a TypedConstant of type `string` but holding a `TypeSymbol` value.
`SynthesizeExtensionErasureAttribute` is used in `MethodSymbol.cs` (for return type) and in `SourceParameterSymbolBase.cs` (for parameters).

The attribute is only synthesized when there is an extension that needs erasure. This is handled by `TypeSymbolExtensions.ContainsErasedExtensionType` and some updates to the `TypeSymbolExtensions.VisitType` infrastructure (we're adding the `asContainer` parameter and switching to use named delegate types `TypePredicate` and `TypeWithAnnotationsPredicate`). 

That `TypedConstant` is converted to a `MetadataSerializedType` in `AttributeDataAdapter.CreateSerializedType` (similar handling to `typeof` constants, which are converted to `MetadataTypeOf`).
We use that `MetadataSerializedType` in two places:
1. in `MetadataWriter.SerializeMetadataExpression`, to be serialized into a string
2. in `MetadataVisitor.Visit(MetadataSerializedType)` to track references/dependencies (part of the `MetadataVisitor`/`ReferenceIndexerBase`/`ReferenceIndexer` machinery).

## Serializing the type to a string
The type serialization is handled by `TypeNameSerializer` and it is extended to allow encoding references to type parameters. We add the `allowTypeParameters` parameter, which is only used for encoding extension erasure attributes.

## Keeping track of references from the attribute
`ReferenceIndexerBase` needs to index the references both for the underlying type and the unerased type (used in attribute with typeof-like format). So `ReferenceIndexerBase.VisitTypeReference` is extended to track both of those separately (see change to `IReferenceOrISignature`).
For example, if the type is `C<E>`, we'll visit the `C<E>` type reference as part of the emitted type (`C<Underlying>`) and as part of the erasure attribute (as `C<E>` unerased).

# Deserializing from string back to type

The type deserialization is handled by `TypeNameDecoder` and it is also extended to allow decoding references to type parameters. We add the `allowTypeParameters` parameter, which is only used for decoding extension erasure attributes.
In `TypeNameDecoder` we therefore need the ability to lookup type parameters by their index given the current context (current method/type). This capability is pulled up in the type hierarchy (see `TypeNameDecoder.GetGenericTypeParamSymbol` and `TypeNameDecoder.GetGenericMethodTypeParamSymbol`).

The actual decoding from `string` to `TypeSymbol` is in `PEParameterSymbol` which calls `PEModuleSymbol.TryDecodeExtensionErasureAttribute`.

# Explicit usage of the attribute

It is disallowed

# Unbounded erasures

Some erasures would be unbounded. For example, `extension E for C<E>`. We add detection for that for source scenarios.
PROTOTYPE comments capture follow-ups for PE/retargeting scenarios.

# Out-of-scope
The erasure, attribute synthesis and decoding are not handled in other locations (field, constraints, etc).
The custom modifiers referring to extensions are not handled (see comment in test `TypeReference_Modopt_ExtensionAsTypeArgument`).
The encoding of tuple names, dynamic, nullability annotations, etc is not handled.
Generating the attribute when it is missing.

Spec update: https://github.com/dotnet/csharplang/pull/8305
Relates to test plan: https://github.com/dotnet/roslyn/issues/66722

### DIFF
--- a/src/Compilers/CSharp/Portable/Emitter/Model/AttributeDataAdapter.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/AttributeDataAdapter.cs
@@ -88,7 +88,8 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
         Cci.ITypeReference Cci.ICustomAttribute.GetType(EmitContext context)
         {
             PEModuleBuilder moduleBeingBuilt = (PEModuleBuilder)context.Module;
-            return moduleBeingBuilt.Translate(this.AttributeClass, syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode, diagnostics: context.Diagnostics);
+            // PROTOTYPE validate once extension types are all allowed in attributes
+            return moduleBeingBuilt.Translate(this.AttributeClass, syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode, diagnostics: context.Diagnostics, keepExtension: false);
         }
 
         bool Cci.ICustomAttribute.AllowMultiple
@@ -107,6 +108,9 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             {
                 case TypedConstantKind.Array:
                     return CreateMetadataArray(argument, context);
+
+                case TypedConstantKind.Primitive when argument.TypeInternal.SpecialType is SpecialType.System_String && argument.ValueInternal is TypeSymbol:
+                    return CreateSerializedType(argument, context);
 
                 case TypedConstantKind.Type:
                     return CreateType(argument, context);
@@ -148,6 +152,20 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             var diagnostics = context.Diagnostics;
             return new MetadataTypeOf(moduleBeingBuilt.Translate((TypeSymbol)argument.ValueInternal, syntaxNodeOpt, diagnostics),
                                       moduleBeingBuilt.Translate((TypeSymbol)argument.TypeInternal, syntaxNodeOpt, diagnostics));
+        }
+
+        private static MetadataSerializedType CreateSerializedType(TypedConstant argument, EmitContext context)
+        {
+            Debug.Assert(argument.ValueInternal is not null);
+            Debug.Assert(((TypeSymbol)argument.TypeInternal).SpecialType == SpecialType.System_String);
+
+            var moduleBeingBuilt = (PEModuleBuilder)context.Module;
+            var syntaxNodeOpt = (CSharpSyntaxNode)context.SyntaxNode;
+            var diagnostics = context.Diagnostics;
+
+            return new MetadataSerializedType(
+                moduleBeingBuilt.Translate((TypeSymbol)argument.ValueInternal, syntaxNodeOpt, diagnostics, keepExtension: true),
+                moduleBeingBuilt.Translate((TypeSymbol)argument.TypeInternal, syntaxNodeOpt, diagnostics));
         }
 
         private static MetadataConstant CreateMetadataConstant(ITypeSymbolInternal type, object value, EmitContext context)
--- a/src/Compilers/CSharp/Portable/Emitter/Model/MethodSymbolAdapter.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/MethodSymbolAdapter.cs
@@ -243,7 +243,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             return ((PEModuleBuilder)context.Module).Translate(AdaptedMethodSymbol.ReturnType,
                 syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode,
                 diagnostics: context.Diagnostics,
-                keepExtension: context.KeepExtensions);
+                keepExtension: false);
         }
 
         IEnumerable<Cci.ITypeReference> Cci.IGenericMethodInstanceReference.GetGenericArguments(EmitContext context)
--- a/src/Compilers/CSharp/Portable/Emitter/Model/MethodSymbolAdapter.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/MethodSymbolAdapter.cs
@@ -87,6 +87,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             return moduleBeingBuilt.Translate(containingType,
                 syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode,
                 diagnostics: context.Diagnostics,
+                keepExtension: true,
                 needDeclaration: AdaptedMethodSymbol.IsDefinition);
         }
 
@@ -241,7 +242,8 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
         {
             return ((PEModuleBuilder)context.Module).Translate(AdaptedMethodSymbol.ReturnType,
                 syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode,
-                diagnostics: context.Diagnostics);
+                diagnostics: context.Diagnostics,
+                keepExtension: context.KeepExtensions);
         }
 
         IEnumerable<Cci.ITypeReference> Cci.IGenericMethodInstanceReference.GetGenericArguments(EmitContext context)
--- a/src/Compilers/CSharp/Portable/Emitter/Model/ParameterSymbolAdapter.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/ParameterSymbolAdapter.cs
@@ -54,7 +54,8 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
         {
             return ((PEModuleBuilder)context.Module).Translate(AdaptedParameterSymbol.Type,
                                                       syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode,
-                                                      diagnostics: context.Diagnostics);
+                                                      diagnostics: context.Diagnostics,
+                                                      keepExtension: context.KeepExtensions);
         }
 
         ushort Cci.IParameterListEntry.Index
--- a/src/Compilers/CSharp/Portable/Emitter/Model/ParameterSymbolAdapter.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/ParameterSymbolAdapter.cs
@@ -55,7 +55,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             return ((PEModuleBuilder)context.Module).Translate(AdaptedParameterSymbol.Type,
                                                       syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode,
                                                       diagnostics: context.Diagnostics,
-                                                      keepExtension: context.KeepExtensions);
+                                                      keepExtension: false);
         }
 
         ushort Cci.IParameterListEntry.Index
--- a/src/Compilers/CSharp/Portable/Symbols/Compilation_WellKnownMembers.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Compilation_WellKnownMembers.cs
@@ -7,9 +7,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Threading;
-using Microsoft.Cci;
 using Microsoft.CodeAnalysis.CSharp.Symbols;
-using Microsoft.CodeAnalysis.Emit;
 using Microsoft.CodeAnalysis.PooledObjects;
 using Microsoft.CodeAnalysis.RuntimeMembers;
 using Microsoft.CodeAnalysis.Symbols;
@@ -814,11 +812,14 @@ namespace Microsoft.CodeAnalysis.CSharp
         internal SynthesizedAttributeData? SynthesizeExtensionErasureAttribute(TypeSymbol type)
         {
             Debug.Assert(type is not null);
-            Debug.Assert(type.ContainsErasedExtensionType());
+            Debug.Assert(type.ContainsExtensionTypeToErase());
 
             var stringType = GetSpecialType(SpecialType.System_String);
             Debug.Assert(stringType is not null);
 
+            // Note: The value is a TypeSymbol despite the type being System.String.
+            //   We'll use the TypeSymbol to index references and to serialize a string for the attribute.
+            //   This odd-shaped TypedConstant is not visible outside the compiler.
             var args = ImmutableArray.Create(new TypedConstant(stringType, TypedConstantKind.Primitive, type));
             return TrySynthesizeAttribute(WellKnownMember.System_Runtime_CompilerServices_ExtensionErasureAttribute__ctorEncodedType, args);
         }
--- a/src/Compilers/CSharp/Portable/Symbols/Compilation_WellKnownMembers.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Compilation_WellKnownMembers.cs
@@ -7,11 +7,14 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Threading;
+using Microsoft.Cci;
 using Microsoft.CodeAnalysis.CSharp.Symbols;
+using Microsoft.CodeAnalysis.Emit;
 using Microsoft.CodeAnalysis.PooledObjects;
 using Microsoft.CodeAnalysis.RuntimeMembers;
 using Microsoft.CodeAnalysis.Symbols;
 using Roslyn.Utilities;
+using static Microsoft.CodeAnalysis.CSharp.Symbols.TypeSymbolExtensions;
 
 namespace Microsoft.CodeAnalysis.CSharp
 {
@@ -808,6 +811,18 @@ namespace Microsoft.CodeAnalysis.CSharp
             return TrySynthesizeAttribute(WellKnownMember.System_Runtime_CompilerServices_TupleElementNamesAttribute__ctorTransformNames, args);
         }
 
+        internal SynthesizedAttributeData? SynthesizeExtensionErasureAttribute(TypeSymbol type)
+        {
+            Debug.Assert(type is not null);
+            Debug.Assert(type.ContainsErasedExtensionType());
+
+            var stringType = GetSpecialType(SpecialType.System_String);
+            Debug.Assert(stringType is not null);
+
+            var args = ImmutableArray.Create(new TypedConstant(stringType, TypedConstantKind.Primitive, type));
+            return TrySynthesizeAttribute(WellKnownMember.System_Runtime_CompilerServices_ExtensionErasureAttribute__ctorEncodedType, args);
+        }
+
         internal SynthesizedAttributeData? SynthesizeAttributeUsageAttribute(AttributeTargets targets, bool allowMultiple, bool inherited)
         {
             var attributeTargetsType = GetWellKnownType(WellKnownType.System_AttributeTargets);
@@ -853,7 +868,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             internal static bool TryGetNames(TypeSymbol type, ArrayBuilder<string?> namesBuilder)
             {
-                type.VisitType((t, builder, _ignore) => AddNames(t, builder), namesBuilder);
+                type.VisitType((t, builder, _, _) => AddNames(t, builder), namesBuilder);
                 return namesBuilder.Any(name => name != null);
             }
 
@@ -926,11 +941,11 @@ namespace Microsoft.CodeAnalysis.CSharp
                 {
                     // Native compiler encodes an extra transform flag, always false, for each custom modifier.
                     HandleCustomModifiers(customModifiersCount, transformFlagsBuilder);
-                    type.VisitType((typeSymbol, builder, isNested) => AddFlags(typeSymbol, builder, isNested, addCustomModifierFlags: true), transformFlagsBuilder);
+                    type.VisitType((typeSymbol, builder, isNested, isContainer) => AddFlags(typeSymbol, builder, isNested, addCustomModifierFlags: true), transformFlagsBuilder);
                 }
                 else
                 {
-                    type.VisitType((typeSymbol, builder, isNested) => AddFlags(typeSymbol, builder, isNested, addCustomModifierFlags: false), transformFlagsBuilder);
+                    type.VisitType((typeSymbol, builder, isNested, isContainer) => AddFlags(typeSymbol, builder, isNested, addCustomModifierFlags: false), transformFlagsBuilder);
                 }
             }
 
@@ -993,8 +1008,9 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                 static void handleFunctionPointerType(FunctionPointerTypeSymbol funcPtr, ArrayBuilder<bool> transformFlagsBuilder, bool addCustomModifierFlags)
                 {
-                    Func<TypeSymbol, (ArrayBuilder<bool>, bool), bool, bool> visitor =
-                        (TypeSymbol type, (ArrayBuilder<bool> builder, bool addCustomModifierFlags) param, bool isNestedNamedType) => AddFlags(type, param.builder, isNestedNamedType, param.addCustomModifierFlags);
+                    TypePredicate<(ArrayBuilder<bool>, bool)> visitor =
+                        (TypeSymbol type, (ArrayBuilder<bool> builder, bool addCustomModifierFlags) param, bool isNestedNamedType, bool isContainer) =>
+                            AddFlags(type, param.builder, isNestedNamedType, param.addCustomModifierFlags);
 
                     // The function pointer type itself gets a false
                     transformFlagsBuilder.Add(false);
@@ -1041,7 +1057,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             internal static void Encode(ArrayBuilder<bool> builder, TypeSymbol type)
             {
                 Debug.Assert(type.ContainingAssembly?.RuntimeSupportsNumericIntPtr != true);
-                type.VisitType((typeSymbol, builder, isNested) => AddFlags(typeSymbol, builder), builder);
+                type.VisitType((typeSymbol, builder, isNested, isContainer) => AddFlags(typeSymbol, builder), builder);
             }
 
             private static bool AddFlags(TypeSymbol type, ArrayBuilder<bool> builder)
--- a/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEParameterSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEParameterSymbol.cs
@@ -11,7 +11,6 @@ using System.Reflection;
 using System.Reflection.Metadata;
 using System.Reflection.Metadata.Ecma335;
 using System.Runtime.InteropServices;
-using System.Text;
 using System.Threading;
 using Microsoft.CodeAnalysis.CSharp.Emit;
 using Microsoft.CodeAnalysis.PooledObjects;
--- a/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEParameterSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEParameterSymbol.cs
@@ -11,6 +11,7 @@ using System.Reflection;
 using System.Reflection.Metadata;
 using System.Reflection.Metadata.Ecma335;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Threading;
 using Microsoft.CodeAnalysis.CSharp.Emit;
 using Microsoft.CodeAnalysis.PooledObjects;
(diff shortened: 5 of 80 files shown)


jcouv:
@AlekseyTs @jjonescz for review. Thanks
There's a remaining PEVerify failure in one test, which I'm investigating today.

src/Compilers/CSharp/Portable/Symbols/Symbol.cs
```diff
@@ -1604,6 +1604,10 @@ internal bool ReportExplicitUseOfReservedAttributes(in DecodeWellKnownAttributeA
                 reportExplicitUseOfReservedAttribute(attribute, arguments, AttributeDescription.RefSafetyRulesAttribute))
             {
             }
+            else if (attribute.IsTargetAttribute(AttributeDescription.ExtensionErasureAttribute))
+            {
+                diagnostics.Add(ErrorCode.ERR_ExplicitExtensionErasureAttr, arguments.AttributeSyntaxOpt.Location);
```
jjonescz, line 1609: It looks like we can reuse `ERR_ExplicitReservedAttr` for this (via `reportExplicitUseOfReservedAttribute`) instead of introducing new error.

src/Compilers/Core/Portable/CodeGen/MetadataSerializedType.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+
+namespace Microsoft.CodeAnalysis.CodeGen
+{
+    /// <summary>
+    /// When erasing an extension type E with underlying type U,
```
jjonescz, line 10: Probably unrelated question: did we want to use just one of the term "underlying type" / "extended type" or both interchangeably?

src/Compilers/Core/Portable/CodeGen/MetadataSerializedType.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// we need to emit `[ExtensionErasure("E")] U`.
+    /// The serialized type encoding the un-erased extension type is
```
jjonescz, line 12:
```suggestion
    /// The serialized type encoding of the un-erased extension type is
```

src/Compilers/Core/Portable/MetadataReader/PEModule.cs
```diff
@@ -1184,6 +1184,20 @@ internal bool HasTupleElementNamesAttribute(EntityHandle token, out ImmutableArr
             return TryExtractStringArrayValueFromAttribute(info.Handle, out tupleElementNames);
         }
 
+        internal bool HasExtensionErasureAttribute(EntityHandle token, out string serializedType)
```
jjonescz, line 1187: Is this method used anywhere?

src/Compilers/Core/Portable/MetadataReader/PEModule.cs
```diff
@@ -1184,6 +1184,20 @@ internal bool HasTupleElementNamesAttribute(EntityHandle token, out ImmutableArr
...
+        {
+            var info = FindTargetAttribute(token, AttributeDescription.ExtensionErasureAttribute);
+            Debug.Assert(!info.HasValue || info.SignatureIndex == 0);
+
+            if (!info.HasValue)
+            {
+                serializedType = null;
```
jjonescz, line 1194: Consider nullable-annotating this method.

src/Compilers/Core/Portable/PEWriter/ReferenceIndexerBase.cs
```diff
@@ -15,8 +15,11 @@ namespace Microsoft.Cci
 {
     internal abstract class ReferenceIndexerBase : MetadataVisitor
     {
-        private readonly HashSet<IReferenceOrISignature> _alreadySeen = new();
-        private readonly HashSet<IReferenceOrISignature> _alreadyHasToken = new();
+        // Note: for the _alreadySeen set, we use the KeepExtensions flag to differentiate between
+        // a visit that erases extension types from one that doesn't.
+        // TODO2
```
jjonescz, line 20: Empty TODO?

src/Compilers/Core/Portable/MetadataReader/TypeNameDecoder.cs
```diff
@@ -181,7 +191,7 @@ internal TypeSymbol GetTypeSymbol(MetadataHelpers.AssemblyQualifiedTypeName full
             if (fullName.TypeArguments != null)
             {
                 ImmutableArray<bool> argumentRefersToNoPiaLocalType;
-                var typeArguments = ResolveTypeArguments(fullName.TypeArguments, out argumentRefersToNoPiaLocalType);
+                var typeArguments = ResolveTypeArguments(fullName.TypeArguments, out argumentRefersToNoPiaLocalType, allowTypeParameters);
```
jjonescz, line 194: It looks like we decode `!n` and `!!n` only when they are in type argument position. Can they be also array element type for example (`!1[]`)?

src/Compilers/Core/Portable/IReferenceOrISignature.cs
```diff
@@ -19,15 +20,29 @@ namespace Microsoft.CodeAnalysis
     internal readonly struct IReferenceOrISignature : IEquatable<IReferenceOrISignature>
     {
         private readonly object _item;
+        internal readonly bool KeepExtensions;
```
jjonescz, line 23: Is this used or can it be `private`?

src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs
```diff
@@ -43548,4 +43551,1989 @@ static void Main()
...
+System.Console.Write(D.M().ToString());
+
+public explicit extension E for U { }
+
+public class U { }
+public class D
+{
+    public static System.Type M()
+    {
+        return typeof(E);
```
jjonescz, line 45092: Should we test `typeof` and `default` that references an extension type in an attribute?

jjonescz (commented):
Thanks for the detailed PR description, it's very helpful.

src/Compilers/Core/Portable/CodeGen/MetadataSerializedType.cs
jcouv: They are interchangeable, but you're correct to call this out. I've been trying to align on "extended type" but old habits die hard ;-)

src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs
jcouv: Good idea. I'll add tests for `typeof` in attribute here, but `default` will be in a follow-up PR (focused on expressions and conversions).

src/Compilers/Core/Portable/MetadataReader/TypeNameDecoder.cs
jcouv: Good catch. Thanks

src/Compilers/Core/Portable/IReferenceOrISignature.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Diagnostics;
```
jcouv, line 6: :memo: All the circularity and self-reference tests fail in UsedAssemblies branch (ie. when emitting). Those scenarios needs to be blocked with an error. I'm working on this now.

src/Compilers/Core/Portable/MetadataReader/TypeNameDecoder.cs
```diff
@@ -153,49 +163,58 @@ internal TypeSymbol GetTypeSymbol(MetadataHelpers.AssemblyQualifiedTypeName full
...
 
-            // Find the top level type
-            Debug.Assert(MetadataHelpers.IsValidMetadataIdentifier(fullName.TopLevelType));
-            var mdName = MetadataTypeName.FromFullName(fullName.TopLevelType);
-            TypeSymbol container = LookupTopLevelTypeDefSymbol(ref mdName, referencedAssemblyIndex, out refersToNoPiaLocalType);
-
-            // Process any nested types
-            if (fullName.NestedTypes != null)
+            TypeSymbol container;
+            if (allowTypeParameters && tryResolveTypeParameterReference(fullName.TopLevelType) is { } typeParameter)
```
AlekseyTs, line 167: >tryResolveTypeParameterReference(fullName.TopLevelType) is { } typeParameter [](http://example.com/codeflow?start=39&length=76)

We probably shouldn't try this if there is an assembly name.

src/Compilers/Core/Portable/MetadataReader/TypeNameDecoder.cs
```diff
@@ -216,18 +235,36 @@ internal TypeSymbol GetTypeSymbol(MetadataHelpers.AssemblyQualifiedTypeName full
...
 
             return container;
+
+            // Decode `!n` (or `!!n`) to a type parameter from the current type (or current method).
+            // This is used for decoding extension erasure attributes
+            TypeSymbol tryResolveTypeParameterReference(string serialized)
+            {
+                switch (serialized)
+                {
+                    case ['!', '!', .. var rest] when Int32.TryParse(rest, out int index):
```
AlekseyTs, line 245: >index [](http://example.com/codeflow?start=83&length=5)

Do we need to check that the value is >= 0?

AlekseyTs, line 245: >.. var rest [](http://example.com/codeflow?start=36&length=11)

We probably should check that the ```rest``` contains only digits to filter out signs and any other weird formats

src/Compilers/Core/Portable/MetadataReader/TypeNameDecoder.cs
```diff
@@ -216,18 +235,36 @@ internal TypeSymbol GetTypeSymbol(MetadataHelpers.AssemblyQualifiedTypeName full
...
+                        return GetGenericMethodTypeParamSymbol(index);
+
+                    case ['!', .. var rest] when Int32.TryParse(rest, out int index):
```
AlekseyTs, line 248: >['!', .. var rest] when Int32.TryParse(rest, out int index): [](http://example.com/codeflow?start=25&length=60)

Similar comments

src/Compilers/Core/Portable/MetadataReader/TypeNameDecoder.cs
```diff
@@ -216,18 +235,36 @@ internal TypeSymbol GetTypeSymbol(MetadataHelpers.AssemblyQualifiedTypeName full
...
         {
             int count = arguments.Length;
             var typeArgumentsBuilder = ArrayBuilder<KeyValuePair<TypeSymbol, ImmutableArray<ModifierInfo<TypeSymbol>>>>.GetInstance(count);
             var refersToNoPiaBuilder = ArrayBuilder<bool>.GetInstance(count);
 
             foreach (var argument in arguments)
             {
-                bool refersToNoPia;
-                typeArgumentsBuilder.Add(new KeyValuePair<TypeSymbol, ImmutableArray<ModifierInfo<TypeSymbol>>>(GetTypeSymbol(argument, out refersToNoPia), ImmutableArray<ModifierInfo<TypeSymbol>>.Empty));
+                bool refersToNoPia = false;
```
AlekseyTs, line 265: > = false; [](http://example.com/codeflow?start=34&length=9)

It looks like the initialization is unnecessary

AlekseyTs:
>                 if (!_alreadyHasToken.Add(new IReferenceOrISignature(typeReference)))

The fact that we are not adjusting this use site as well looks suspicious because it look like we might share tokens between different type forms. Similar for other usages of the same constructor in this file. I think the alternative approach should eliminate this problem, so a prototype comment should probably be sufficient for now unless we can find an simple scenario that is broken.

---
Refers to: src/Compilers/Core/Portable/PEWriter/ReferenceIndexerBase.cs:433 in 3c181e6. [](commit_id = 3c181e621e4af2fe116c7176cd0f07e74eda5d08, deletion_comment = False)

src/Compilers/Core/Portable/PEWriter/TypeNameSerializer.cs
```diff
@@ -88,6 +88,25 @@ internal static string GetSerializedTypeName(this ITypeReference typeReference,
...
                 }
 
+                if (allowTypeParameters)
+                {
+                    // When encoding types into extension erasure attributes we allow type parameters (encoded as `!N` and `!!N`).
+                    if (typeReference.AsGenericTypeParameterReference is { } genericTypeParameter)
+                    {
+                        sb.Append("!");
+                        var index = MetadataWriter.GetNumberOfInheritedTypeParameters(genericTypeParameter.DefiningType, context) + genericTypeParameter.Index;
+                        sb.Append(index);
```
AlekseyTs, line 98: >Append(index) [](http://example.com/codeflow?start=27&length=13)

Is there a way to specify specific format here? I would be good to avoid dependency on default format which could be affected by current locale, etc.

src/Compilers/Core/Portable/PEWriter/TypeNameSerializer.cs
```diff
@@ -88,6 +88,25 @@ internal static string GetSerializedTypeName(this ITypeReference typeReference,
...
+                        goto done;
+                    }
+                    else if (typeReference.AsGenericMethodParameterReference is { } genericMethodParameter)
+                    {
+                        sb.Append("!!");
+                        var index = genericMethodParameter.Index;
+                        sb.Append(index);
```
AlekseyTs, line 105: >sb.Append(index); [](http://example.com/codeflow?start=24&length=17)

Similar comment

src/Compilers/Core/Portable/Symbols/Attributes/AttributeDescription.cs
```diff
@@ -448,6 +449,7 @@ static AttributeDescription()
         internal static readonly AttributeDescription TypeLibTypeAttribute = new AttributeDescription("System.Runtime.InteropServices", "TypeLibTypeAttribute", s_signaturesOfTypeLibTypeAttribute);
         internal static readonly AttributeDescription DynamicAttribute = new AttributeDescription("System.Runtime.CompilerServices", "DynamicAttribute", s_signaturesOfDynamicAttribute);
         internal static readonly AttributeDescription TupleElementNamesAttribute = new AttributeDescription("System.Runtime.CompilerServices", "TupleElementNamesAttribute", s_signaturesOfTupleElementNamesAttribute);
+        internal static readonly AttributeDescription ExtensionErasureAttribute = new AttributeDescription("System.Runtime.CompilerServices", "ExtensionErasureAttribute", s_signaturesOfExtensionErasureAttribute);
```
AlekseyTs, line 452: >s_signaturesOfExtensionErasureAttribute [](http://example.com/codeflow?start=171&length=39)

```s_signatures_HasThis_Void_String_Only```? And remove the new field

File: src/ExpressionEvaluator/CSharp/Source/ExpressionCompiler/Symbols/EENamedTypeSymbol.cs [hunk 18]
```diff
@@ -362,9 +362,7 @@ internal override bool IsInterface
         internal override bool IsExtension => false;
         internal override bool IsExplicitExtension => false;
         internal override TypeSymbol GetDeclaredExtensionUnderlyingType() => throw ExceptionUtilities.Unreachable();
-
-        internal override TypeSymbol GetExtendedTypeNoUseSiteDiagnostics(ConsList<TypeSymbol> basesBeingResolved)
-            => throw ExceptionUtilities.Unreachable();
+        internal override TypeSymbol GetExtendedTypeNoUseSiteDiagnostics(ConsList<TypeSymbol> basesBeingResolved) => null;
```
AlekseyTs, line 365: >null [](http://example.com/codeflow?start=117&length=4)

Is this always the right value? Even if we are debugging code inside an extension type?

src/Compilers/CSharp/Portable/Emitter/Model/AttributeDataAdapter.cs
```diff
@@ -108,6 +109,9 @@ private Cci.IMetadataExpression CreateMetadataExpression(TypedConstant argument,
                 case TypedConstantKind.Array:
                     return CreateMetadataArray(argument, context);
 
+                case TypedConstantKind.Primitive when argument.TypeInternal.SpecialType is SpecialType.System_String && argument.ValueInternal is TypeSymbol:
```
AlekseyTs, line 112: >argument.ValueInternal is TypeSymbol [](http://example.com/codeflow?start=120&length=36)

```TypedConstant``` is a public API. Are we certain we never expose instances like this to the public? Otherwise, ```Value``` has unexpected type (not a string).

File: src/Compilers/CSharp/Portable/Emitter/Model/MethodSymbolAdapter.cs [hunk 20]
```diff
@@ -241,7 +242,8 @@ Cci.ITypeReference Cci.ISignature.GetType(EmitContext context)
         {
             return ((PEModuleBuilder)context.Module).Translate(AdaptedMethodSymbol.ReturnType,
                 syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode,
-                diagnostics: context.Diagnostics);
+                diagnostics: context.Diagnostics,
+                keepExtension: context.KeepExtensions);
```
AlekseyTs, line 246: >context.KeepExtensions [](http://example.com/codeflow?start=31&length=22)

Shouldn't we always erase extensions from return type of a method?

src/Compilers/CSharp/Portable/Emitter/Model/ParameterSymbolAdapter.cs
```diff
@@ -54,7 +54,8 @@ Cci.ITypeReference Cci.IParameterTypeInformation.GetType(EmitContext context)
         {
             return ((PEModuleBuilder)context.Module).Translate(AdaptedParameterSymbol.Type,
                                                       syntaxNodeOpt: (CSharpSyntaxNode)context.SyntaxNode,
-                                                      diagnostics: context.Diagnostics);
+                                                      diagnostics: context.Diagnostics,
+                                                      keepExtension: context.KeepExtensions);
```
AlekseyTs, line 58: >context.KeepExtensions [](http://example.com/codeflow?start=69&length=22)

Shouldn't we always erase extensions here?

src/Compilers/CSharp/Portable/Symbols/Compilation_WellKnownMembers.cs
```diff
@@ -7,11 +7,14 @@
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Threading;
+using Microsoft.Cci;
 using Microsoft.CodeAnalysis.CSharp.Symbols;
+using Microsoft.CodeAnalysis.Emit;
```
AlekseyTs, line 12: Are the new usings really necessary?

src/Compilers/CSharp/Portable/Symbols/TypeSymbolExtensions.cs
```diff
@@ -1235,32 +1246,38 @@ internal static bool ContainsNativeIntegerWrapperType(this TypeWithAnnotations t
             return type.Type?.ContainsNativeIntegerWrapperType() == true;
         }
 
+        internal static bool ContainsErasedExtensionType(this TypeSymbol type)
```
AlekseyTs, line 1249: >ContainsErasedExtensionType [](http://example.com/codeflow?start=29&length=27)

The name feels somewhat misleading/confusing because , if the extension type is erased, then how it can be contained. Consider renaming to something like "ContainsExtensionTypeToErase"

AlekseyTs:
>                         var result = VisitType(

We probably never want to detect extension types at this position (on any level) for the purpose of ```ContainsErasedExtensionType```

---
Refers to: src/Compilers/CSharp/Portable/Symbols/TypeSymbolExtensions.cs:810 in 3c181e6. [](commit_id = 3c181e621e4af2fe116c7176cd0f07e74eda5d08, deletion_comment = False)

src/Compilers/CSharp/Portable/Symbols/TypeSymbolExtensions.cs
```diff
@@ -1235,32 +1246,38 @@ internal static bool ContainsNativeIntegerWrapperType(this TypeWithAnnotations t
...
+        {
+            var result = type.VisitType((type, _, _, isContainer) => !isContainer && type.IsExtension, arg: (object?)null, canDigThroughNullable: true, asContainer: false);
```
AlekseyTs, line 1251: >VisitType [](http://example.com/codeflow?start=30&length=9)

Consider explicitly passing ```visitCustomModifiers: false```

src/Compilers/CSharp/Portable/Symbols/Source/SourceParameterSymbolBase.cs
```diff
@@ -127,6 +127,11 @@ internal override void AddSynthesizedAttributes(PEModuleBuilder moduleBuilder, r
             {
                 AddSynthesizedAttribute(ref attributes, moduleBuilder.SynthesizeNullableAttributeIfNecessary(this, GetNullableContextValue(), type));
             }
+
+            if (type.Type.ContainsErasedExtensionType())
+            {
+                AddSynthesizedAttribute(ref attributes, compilation.SynthesizeExtensionErasureAttribute(type.Type));
```
AlekseyTs, line 133: >AddSynthesizedAttribute(ref attributes, compilation.SynthesizeExtensionErasureAttribute(type.Type)); [](http://example.com/codeflow?start=16&length=100)

If we are not dealing with adding this attribute definition, perhaps we should check its presence and report an error when it is unavailable?

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEParameterSymbol.cs
```diff
@@ -305,17 +305,26 @@ private enum IsParamsValues : byte
...
-                // NamedTypeSymbol and TupleTypeSymbol unnecessarily.
-
-                // The containing type is passed to NullableTypeDecoder.TransformType to determine access
-                // for property parameters because the property does not have explicit accessibility in metadata.
-                var accessSymbol = containingSymbol.Kind == SymbolKind.Property ? containingSymbol.ContainingSymbol : containingSymbol;
-                typeWithAnnotations = NullableTypeDecoder.TransformType(typeWithAnnotations, handle, moduleSymbol, accessSymbol: accessSymbol, nullableContext: nullableContext);
-                typeWithAnnotations = TupleTypeDecoder.DecodeTupleTypesIfApplicable(typeWithAnnotations, handle, moduleSymbol);
+                if (moduleSymbol.TryDecodeExtensionErasureAttribute(handle, (PENamedTypeSymbol)containingSymbol.ContainingType, containingSymbol as PEMethodSymbol) is { } typeWithExtensions)
+                {
+                    typeWithAnnotations = TypeWithAnnotations.Create(typeWithExtensions);
```
AlekseyTs, line 310: >typeWithAnnotations = TypeWithAnnotations.Create(typeWithExtensions) [](http://example.com/codeflow?start=20&length=68)

Should we drop the attribute from the symbol?

AlekseyTs:

src/Compilers/CSharp/Portable/Symbols/Source/SourceParameterSymbolBase.cs
jcouv: I left that as a PROTOTYPE issue in test for now. It doesn't seem worth it to add a check that will be removed later. I could synthesize the attribute definition now if you prefer.

jcouv:
>                 if (!_alreadyHasToken.Add(new IReferenceOrISignature(typeReference)))

The non-erased type is only used as a string in ExtensionErasure attribute, so we don't need a separate token for it. Un-erased types with extensions should not exist in metadata as they break constraints, so I don't expect any other scenario to need it.
I'll leave a PROTOTYPE comment regardless to be safe

---
In reply to: [2249374954](https://github.com/dotnet/roslyn/pull/74313#issuecomment-2249374954) [](http://example.com/codeflow?ancestors=2249374954)

---
Refers to: src/Compilers/Core/Portable/PEWriter/ReferenceIndexerBase.cs:433 in 3c181e6. [](commit_id = 3c181e621e4af2fe116c7176cd0f07e74eda5d08, deletion_comment = False)

AlekseyTs:
>                 if (!_alreadyHasToken.Add(new IReferenceOrISignature(typeReference)))

Think custom modifiers, for example. And BTW, the references don't have to break constraints. They might, but don't have to

---
In reply to: [2250956888](https://github.com/dotnet/roslyn/pull/74313#issuecomment-2250956888) [](http://example.com/codeflow?ancestors=2250956888,2249374954)

---
Refers to: src/Compilers/Core/Portable/PEWriter/ReferenceIndexerBase.cs:433 in 3c181e6. [](commit_id = 3c181e621e4af2fe116c7176cd0f07e74eda5d08, deletion_comment = False)

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEParameterSymbol.cs
jcouv: That would make sense. I think we should keep it in the PEParameterSymbol which represents what is actually in metadata, but remove it from PEExtensionInstanceMethodParameterSymbol.
Would that work?

src/Compilers/CSharp/Portable/Emitter/Model/AttributeDataAdapter.cs
jcouv: Tracing the usages, I couldn't find any way this weird `TypedConstant` could be seen outside the compiler.
It is synthesized in `AddSynthesizedReturnTypeAttributes` (internal) which is called as part of CCI API (Cci.IMethodDefinition.GetReturnValueAttributes, internal). Tracing the usages further we stay within CCI (writing metadata tables, indexing references) which also don't expose the attribute/constant publicly.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEParameterSymbol.cs
AlekseyTs: Why would PEExtensionInstanceMethodParameterSymbol be special here? We are actually removing a bunch of attributes that we synthesize

src/Compilers/Core/Portable/MetadataReader/TypeNameDecoder.cs
AlekseyTs: It doesn't look like assembly name is checked

src/Compilers/Core/Portable/MetadataReader/TypeNameDecoder.cs
```diff
@@ -227,8 +283,7 @@ internal TypeSymbol GetTypeSymbol(MetadataHelpers.AssemblyQualifiedTypeName full
             foreach (var argument in arguments)
             {
                 bool refersToNoPia;
-                typeArgumentsBuilder.Add(new KeyValuePair<TypeSymbol, ImmutableArray<ModifierInfo<TypeSymbol>>>(GetTypeSymbol(argument, out refersToNoPia), ImmutableArray<ModifierInfo<TypeSymbol>>.Empty));
-                refersToNoPiaBuilder.Add(refersToNoPia);
```
AlekseyTs, line 231: >refersToNoPiaBuilder.Add(refersToNoPia); [](http://example.com/codeflow?start=16&length=40)

Is removal of this line intentional?

AlekseyTs:

src/Compilers/CSharp/Portable/Symbols/Source/SourceExtensionTypeSymbol.cs
```diff
@@ -203,6 +212,88 @@ static bool collectTypeParameters(TypeSymbol type, PooledHashSet<TypeParameterSy
...
+            {
+                Debug.Assert(definition.IsDefinition);
+                PooledHashSet<TypeSymbol> alreadyVisited = PooledHashSet<TypeSymbol>.GetInstance();
+                var result = foundSelfReferenceInErasure(underlyingType, definition, alreadyVisited: alreadyVisited, basesBeingResolved: basesBeingResolved, isContainer: false);
+                alreadyVisited.Free();
+                return result;
+            }
+
+            // Returns true if any erasure in the visited type is the given definition
+            static bool foundSelfReferenceInErasure(TypeSymbol type, SourceExtensionTypeSymbol definition, PooledHashSet<TypeSymbol> alreadyVisited, ConsList<TypeSymbol>? basesBeingResolved, bool isContainer = false)
```
AlekseyTs, line 225: >foundSelfReferenceInErasure [](http://example.com/codeflow?start=24&length=27)

This looks unnecessary complicated. What advantage do we get from bundling two unrelated checks? It feels to me that the check could be performed later and with much simpler ```VisitType``` invocation. Not at the time we are resolving the extended type. For example, in ```AfterMembersChecks``` where we already perform some checks involving extended type.

jcouv: I agree with the first part. Moved this check to `AfterMembersChecks`.
For the second part (using a simpler `VisitType` invocation), I don't see how that would work, even if `VisitType` was enhanced to erase extension types.

src/Compilers/CSharp/Portable/Symbols/Source/SourceExtensionTypeSymbol.cs
```diff
@@ -140,7 +137,10 @@ static bool collectTypeParameters(TypeSymbol type, PooledHashSet<TypeParameterSy
         }
 
         internal sealed override TypeSymbol? GetDeclaredExtensionUnderlyingType()
-            => GetDeclaredExtensionInfo(basesBeingResolved: null).UnderlyingType;
+        {
+            var basesBeingResolved = ConsList<TypeSymbol>.Empty.Prepend(this.OriginalDefinition);
+            return GetDeclaredExtensionInfo(basesBeingResolved).UnderlyingType;
```
AlekseyTs, line 142: > return GetDeclaredExtensionInfo(basesBeingResolved).UnderlyingType; [](http://example.com/codeflow?start=11&length=68)

Are the changes around ```basesBeingResolved``` necessary?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -1922,12 +1923,89 @@ static bool needsTupleElementNamesAttribute(TypeSymbol type)
...
+                    if (!isContainer && type.IsExtension)
+                    {
+                        if (extensionsBeingErased.Contains(type.OriginalDefinition))
+                        {
+                            return true;
+                        }
+
+                        if (type.OriginalDefinition.GetExtendedTypeNoUseSiteDiagnostics(null) is not { } extendedType)
+                        {
+                            return true;
```
AlekseyTs, line 1954: >return true; [](http://example.com/codeflow?start=28&length=12)

I guess it is true that this is an error situation, but do we have to trat it as a circular dependency? Could we just ignore such type or treat it as a regular type for the purpose of the check?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -1922,12 +1923,89 @@ static bool needsTupleElementNamesAttribute(TypeSymbol type)
...
+            }
+            return;
+
+            static bool foundUnboundedErasure(TypeSymbol type, ConsList<TypeSymbol> extensionsBeingErased, bool isContainer = false)
+            {
+                if (type is NamedTypeSymbol)
+                {
+                    if (!isContainer && type.IsExtension)
+                    {
+                        if (extensionsBeingErased.Contains(type.OriginalDefinition))
```
AlekseyTs, line 1947: >type.OriginalDefinition [](http://example.com/codeflow?start=59&length=23)

Do we have a test demonstrating that checking for specific extension constructions instead is insufficient? I.e. the one that involves infinite generic expansion, where going through erasure process we would be running into more and more unique constructions of the same extension type without an end.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -1922,12 +1923,89 @@ static bool needsTupleElementNamesAttribute(TypeSymbol type)
...
+                            return true;
+                        }
+
+                        var newExtensionsBeingErased = extensionsBeingErased.Prepend(type.OriginalDefinition);
+                        if (foundUnboundedErasure(extendedType, newExtensionsBeingErased))
+                        {
+                            return true;
+                        }
+
+                        type = type.GetExtendedTypeNoUseSiteDiagnostics(null)!;
```
AlekseyTs, line 1963: >type [](http://example.com/codeflow?start=24&length=4)

It looks like we are missing a call to ```foundUnboundedErasure``` for this type when it is not a named type.

AlekseyTs:
Done with review pass (commit 13)

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEParameterSymbol.cs
```diff
@@ -305,17 +305,27 @@ private enum IsParamsValues : byte
...
```
AlekseyTs, line 310: >typeWithAnnotations = TypeWithAnnotations.Create(typeWithExtensions); [](http://example.com/codeflow?start=20&length=69)

We might want to verify that our erasure process yields the type from the signature, otherwise IL that we emit at the call site won't match the method. Consider adding a prototype comment about that.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
jcouv: Yes, I think `TypeDepends_SelfReference_Expanding` because ever new type that result from erasure/expansion is new and larger than previous.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
jcouv: Changed to ignore the type as we discussed. That works

AlekseyTs (approved):
LGTM (commit 17)

jjonescz (approved):

PR #65398, opened 2022-11-12
Correct `Nullable<T>.Equals()` calls in `BoundNode.Update()`

cston:

PR #74317, opened 2024-07-09
Support long chains of `else if` statements

cston:
Avoid recursion when visiting nested `else if` statements.

See https://github.com/dotnet/roslyn/issues/72393.


src/Compilers/CSharp/Portable/BoundTree/BoundNode.cs
```diff
@@ -654,6 +654,30 @@ private void CheckDeclared(LocalSymbol local)
                 return null;
             }
 
+            public override BoundNode? VisitIfStatement(BoundIfStatement node)
```
AlekseyTs, line 657: >public override BoundNode? VisitIfStatement(BoundIfStatement node) [](http://example.com/codeflow?start=12&length=66)

Move this to ```BoundTreeWalkerWithStackGuardWithoutRecursionOnTheLeftOfBinaryOperator```?

cston: Sounds good.

src/Compilers/CSharp/Portable/Lowering/DiagnosticsPass_ExpressionTrees.cs
```diff
@@ -1045,5 +1045,32 @@ public override BoundNode VisitCollectionExpression(BoundCollectionExpression no
 
             return base.VisitCollectionExpression(node);
         }
+
+        public override BoundNode VisitIfStatement(BoundIfStatement node)
```
cston, line 1049: >VisitIfStatement [](http://example.com/codeflow?start=34&length=16)

Consider deriving `DiagnosticsPass` from `BoundTreeWalkerWithStackGuardWithoutRecursionOnTheLeftOfBinaryOperator`. That should allow removing this override and simplifying `VisitBinaryOperator`.

cston:
>             BoundStatement? rewrittenAlternativeOpt,

Is `RewriteIfStatement()` still called with non-`null` alternative? If not, let's remove this.

---
Refers to: src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_IfStatement.cs:123 in 487db6e. [](commit_id = 487db6e9101a8e58384114b5bff05d877542fefd, deletion_comment = False)

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_IfStatement.cs
```diff
@@ -16,28 +14,106 @@ internal sealed partial class LocalRewriter
...
+
+                if (elseIfStatement is null)
+                {
+                    rewrittenAlternative = VisitStatement(node.AlternativeOpt);
+                }
+
+                var afterif = new GeneratedLabelSymbol("afterif");
+                stack.Push((node, afterif, builder.Count));
+
+                if (elseIfStatement is null && rewrittenAlternative is null)
```
cston, line 45: >elseIfStatement is null && rewrittenAlternative is null [](http://example.com/codeflow?start=20&length=55)

Can this be simplified to `rewrittenAlternative is null`?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_IfStatement.cs
```diff
@@ -16,28 +14,106 @@ internal sealed partial class LocalRewriter
...
+                    builder.Add(new BoundConditionalGoto(rewrittenCondition.Syntax, rewrittenCondition, false, afterif));
+                    builder.Add(rewrittenConsequence);
+                    break;
+                }
+                else
+                {
+                    // if (condition)
+                    //     consequence;
+                    // else 
+                    //     alternative
```
cston, line 65: >alternative [](http://example.com/codeflow?start=27&length=11)

Minor: missing `;`

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -8016,6 +8016,16 @@ private StatementSyntax ParseStatementCore(SyntaxList<AttributeListSyntax> attri
...
+        private StatementSyntax TryReuseStatement(SyntaxList<AttributeListSyntax> attributes, bool isGlobal)
+        {
+            if (canReuseStatement(attributes, isGlobal))
+            {
+                return (StatementSyntax)this.EatNode();
+            }
+
+            return null;
 
             bool canReuseStatement(SyntaxList<AttributeListSyntax> attributes, bool isGlobal)
```
cston, line 8030: >canReuseStatement [](http://example.com/codeflow?start=17&length=17)

Minor: Consider inlining this method.

src/Compilers/CSharp/Portable/Lowering/DiagnosticsPass_ExpressionTrees.cs
AlekseyTs: > Consider deriving `DiagnosticsPass` from `BoundTreeWalkerWithStackGuardWithoutRecursionOnTheLeftOfBinaryOperator`. That should allow removing this override and simplifying `VisitBinaryOperator`.

I'll open a follow-up issue for this refactoring. There is another PR coming about binary patterns. Will take care of all of them together.

AlekseyTs:

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter.cs
```diff
@@ -1143,6 +1143,17 @@ private CompoundUseSiteInfo<AssemblySymbol> GetNewCompoundUseSiteInfo()
         /// </summary>
         private sealed class LocalRewritingValidator : BoundTreeWalkerWithStackGuardWithoutRecursionOnTheLeftOfBinaryOperator
         {
+            public override BoundNode? Visit(BoundNode? node)
+            {
+                if (node is BoundIfStatement)
```
333fred, line 1148: I don't understand what this changes. What does this get us?

src/Compilers/CSharp/Portable/Operations/CSharpOperationFactory.cs
```diff
@@ -1801,15 +1801,49 @@ private IEmptyOperation CreateBoundNoOpStatementOperation(BoundNoOpStatement bou
...
+
+                if (alternative is BoundIfStatement elseIfStatement)
+                {
+                    boundIfStatement = elseIfStatement;
+                }
+                else
+                {
+                    whenFalse = Create(alternative);
+                    break;
+                }
```
333fred, line 1826: Nit: `Create` is null-resileint. Consider simplifying this to just check if `alternative` is a `BoundIfStatement`, and unconditionally calling `Create` and break in the else.

333fred (commented):
A couple of minor comments

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter.cs
AlekseyTs: > I don't understand what this changes. What does this get us?

It gets us a validation that we don't have ```BoundIfStatement```s post lowering. We no longer can enforce that by overriding ```VisitIfStatement``` because it is now sealed in base. I don't think it makes sense to "unseal" it for the purpose of debug-only visitor.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -2180,7 +2180,7 @@ private static bool IsEmptyRewritePossible(BoundNode node)
             }
         }
 
-        private sealed class EmptyRewriter : BoundTreeRewriterWithStackGuard
+        private sealed class EmptyRewriter : BoundTreeRewriterWithStackGuardWithoutRecursionOnTheLeftOfBinaryOperator
```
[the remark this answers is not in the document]
AlekseyTs: This visitor and its usage were added in #65398. It looks there was no specific motivation around verifying stack guard behavior. Rather it looks like it was about asserting that ```Update``` implementations are working for all possible nodes. The purpose is still preserved, in my opinion.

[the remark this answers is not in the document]
AlekseyTs: By the way, the change made to ```BoundTreeRewriterWithStackGuard``` around binary patterns probably makes sense since they are used in expression context. However, it is probably not that useful because patterns do not survive the very first rewrite phase (lowering).

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter.cs
333fred: > We no longer can enforce that by overriding VisitIfStatement because it is now sealed in base.

Ah, that's what I was missing. Thanks.

333fred (approved):

cston:
LGTM, although it looks like I cannot approve the PR, perhaps because I created the PR.

AlekseyTs:
Chuck reviewed my changes, and, I reviewed his initial changes. In my opinion, that should count for a second sign-off. Merging.

src/Compilers/CSharp/Portable/Lowering/DiagnosticsPass_ExpressionTrees.cs
AlekseyTs: Opened #75319.

PR #74329, opened 2024-07-10
Update VSSDK

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/74156.


jjonescz:
Tests are failing with

> System.TypeLoadException : Method 'NormalizeNewlines' in type 'Microsoft.VisualStudio.Text.Document.WhitespaceManager' from assembly 'Microsoft.VisualStudio.Platform.VSEditor, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' does not have an implementation.

I guess that makes sense since I updated some VS packages and this might be running on older VS version?

genlu (approved):

Cosifne:
> Tests are failing with
> 
> > System.TypeLoadException : Method 'NormalizeNewlines' in type 'Microsoft.VisualStudio.Text.Document.WhitespaceManager' from assembly 'Microsoft.VisualStudio.Platform.VSEditor, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' does not have an implementation.
> 
> I guess that makes sense since I updated some VS packages and this might be running on older VS version?

Yeah but this would end up break the CI in main. Either we need to introducing a vs-deps branch or waiting the CI machine update

Issue #74297, opened 2024-07-09
Hitting Debug.Assert in the compiler (TypeWithAnnotations.cs)


davidwengier:
Hitting an assert here locally, when opening Razor.sln on my branch: https://github.com/dotnet/roslyn/blob/0971b49f5852a2ed2d091c3e5343a43e909b10ec/src/Compilers/CSharp/Portable/Symbols/TypeWithAnnotations.cs#L55

Looks like it's caused by having an error type with a nullable annotation, in this case `SignatureHelp?`. Dump file: https://microsoftapc-my.sharepoint.com/:u:/g/personal/dawengie_microsoft_com/Ed8PckcjQGJAjwWdgREarC0BgbhVRrmPfFAXrJ3JOBHpyA?e=9kpYEH

jjonescz:
Looks like the `System.Nullable<T>` type itself cannot be found. I don't know why that is happening in the first place - maybe the project is loading incorrectly in the IDE somehow? But I can at least modify the assert to not fail when that happens.

PR #74330, opened 2024-07-10
Handle missing System.Nullable type

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/74297.

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/TypeWithAnnotations.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/TypeWithAnnotations.cs
@@ -739,7 +739,8 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
 
             if (NullableAnnotation != NullableAnnotation.Oblivious)
             {
-                if (!typeSymbol.IsValueType)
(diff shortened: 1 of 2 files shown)


333fred (approved):

cston (approved):

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -159033,5 +159033,78 @@ class C
                 //                 c1.
                 Diagnostic(ErrorCode.ERR_SemicolonExpected, "").WithLocation(15, 20));
         }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/74297")]
+        public void MissingSystemNullable()
```
huoyaoyuan, line 159038: Nit: should the test be in `NullableTests` instead of `NullableReferenceTypesTests`? It's about nullable value types.

jjonescz: Yeah, sounds good, I will move them, thanks.

PR #74337, opened 2024-07-10
Always check get method parameters in override indexer

### LINKED ISSUE #73384: Errors for scoped differences in override indexers are missing in some cases
```cs
public ref struct RS { }

public class Base
{
    public virtual RS this[scoped RS rs] { get => default; }
    public virtual RS this[scoped RS rs, int _] { get => default; set { }}
}

public class C : Base
{
    public override RS this[RS rs] { get => default; } // error CS8987: The 'scoped' modifier of parameter 'rs' doesn't match overridden or implemented member.
    public override RS this[RS rs, int _] { get => default; set { } } // no error
}
```

**Expected**: Both commented lines have errors
**Actual**: Only the first commented line has an error. It seems like adding a set accessor to a similar indexer declaration removes the error.

RikkiGibson:
Closes #73384

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol_ImplementationChecks.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol_ImplementationChecks.cs
@@ -1028,9 +1028,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                                 overridingMemberLocation,
                                 overriddenMethod,
                                 overridingMethod,
-                                diagnostics,
-                                checkReturnType: true,
-                                checkParameters: true);
+                                diagnostics);
                         }
                     }
 
@@ -1101,13 +1099,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                                 overridingProperty.GetMethod.GetFirstLocation(),
                                 overriddenGetMethod,
                                 overridingProperty.GetMethod,
-                                diagnostics,
-                                checkReturnType: true,
-                                // Don't check parameters on the getter if there is a setter
-                                // because they will be a subset of the setter
-                                checkParameters: overridingProperty.SetMethod is null ||
-                                                 overriddenGetMethod?.AssociatedSymbol != overriddenProperty ||
-                                                 overriddenProperty.GetOwnOrInheritedSetMethod()?.AssociatedSymbol != overriddenProperty);
+                                diagnostics);
                         }
 
                         if (overridingProperty.SetMethod is object)
@@ -1117,9 +1109,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                                 overridingProperty.SetMethod.GetFirstLocation(),
                                 ownOrInheritedOverriddenSetMethod,
                                 overridingProperty.SetMethod,
-                                diagnostics,
-                                checkReturnType: false,
-                                checkParameters: true);
+                                diagnostics);
 
                             if (ownOrInheritedOverriddenSetMethod is object &&
                                 overridingProperty.SetMethod.IsInitOnly != ownOrInheritedOverriddenSetMethod.IsInitOnly)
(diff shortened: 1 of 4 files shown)


src/Compilers/CSharp/Test/Semantic/Semantics/RefFieldTests.cs
```diff
@@ -16929,6 +16929,35 @@ class B2 : A<string>
                 Diagnostic(ErrorCode.ERR_ScopedMismatchInParameterOfOverrideOrImplementation, "F6").WithArguments("s").WithLocation(17, 32));
         }
 
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/73384")]
+        public void Indexer_EscapeInGetter()
```
jaredpar, line 16933: Do we have similar tests for interfaces?

src/Compilers/CSharp/Test/Emit2/Semantics/RefReadonlyParameterTests.cs
```diff
@@ -5975,12 +5975,15 @@ static void Main()
             C123
             """, sourceSymbolValidator: verify, symbolValidator: verify);
         verifier.VerifyDiagnostics(
-            // (22,9): warning CS9196: Reference kind modifier of parameter 'ref readonly int x' doesn't match the corresponding parameter 'in int x' in overridden or implemented member.
-            //         set { }
-            Diagnostic(ErrorCode.WRN_OverridingDifferentRefness, "set").WithArguments("ref readonly int x", "in int x").WithLocation(22, 9),
-            // (28,19): warning CS9191: The 'ref' modifier for argument 1 corresponding to 'in' parameter is equivalent to 'in'. Consider using 'in' instead.
-            //         _ = c[ref x];
-            Diagnostic(ErrorCode.WRN_BadArgRef, "x").WithArguments("1").WithLocation(28, 19));
+                // (17,9): warning CS9196: Reference kind modifier of parameter 'ref readonly int x' doesn't match the corresponding parameter 'in int x' in overridden or implemented member.
```
jaredpar, line 5978: Do we have the inverse of this scenario in a test: override `ref readonly` with `in`?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol_ImplementationChecks.cs
```diff
@@ -1028,9 +1028,7 @@ void checkSingleOverriddenMember(Symbol overridingMember, Symbol overriddenMembe
                                 overridingMemberLocation,
                                 overriddenMethod,
                                 overridingMethod,
```
jaredpar, line 1030: Was the `checkReturnType` value superfluous before?

RikkiGibson: It was only used to suppress nullable variance diagnostics related to the return type. The only time false was passed for it was for a property setter, which is going to return void anyway, so we don't really expect any nullable diagnostics related to the return type to occur.

src/Compilers/CSharp/Test/Emit2/Semantics/RefReadonlyParameterTests.cs
RikkiGibson: Yes, `Implementation_RefReadonly_In_Indexer` for example tests this scenario in an interface and implementation.

src/Compilers/CSharp/Test/Semantic/Semantics/RefFieldTests.cs
RikkiGibson: No and we had a dupe of almost the same check which causes the error to get dropped. Fixing.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol_ImplementationChecks.cs
333fred: So, I hate to be the voice of insanity, but IL doesn't require that `set` be a `void`-returning method. What do we do if we're overriding a non-`void` setter?

RikkiGibson: This change may cause redundant or strange nullable diagnostics to appear in that case. I am OK with that.

333fred (approved):

RikkiGibson:

jaredpar (approved):

PR #74339, opened 2024-07-10
Do not use memory mapped files on non-windows

dibarbet:
Different version of https://github.com/dotnet/roslyn/pull/73628
Resolves https://github.com/dotnet/vscode-csharp/issues/7119

Memory mapped files on linux are saved to /dev/shm/ which has a tiny size by default in docker containers.  This causes the server to crash.  Fix is to only use memory mapped files when OOP is required (Windows).

### DIFF
--- a/src/Workspaces/Core/Portable/Serialization/SerializerService.cs
+++ b/src/Workspaces/Core/Portable/Serialization/SerializerService.cs
@@ -43,7 +43,7 @@ internal partial class SerializerService : ISerializerService
 
     private readonly SolutionServices _workspaceServices;
 
-    private readonly TemporaryStorageService _storageService;
+    private readonly Lazy<TemporaryStorageService> _storageService;
     private readonly ITextFactoryService _textService;
     private readonly IDocumentationProviderService? _documentationService;
     private readonly IAnalyzerAssemblyLoaderProvider _analyzerLoaderProvider;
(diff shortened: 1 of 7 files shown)


src/Workspaces/Core/Portable/Serialization/SerializerService.cs
```diff
@@ -55,9 +55,10 @@ private protected SerializerService(SolutionServices workspaceServices)
     {
         _workspaceServices = workspaceServices;
 
-        // Serialization is only involved when we have a remote process.  Which is only in VS. So the type of the
-        // storage service here is well known.
-        _storageService = (TemporaryStorageService)workspaceServices.GetRequiredService<ITemporaryStorageServiceInternal>();
+        // Serialization to temporary storage is only involved when we have a remote process.  Which is only in VS. So the type of the
+        // storage service here is well known.  However the serializer is created in other cases (e.g. to compute project state checksums).
+        // So lazily instantiate the storage service to avoid attempting to get the TemporaryStorageService when not available.
+        _storageService = new Lazy<TemporaryStorageService>(() => (TemporaryStorageService)workspaceServices.GetRequiredService<ITemporaryStorageServiceInternal>());
```
dibarbet, line 61: A bit weird.  the serializer service is almost only used for oop serialization - except its also used for create project parse option checksums on all platforms.

The TemporaryStorageService is not used in that case, so instantiating it lazily to avoid throwing when it attempts to convert to TemporaryStorageService

src/Workspaces/Core/Portable/TemporaryStorage/TrivialTemporaryStorageService.cs
```diff
@@ -0,0 +1,81 @@
...
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.CodeAnalysis;
+
+internal sealed class TrivialTemporaryStorageService : ITemporaryStorageServiceInternal
```
dibarbet, line 14: adapted from the old version
https://github.com/dotnet/roslyn/blob/ea35f7ce524e9424adb16fd6b8a2308e55e6de0e/src/Workspaces/Core/Portable/Workspace/Host/TemporaryStorage/TrivialTemporaryStorageService.cs

CyrusNajmabadi (approved):

PR #74349, opened 2024-07-11
Handle span conversions in expression trees

jjonescz:

### DIFF
--- a/src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
+++ b/src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
@@ -1827,6 +1827,196 @@ public class FirstClassSpanTests : CSharpTestBase
     }
 
     [Theory, CombinatorialData]
+    public void Conversion_Array_Span_Implicit_ExpressionTree_01(
+        [CombinatorialValues("Span", "ReadOnlySpan")] string type)
+    {
+        var source = $$"""
+            using System;
+            using System.Linq.Expressions;
+
+            C.R(a => C.M(a));
+
+            static class C
+            {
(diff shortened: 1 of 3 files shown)


333fred (approved):

jjonescz:

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -3352,6 +3542,136 @@ static class C
         CompileAndVerify(comp, expectedOutput: "11").VerifyDiagnostics();
     }
 
+    [Theory, MemberData(nameof(LangVersions))]
+    public void OverloadResolution_ReadOnlySpanVsArray_06(LanguageVersion langVersion)
+    {
+        var source = """
+            using System;
+
+            C.M(default(object[]));
```
cston, line 3551: > C.M(default(object[])); [](http://example.com/codeflow?start=11&length=24)

Minor: Consider adding this case to the previous test rather than creating a separate test.

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -3352,6 +3542,136 @@ static class C
...
+    }
+
+    [Theory, MemberData(nameof(LangVersions))]
+    public void OverloadResolution_ReadOnlySpanVsArray_ExpressionTree_05(LanguageVersion langVersion)
+    {
+        var source = """
+            using System;
+            using System.Linq.Expressions;
+
+            C.R(() => C.M(default(object[])));
```
cston, line 3662: >C.R(() => C.M(default(object[]))); [](http://example.com/codeflow?start=12&length=34)

Minor: Consider adding this case to the previous test rather than creating a separate test.

cston (approved):

PR #74361, opened 2024-07-11
Add IsEnabled property to IEditorInlineRenameService

MariaSolOs:
In order for TypeScript to enable LSP rename, we need the rename command handler to no-op when the respective language service has indicated so.

### DIFF
--- a/src/EditorFeatures/Core/ExternalAccess/VSTypeScript/VSTypeScriptEditorInlineRenameService.cs
+++ b/src/EditorFeatures/Core/ExternalAccess/VSTypeScript/VSTypeScriptEditorInlineRenameService.cs
@@ -24,6 +24,19 @@ internal sealed class VSTypeScriptEditorInlineRenameService(
 {
     private readonly Lazy<VSTypeScriptEditorInlineRenameServiceImplementation>? _service = service;
 
+    public bool IsEnabled
+    {
+        get
+        {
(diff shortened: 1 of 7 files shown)


dibarbet (approved):

src/EditorFeatures/Core/ExternalAccess/VSTypeScript/VSTypeScriptEditorInlineRenameService.cs
```diff
@@ -24,6 +24,19 @@ internal sealed class VSTypeScriptEditorInlineRenameService(
...
+     public bool IsEnabled
+     {
+        get
+        {
+            if (_service != null)
+            {
+                return _service.Value.IsEnabled();
+            }
+
+            return AbstractEditorInlineRenameService.DefaultEnablement;
```
dibarbet, line 36: `DefaultEnablement` is only used here right?  If so I would just inline it.

dibarbet (approved):

src/EditorFeatures/Core/ExternalAccess/VSTypeScript/VSTypeScriptEditorInlineRenameService.cs
MariaSolOs: Will do

PR #74363, opened 2024-07-11
Fix interceptors bugs involving ref this parameters

### LINKED ISSUE #71657: Intercepting a struct rvalue instance method call with an extension method trips a debug assertion in the compiler
We should resolve by either emitting an error at the interception (which feels unfortunate, but consistent with extension method calls), or by creating a temp during lowering.

Note that the shipped release-mode behavior here is to create a temp in the emit layer and do the call.
 
https://github.com/dotnet/roslyn/blob/1c88444212854f6b6be43d9901f6c3d541bdcafa/src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs#L722

```cs
[Fact]
public void InterceptStructInstanceMethod_RvalueReceiver()
{
    // TODO2: this test crashes in debug mode due to assert in EmitExpression.cs:722.
    var source = ($$"""
        using System;

        struct Program
        {
            public void InterceptableMethod() => Console.Write("Original");

            public static void Main()
            {
                new Program().InterceptableMethod();
            }
        }
        """, "Program.cs");

    var interceptor = ("""
        using System.Runtime.CompilerServices;
        using System;

        static class D
        {
            [InterceptsLocation("Program.cs", 9, 23)]
            public static void Interceptor(this ref Program x) => Console.Write("Intercepted");
        }
        """, "Interceptor.cs");

    var verifier = CompileAndVerify(new[] { source, s_attributesSource }, parseOptions: RegularWithInterceptors);
    verifier.VerifyDiagnostics();
    verifier.VerifyIL("Program.Main", """
        {
            // Code size       15 (0xf)
            .maxstack  2
            .locals init (Program V_0)
            IL_0000:  ldloca.s   V_0
            IL_0002:  dup
            IL_0003:  initobj    "Program"
            IL_0009:  call       "void Program.InterceptableMethod()"
            IL_000e:  ret
        }
        """);

    verifier = CompileAndVerify(new[] { source, interceptor, s_attributesSource }, parseOptions: RegularWithInterceptors);
    verifier.VerifyDiagnostics();
    verifier.VerifyIL("Program.Main", """

        """);
}
```

### LINKED ISSUE #71714: Interceptor with 'ref readonly this' parameter should have compatibility with 'readonly' original method
Punted out of #71655 because I wanted to take the proper time to figure out the best way to "normalize"/compare ref kinds, particularly when combinations of `readonly` methods, `in` parameters, and `ref readonly` parameters are in use.

Note that the below scenario compiles successfully when the interceptor has a `this in` parameter. Most likely, we want `this ref readonly` to "just work" where `this in` would work, while `this ref` will be incompatible.
 
```cs
    [Theory]
    [InlineData("ref readonly")]
    [InlineData("ref")]
    public void SignatureMismatch_12(string interceptorRefKind)
    {
        var source = ("""
            using System;

            struct Program
            {
                public readonly void InterceptableMethod() => Console.Write("Original");

                public static void Main()
                {
                    new Program().InterceptableMethod();
                }
            }
            """, "Program.cs");

        var interceptor = ($$"""
            using System.Runtime.CompilerServices;
            using System;

            static class D
            {
                [InterceptsLocation("Program.cs", 9, 23)]
                public static void Interceptor(this {{interceptorRefKind}} Program x) => Console.Write("Intercepted");
            }
            """, "Interceptor.cs");
        var verifier = CompileAndVerify(new[] { source, s_attributesSource }, parseOptions: RegularWithInterceptors, expectedOutput: "Original");
        verifier.VerifyDiagnostics();

        var comp = CreateCompilation(new[] { source, interceptor, s_attributesSource }, parseOptions: RegularWithInterceptors);
        comp.VerifyEmitDiagnostics(
            // Interceptor.cs(6,6): error CS9148: Interceptor must have a 'this' parameter matching parameter 'in Program this' on 'Program.InterceptableMethod()'.
            //     [InterceptsLocation("Program.cs", 9, 23)]
            Diagnostic(ErrorCode.ERR_InterceptorMustHaveMatchingThisParameter, @"InterceptsLocation(""Program.cs"", 9, 23)").WithArguments("in Program this", "Program.InterceptableMethod()").WithLocation(6, 6));
    }
```

RikkiGibson:
Closes #71714
Closes #71657

### DIFF
--- a/docs/features/interceptors.md
+++ b/docs/features/interceptors.md
@@ -180,7 +180,7 @@ This allows generator authors to avoid *polluting lookup* with interceptors, hel
 
 We may also want to consider adjusting behavior of `[EditorBrowsable]` to work in the same compilation.
 
-### Struct rvalue receivers
+### Struct receiver capture
 
 An interceptor whose `this` parameter takes a struct by-reference can generally be used to intercept a struct instance method call, assuming the methods are compatible per [Signature matching](#signature-matching). This includes a specific situation where the interceptor wouldn't be directly usable as an extension method, due to the receiver being an rvalue. See also [12.8.9.3 Extension method invocations
 ](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12893-extension-method-invocations) in the standard.
--- a/docs/features/interceptors.md
+++ b/docs/features/interceptors.md
@@ -195,7 +195,7 @@ struct S
 
 static class Program
 {
-    public static void Interceptor()
+    public static void Main()
     {
         new S().Original(); // L1: interception is valid, no errors.
         new S().Interceptor(); // error CS1510: A ref or out value must be an assignable variable
--- a/docs/features/interceptors.md
+++ b/docs/features/interceptors.md
@@ -151,7 +151,7 @@ static class D
 When a call is intercepted, the interceptor and interceptable methods must meet the signature matching requirements detailed below:
 - When an interceptable instance method is compared to a static interceptor method (including a classic extension method), we use the method as if it is an extension in reduced form for comparison. The first parameter of the static method is compared to the instance method `this` parameter.
     - The implementation currently requires the interceptor to be an extension method for this comparison to work. We plan on addressing this before releasing .NET 8.
-- The returns and parameters of the respective methods, including the `this` parameter, must have the same ref kinds and types.
+- The returns and parameters of the respective methods must have the same ref kinds and types.
 - The `this` parameter of the respective methods must have the same ref kinds and types, except that when a `readonly` struct instance method is intercepted with a static method, it is permitted for the interceptor `this` parameter to be either `in` or `ref readonly`.
 - A warning is reported instead of an error if a type difference is found where the types are not distinct to the runtime. For example, `object` and `dynamic`.
 - No warning or error is reported for a *safe* nullability difference, such as when the interceptable method accepts a `string` parameter, and the interceptor accepts a `string?` parameter.
--- a/docs/features/interceptors.md
+++ b/docs/features/interceptors.md
@@ -182,8 +182,7 @@ We may also want to consider adjusting behavior of `[EditorBrowsable]` to work i
 
 ### Struct receiver capture
 
-An interceptor whose `this` parameter takes a struct by-reference can generally be used to intercept a struct instance method call, assuming the methods are compatible per [Signature matching](#signature-matching). This includes a specific situation where the interceptor wouldn't be directly usable as an extension method, due to the receiver being an rvalue. See also [12.8.9.3 Extension method invocations
-](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12893-extension-method-invocations) in the standard.
+An interceptor whose `this` parameter takes a struct by-reference can generally be used to intercept a struct instance method call, assuming the methods are compatible per [Signature matching](#signature-matching). This includes situations where the receiver must be implicitly captured to temp before the invocation, even if such capture is not permitted when the interceptor is called directly. See also [12.8.9.3 Extension method invocations](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12893-extension-method-invocations) in the standard.
 
 
 ```cs
--- a/docs/features/interceptors.md
+++ b/docs/features/interceptors.md
@@ -151,7 +151,8 @@ static class D
 When a call is intercepted, the interceptor and interceptable methods must meet the signature matching requirements detailed below:
 - When an interceptable instance method is compared to a static interceptor method (including a classic extension method), we use the method as if it is an extension in reduced form for comparison. The first parameter of the static method is compared to the instance method `this` parameter.
     - The implementation currently requires the interceptor to be an extension method for this comparison to work. We plan on addressing this before releasing .NET 8.
-- The returns and parameters, including the `this` parameter, must have the same ref kinds and types.
+- The returns and parameters of the respective methods, including the `this` parameter, must have the same ref kinds and types.
+- The `this` parameter of the respective methods must have the same ref kinds and types, except that when a `readonly` struct instance method is intercepted with a static method, it is permitted for the interceptor `this` parameter to be either `in` or `ref readonly`.
 - A warning is reported instead of an error if a type difference is found where the types are not distinct to the runtime. For example, `object` and `dynamic`.
 - No warning or error is reported for a *safe* nullability difference, such as when the interceptable method accepts a `string` parameter, and the interceptor accepts a `string?` parameter.
 - Method names and parameter names are not required to match.
@@ -178,6 +179,38 @@ An interceptor contained in a file-local type is permitted to intercept a call i
 This allows generator authors to avoid *polluting lookup* with interceptors, helps avoid name conflicts, and prevents use of interceptors in *unintended positions* from the interceptor author's point-of-view.
 
 We may also want to consider adjusting behavior of `[EditorBrowsable]` to work in the same compilation.
+
+### Struct rvalue receivers
+
+An interceptor whose `this` parameter takes a struct by-reference can generally be used to intercept a struct instance method call, assuming the methods are compatible per [Signature matching](#signature-matching). This includes a specific situation where the interceptor wouldn't be directly usable as an extension method, due to the receiver being an rvalue. See also [12.8.9.3 Extension method invocations
+](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12893-extension-method-invocations) in the standard.
+
+
+```cs
+using System.Runtime.CompilerServices;
+
+struct S
+{
+    public void Original() { }
+}
+
+static class Program
+{
+    public static void Interceptor()
+    {
+        new S().Original(); // L1: interception is valid, no errors.
+        new S().Interceptor(); // error CS1510: A ref or out value must be an assignable variable
+    }
+}
+
+static class D
+{
+    [InterceptsLocation(1, "..(refers to call at L1)")]
+    public static void Interceptor(this ref S s)
+}
+```
+
+The goal of this is to avoid a "decoder ring" experience where an interceptor author needs to copy exactly the same ref kind sometimes and use a different ref kind other times, depending on the call arguments. We generally want that if your interceptor is accessible at the call site and uses all the same parameter ref kinds and types that you will be allowed to intercept.
 
 ### Editor experience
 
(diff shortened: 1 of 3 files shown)


docs/features/interceptors.md
RikkiGibson, line 1: Particularly looking for feedback on the design doc changes here.

docs/features/interceptors.md
```diff
@@ -151,7 +151,8 @@ static class D
 When a call is intercepted, the interceptor and interceptable methods must meet the signature matching requirements detailed below:
 - When an interceptable instance method is compared to a static interceptor method (including a classic extension method), we use the method as if it is an extension in reduced form for comparison. The first parameter of the static method is compared to the instance method `this` parameter.
     - The implementation currently requires the interceptor to be an extension method for this comparison to work. We plan on addressing this before releasing .NET 8.
-- The returns and parameters, including the `this` parameter, must have the same ref kinds and types.
+- The returns and parameters of the respective methods, including the `this` parameter, must have the same ref kinds and types.
```
RikkiGibson, line 154:
```suggestion
- The returns and parameters of the respective methods must have the same ref kinds and types.
```

docs/features/interceptors.md
```diff
@@ -179,6 +180,38 @@ This allows generator authors to avoid *polluting lookup* with interceptors, hel
 
 We may also want to consider adjusting behavior of `[EditorBrowsable]` to work in the same compilation.
 
+### Struct receiver capture
+
+An interceptor whose `this` parameter takes a struct by-reference can generally be used to intercept a struct instance method call, assuming the methods are compatible per [Signature matching](#signature-matching). This includes a specific situation where the interceptor wouldn't be directly usable as an extension method, due to the receiver being an rvalue. See also [12.8.9.3 Extension method invocations
+](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12893-extension-method-invocations) in the standard.
```
RikkiGibson, line 186:
```suggestion
An interceptor whose `this` parameter takes a struct by-reference can generally be used to intercept a struct instance method call, assuming the methods are compatible per [Signature matching](#signature-matching). This includes situations where the receiver must be implicitly captured to temp before the invocation, even if such capture were not permitted in the case where the interceptor was called directly. See also [12.8.9.3 Extension method invocations](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12893-extension-method-invocations) in the standard.
```

src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs
```diff
@@ -3784,8 +3784,10 @@ static class D
             Diagnostic(ErrorCode.ERR_InterceptorMustHaveMatchingThisParameter, @"InterceptsLocation(""Program.cs"", 10, 23)").WithArguments("ref Program this", "Program.InterceptableMethod()").WithLocation(16, 6));
     }
 
-    [Fact]
-    public void SignatureMismatch_11()
+    [Theory, WorkItem("https://github.com/dotnet/roslyn/issues/71714")]
+    [InlineData("ref readonly")]
```
jjonescz, line 3788: I'm wondering if we should warn when an rvalue is passed to a `ref readonly` parameter (we do that everywhere else except for interceptors it seems). The interceptor should just use `in` if it wants to intercept an rvalue receiver.

src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs
```diff
@@ -3801,14 +3803,14 @@ public static void Main()
...
-        var interceptor = ("""
+        var interceptor = ($$"""
             using System.Runtime.CompilerServices;
             using System;
 
             static class D
             {
                 [InterceptsLocation("Program.cs", 9, 23)]
-                public static void Interceptor(this in Program x) => Console.Write("Intercepted");
+                public static void Interceptor(this {{refKind}} Program x) => Console.Write("Intercepted");
```
jjonescz, line 3813: Do we have a similar test where the intereptor has a `ref readonly` parameter (but it's not the receiver) and the callsite passes an rvalue to it? And similarly when an lvalue is passed but without any `ref`/`in` callsite modifier. Both would be warnings if those were non-intercepted invocations. I'm wondering if there should be a warning for the intercepted call as well (the interceptor can just use `in` to avoid the warning)?

docs/features/interceptors.md
```diff
@@ -179,6 +180,37 @@ This allows generator authors to avoid *polluting lookup* with interceptors, hel
...
+    public static void Interceptor()
+    {
+        new S().Original(); // L1: interception is valid, no errors.
+        new S().Interceptor(); // error CS1510: A ref or out value must be an assignable variable
+    }
+}
+
+static class D
+{
+    [InterceptsLocation(1, "..(refers to call at L1)")]
```
jaredpar, line 207: What is L1?

docs/features/interceptors.md
```diff
@@ -179,6 +180,37 @@ This allows generator authors to avoid *polluting lookup* with interceptors, hel
...
+    public static void Interceptor(this ref S s)
+}
+```
+
+The goal of this is to avoid a "decoder ring" experience where an interceptor author needs to copy exactly the same ref kind sometimes and use a different ref kind other times, depending on the call arguments. We generally want that if your interceptor is accessible at the call site and uses all the same parameter ref kinds and types that you will be allowed to intercept.
```
jaredpar, line 212: Not following the decoder ring bit because the `ref` kind is the same here.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_Call.cs
```diff
@@ -401,7 +420,7 @@ BoundExpression visitArgumentsAndFinishRewrite(BoundCall node, BoundExpression?
                     ref temps,
                     invokedAsExtensionMethod);
 
-                InterceptCallAndAdjustArguments(ref method, ref rewrittenReceiver, ref rewrittenArguments, ref argRefKindsOpt, invokedAsExtensionMethod, node.InterceptableNameSyntax);
+                InterceptCallAndAdjustArguments(ref method, ref rewrittenReceiver, ref rewrittenArguments, ref argRefKindsOpt, ref temps, invokedAsExtensionMethod, node.InterceptableNameSyntax);
```
AlekseyTs, line 423: >InterceptCallAndAdjustArguments [](http://example.com/codeflow?start=16&length=31)

Is implementation of ```Instrumenter.InterceptCallAndAdjustArguments``` called below subject to the same issue?

docs/features/interceptors.md
RikkiGibson: L1 is a commented line earlier in the sample

docs/features/interceptors.md
```diff
@@ -179,6 +180,37 @@ This allows generator authors to avoid *polluting lookup* with interceptors, hel
...
+{
+    public void Original() { }
+}
+
+static class Program
+{
+    public static void Interceptor()
+    {
+        new S().Original(); // L1: interception is valid, no errors.
+        new S().Interceptor(); // error CS1510: A ref or out value must be an assignable variable
```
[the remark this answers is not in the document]
RikkiGibson: What this sample is meant to demonstrate is that even though directly calling `Interceptor()` would fail here, using it to intercept call to instance method `Original()` will succeed.

docs/features/interceptors.md
RikkiGibson: I unintentionally made this point confusing. Let me try to re-state it.

If we were to disallow intercepting a call like `new S().Original()` with an interceptor like `public static void Interceptor(this ref S s)`, it means one of two things:

1. If you don't own type `S`, you have no ability to intercept calls to `Original` (unless user happens to always be calling it on a compatible lvalue). This seems like an unreasonable limitation.
2. Or, we would devise some other rule which allows intercepting a call like `new S().Original()` with an extension method. For example, by permitting it if the interceptor takes its `this` parameter by-value, since a direct call to the interceptor method would be valid in that case. The "decoder ring" would be the interceptor author's decision of which ref kind to use, based on the call receiver.

src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs
RikkiGibson: I think if we do end up permitting this scenario, we actually shouldn't warn. We already do not warn, by design, for things like `[Obsolete]` on an interceptor. This seems to fall in a similar category.

Do you know if we consider the `this` parameter of a readonly struct method to be `in` or `ref readonly` on the language level? Or maybe it doesn't fall into that categorization at all?

It seems like the compiler treats such a `this` parameter as `in`, but I'm not sure of the full significance of that.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_Call.cs
RikkiGibson: Based on the instrumenters which currently override this method in the compiler, I think it is not subject to #71657 because it will not substitute a call to instance method with a call to static method.

src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs
jjonescz: > Do you know if we consider the `this` parameter of a readonly struct method to be `in` or `ref readonly` on the language level?
> It seems like the compiler treats such a `this` parameter as `in`

Yes, I think that's like an `in` parameter. It's fine to call readonly struct methods on rvalues, that would warn if they were considered `ref readonly`.

jjonescz: > I think if we do end up permitting this scenario, we actually shouldn't warn. We already do not warn, by design, for things like `[Obsolete]` on an interceptor. This seems to fall in a similar category.

Makes sense to me, thanks.

docs/features/interceptors.md
```diff
@@ -179,6 +180,37 @@ This allows generator authors to avoid *polluting lookup* with interceptors, hel
...
+using System.Runtime.CompilerServices;
+
+struct S
+{
+    public void Original() { }
+}
+
+static class Program
+{
+    public static void Interceptor()
```
jaredpar, line 198: Is this meant to be an instance method? If not having a `static` method named `Interceptor` here is a bit confusing. It's not a part of the `Interceptor()` resolution below (unless I'm missing somtehnig).

RikkiGibson: It's not, I should have just named this Main or something.

src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs
RikkiGibson: I believe the answer here is the same as for https://github.com/dotnet/roslyn/pull/74363#discussion_r1677358306, i.e. only the original method parameters are used to decide to report a warning about a wrong argument value kind or callsite modifier.

PR #74382, opened 2024-07-14
Remove Overload Resolution Priority from working set

Youssef1313:
It looks like the feature is already merged to main and is listed in C# 13 in the same document.


jcouv:
I believe this is by-design. There's VB work remaining. Tagging @333fred to confirm and potentially close

333fred:
It is not by design, this was probably a merge conflict.

333fred (approved):

PR #74387, opened 2024-07-15
Improve span conversion tests

jjonescz:
Follow up on https://github.com/dotnet/roslyn/pull/74232#issuecomment-2215489010, i.e., switching to `CreateCompilationWithSpanAndMemoryExtensions` so Span from the .NET runtime library is used in tests when not running on .NET Framework. The rest are modifications needed because of that.

I'm doing these modifications before merging main into the feature branch so that the merge PR doesn't need to contain any non-trivial changes like this. (Some changes would be needed even if we didn't switch to `CreateCompilationWithSpanAndMemoryExtensions` because of the Span sources consolidation in https://github.com/dotnet/roslyn/pull/74281.)

Commit-by-commit review might be useful. (EDIT: Skip the fourth one https://github.com/dotnet/roslyn/pull/74387/commits/2e98bd221882d987a91387f1486cc120b1cb8ed6 as it's been reverted.)

### DIFF
--- a/src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
+++ b/src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
@@ -55,7 +55,7 @@ public class FirstClassSpanTests : CSharpTestBase
             }
             """;
 
-        CreateCompilationWithSpan(source, parseOptions: TestOptions.Regular12).VerifyDiagnostics(
+        CreateCompilationWithSpanAndMemoryExtensions(source, parseOptions: TestOptions.Regular12).VerifyDiagnostics(
             // (7,65): error CS0121: The call is ambiguous between the following methods or properties: 'StringExtensions.Join(string, params string[])' and 'StringExtensions.Join(string, params ReadOnlySpan<string>)'
             //     public static string M(StringValues sv) => StringExtensions.Join(",", sv);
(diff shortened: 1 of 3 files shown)


jjonescz:

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -1957,7 +1981,9 @@ static class C
                 public static void M({{type}}<string> x) { }
             }
             """;
-        var comp = CreateCompilationWithSpan(source, parseOptions: TestOptions.Regular.WithLanguageVersion(langVersion));
+        var comp = CreateCompilationWithSpanAndMemoryExtensions(source,
+            parseOptions: TestOptions.Regular.WithLanguageVersion(langVersion),
+            noStringToReadOnlySpanConversionInSource: true);
```
cston, line 1986: >noStringToReadOnlySpanConversionInSource: true [](http://example.com/codeflow?start=12&length=46)

Is this the only place `noStringToReadOnlySpanConversionInSource` is used? If so, could we just special-case this use? And is this a scenario where we expect to support the non-standard definition? If not, should we test that we're *not* supporting it?

jjonescz: > Is this the only place `noStringToReadOnlySpanConversionInSource` is used? If so, could we just special-case this use?

Yes, you're right, I will add only what's needed from the span source into the test and revert the complicated workaround, thanks.

> And is this a scenario where we expect to support the non-standard definition?

Yes, the non-standard user-defined operator from `string` to `ROS<T>` is expected to be supported here - because we do not ignore UDCs from `null` to `ROS<T>`, so we use the UDC.

cston (approved):

PR #74402, opened 2024-07-16
Razor EA changes to allow better tests in Razor

davidwengier:
* Moves things used for testing into a Testing folder so we know its safe to change
* Allows creation of the test language server with a different language server kind (for cohosting)
* Allows tests to use the RazorPinnedSolutionInfoWrapper
* Tiny bit of clean up


src/Tools/ExternalAccess/Razor/Testing/RazorTestLanguageServerFactory.cs
```diff
@@ -15,28 +15,29 @@
 
 namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
 {
-    [Export(typeof(AbstractRazorLanguageServerFactoryWrapper))]
     [Shared]
-    internal class RazorLanguageServerFactoryWrapper : AbstractRazorLanguageServerFactoryWrapper
+    [Export(typeof(RazorTestLanguageServerFactory))]
+    [Export(typeof(AbstractRazorLanguageServerFactoryWrapper))]
```
davidwengier, line 20: The abstract class is only used for this export, and was probably just copied from Razor where we used to use abstract classes a lot. I'm exporting it as itself now, so that we can remove the abstract class in future, but we have to wait until these changes make it to VS Preview images for that to happen, or the Roslyn integration tests will fail.

JoeRobich (approved):

PR #74411, opened 2024-07-16
Reduce allocations in AnalyzerFileReference.GetAnalyzerTypeNameMap

ToddGrun:
This method showed up when investigating a recent regression in allocations due to a change in lightbulb behaviors. Although this method hasn't changed recently, it is one of the easiest ways to claw back some memory during this scenario.

Locally, opening the vs-platform solution with this change reduces allocations under GetAnalyzerTypeNameMap from about 1.9% (63 MB) to about 0.1% (5 MB)

*** previous allocations during vs-platform solution open ***
image

*** allocations with these changes during vs-platform solution open ***
image

### DIFF
--- a/src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs
+++ b/src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs
@@ -31,7 +31,7 @@ namespace Microsoft.CodeAnalysis.Diagnostics
     /// </remarks>
     public sealed class AnalyzerFileReference : AnalyzerReference, IEquatable<AnalyzerReference>
     {
-        private delegate IEnumerable<string> AttributeLanguagesFunc(PEModule module, CustomAttributeHandle attribute);
+        private delegate ImmutableArray<string> AttributeLanguagesFunc(PEModule module, CustomAttributeHandle attribute);
 
         public override string FullPath { get; }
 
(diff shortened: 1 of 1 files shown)


src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs
```diff
@@ -224,48 +224,63 @@ private static AnalyzerLoadFailureEventArgs CreateAnalyzerFailedArgs(Exception e
...
-                              from typeDefHandle in module.MetadataReader.TypeDefinitions
-                              let typeDef = module.MetadataReader.GetTypeDefinition(typeDefHandle)
-                              let supportedLanguages = GetSupportedLanguages(typeDef, module.Module, attributeType, languagesFunc)
-                              where supportedLanguages.Any()
-                              let typeName = GetFullyQualifiedTypeName(typeDef, module.Module)
-                              from supportedLanguage in supportedLanguages
-                              group typeName by supportedLanguage;
-
-            return typeNameMap.ToImmutableSortedDictionary(g => g.Key, g => g.ToImmutableHashSet(), StringComparer.OrdinalIgnoreCase);
+            Dictionary<string, ImmutableHashSet<string>.Builder> typeNameMap = new Dictionary<string, ImmutableHashSet<string>.Builder>(StringComparer.OrdinalIgnoreCase);
```
333fred, line 227: Should we use `ImmutableSegmentedHashSet` instead?

ToddGrun: I didn't see these hitting the LOH in the profile that I looked at, so I'd prefer to keep as-is, without that evidence to back up such a change.

ToddGrun:
@dotnet/roslyn-compiler -- ptal

chsienki (approved):

333fred (approved):

PR #74420, opened 2024-07-17
Implement remaining span conversions

jjonescz:

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
@@ -623,14 +623,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             var sameElementTypes = sourceElementType.Equals(destinationElementType, TypeCompareKind.AllIgnoreOptions);
 
-            if (source.OriginalDefinition.IsReadOnlySpan())
-            {
-                Debug.Assert(!sameElementTypes);
-            }
-            else
-            {
-                Debug.Assert(source.OriginalDefinition.IsSpan());
-            }
+            Debug.Assert(!source.OriginalDefinition.IsReadOnlySpan() || !sameElementTypes);
 
             return !sameElementTypes;
         }
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
@@ -663,7 +663,8 @@ namespace Microsoft.CodeAnalysis.CSharp
             MethodSymbol? result = null;
             foreach (var memoryExtensionsType in compilation.GetTypesByMetadataName(WellKnownMemberNames.MemoryExtensionsTypeFullName))
             {
-                if (TryFindSingleMethod(memoryExtensionsType.GetSymbol<NamedTypeSymbol>(), WellKnownMemberNames.AsSpanMethodName, 0,
+                if (memoryExtensionsType.DeclaredAccessibility == Accessibility.Public &&
+                    TryFindSingleMethod(memoryExtensionsType.GetSymbol<NamedTypeSymbol>(), WellKnownMemberNames.AsSpanMethodName, 0,
                     static (_, method) => method is
                     {
                         ParameterCount: 1,
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
@@ -488,7 +488,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 else if (conversion.IsSpan)
                 {
                     Debug.Assert(source.Type is not null);
-                    Debug.Assert(destination.OriginalDefinition.IsSpan() || destination.OriginalDefinition.IsReadOnlySpan());
+                    Debug.Assert(destination.IsSpan() || destination.IsReadOnlySpan());
 
                     CheckFeatureAvailability(syntax, MessageID.IDS_FeatureFirstClassSpan, diagnostics);
 
@@ -507,9 +507,9 @@ namespace Microsoft.CodeAnalysis.CSharp
                     }
 
                     // ReadOnlySpan<T> Span<T>.op_Implicit(Span<T>)
-                    if (source.Type.OriginalDefinition.IsSpan())
+                    if (source.Type.IsSpan())
                     {
-                        Debug.Assert(destination.OriginalDefinition.IsReadOnlySpan());
+                        Debug.Assert(destination.IsReadOnlySpan());
                         reportUseSiteOrMissing(
                             TryFindImplicitOperatorFromSpan(source.Type.OriginalDefinition, destination.OriginalDefinition),
                             source.Type.OriginalDefinition,
@@ -519,19 +519,19 @@ namespace Microsoft.CodeAnalysis.CSharp
                     }
 
                     // ReadOnlySpan<T> ReadOnlySpan<T>.CastUp<TDerived>(ReadOnlySpan<TDerived>)
-                    if (source.Type.OriginalDefinition.IsSpan() || source.Type.OriginalDefinition.IsReadOnlySpan())
+                    if (source.Type.IsSpan() || source.Type.IsReadOnlySpan())
                     {
-                        Debug.Assert(destination.OriginalDefinition.IsReadOnlySpan());
+                        Debug.Assert(destination.IsReadOnlySpan());
                         if (NeedsSpanCastUp(source.Type, destination))
                         {
                             // If converting Span<TDerived> -> ROS<TDerived> -> ROS<T>,
                             // the source of the CastUp is the return type of the op_Implicit (i.e., the ROS<TDerived>)
                             // which has the same original definition as the destination ROS<T>.
-                            TypeSymbol sourceForCastUp = source.Type.OriginalDefinition.IsSpan()
+                            TypeSymbol sourceForCastUp = source.Type.IsSpan()
                                 ? destination.OriginalDefinition
                                 : source.Type.OriginalDefinition;
 
-                            MethodSymbol? castUpMethod = TryFindCastUpMethod(sourceForCastUp.OriginalDefinition, destination.OriginalDefinition);
+                            MethodSymbol? castUpMethod = TryFindCastUpMethod(sourceForCastUp, destination.OriginalDefinition);
                             reportUseSiteOrMissing(
                                 castUpMethod,
                                 destination.OriginalDefinition,
@@ -581,8 +581,8 @@ namespace Microsoft.CodeAnalysis.CSharp
             Debug.Assert(type.IsSpan() || type.IsReadOnlySpan());
             Debug.Assert(type.IsDefinition);
 
-            return TryFindImplicitOperator(type, 0, static (_, parameterType) =>
-                parameterType is ArrayTypeSymbol { IsSZArray: true, ElementType: TypeParameterSymbol });
+            return TryFindImplicitOperator(type, 0, static (_, method) =>
+                method.Parameters[0].Type is ArrayTypeSymbol { IsSZArray: true, ElementType: TypeParameterSymbol });
         }
 
         // ReadOnlySpan<T> Span<T>.op_Implicit(Span<T>)
@@ -592,30 +592,27 @@ namespace Microsoft.CodeAnalysis.CSharp
             Debug.Assert(spanType.IsDefinition && readonlySpanType.IsDefinition);
 
             return TryFindImplicitOperator(spanType, readonlySpanType,
-                parameterPredicate: static (_, parameterType) => parameterType.IsSpan(),
-                returnTypePredicate: static (readonlySpanType, returnType) => readonlySpanType.Equals(returnType.OriginalDefinition, TypeCompareKind.ConsiderEverything));
+                static (readonlySpanType, method) => method.Parameters[0].Type.IsSpan() &&
+                    readonlySpanType.Equals(method.ReturnType.OriginalDefinition, TypeCompareKind.ConsiderEverything));
         }
 
         private static MethodSymbol? TryFindImplicitOperator<TArg>(TypeSymbol type, TArg arg,
-            Func<TArg, TypeSymbol, bool> parameterPredicate,
-            Func<TArg, TypeSymbol, bool>? returnTypePredicate = null)
+            Func<TArg, MethodSymbol, bool> predicate)
         {
-            return TryFindSingleMember(type, WellKnownMemberNames.ImplicitConversionName, (parameterPredicate, returnTypePredicate, arg),
+            return TryFindSingleMethod(type, WellKnownMemberNames.ImplicitConversionName, (predicate, arg),
                 static (arg, method) => method is
                 {
                     ParameterCount: 1,
                     Arity: 0,
                     IsStatic: true,
                     DeclaredAccessibility: Accessibility.Public,
-                    Parameters: [{ Type: { } parameterType }],
-                } && arg.parameterPredicate(arg.arg, parameterType) &&
-                    arg.returnTypePredicate?.Invoke(arg.arg, method.ReturnType) != false);
+                } && arg.predicate(arg.arg, method));
         }
 
         internal static bool NeedsSpanCastUp(TypeSymbol source, TypeSymbol destination)
         {
-            Debug.Assert(source.OriginalDefinition.IsSpan() || source.OriginalDefinition.IsReadOnlySpan());
-            Debug.Assert(destination.OriginalDefinition.IsReadOnlySpan());
+            Debug.Assert(source.IsSpan() || source.IsReadOnlySpan());
+            Debug.Assert(destination.IsReadOnlySpan());
             Debug.Assert(!source.IsDefinition && !destination.IsDefinition);
 
             var sourceElementType = ((NamedTypeSymbol)source).TypeArgumentsWithAnnotationsNoUseSiteDiagnostics[0].Type;
@@ -623,7 +620,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             var sameElementTypes = sourceElementType.Equals(destinationElementType, TypeCompareKind.AllIgnoreOptions);
 
-            Debug.Assert(!source.OriginalDefinition.IsReadOnlySpan() || !sameElementTypes);
+            Debug.Assert(!source.IsReadOnlySpan() || !sameElementTypes);
 
             return !sameElementTypes;
         }
@@ -634,7 +631,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             Debug.Assert(source.IsReadOnlySpan() && destination.IsReadOnlySpan());
             Debug.Assert(source.IsDefinition && destination.IsDefinition);
 
-            return TryFindSingleMember(destination, WellKnownMemberNames.CastUpMethodName, (source, destination),
+            return TryFindSingleMethod(destination, WellKnownMemberNames.CastUpMethodName, (source, destination),
                 static (arg, method) => method is
                 {
                     ParameterCount: 1,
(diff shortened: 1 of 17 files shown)


src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -1446,7 +3417,90 @@ .maxstack  1
...
+              IL_0001:  call       "System.Span<int> System.ReadOnlySpan<int>.op_Explicit(System.ReadOnlySpan<int>)"
+              IL_0006:  ret
+            }
+            """);
+
+        var expectedDiagnostics = new[]
+        {
+            // (4,44): error CS0030: Cannot convert type 'System.ReadOnlySpan<int>' to 'System.Span<int>'
+            //      Span<int> M(ReadOnlySpan<int> arg) => (Span<int>)arg;
+            Diagnostic(ErrorCode.ERR_NoExplicitConv, "(Span<int>)arg").WithArguments("System.ReadOnlySpan<int>", "System.Span<int>").WithLocation(4, 44)
```
333fred, line 3454: I would expect this to be legal.

333fred (commented):
Done review pass

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
jjonescz: Why? There is no implicit nor explicit conversion from ReadOnlySpan to Span. There is an explicit span conversion only from array to (ReadOnly)Span. Similarly we could propose other explicit conversions but I'm not sure how we would emit them anyway.

333fred: Because the type defines a UDC that permits it.

jjonescz: Ah I missed that. Then you're right that looks a bit unexpected, let me check what's happening.

jjonescz: Okay, this UDC is ignored - per [the speclet](https://github.com/dotnet/csharplang/blob/main/proposals/first-class-span-types.md#user-defined-conversions):

> User-defined conversions are not considered when converting between
> - any single-dimensional `array_type` and `System.Span<T>`/`System.ReadOnlySpan<T>`,
> - any combination of `System.Span<T>`/`System.ReadOnlySpan<T>`,
> - `string` and `System.ReadOnlySpan<char>`.

This case is the second bullet point.

333fred: I'm not sure why the speclet says that. The only things the speclet should talk about are the conversions that it actually cares about. It does not care about converting from `ReadOnlySpan` to `Span.

333fred: The spec should certainly say that no ROS->Span standard explicit conversion exists, but that isn't what we have here. We have a UDC. We shouldn't be disallowing this, as the feature doesn't care about it. It of course can't fall in the bucket of a standard implicit conversion; it can't contribute to the things that only "standard" conversions can. But it can absolutely still exist and be used, and be called by user code as any UDC can.

jjonescz: Okay, works for me. Updating the speclet in https://github.com/dotnet/csharplang/pull/8312 (the second commit is relevant to this discussion).

333fred (approved):

jjonescz:
@cston for a second review, thanks

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -487,51 +487,216 @@ void checkConstraintLanguageVersionAndRuntimeSupportForConversion(SyntaxNode syn
...
+                        if (NeedsSpanCastUp(source.Type, destination))
+                        {
+                            // If converting Span<TDerived> -> ROS<TDerived> -> ROS<T>,
+                            // the source of the CastUp is the return type of the op_Implicit (i.e., the ROS<TDerived>)
+                            // which has the same original definition as the destination ROS<T>.
+                            TypeSymbol sourceForCastUp = source.Type.OriginalDefinition.IsSpan()
+                                ? destination.OriginalDefinition
+                                : source.Type.OriginalDefinition;
+
+                            MethodSymbol? castUpMethod = TryFindCastUpMethod(sourceForCastUp.OriginalDefinition, destination.OriginalDefinition);
```
cston, line 534: >OriginalDefinition [](http://example.com/codeflow?start=93&length=18)

Isn't `sourceForCastUp` already an original definition?

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -487,51 +487,216 @@ void checkConstraintLanguageVersionAndRuntimeSupportForConversion(SyntaxNode syn
...
+                            TryFindImplicitOperatorFromArray(destination.OriginalDefinition),
+                            destination.OriginalDefinition,
+                            WellKnownMemberNames.ImplicitConversionName,
+                            syntax,
+                            diagnostics);
                     }
-                    else
+
+                    // ReadOnlySpan<T> Span<T>.op_Implicit(Span<T>)
+                    if (source.Type.OriginalDefinition.IsSpan())
```
cston, line 510: >OriginalDefinition.IsSpan() [](http://example.com/codeflow?start=36&length=27)

Is `.OriginalDefinition` necessary for the calls to `IsSpan()` and `IsReadOnlySpan()`? If not, consider removing `.OriginalDefinition` in those cases, since there are many uses of that here.

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -487,51 +487,216 @@ void checkConstraintLanguageVersionAndRuntimeSupportForConversion(SyntaxNode syn
...
+            return TryFindImplicitOperator(spanType, readonlySpanType,
+                parameterPredicate: static (_, parameterType) => parameterType.IsSpan(),
+                returnTypePredicate: static (readonlySpanType, returnType) => readonlySpanType.Equals(returnType.OriginalDefinition, TypeCompareKind.ConsiderEverything));
+        }
 
-            return TryFindSingleMember(type, WellKnownMemberNames.ImplicitConversionName,
-                static (method) => method is
+        private static MethodSymbol? TryFindImplicitOperator<TArg>(TypeSymbol type, TArg arg,
+            Func<TArg, TypeSymbol, bool> parameterPredicate,
+            Func<TArg, TypeSymbol, bool>? returnTypePredicate = null)
```
cston, line 601: Consider simplifying this to a single `Func<TArg, MethodSymbol>` instead, and let the caller decide whether to check the type of the parameter and/or the return type.

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -487,51 +487,216 @@ void checkConstraintLanguageVersionAndRuntimeSupportForConversion(SyntaxNode syn
...
+                    Arity: 1,
+                    IsStatic: true,
+                    DeclaredAccessibility: Accessibility.Public,
+                    Parameters: [{ } parameter],
+                    TypeArgumentsWithAnnotations: [{ } typeArgument],
+                } &&
+                    // parameter type is the source ReadOnlySpan<>
+                    arg.source.Equals(parameter.Type.OriginalDefinition, TypeCompareKind.ConsiderEverything) &&
+                    // return type is the destination ReadOnlySpan<>
+                    arg.destination.Equals(method.ReturnType.OriginalDefinition, TypeCompareKind.ConsiderEverything) &&
```
cston, line 650: It looks like these two checks are just checking that the parameter type and return type are some constructed instances of `ReadOnlySpan<T>` since `arg.source` and `arg.destination` are both definitions.

If that's the case, can we simplify those checks to `parameter.Type.IsReadOnlySpan() && method.ReturnType.IsReadOnlySpan()`?

And if so, do we need to pass in `(source, destination)` as the `arg` for `TryFindSingleMember()` or could we just pass `destination.TypeParameters[0]` instead?

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -487,51 +487,216 @@ void checkConstraintLanguageVersionAndRuntimeSupportForConversion(SyntaxNode syn
...
+                        Arity: 0,
+                        IsStatic: true,
+                        DeclaredAccessibility: Accessibility.Public,
+                        Parameters: [{ Type.SpecialType: SpecialType.System_String }]
+                    }) is { } method &&
+                    method.ReturnType.Equals(readOnlySpanType, TypeCompareKind.ConsiderEverything))
+                {
+                    if (result is not null)
+                    {
+                        // Ambiguous member found.
```
cston, line 682: >Ambiguous [](http://example.com/codeflow?start=27&length=9)

Will we hit an ambiguity if one of the referenced assemblies in the compilation redefines `System.MemoryExtensions`, even as `internal`?

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -487,51 +487,216 @@ void checkConstraintLanguageVersionAndRuntimeSupportForConversion(SyntaxNode syn
...
+
+                    result = method;
+                }
+            }
+
+            return result;
         }
 
-        private static MethodSymbol? TryFindSingleMember(TypeSymbol type, string name, Func<MethodSymbol, bool> predicate)
+        private static MethodSymbol? TryFindSingleMember<TArg>(TypeSymbol type, string name, TArg arg, Func<TArg, MethodSymbol, bool> predicate)
```
cston, line 693: >TryFindSingleMember [](http://example.com/codeflow?start=37&length=19)

Consider renaming to `TryFindSingleMethod`.

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
```diff
@@ -3975,6 +3975,26 @@ private bool HasImplicitSpanConversion(TypeSymbol? source, TypeSymbol destinatio
                     return hasCovariantConversion(elementType, spanElementType, ref useSiteInfo);
                 }
             }
+            // SPEC: From `System.Span<Ti>` to `System.ReadOnlySpan<Ui>`, provided that `Ti` is covariance-convertible to `Ui`.
+            // SPEC: From `System.ReadOnlySpan<Ti>` to `System.ReadOnlySpan<Ui>`, provided that `Ti` is covariance-convertible to `Ui`.
+            else if (source.OriginalDefinition.IsSpan() || source.OriginalDefinition.IsReadOnlySpan())
```
cston, line 3980: >OriginalDefinition [](http://example.com/codeflow?start=28&length=18)

Can `.OriginalDefinition` be dropped for the calls to `IsSpan()` and `IsReadOnlySpan()`?

cston:
>     public void Conversion_Span_ReadOnlySpan_Implicit(

It looks like these tests use implicit and explicit casts.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:327 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>     [InlineData("ReadOnlySpan<char> AsSpan(object o)", false)]

Consider testing `Span<char> AsSpan(string s)`.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:820 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>     }

Consider testing:
```
ReadOnlySpan<string> x = new Span<string?>();
ReadOnlySpan<string?> y = new Span<string>();
```

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2305 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>                     public static implicit operator ReadOnlySpan<string>(ReadOnlySpan<T> span) => default;

It seems surprising to me that we allow user-defined conversions between spans and arrays, spans, or strings.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2320 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>                     public static implicit operator Span<T>(ReadOnlySpan<T> span) => default;

It seems surprising to me that we allow user-defined conversions from `ReadOnlySpan<T>` to `Span<T>`.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2434 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>                     public static implicit operator Span<T>(string s) => default;

I'm surprised we allow user-defined conversions from `string` to `Span<T>`.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2579 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>                 ReadOnlySpan<string> M7(Span<object?> arg) => (ReadOnlySpan<string>)arg;

Should this be ... ?
```
ReadOnlySpan<object> M7(Span<string?> arg) => (ReadOnlySpan<object>)arg;
```

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2754 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>     public void Conversion_Span_ReadOnlySpan_Implicit_NullableAnalysis_NestedNullability_Covariant()

Is this case different than `Conversion_Span_ReadOnlySpan_Implicit_NullableAnalysis_NestedNullability`?

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:3143 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>     public void Conversion_ReadOnlySpan_ReadOnlySpan_Implicit_NullableAnalysis_NestedNullability_Covariant()

Is this case different than `Conversion_ReadOnlySpan_ReadOnlySpan_Implicit_NullableAnalysis_NestedNullability`?

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:3172 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

cston:
>                     public static explicit operator Span<T>(ReadOnlySpan<T> span) => throw null;

I'm surprised we allow explicit user-defined conversions between spans.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:3684 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

jjonescz:
>                     public static implicit operator ReadOnlySpan<string>(ReadOnlySpan<T> span) => default;

We didn't, but Fred said we should. See [https://github.com/dotnet/roslyn/pull/74420#discussion_r1683455091](https://github.com/dotnet/roslyn/pull/74420#discussion_r1683455091). The change is in [https://github.com/dotnet/roslyn/pull/74420/commits/d9e4b719f83b29c813db1e873475a3870f654400](https://github.com/dotnet/roslyn/pull/74420/commits/d9e4b719f83b29c813db1e873475a3870f654400).

---
In reply to: [2259370393](https://github.com/dotnet/roslyn/pull/74420#issuecomment-2259370393) [](http://example.com/codeflow?ancestors=2259370393)

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2320 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
jjonescz: This check also makes sure the ReadOnlySpan type is not different than the one used in the conversion. For example test `Conversion_ReadOnlySpan_ReadOnlySpan_Implicit_MultipleSpans_02` (and _03, _04) would behave differently if we checked just `IsReadOnlySpan()`.

jjonescz:
>                 ReadOnlySpan<string> M7(Span<object?> arg) => (ReadOnlySpan<string>)arg;

It was modeled on the previous test (only with a Span instead of an array). But I will add the test you are suggesting.

---
In reply to: [2259378657](https://github.com/dotnet/roslyn/pull/74420#issuecomment-2259378657) [](http://example.com/codeflow?ancestors=2259378657)

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2754 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

jjonescz:
>     }

Yes, the tests have suffix `_UnrelatedElementType`.

---
In reply to: [2259350562](https://github.com/dotnet/roslyn/pull/74420#issuecomment-2259350562) [](http://example.com/codeflow?ancestors=2259350562)

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:487 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

jjonescz:
>     }

These conversions are already tested in `Conversion_Span_ReadOnlySpan_Implicit_NullableAnalysis`.

---
In reply to: [2259368438](https://github.com/dotnet/roslyn/pull/74420#issuecomment-2259368438) [](http://example.com/codeflow?ancestors=2259368438)

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2305 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

jjonescz:
>     public void Conversion_Span_ReadOnlySpan_Implicit(

The `_Implicit` refers to the fact that this is the "implicit span conversion", as opposed to the "explicit span conversion" which exists only from array to (ReadOnly)Span if there is explicit conversion between the element types. Even with the cast, it's only testing the implicit conversion in a sense.

I agree this is confusing, and I thought about removing this `_Implicit` from most tests but at this point it would be too many changes which seems unnecessary.

---
In reply to: [2259344500](https://github.com/dotnet/roslyn/pull/74420#issuecomment-2259344500) [](http://example.com/codeflow?ancestors=2259344500)

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:327 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
jjonescz: No, we shouldn't, only if they also redefine the method `AsSpan`. I will add a test. Thanks.

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
cston: > I'm not sure why the speclet says that. The only things the speclet should talk about are the conversions that it actually cares about. It does not care about converting from `ReadOnlySpan` to `Span`.

My expectation is that if some conversions between spans and arrays, spans, and strings are handled directly in the compiler, then all such conversions should be handled in the compiler, and the compiler should not fallback to user-defined conversions between these types. Was this covered at LDM?

cston:
>                     public static implicit operator ReadOnlySpan<string>(ReadOnlySpan<T> span) => default;

Added a comment here. We should discuss whether to allow user-defined conversions at LDM, if not already discussed.

---
In reply to: [2262345108](https://github.com/dotnet/roslyn/pull/74420#issuecomment-2262345108) [](http://example.com/codeflow?ancestors=2262345108,2259370393)

---
Refers to: src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs:2320 in d9e4b71. [](commit_id = d9e4b719f83b29c813db1e873475a3870f654400, deletion_comment = False)

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
jjonescz: > Was this covered at LDM?

It wasn't, adding an open question in https://github.com/dotnet/csharplang/pull/8312.

cston (approved):

jjonescz:

333fred (approved):

PR #74428, opened 2024-07-17
Adjust overload resolution priority from LDM

333fred:
LDM today decided that on older language versions, we shouldn't influence overload resolution from OverloadResolutionPriority. It also decided that we should disallow the attribute on local functions.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
@@ -1712,10 +1712,15 @@ outerDefault:
             return currentBestIndex;
         }
 
-        private static void RemoveLowerPriorityMembers<TMemberResolution, TMember>(ArrayBuilder<TMemberResolution> results)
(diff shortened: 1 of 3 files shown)


333fred:
@jcouv @cston for review.

cston (approved):

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1712,10 +1712,15 @@ private int GetTheBestCandidateIndex<TMember>(ArrayBuilder<MemberResolutionResul
             return currentBestIndex;
         }
 
-        private static void RemoveLowerPriorityMembers<TMemberResolution, TMember>(ArrayBuilder<TMemberResolution> results)
+        private void RemoveLowerPriorityMembers<TMemberResolution, TMember>(ArrayBuilder<TMemberResolution> results)
```
jcouv, line 1715: Let's also incorporate the change into the spec

jcouv (approved):
LGTM Thanks (iteration 1)

PR #74431, opened 2024-07-17
Increase driver cache size

chsienki:
This ups the size of the generator cache based on observed 'big' projects in the wild. We also only cache a generator driver if it actually produced any results, based on observations that its generally fast to produce nothing. 

Also updates the replay tool to support multiple iterations for testing scenarios like this (second commit)

### DIFF
--- a/src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
+++ b/src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
@@ -838,7 +838,11 @@ namespace Microsoft.CodeAnalysis
             driver = driver.RunGeneratorsAndUpdateCompilation(input, out var compilationOut, out var diagnostics);
             generatorDiagnostics.AddRange(diagnostics);
 
-            if (!disableCache)
+            // We only cache the generator driver if it produced any generated files. While its possible that it was expensive
+            // to calculate that nothing needed to be generated, real world usage has found that generators are generally only
+            // expensive when actually producing source. By only caching those with results, we help to keep memory usage down
+            // when it probably wouldn't improve the performance anyway.
+            if (!disableCache && driver.GetRunResult().GeneratedTrees.Any())
             {
                 this.GeneratorDriverCache?.CacheGenerator(cacheKey, driver);
             }
--- a/src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
+++ b/src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
@@ -838,7 +838,7 @@ namespace Microsoft.CodeAnalysis
             driver = driver.RunGeneratorsAndUpdateCompilation(input, out var compilationOut, out var diagnostics);
             generatorDiagnostics.AddRange(diagnostics);
 
-            // We only cache the generator driver if it produced any generated files. While its possible that it was expensive
+            // We only cache the generator driver if it produced any generated files. While it's possible that it was expensive
             // to calculate that nothing needed to be generated, real world usage has found that generators are generally only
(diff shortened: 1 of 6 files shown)


chsienki:
@dotnet/roslyn-compiler @jaredpar for review please :)

333fred (approved):

src/Compilers/Core/Portable/SourceGeneration/GeneratorDriverCache.cs
```diff
@@ -10,11 +10,11 @@
 
 namespace Microsoft.CodeAnalysis
 {
-    internal sealed class GeneratorDriverCache
+    internal sealed class GeneratorDriverCache(int maxCacheSize = 100)
     {
-        internal const int MaxCacheSize = 10;
+        internal readonly int MaxCacheSize = maxCacheSize;
```
jjonescz, line 15: Can this be private now?

src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
```diff
@@ -838,7 +838,11 @@ public virtual int Run(TextWriter consoleOutput, CancellationToken cancellationT
             driver = driver.RunGeneratorsAndUpdateCompilation(input, out var compilationOut, out var diagnostics);
             generatorDiagnostics.AddRange(diagnostics);
 
-            if (!disableCache)
+            // We only cache the generator driver if it produced any generated files. While its possible that it was expensive
```
jjonescz, line 841:
```suggestion
            // We only cache the generator driver if it produced any generated files. While it's possible that it was expensive
```

src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
```diff
@@ -838,7 +838,11 @@ public virtual int Run(TextWriter consoleOutput, CancellationToken cancellationT
...
+            // to calculate that nothing needed to be generated, real world usage has found that generators are generally only
+            // expensive when actually producing source. By only caching those with results, we help to keep memory usage down
+            // when it probably wouldn't improve the performance anyway.
+            if (!disableCache && driver.GetRunResult().GeneratedTrees.Any())
```
jjonescz, line 845: I'm curious about this - caching empty results increases memory usage? Shouldn't empty results be small?

jaredpar: What is the ballpark for how much the cache costs in terms of memory usage?

src/Tools/Replay/Replay.cs
```diff
@@ -263,25 +272,9 @@ static void RewriteOutputPaths(string outputDirectory, string[] args)
...
-    internal bool Wait { get; } = wait;
+    internal string TempDirectory { get; } = Path.Combine(OutputDirectory, "temp");
 }
 
-internal readonly struct BuildData(CompilerCall compilerCall, BuildResponse response)
-{
-    internal CompilerCall CompilerCall { get; } = compilerCall;
-    internal BuildResponse BuildResponse { get; } = response;
-}
+internal readonly record struct BuildData(CompilerCall CompilerCall, BuildResponse BuildResponse);
```
jaredpar, line 280: `BuildResponse` isn't equitable which is why I aovided a `record` here.

src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
chsienki: The driver holds onto copies of things like the Compilation, as well as all the intermediate nodes + state tables. So even if it doesn't produce anything, with enough files it can be a non-trivial amount of state.

PR #74441, opened 2024-07-18
Unify code paths that update ordinary/additional/analyzerconfig documents

tmat:
Refactoring of ProjectState and SolutionCompilationState.

### DIFF
--- a/src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs
+++ b/src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs
@@ -129,6 +129,14 @@ internal partial class ProjectState
         _lazyContentHashToDocumentId = AsyncLazy.Create(static (self, cancellationToken) => self.ComputeContentHashToDocumentIdAsync(cancellationToken), arg: this);
     }
 
+    public TextDocumentStates<TDocumentState> GetDocumentStates<TDocumentState>()
+        where TDocumentState : TextDocumentState
+        => (TextDocumentStates<TDocumentState>)(object)(
+            typeof(TDocumentState) == typeof(DocumentState) ? DocumentStates :
+            typeof(TDocumentState) == typeof(AdditionalDocumentState) ? AdditionalDocumentStates :
+            typeof(TDocumentState) == typeof(AnalyzerConfigDocumentState) ? AnalyzerConfigDocumentStates :
+            throw ExceptionUtilities.UnexpectedValue(typeof(TDocumentState)));
+
     private async Task<Dictionary<ImmutableArray<byte>, DocumentId>> ComputeContentHashToDocumentIdAsync(CancellationToken cancellationToken)
     {
         var result = new Dictionary<ImmutableArray<byte>, DocumentId>(ImmutableArrayComparer<byte>.Instance);
@@ -275,6 +283,14 @@ internal partial class ProjectState
 
         return doc;
     }
+
+    internal TDocumentState CreateDocument<TDocumentState>(DocumentInfo documentInfo)
+        where TDocumentState : TextDocumentState
+        => (TDocumentState)(object)(
+            typeof(TDocumentState) == typeof(DocumentState) ? CreateDocument(documentInfo, ParseOptions, new LoadTextOptions(ChecksumAlgorithm)) :
+            typeof(TDocumentState) == typeof(AdditionalDocumentState) ? new AdditionalDocumentState(LanguageServices.SolutionServices, documentInfo, new LoadTextOptions(ChecksumAlgorithm)) :
+            typeof(TDocumentState) == typeof(AnalyzerConfigDocumentState) ? new AnalyzerConfigDocumentState(LanguageServices.SolutionServices, documentInfo, new LoadTextOptions(ChecksumAlgorithm)) :
+            throw ExceptionUtilities.UnexpectedValue(typeof(TDocumentState)));
 
     public AnalyzerOptions AnalyzerOptions
         => _lazyAnalyzerOptions ??= new AnalyzerOptions(
@@ -828,26 +844,25 @@ internal partial class ProjectState
     }
 
     public ProjectState UpdateDocument(DocumentState newDocument, bool contentChanged)
-        => UpdateDocuments([newDocument], contentChanged);
+        => UpdateDocuments([DocumentStates.GetRequiredState(newDocument.Id)], [newDocument], contentChanged);
 
-    public ProjectState UpdateDocuments(ImmutableArray<DocumentState> newDocuments, bool contentChanged)
+    public ProjectState UpdateDocuments(ImmutableArray<DocumentState> oldDocuments, ImmutableArray<DocumentState> newDocuments, bool contentChanged)
     {
-        var oldDocuments = newDocuments.SelectAsArray(d => DocumentStates.GetRequiredState(d.Id));
-        if (oldDocuments.SequenceEqual(newDocuments))
-            return this;
-
-        // Must not be empty as we would have otherwise bailed out in the check above.
-        Contract.ThrowIfTrue(newDocuments.IsEmpty);
+        Contract.ThrowIfTrue(oldDocuments.IsEmpty);
+        Contract.ThrowIfFalse(oldDocuments.Length == newDocuments.Length);
+        Debug.Assert(!oldDocuments.SequenceEqual(newDocuments));
 
         var newDocumentStates = DocumentStates.SetStates(newDocuments);
 
         // When computing the latest dependent version, we just need to know how 
         GetLatestDependentVersions(
-            newDocumentStates, AdditionalDocumentStates,
+            newDocumentStates,
+            AdditionalDocumentStates,
             oldDocuments.CastArray<TextDocumentState>(),
             newDocuments.CastArray<TextDocumentState>(),
             contentChanged,
-            out var dependentDocumentVersion, out var dependentSemanticVersion);
+            out var dependentDocumentVersion,
+            out var dependentSemanticVersion);
 
         return With(
             documentStates: newDocumentStates,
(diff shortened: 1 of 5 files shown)


src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.TranslationAction_Actions.cs
```diff
@@ -84,26 +85,44 @@ public override Task<Compilation> TransformCompilationAsync(Compilation oldCompi
...
+                {
+                    generatorDriver = generatorDriver.ReplaceAdditionalText(_oldStates[i].AdditionalText, _newStates[i].AdditionalText);
+                }
 
-                return generatorDriver.ReplaceAdditionalText(oldText, newText);
+                return generatorDriver;
             }
         }
 
+        internal sealed class TouchAnalyzerConfigDocumentsAction(
```
tmat, line 110: Separate this action from `ProjectCompilationOptionsAction`.

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.TranslationAction_Actions.cs
```diff
@@ -57,22 +58,22 @@ public override GeneratorDriver TransformGeneratorDriver(GeneratorDriver generat
...
-                    return new TouchDocumentsAction(priorAction.OldProjectState, this.NewProjectState, _newStates);
+                    return new TouchDocumentsAction(priorAction.OldProjectState, NewProjectState, priorTouchAction._oldStates, _newStates);
                 }
 
                 return null;
             }
         }
 
-        internal sealed class TouchAdditionalDocumentAction(
+        internal sealed class TouchAdditionalDocumentsAction(
```
tmat, line 68: Allow multiple documents to be changed at the same time, like TouchDocumentsAction.

src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs
```diff
@@ -856,34 +869,36 @@ public ProjectState UpdateDocuments(ImmutableArray<DocumentState> newDocuments,
     }
 
     public ProjectState UpdateAdditionalDocument(AdditionalDocumentState newDocument, bool contentChanged)
+        => UpdateAdditionalDocuments([AdditionalDocumentStates.GetRequiredState(newDocument.Id)], [newDocument], contentChanged);
+
+    public ProjectState UpdateAdditionalDocuments(ImmutableArray<AdditionalDocumentState> oldDocuments, ImmutableArray<AdditionalDocumentState> newDocuments, bool contentChanged)
```
tmat, line 874: Follow the same pattern as `UpdateDocuments`

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
```diff
@@ -7,6 +7,7 @@
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Dynamic;
```
tmat, line 10: Will remove. The IDE keeps adding random usings :(

tmat:
@ToddGrun ptal

src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs
```diff
@@ -828,26 +844,25 @@ public ProjectState RemoveAllNormalDocuments()
...
-    public ProjectState UpdateDocuments(ImmutableArray<DocumentState> newDocuments, bool contentChanged)
+    public ProjectState UpdateDocuments(ImmutableArray<DocumentState> oldDocuments, ImmutableArray<DocumentState> newDocuments, bool contentChanged)
     {
-        var oldDocuments = newDocuments.SelectAsArray(d => DocumentStates.GetRequiredState(d.Id));
-        if (oldDocuments.SequenceEqual(newDocuments))
-            return this;
-
-        // Must not be empty as we would have otherwise bailed out in the check above.
-        Contract.ThrowIfTrue(newDocuments.IsEmpty);
+        Contract.ThrowIfTrue(oldDocuments.IsEmpty);
```
ToddGrun, line 851: >Contract.ThrowIfTrue(oldDocuments.IsEmpty); [](http://example.com/codeflow?start=8&length=43)

Just wanted to make sure this is a new requirement of the callers

src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs
```diff
@@ -856,34 +871,38 @@ public ProjectState UpdateDocuments(ImmutableArray<DocumentState> newDocuments,
...
-        {
-            return this;
-        }
-
-        var newDocumentStates = AnalyzerConfigDocumentStates.SetState(newDocument);
+        => UpdateAnalyzerConfigDocuments([AnalyzerConfigDocumentStates.GetRequiredState(newDocument.Id)], [newDocument]);
 
+    public ProjectState UpdateAnalyzerConfigDocuments(ImmutableArray<AnalyzerConfigDocumentState> oldDocuments, ImmutableArray<AnalyzerConfigDocumentState> newDocuments)
+    {
+        Debug.Assert(!oldDocuments.SequenceEqual(newDocuments));
```
ToddGrun, line 904: >Debug [](http://example.com/codeflow?start=8&length=5)

Intentionally not doing the contract.Throw calls like the other two methods?

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
```diff
@@ -699,48 +697,128 @@ public SolutionCompilationState RemoveAnalyzerReference(ProjectId projectId, Ana
...
+
+                        if (ReferenceEquals(oldDocumentState, newDocumentState))
+                            continue;
+
+                        newDocumentStates.Add(newDocumentState);
+                    }
+
+                    return (projectId, newDocumentStates.ToImmutableAndClear());
+                }),
+            (oldProjectState, newDocumentStates) => GetUpdateDocumentsTranslationAction(oldProjectState, newDocumentStates, contentChanged));
```
ToddGrun, line 748: >(oldProjectState, newDocumentStates) => GetUpdateDocumentsTranslationAction(oldProjectState, newDocumentStates, contentChanged)); [](http://example.com/codeflow?start=12&length=129)

Should WithDocumentStatesOfMultipleProjects take in an arg to avoid this closure?

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
```diff
@@ -1534,48 +1565,18 @@ public SolutionCompilationState RemoveAnalyzerConfigDocuments(ImmutableArray<Doc
...
+        return WithDocumentStatesOfMultipleProjects(
             documentInfos.GroupBy(d => d.Id.ProjectId).Select(g =>
             {
                 var projectId = g.Key;
-                this.SolutionState.CheckContainsProject(projectId);
-                var projectState = this.SolutionState.GetRequiredProjectState(projectId);
-                return (projectId, newDocumentStates: g.SelectAsArray(di => createDocumentState(di, projectState)));
+                SolutionState.CheckContainsProject(projectId);
+                var projectState = SolutionState.GetRequiredProjectState(projectId);
+                return (projectId, newDocumentStates: g.SelectAsArray(projectState.CreateDocument<TDocumentState>));
```
ToddGrun, line 1574: >SelectAsArray [](http://example.com/codeflow?start=56&length=13)

nit: would be nice if this used the arg version of SelectAsArray

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
```diff
@@ -1591,39 +1592,59 @@ public SolutionCompilationState RemoveAnalyzerConfigDocuments(ImmutableArray<Doc
...
 
-            if (oldProjectState == null)
-            {
-                throw new InvalidOperationException(string.Format(WorkspacesResources._0_is_not_part_of_the_workspace, documentIdsInProject.Key));
-            }
+        return newCompilationState;
+    }
 
-            using var _ = ArrayBuilder<T>.GetInstance(out var removedDocumentStates);
+    private SolutionCompilationState RemoveDocumentsFromSingleProject<T>(ProjectId projectId, ImmutableArray<DocumentId> documentIds)
```
ToddGrun, line 1601: >ImmutableArray [](http://example.com/codeflow?start=94&length=14)

nice, it looks like you got rid of a ToLookup call that was being executed twice.

ToddGrun (approved):
:shipit:

src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs
ToddGrun: I was referring to the difference between the validation done in this method and UpdateDocuments/UpdateAdditionalDocuments where those other two both do a SequenceEqual validation in addition to two Contract checks.

src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs
tmat: Not sure I follow. The contract has been there before.

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
tmat: We actually do not have TArg version of IEnumerable.SelectAsArray. If needed let's add it in another PR.

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
tmat: Not sure it's worth the extra noise. We already allocate enumerators in `GroupBy` and `Select` here. One more allocation for the closure shouldn't make much difference.

src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs
ToddGrun: But it would have returned before hitting the contract failure previously, right?

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
ToddGrun: Up to you, but this is an area that definitely shows up in allocation profiles. There's a reason almost all code in this file uses static lambdas.

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
ToddGrun: oh, is that only available at the compiler level? I'm a bit surprised if so.

ToddGrun: Yep, you're right that we don't have an IEnumerable arg version.

PR #74447, opened 2024-07-18
Add monthly smoke test to release branch official builds

RikkiGibson:
Run official builds each month on all active servicing branches starting with 17.6. This helps us catch if these infrequently run builds have been broken by infrastructure changes, and avoid scrambling when we need to build a servicing fix before a deadline.

Copied schedule from dotnet/roslyn-analyzers#7316. We are expecting if merged the job will run this Sunday so we can verify that everything is working as we expect.

@arkalyanms @phil-allen-msft

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -36,6 +36,18 @@ parameters:
   type: boolean
   default: true
 
+schedules:
+  - cron: "0 8 23-29 * 0"
+    displayName: "Monthly smoke test"
+    branches:
+      include: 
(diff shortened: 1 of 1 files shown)


333fred (approved):

azure-pipelines-official.yml
```diff
@@ -36,6 +36,18 @@ parameters:
   type: boolean
   default: true
 
+schedules:
```
jjonescz, line 39: Do we want to do this in razor as well?

PR #74461, opened 2024-07-19
Use Microsoft.Build.NoTargets instead of Microsoft.Build.Traversal

sharwell:


global.json
```diff
@@ -13,6 +13,7 @@
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "9.0.0-beta.24352.2",
     "Microsoft.DotNet.Helix.Sdk": "9.0.0-beta.24352.2",
+    "Microsoft.Build.NoTargets": "3.7.56",
     "Microsoft.Build.Traversal": "3.4.0"
```
JoeRobich, line 17: Can this entry be removed?

```suggestion
```

JoeRobich (approved):

jasonmalinowski:
Why? Can we add a comment to the PR here?

dibarbet:
Also interested - I originally wrote as a traversal project because I was under the impression that is what should be used to specify a certain set of projects to build.  Would like to know what is going wrong with it.

sharwell:
I was under the impression that Microsoft.Build.Traversal was for building projects in subfolders of the traversal project, but it seems to work for projects in arbitrary other locations as well.

PR #74463, opened 2024-07-19
Flesh out capturing of `this` into a closure in an instance extension member.

AlekseyTs:

### DIFF
--- a/src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs
+++ b/src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs
@@ -41268,6 +41268,84 @@ public interface I1
     }
 
     [Fact]
+    public void InstanceMethod_Metadata_25_LambdaCapturing()
+    {
+        var src1 = """
+public implicit extension E<T> for T where T : class, I1
+{
+    public void Method()
+    {
+        var d1 = () => this.Increment();
+        d1();
+        var d2 = () => this.EIncrement();
+        d2();
+    }
+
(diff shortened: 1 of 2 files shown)


AlekseyTs:
@jcouv, @jjonescz, @dotnet/roslyn-compiler Please review

src/Compilers/CSharp/Test/Emit3/ExtensionTypeTests.cs
```diff
@@ -41117,6 +41117,156 @@ static void Main()
...
+            //         var d2 = () => this.EIncrement();
+            Diagnostic(ErrorCode.ERR_ThisStructNotInAnonMeth, "this").WithLocation(7, 24)
+            );
+    }
+
+    [Fact]
+    public void InstanceMethod_Metadata_24_LambdaCapturing()
+    {
+        var src1 = """
+public implicit extension E<T> for T where T : I1
```
jjonescz, line 41238: What happens when when this is `extension E<T> for T where T : class, I1` (i.e., we know `T` is a reference type)?

AlekseyTs: I think in this case the code would compile successfully. The error is based on RefKind of `this` parameter and it will be None for an extended type known to be a reference type.

AlekseyTs:
@jcouv, @jjonescz, @dotnet/roslyn-compiler Please review

jcouv (approved):
LGTM Thanks (iteration 2)

PR #74470, opened 2024-07-20
Address issues with VSIX item publishing projects and FUTDC

tmat:
Workaround missing dependency publishing feature in the VS SDK.

Introduces the following convention: When a project reference publish outputs (i.e. more than standard build outputs) need to be packaged into a VSIX,
the project should define `VsixItemsOutputGroup` target and include it in `ProjectRefrence.IncludeOutputGroupsInVSIX`.

`VsixItemsOutputGroup` target may gather all the publishing items to be packaged into the VSIX using `PublishItemsOutputGroup` target,
but should not depend on `Publish` target (i.e. invoking `VsixItemsOutputGroup` target should not trigger actual publishing).

A custom target run before `CreateVsixContainer` will take care of publishing for all `ProjectReferences` with `VsixItemsOutputGroup`.

### DIFF
--- a/eng/targets/VisualStudio.targets
+++ b/eng/targets/VisualStudio.targets
@@ -151,12 +151,12 @@
     Workaround for missing dependency publishing feature in the VS SDK.
     
     Relies on the following convention: When a project reference publish outputs (i.e. more than standard build outputs) need to be packaged into a VSIX,
-    the project should define VsixItemsOutputGroup target and include it in ProjectRefrence.IncludeOutputGroupsInVSIX.
+    the project should define PublishedProjectOutputGroup target and include it in ProjectRefrence.IncludeOutputGroupsInVSIX.
 
-    VsixItemsOutputGroup target may gather all the publishing items to be packaged into the VSIX using PublishItemsOutputGroup target,
-    but should not depend on Publish target (i.e. invoking VsixItemsOutputGroup target should not trigger actual publishing).
+    PublishedProjectOutputGroup target may gather all the publishing items to be packaged into the VSIX using PublishItemsOutputGroup target,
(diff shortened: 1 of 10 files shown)


eng/targets/VisualStudio.targets
```diff
@@ -143,4 +143,32 @@
   </Target>
 
   <Import Project="GeneratePkgDef.targets" Condition="'$(MSBuildRuntimeType)' != 'Core' and '$(GeneratePkgDefFile)' == 'true' and '$(DesignTimeBuild)' != 'true' and '$(BuildingForLiveUnitTesting)' != 'true'" />
+
+  <!--
+    Workaround for missing dependency publishing feature in the VS SDK.
+    
+    Relies on the following convention: When a project reference publish outputs (i.e. more than standard build outputs) need to be packaged into a VSIX,
+    the project should define VsixItemsOutputGroup target and include it in ProjectRefrence.IncludeOutputGroupsInVSIX.
```
sharwell, line 151: I'm not a fan of the name `VsixItemsOutputGroup` here, because the actual VSIX is produced by the Build step and not the Publish step, so it's confusing to use the same name to refer specifically to Publish outputs.

tmat: Suggestions for a better name? `VsixPublishedItemsOutputGroup`?

sharwell: I guess `PublishedProjectOutputGroup` would be the closest match to `BuiltProjectOutputGroup`, unless you only intended to include a subset of the published items in the VSIX. Any of these is fine though.

arkalyanms (approved):

PR #74483, opened 2024-07-22
Move to new version of MS.DevDiv.Optimization.Data.Powershell

jaredpar:
This will _hopefully_ get us past a CG alert.


333fred (approved):

Cosifne (approved):

JoeRobich:
We should see if we can remove any of these package references now that there is a newer release of MS.DD.O.D.PS https://github.com/dotnet/roslyn/blob/9d7489b84a998578f62fadb382550bf71c078397/eng/InternalTools.props#L16-L21

PR #74486, opened 2024-07-22
Include PrepareResources before ExtensionJsonOutputGroup

Cosifne:
`ExtensionJsonOutputGroup` depends on `CoreCompile`, but xaml compilation is not included, and it might cause build error in IDE.

Include `PrepareResources ` to make sure xaml is correctly compiled.

### DIFF
--- a/src/VisualStudio/CSharp/Impl/Microsoft.VisualStudio.LanguageServices.CSharp.csproj
+++ b/src/VisualStudio/CSharp/Impl/Microsoft.VisualStudio.LanguageServices.CSharp.csproj
@@ -16,8 +16,6 @@
     <DeployExtension>false</DeployExtension>
     <VssdkCompatibleExtension>true</VssdkCompatibleExtension>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


src/VisualStudio/CSharp/Impl/Microsoft.VisualStudio.LanguageServices.CSharp.csproj
```diff
@@ -16,6 +16,8 @@
     <DeployExtension>false</DeployExtension>
     <VssdkCompatibleExtension>true</VssdkCompatibleExtension>
   </PropertyGroup>
+    <!-- Workaround to fix https://github.com/dotnet/msbuild/issues/10306 -->
+  <Target Name="ExtensionJsonOutputGroupFixed" Returns="@(ExtensionJsonOutputGroupOutput)" DependsOnTargets="PrepareResources;ExtensionJsonOutputGroup" />
```
sharwell, line 20: 💭 What about putting this in eng/targets/VisualStudio.targets?

arkalyanms (approved):

Issue #73928, opened 2024-06-11
Can't suppress NRT warnings on ternary with refs


MihaZupan:
**Version Used**: 

SDK 9.0.100-preview.6.24303.15

**Steps to Reproduce**:

While you can suppress NRT warnings on simple assignments like so
```c#
ref object myRef = ref GetNullableObjectRef()!;
ref object? myRef = ref GetObjectRef()!;
```

I've been unable to find a combination of `!`/`?`/`()` to suppress the warnings on this ternary:
```c#
void Problem(bool condition)
{
    ref object myRef = ref (condition
        ? ref GetObjectRef()!
        : ref GetNullableObjectRef()!)!;

    // Assuming these are methods whose shape I can't change
    static ref object GetObjectRef() => ref Unsafe.NullRef<object>();
    static ref object? GetNullableObjectRef() => ref Unsafe.NullRef<object?>();
}
```

[SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABADAAgwRgDoAlAVwDsAXASwFsZCBhCWgB2oBsYoBlbgN2pgYAZwDcAWABQ0gkjwocABSgRgXWgApgECBxyRyAE2o0I5AJTSA3tJz2csAGY41AKxhhKOWgE9iMC4AvI6BOJqGJmbkdg5xAPyhLgDiMJQA8sAeXgFOmhYAhLFx9iBJOKmUAHKkHBwAhuowmdmUufkFhZIyUnEA9H04AIIiIqS01OQA5jiUABaiMDj1sD5pcxBGIjgA7hsiSyJz9axLAJIG9eQA5N5gx9MwxXj48s6uWZ7elS1f7RY4IIAPnKAFVyCJ6k4GDU6rkADzuL5A/LdOJycpIryJSqwhpNX45QL5QEg97gyHQwh4hFYyjxFEWboAXyAA=)

**Diagnostic Id**:

```
warning CS8619: Nullability of reference types in value of type 'object' doesn't match target type 'object?'.
```

**Expected Behavior**:

I should be able to suppress warnings with some combination of `!`s.

**Actual Behavior**:

I can only avoid warnings via `#nullable disable`.

TessenR:
@jjonescz Is it expected that now you can suppress nullability mismatch for `foreach` with `ref` variables but not for `foreach` with regular variables?

[...]

PR #74498, opened 2024-07-23
Allow suppressing nullability warnings in more ref scenarios

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/73928.

### DIFF
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -5805,14 +5805,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                     // Prefer "not null" (correct if the operand is suppressed, otherwise we have warned above).
                     lValueAnnotation = consequenceLValue.NullableAnnotation.Meet(alternativeLValue.NullableAnnotation);
                     rValueState = consequenceRValue.State.Meet(alternativeRValue.State);
-
-                    void reportMismatchIfNecessary(BoundExpression node, TypeWithAnnotations source, TypeWithAnnotations destination)
-                    {
-                        if (!node.IsSuppressed && IsNullabilityMismatch(source, destination))
-                        {
-                            ReportNullabilityMismatchInAssignment(node.Syntax, source, destination);
-                        }
-                    }
                 }
                 else if (!node.HasErrors)
                 {
@@ -5981,6 +5973,14 @@ namespace Microsoft.CodeAnalysis.CSharp
                 Debug.Assert(isRef);
                 TypeWithAnnotations lValueType = VisitLvalueWithAnnotations(operand);
                 return (lValueType, ResultType);
+            }
+
+            void reportMismatchIfNecessary(BoundExpression node, TypeWithAnnotations source, TypeWithAnnotations destination)
+            {
+                if (!node.IsSuppressed && IsNullabilityMismatch(source, destination))
+                {
+                    ReportNullabilityMismatchInAssignment(node.Syntax, source, destination);
+                }
             }
         }
 
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -5801,10 +5801,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                     var refResultTypeWithAnnotations = TypeWithAnnotations.Create(refResultType, lValueAnnotation);
                     reportMismatchIfNecessary(originalConsequence, consequenceLValue, refResultTypeWithAnnotations);
                     reportMismatchIfNecessary(originalAlternative, alternativeLValue, refResultTypeWithAnnotations);
-
-                    // Prefer "not null" (correct if the operand is suppressed, otherwise we have warned above).
-                    lValueAnnotation = consequenceLValue.NullableAnnotation.Meet(alternativeLValue.NullableAnnotation);
-                    rValueState = consequenceRValue.State.Meet(alternativeRValue.State);
                 }
                 else if (!node.HasErrors)
                 {
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -10891,7 +10891,8 @@ namespace Microsoft.CodeAnalysis.CSharp
                         if (iterationVariable.IsRef)
                         {
                             // foreach (ref DestinationType variable in collection)
-                            if (IsNullabilityMismatch(sourceType, destinationType))
+                            if (node.Expression is not BoundConversion { Operand.IsSuppressed: true } &&
+                                IsNullabilityMismatch(sourceType, destinationType))
                             {
                                 var foreachSyntax = (ForEachStatementSyntax)node.Syntax;
                                 ReportNullabilityMismatchInAssignment(foreachSyntax.Type, sourceType, destinationType);
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -5781,7 +5781,8 @@ namespace Microsoft.CodeAnalysis.CSharp
                 Join(ref this.State, ref consequenceState);
 
                 TypeSymbol? refResultType = node.Type?.SetUnknownNullabilityForReferenceTypes();
-                if (IsNullabilityMismatch(consequenceLValue, alternativeLValue))
+                if (!originalConsequence.IsSuppressed && !originalAlternative.IsSuppressed &&
+                    IsNullabilityMismatch(consequenceLValue, alternativeLValue))
                 {
                     // l-value types must match
                     ReportNullabilityMismatchInAssignment(node.Syntax, consequenceLValue, alternativeLValue);
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -5802,9 +5802,9 @@ namespace Microsoft.CodeAnalysis.CSharp
                     reportMismatchIfNecessary(originalConsequence, consequenceLValue, refResultTypeWithAnnotations);
                     reportMismatchIfNecessary(originalAlternative, alternativeLValue, refResultTypeWithAnnotations);
 
-                    // Set "not null" top-level nullability. Consistent with equivalent method type inference or non-ref ternary scenarios.
-                    lValueAnnotation = NullableAnnotation.NotAnnotated;
-                    rValueState = NullableFlowState.NotNull;
+                    // Prefer "not null" (correct if the operand is suppressed, otherwise we have warned above).
+                    lValueAnnotation = consequenceLValue.NullableAnnotation.Meet(alternativeLValue.NullableAnnotation);
+                    rValueState = consequenceRValue.State.Meet(alternativeRValue.State);
 
                     void reportMismatchIfNecessary(BoundExpression node, TypeWithAnnotations source, TypeWithAnnotations destination)
                     {
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -5772,6 +5772,11 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             if (isRef)
             {
+                Debug.Assert(node is not BoundConditionalOperator { WasTargetTyped: true }, """
+                    Unexpected ref target typed conditional operator.
+                    Should not do type inference below in this case.
+                    """);
+
                 TypeWithAnnotations consequenceLValue;
                 TypeWithAnnotations alternativeLValue;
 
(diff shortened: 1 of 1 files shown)


src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -5781,7 +5781,8 @@ void makeAndAdjustReceiverSlot(BoundExpression receiver)
                 Join(ref this.State, ref consequenceState);
 
                 TypeSymbol? refResultType = node.Type?.SetUnknownNullabilityForReferenceTypes();
-                if (IsNullabilityMismatch(consequenceLValue, alternativeLValue))
+                if (!originalConsequence.IsSuppressed && !originalAlternative.IsSuppressed &&
+                    IsNullabilityMismatch(consequenceLValue, alternativeLValue))
```
jcouv, line 5785: Taking a look at other locations that use `IsNullabilityMismatch`, it looks like there's an other one with a similar problem (doesn't first check for suppression), in `VisitForEachIterationVariables`.
The scenario is `foreach (ref object o in refCollectionOfNullableObjects!)`. More details below.
Feels like that should work. But since that problem was not reported by any users yet,  I'm fine to punt. 
Just wanted to mention it.

```
#nullable enable

using System;
using System.Collections;

public class RefList : IEnumerable<object>
{
    private object[] _items;

    public RefList(object[] items)
    {
        _items = items;
    }

    public RefEnumerator GetEnumerator() => new RefEnumerator(_items);

    IEnumerator<object> IEnumerable<object>.GetEnumerator() => GetEnumerator();

    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();

    public struct RefEnumerator : IEnumerator<object?>
    {
        private object?[] _items;
        private int _index;

        public RefEnumerator(object?[] items)
        {
            _items = items;
            _index = -1;
        }

        public ref object? Current => ref _items[_index];

        object? IEnumerator<object?>.Current => Current;

        object? IEnumerator.Current => Current;

        public bool MoveNext()
        {
            _index++;
            return _index < _items.Length;
        }

        public void Reset() => _index = -1;

        public void Dispose() { }
    }
}

class Program
{
    static void M(RefList refList)
    {
        foreach (ref object item in refList!) // suppression doesn't affect warning
        {
        }
    }
}

```

jcouv (approved):
LGTM Thanks (iteration 1)

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -55176,6 +55176,124 @@ static void F3(bool b, IOut<object> x3, IOut<object?> y3)
...
+                //     ref object M1(bool b, ref object? x, ref object y) => ref b ? ref x : ref y;
+                Diagnostic(ErrorCode.WRN_NullabilityMismatchInAssignment, "b ? ref x : ref y").WithArguments("object?", "object").WithLocation(4, 63),
+                // (4,63): warning CS8601: Possible null reference assignment.
+                //     ref object M1(bool b, ref object? x, ref object y) => ref b ? ref x : ref y;
+                Diagnostic(ErrorCode.WRN_NullReferenceAssignment, "b ? ref x : ref y").WithLocation(4, 63),
+                // (6,63): warning CS8601: Possible null reference assignment.
+                //     ref object M3(bool b, ref object? x, ref object y) => ref b ? ref x : ref y!;
+                Diagnostic(ErrorCode.WRN_NullReferenceAssignment, "b ? ref x : ref y!").WithLocation(6, 63),
+                // (8,64): warning CS8619: Nullability of reference types in value of type 'object?' doesn't match target type 'object'.
+                //     ref object M5(bool b, ref object? x, ref object y) => ref (b ? ref x : ref y)!;
```
333fred, line 55233: Feels like this one should be suppressed too, or we should be reporting the error on `ref x` specifically.

jjonescz: > we should be reporting the error on `ref x` specifically.

It's not a problem of the `ref x` specifically though, it's simply that `ref x` and `ref y` are mismatched. Putting suppression on any of them would make the warning go away.

> Feels like this one should be suppressed too

Yeah, I think this can be done, thanks.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
jjonescz: That's a good catch, thanks. I can try to fix that as part of this PR too.

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
jcouv: This one didn't seem obvious to me. May be worth an email to LDM

jcouv (approved):
LGTM Thanks (iteration 3) but would like to get more opinions/confirmation on expected behavior for suppression on the whole ternary

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
333fred: The syntax that we gave an error on was the entire ternary. Putting a suppression on the entire ternary didn't then suppress the warning. I don't much care which of those facts need to change, but having them both be true is the root of the bug Jan is fixing in the first place.

333fred (approved):

333fred (approved):

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -5780,20 +5785,40 @@ void makeAndAdjustReceiverSlot(BoundExpression receiver)
...
+                    // Report warning for each operand that is not convertible to the target type.
+                    var refResultTypeWithAnnotations = TypeWithAnnotations.Create(refResultType, lValueAnnotation);
+                    reportMismatchIfNecessary(originalConsequence, consequenceLValue, refResultTypeWithAnnotations);
+                    reportMismatchIfNecessary(originalAlternative, alternativeLValue, refResultTypeWithAnnotations);
+
+                    // Set "not null" top-level nullability. Consistent with equivalent method type inference or non-ref ternary scenarios.
+                    lValueAnnotation = NullableAnnotation.NotAnnotated;
+                    rValueState = NullableFlowState.NotNull;
+
+                    void reportMismatchIfNecessary(BoundExpression node, TypeWithAnnotations source, TypeWithAnnotations destination)
```
jcouv, line 5809: Consider moving the local function towards the end for method. We don't follow that guideline 100% but generally do. It helps keep the main flow of logic uninterrupted.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -5780,20 +5785,40 @@ void makeAndAdjustReceiverSlot(BoundExpression receiver)
...
+                    var discardedUseSiteInfo = CompoundUseSiteInfo<AssemblySymbol>.Discarded;
+                    refResultType = BestTypeInferrer.InferBestTypeForConditionalOperator(consequencePlaceholder, alternativePlaceholder, _conversions, out _, ref discardedUseSiteInfo);
+
+                    // Report warning for each operand that is not convertible to the target type.
+                    var refResultTypeWithAnnotations = TypeWithAnnotations.Create(refResultType, lValueAnnotation);
+                    reportMismatchIfNecessary(originalConsequence, consequenceLValue, refResultTypeWithAnnotations);
+                    reportMismatchIfNecessary(originalAlternative, alternativeLValue, refResultTypeWithAnnotations);
+
+                    // Set "not null" top-level nullability. Consistent with equivalent method type inference or non-ref ternary scenarios.
+                    lValueAnnotation = NullableAnnotation.NotAnnotated;
```
jcouv, line 5806: This feels strange. In a scenario where we have `List<object?>?` vs. `List<object>?`, why would we have a not-null annotation or state?

jcouv (commented):
Done with review pass (iteration 4)

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -5780,20 +5785,32 @@ void makeAndAdjustReceiverSlot(BoundExpression receiver)
...
+                    refResultType = BestTypeInferrer.InferBestTypeForConditionalOperator(consequencePlaceholder, alternativePlaceholder, _conversions, out _, ref discardedUseSiteInfo);
+
+                    // Report warning for each operand that is not convertible to the target type.
+                    var refResultTypeWithAnnotations = TypeWithAnnotations.Create(refResultType, lValueAnnotation);
+                    reportMismatchIfNecessary(originalConsequence, consequenceLValue, refResultTypeWithAnnotations);
+                    reportMismatchIfNecessary(originalAlternative, alternativeLValue, refResultTypeWithAnnotations);
+
+                    // Prefer "not null" (correct if the operand is suppressed, otherwise we have warned above).
+                    lValueAnnotation = consequenceLValue.NullableAnnotation.Meet(alternativeLValue.NullableAnnotation);
+                    rValueState = consequenceRValue.State.Meet(alternativeRValue.State);
```
jcouv, line 5807: The usage of `Meet` for result state is surprising, I would have expected a `Join` which is used when two branches join. 
For the annotation, I think `EnsureCompatible` is correct (that's what we use when merging type arguments when there is no variance).
So I think we could remove those two lines and just rely on the logic above (initialization of `lValueAnnotation` and `rValueState` at lines 5788 and 5789). Anything wrong happens if we do that?

jcouv (commented):
Done with review pass (iteration 6). Tests not fully looked at yet

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
jjonescz: That makes sense, thanks. I must have had a mistake in the code previously when I tried that (I was seeing some weird nullability warnings).

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -53204,67 +53205,73 @@ static void F1(bool b, ref string? x1, ref string y1)
...
+        (b ? ref x3 : ref y3)/*T:IIn<string!>!*/.ToString(); // 10
         (b ? ref y3 : ref y3)/*T:IIn<string!>!*/.ToString();
     }
     static void F4(bool b, ref IOut<string?> x4, ref IOut<string> y4)
     {
-        (b ? ref x4 : ref x4)/*T:IOut<string?>!*/.P.ToString(); // 12
-        (b ? ref y4 : ref x4)/*T:IOut<string>!*/.P.ToString(); // 13
-        (b ? ref x4 : ref y4)/*T:IOut<string>!*/.P.ToString(); // 14
+        (b ? ref x4 : ref x4)/*T:IOut<string?>!*/.P.ToString(); // 11
+        (b ? ref y4 : ref x4)/*T:IOut<string?>!*/.P.ToString(); // 12, 13
```
jcouv, line 53222: I compared this test to an equivalent test with type inference and I'm seeing different warnings. It would be good to understand why and whether it is unexpected. My expectation is that they would match.

```
interface I<T> { T P { get; } }
interface IIn<in T> { }
interface IOut<out T> { T P { get; } }
class C
{
    static void F1(bool b, ref string? x1, ref string y1)
    {
        M(ref x1, ref x1).ToString(); // 1 on M
        M(ref x1, ref y1).ToString(); // 2 on x1
        M(ref y1, ref x1).ToString();
        M(ref y1, ref y1).ToString();
    }
    static void F2(bool b, ref I<string?> x2, ref I<string> y2)
    {
        M(ref x2, ref x2).P.ToString(); // 6 on P
        M(ref y2, ref x2).P.ToString(); // 7 on x2
        M(ref x2, ref y2).P.ToString(); // 8 on x2
        M(ref y2, ref y2).P.ToString();
    }
    static void F3(bool b, ref IIn<string?> x3, ref IIn<string> y3)
    {
        M(ref x3, ref x3).ToString();
        M(ref y3, ref x3).ToString(); // 9 on x3
        M(ref x3, ref y3).ToString(); // 10 on x3
        M(ref y3, ref y3).ToString();
    }
    static void F4(bool b, ref IOut<string?> x4, ref IOut<string> y4)
    {
        M(ref x4, ref x4).P.ToString(); // 11 on M
        M(ref y4, ref x4).P.ToString(); // 12 on x4
        M(ref x4, ref y4).P.ToString(); // 14 on x4
        M(ref y4, ref y4).P.ToString();
    }
    static T M<T>(ref T t1, ref T t2) => throw null!;
}
```
[sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAFxANwKYCcoDMCecAmIA1AD6wIZYQDGqABAJIA8AKgHx0DedLdAClzoBzVAgDcdAL5SAsAChylGvQYMYTWDw7dJ8xeiq1GAeQCuCJgHtzWwbwHcR4qbLkABAEx0AwvM7y6QLo3AEYANmCAFjoAMRCACmBLSwAbOmA4OnRULGCQgAYAfjoADxDM7NzQ/LocEIBKAKD/OSC2ugBZeMrS8qyc3vqAOhZLAGUETBghePqJAHp5uhC6SxhOpvbO7oGyioG64dGJqZm5ukW6LzXezfaunrr93LKj8cnYM7E7toeDvseDRG71Os2+rSCughgVCETc0RiHkSyTSGX6uWY1UKHBKHmejCY1Q4OA8jWhXB+QT+Lzx6NKpKGfGBJ0+YIuSwiNz4lMC1NqtJ6uOGTOOH2mbMuAHZVutcTztoKBQcGSKQazzpcABwy+nyvkk/Ek4XMsVfO5QtqwqKxADMyNS6XxqnUWJxNqdakJBWJNrJbRaW15Oxe7rpJV9JtBcz1wdqocFEdFUYWSwAnDrwzGE4bE2rxRqlgUMzaswd42W3iz8+C2haglb4bFIvbUU6zBZXaVIm3zF78sTIn7mqWXt2w4PGZH1Snlisbh0R7Ux4KJ6qq192ctrrLIouSsuDqup9XNyFojd94ucAfctfjUnp+a7lbeB1WGxY7wEACBl/SXQAF4OAQAALdBLAAdzoGBTBSFIAEJwUkIA)

jjonescz: That's related to the previous discussion. When there is a mismatch (`ref string` vs `ref string?` arguments), type inference infers the method as `M<string>`. But for the ternary, we combine the annotations via `EnsureCompatible`, i.e., it becomes `ref string?`. The flow state might be also different for similar reasons.

I'm not sure we should strive for precise equivalence between type inference and the ternary. The ternary is a conditional construct whose branches are joined together which the method call is not.

Btw, I don't see any difference on this line (there is no warning either way):
> M(ref x3, ref x3).ToString(); // (difference)

jcouv: It took me a while to understand, but I think you're correct. The difference in warnings is expected.

Here's what I think is happening for the following ref ternary scenario:
```
static void F1(bool b, ref string? x1, ref string y1)
{
    (b ? ref x1 : ref y1)/*T:string?*/.ToString(); // 2 on x1, 3 on dereference
}
```
The type for the ternary is `string!` (result from `EnsureCompatible` between `!` and `?`).
But when we analyze the dereference, we look at the state (maybe-null) so we warn on the dereference.

nit: Consider adding scenarios showing that result type. For example, doing a ref assignment to the ternary `(b ? ref x1 : ref y1) = ref ...;`.

In contrast for the method inference scenario:
```
static void F1(bool b, ref string? x1, ref string y1)
{
    M(ref x1, ref y1).ToString(); // 2 on x1 (which is correct in my opinion)
}
```
We also produce the same result type (`string!`, from `EnsureCompatible`) but we don't have a tracked state so we just derive a state from the type (not-null) so there is no warning.

In short, I think the logic is good (we infer the same best type in both cases).

> The ternary is a conditional construct whose branches are joined together which the method call is not.

This comment made me think of another scenario.

```
#nullable enable 

var x = (bool b, ref object o1, ref object? o2) =>
{
    if (b)
    {
        return ref o1;
    }
    return ref o2; // bug: missing warning
};

object x1 = new object();
object? x2 = new object();
bool b = true;

x(b, ref x1, ref x2) = null; // bug: missing warning (inferred delegate should return `ref string!` but actually returns `ref string~`)

ref object? y1 = ref x(b, ref x1, ref x2); // bug: missing warning (inferred delegate should return `ref string!` but actually returns `ref string~`)

ref object y2 = ref x(b, ref x1, ref x2);
```

I would expect that scenario (multiple ref-returns) to behave the same as type inference with multiple ref arguments and as ref ternaries. The return type from `ref string?` and `ref string!` should be `ref string!`.
Unfortunately, there seems to be a bug with `var` local analysis (we're using an anonymous delegate type from initial binding, with `objet~` return) so the current implementation is not very useful at the moment for making comparisons.
Filed https://github.com/dotnet/roslyn/issues/75009

jcouv (approved):
LGTM Thanks (iteration 7)

jjonescz:

333fred (approved):

PR #74507, opened 2024-07-23
Don't try and add the result diagnostics if they are default or empty

chsienki:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2152389 
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2170891

This is a trivial fix and follow up from https://github.com/dotnet/roslyn/pull/74034

Don't know the IDE code well enough to craft a test to cover this though


chsienki:
@dotnet/roslyn-ide for review please. Not sure where / how to add tests for this case.

PR #74509, opened 2024-07-23
Avoid assert in emit layer when intercepted call implicitly captures receiver to temp

### LINKED ISSUE #71657: Intercepting a struct rvalue instance method call with an extension method trips a debug assertion in the compiler

RikkiGibson:
Closes #71657

Extracted from #74363. This part of the change I feel confident is one we should be making. The other part I want to revisit and PR separately.

This PR is preserving an already-shipped behavior without changing it. It is just changing the exact way we get that behavior by adding the temp in lowering, when we know we are in the specific situation we want to permit, rather than in the more general code path in emit, which was tripping a debug assertion.

### DIFF
--- a/src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs
+++ b/src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs
@@ -7842,4 +7842,119 @@ partial struct CustomHandler
         var method = model.GetInterceptorMethod(node);
         Assert.Equal("void SC.M1(this C c)", method.ToTestDisplayString());
     }
+
+    [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/71657")]
+    public void Receiver_RefReturn_NotCapturedToTemp()
+    {
+        var source = CSharpTestSource.Parse("""
+            using System;
+
+            public struct S
+            {
+                public int F;
+
+                static S s;
+                static ref S RS() => ref s;
+
+                void M() => throw null!;
+
+                public static void Main()
+                {
+                    RS().F = 1;
+                    Console.Write(RS().F);
+                    RS().M();
+                    Console.Write(RS().F);
+                }
+            }
+            """, "Program.cs", RegularWithInterceptors);
+
+        var comp = CreateCompilation(source);
+        var model = comp.GetSemanticModel(source);
+        var node = source.GetRoot().DescendantNodes().OfType<InvocationExpressionSyntax>().Single(i => i.ToString() == "RS().M()");
+        var locationSpecifier = model.GetInterceptableLocation(node)!;
+
+        var interceptors = CSharpTestSource.Parse($$"""
+            public static class SC
+            {
+                {{locationSpecifier.GetInterceptsLocationAttributeSyntax()}}
+                public static void M1(this ref S s) => s.F = 2;
+            }
+            """, "Interceptors.cs", RegularWithInterceptors);
+
+        var verifier = CompileAndVerify([source, interceptors, s_attributesTree], expectedOutput: "12");
+        verifier.VerifyDiagnostics();
+        verifier.VerifyIL("S.Main", """
+            {
+              // Code size       52 (0x34)
+              .maxstack  2
+              IL_0000:  call       "ref S S.RS()"
+              IL_0005:  ldc.i4.1
+              IL_0006:  stfld      "int S.F"
+              IL_000b:  call       "ref S S.RS()"
+              IL_0010:  ldfld      "int S.F"
+              IL_0015:  call       "void System.Console.Write(int)"
+              IL_001a:  call       "ref S S.RS()"
+              IL_001f:  call       "void SC.M1(ref S)"
+              IL_0024:  call       "ref S S.RS()"
+              IL_0029:  ldfld      "int S.F"
+              IL_002e:  call       "void System.Console.Write(int)"
+              IL_0033:  ret
+            }
+            """);
+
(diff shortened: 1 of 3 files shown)


src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs
```diff
@@ -7513,4 +7513,333 @@ static class Interceptors
...
+        var node = source.GetRoot().DescendantNodes().OfType<InvocationExpressionSyntax>().Last();
+        var locationSpecifier = model.GetInterceptableLocation(node)!;
+
+        var interceptors = CSharpTestSource.Parse($$"""
+            using System;
+
+            public static class C
+            {
+                {{locationSpecifier.GetInterceptsLocationAttributeSyntax()}}
+                public static void M1(this ref S s) => Console.WriteLine(1);
```
333fred, line 7545: Consider using `s` in a way that mutates, to show that we are actually capturing by ref and that the results are observable. The same for the rest of the test methods.

RikkiGibson: I am wondering how exactly should I demonstrate this? Do you want me to try and read the managed reference at the call site (e.g. by returning 'ref S s' by-reference)?

333fred: Yeah, that would work.

RikkiGibson: > read the field from the local

The lowering temp which is being passed by-ref to the call can't be directly used in user code. We would have to return a ref to that temp if we wanted the call site to be able to read it.

333fred: Am I misunderstanding the original bug? I expect something like this would be accessible, and would have triggered the emit assert:

```cs
var s = new S();
s.M(); // Intercepted, and `s` was mutated
```

RikkiGibson: That case would not trigger the assert. The assert occurs when the compiler needs to create a temp for the receiver to use as an argument for a by-ref `this` parameter, such as `new S().M()`. It won't occur when the receiver argument is already something we can pass by-reference for the `this` parameter.

The emit layer is fine with creating a temp for an instance call receiver in the `new S().M()` case. However, when we intercept the call with a method like `static Extension(this ref S s)`, it basically sees the call as `Extension(new S())`, doesn't expect to create a temp for `new S()`, and will assert.

333fred: Could we observe this with a `ref`-returning method? IE, something like this?

```cs
var s = new S();
Identity(ref s).M();
s.ObserveMutated();

ref S Identity(ref S s) => ref s;
```

RikkiGibson: I think in this case the compiler also doesn't need to capture the receiver 'Identity(ref s)' to temp. But we can go ahead and add the test to show that we use the reference properly here.

jaredpar: Think we could observe it this way 

```csharp
using System;

Span<int> Test() {
  // should error in both intercepted and non intercepted case
  return new S().M();
}

ref struct S {
    [UnscopedRef]
    public Span<int> M() => default;
}

static class D
{
    [InterceptsLocation(1, "...")]
    public static Span<int> Interceptor(this ref S s)
}
```

333fred (approved):

jaredpar (approved):

Issue #74427, opened 2024-07-17
Debug assert triggered for interpolated string convserion


bernd5:
**Version Used**: 

Compile the following code with debug asserts enabled:

```csharp
class C
{
    static void Main()
    {
        var y = (C1)$"dog"; // works
        System.Console.WriteLine(y);
    }
}

class C1
{
    System.FormattableString x;
    public C1(System.FormattableString x)
    {
        this.x = x;
    }

    public static implicit operator C1(System.FormattableString x) => new C1(x);

    public override string ToString()
    {
        return ("C1:") + x.ToString();
    }
}
```

**Expected Behavior**:
Compile without error

**Actual Behavior**:
Debug.Assert triggered at: [ConversionsBase.cs#L674](https://github.com/dotnet/roslyn/blob/b64b16059bc2856a11003d6762f59b6079582297/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs#L674)

PR #74529, opened 2024-07-23
Include InterpolatedString conversions in standard implicit conversion from expression assert

333fred:
We extend the spec to replicate native compiler bugs for implicit conversions from expression; FormattableString is one that we've long included, but haven't actually tested, apparently. Commit 1 is a copy/paste to move a regular function to a local function. Commit 2 is the actual change. Fixes https://github.com/dotnet/roslyn/issues/74427.

### DIFF
--- a/src/Compilers/CSharp/Test/Semantic/Semantics/InterpolationTests.cs
+++ b/src/Compilers/CSharp/Test/Semantic/Semantics/InterpolationTests.cs
@@ -18917,7 +18917,7 @@ literal:literal
         [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/74427")]
         public void InterpolatedStringAsInputToUserDefinedConversion_01()
         {
-            var source = $$"""
+            var source = """
                 class C
                 {
                     static void Main()
@@ -18950,12 +18950,12 @@ literal:literal
         [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/74427")]
         public void InterpolatedStringAsInputToUserDefinedConversion_02()
         {
-            var source = $$"""
+            var source = """
                 class C
                 {
                     static void Main()
                     {
-                        var y = (C1)$"dog"; // works
+                        var y = (C1)$"dog";
                         System.Console.WriteLine(y);
                     }
                 }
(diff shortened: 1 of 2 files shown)


chsienki (approved):

src/Compilers/CSharp/Test/Semantic/Semantics/InterpolationTests.cs
```diff
@@ -18913,5 +18913,75 @@ .maxstack  6
 }
 """);
         }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/74427")]
+        public void InterpolatedStringAsInputToUserDefinedConversion_01()
+        {
+            var source = $$"""
```
jjonescz, line 18920: Nit: 
```suggestion
            var source = """
```

src/Compilers/CSharp/Test/Semantic/Semantics/InterpolationTests.cs
```diff
@@ -18913,5 +18913,75 @@ .maxstack  6
...
+                }
+                """;
+
+            CompileAndVerify(source, expectedOutput: "C1:dog");
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/74427")]
+        public void InterpolatedStringAsInputToUserDefinedConversion_02()
+        {
+            var source = $$"""
```
jjonescz, line 18953:
```suggestion
            var source = """
```

src/Compilers/CSharp/Test/Semantic/Semantics/InterpolationTests.cs
```diff
@@ -18913,5 +18913,75 @@ .maxstack  6
...
+                class C
+                {
+                    static void Main()
+                    {
+                        var y = (C1)$"dog"; // works
```
jjonescz, line 18958: This `// works` comment seems incorrect since the source actually doesn't compile.

jjonescz (approved):

src/Compilers/CSharp/Test/Semantic/Semantics/InterpolationTests.cs
333fred:
```suggestion
                        var y = (C1)$"dog";
```

PR #74530, opened 2024-07-23
Cleanup LSP error reporting

dibarbet:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1922370

We were reporting connection lost exceptions when the server shutdown due to refresh notifications being sent during shutdown (they are triggered by events that are not a part of the shutdown handling in the queue).

This modifies the code to catch connection lost exceptions from refresh notifications.  Other parts of the system already handle connection loss issues and report failures.

Additionally fixes an issue where we would not report NFW for mutating LSP requests.

### DIFF
--- a/src/LanguageServer/Protocol/RoslynRequestExecutionQueue.cs
+++ b/src/LanguageServer/Protocol/RoslynRequestExecutionQueue.cs
@@ -2,9 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Globalization;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.ErrorReporting;
 using Microsoft.CodeAnalysis.LanguageServer.Handler;
+using Microsoft.CodeAnalysis.Shared.TestHooks;
 using Microsoft.CommonLanguageServerProtocol.Framework;
 using Roslyn.Utilities;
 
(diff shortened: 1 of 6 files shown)


src/LanguageServer/Protocol/RoslynRequestExecutionQueue.cs
```diff
@@ -25,15 +27,24 @@ public RoslynRequestExecutionQueue(AbstractLanguageServer<RequestContext> langua
             _initializeManager = languageServer.GetLspServices().GetRequiredService<IInitializeManager>();
         }
 
-        public override Task WrapStartRequestTaskAsync(Task nonMutatingRequestTask, bool rethrowExceptions)
+        public override async Task WrapStartRequestTaskAsync(Task nonMutatingRequestTask, bool rethrowExceptions)
         {
             if (rethrowExceptions)
             {
-                return nonMutatingRequestTask;
+                try
```
dibarbet, line 34: fixes a bug here where we would only report NFW for non-mutating requests.  If a mutating request failed we would never get fault telemetry for it.

src/LanguageServer/Protocol/RoslynRequestExecutionQueue.cs
```diff
@@ -25,15 +27,24 @@ public RoslynRequestExecutionQueue(AbstractLanguageServer<RequestContext> langua
...
+                {
+                    await nonMutatingRequestTask.ConfigureAwait(false);
+                }
+                // If we had an exception, we want to record a NFW for it AND propogate it out to the queue so it can be handled appropriately.
+                catch (Exception ex) when (FatalError.ReportAndPropagateUnlessCanceled(ex, ErrorSeverity.Critical))
```
dibarbet, line 39: i believe this is the correct pattern to report and throw, but let me know if not.

src/LanguageServer/ProtocolUnitTests/HandlerTests.cs
```diff
@@ -135,6 +137,62 @@ public async Task ShutsdownIfDeserializationFailsOnMutatingRequest(bool mutating
...
+        {
+            await using var server = await CreateTestLspServerAsync("", mutatingLspWorkspace);
+
+            var request = new TestRequestWithDocument(new TextDocumentIdentifier
+            {
+                Uri = ProtocolConversions.CreateAbsoluteUri(@"C:\test.cs")
+            });
+
+            var didReport = false;
+            FatalError.OverwriteHandler((exception, severity, dumps) =>
```
dibarbet, line 151: tests to verify NFW handler is called for mutating and non-mutating requests

JoeRobich (approved):

PR #74541, opened 2024-07-24
Report fading locations as part of the main diagnostic

sharwell:
* Add assertions that `WellKnownDiagnosticTags.Unnecessary` is used correctly.
* Include fading locations as part of the main diagnostic instead of reporting them separately.

Submitted as draft to get a list of tests that need to be updated.


DoctorKrolic:
Fixes https://github.com/dotnet/roslyn/issues/71234?

sharwell:
@DoctorKrolic Yes, once complete this should fix most of the remaining cases like #71234. Note that the light bulb may not be visible at fading locations outside the actual diagnostic span; this is acceptable since that is the way the engine is currently designed.

Issue #73215, opened 2024-04-24
Roslyn codegen impacting array variance exception behavior


stephentoub:
This code throws an ArrayTypeMismatchException:
```C#
Set(new Program[1]);

static void Set(object[] array)
{
    ref var o = ref array[0];
    Use(ref o);

    static void Use(ref object o) { }
}
```
[SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwGUYMAKPGAdwAIABARgHYBtBgXQEoBuAWACgBjJPRQ0S5CMABWMMBhbsaAQyhRlAT04CA3gJoGasAGY0AbqpoQaAXiMxTq9RpYAGdn36GaAVQDOMGQmVjwC+obCor4BQQ5WMnIYITQ6NAC+AmlAA)
```asm
; Method Program:<<Main>$>g__Set|0_0(System.Object[]) (FullOpts)
G_M000_IG01:                ;; offset=0x0000
       sub      rsp, 40

G_M000_IG02:                ;; offset=0x0004
       xor      edx, edx
       mov      r8, 0x7FFA8E1A5FA8
       call     CORINFO_HELP_LDELEMA_REF
       nop      

G_M000_IG03:                ;; offset=0x0016
       add      rsp, 40
       ret      
; Total bytes of code: 27
```

But this code does not:
```C#
Set(new Program[1]);

static void Set(object[] array)
{
    ref var o = ref array[0];
}
```
[SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwGUYMAKPGAdwAIABARgHYBtBgXQEoBuAWACgBjJPRQ0S5CMABWMMBhbsaAQyhRlAT04CA3gJoGasAGY0AbqpoQaAXiMxTq9RpYAGdn34BfIA)
```asm
; Method Program:<<Main>$>g__Set|0_0(System.Object[]) (FullOpts)
G_M000_IG01:                ;; offset=0x0000
       sub      rsp, 40

G_M000_IG02:                ;; offset=0x0004
       cmp      dword ptr [rcx+0x08], 0
       jbe      SHORT G_M000_IG04

G_M000_IG03:                ;; offset=0x000A
       add      rsp, 40
       ret      

G_M000_IG04:                ;; offset=0x000F
       call     CORINFO_HELP_RNGCHKFAIL
       int3     
; Total bytes of code: 21
```

[...]

PR #74553, opened 2024-07-25
Avoid optimizing away ref array access

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/73215.

### DIFF
--- a/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
+++ b/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
@@ -659,13 +659,15 @@
 
     <Field Name="Expression" Type="BoundExpression"/>
     <Field Name="Indices" Type="ImmutableArray&lt;BoundExpression&gt;"/>
+  </Node>
 
-    <!--
-      Ref array access has side effects (element type mismatch check)
-      so if it remains standalone (e.g., because the result is discarded)
-      we need to know if the original assignment was ref to emit it correctly.
-    -->
-    <Field Name="IsRef" Type="bool" />
+  <!--
+    Represents an array access that comes from the RHS of an unused ref (non-readonly) assignment.
+
+    Needed so we can emit with correct side effects (runtime element type check).
+  -->
+  <Node Name="BoundRefArrayAccess" Base="BoundExpression">
+    <Field Name="ArrayAccess" Type="BoundArrayAccess" Null="disallow"/>
   </Node>
 
   <!--
--- a/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
+++ b/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
@@ -663,10 +663,10 @@
 
   <!--
     Represents an array access that comes from the RHS of an unused ref (non-readonly) assignment.
-
     Needed so we can emit with correct side effects (runtime element type check).
   -->
   <Node Name="BoundRefArrayAccess" Base="BoundExpression">
+    <Field Name="Type" Type="TypeSymbol?" Override="true" Null="always"/>
     <Field Name="ArrayAccess" Type="BoundArrayAccess" Null="disallow"/>
   </Node>
 
--- a/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
+++ b/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
@@ -659,6 +659,13 @@
 
     <Field Name="Expression" Type="BoundExpression"/>
     <Field Name="Indices" Type="ImmutableArray&lt;BoundExpression&gt;"/>
+
+    <!--
+      Ref array access has side effects (element type mismatch check)
+      so if it remains standalone (e.g., because the result is discarded)
+      we need to know if the original assignment was ref to emit it correctly.
+    -->
+    <Field Name="IsRef" Type="bool" />
   </Node>
 
   <!--
--- a/src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
+++ b/src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
@@ -1003,6 +1003,14 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
 
         private void EmitArrayElementLoad(BoundArrayAccess arrayAccess, bool used)
         {
+            if (arrayAccess.IsRef)
+            {
+                Debug.Assert(!used);
+                EmitArrayElementAddress(arrayAccess, AddressKind.Writeable);
+                EmitPopIfUnused(used);
+                return;
+            }
+
             EmitExpression(arrayAccess.Expression, used: true);
             EmitArrayIndices(arrayAccess.Indices);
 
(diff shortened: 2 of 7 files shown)


src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_AssignmentOperator.cs
```diff
@@ -229,6 +229,11 @@ private BoundExpression VisitAssignmentOperator(BoundAssignmentOperator node, bo
 
                 case BoundKind.DiscardExpression:
                     {
+                        if (isRef && rewrittenRight is BoundArrayAccess arrayAccess)
+                        {
+                            return arrayAccess.Update(isRef: true);
```
AlekseyTs, line 234: >return arrayAccess.Update(isRef: true); [](http://example.com/codeflow?start=28&length=39)

Is there a good reason to adjust behavior for this scenario? Do we have any debug/release difference in behavior without the change?

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
```diff
@@ -1003,6 +1003,14 @@ parameters[i].RefKind switch
 
         private void EmitArrayElementLoad(BoundArrayAccess arrayAccess, bool used)
         {
+            if (arrayAccess.IsRef)
+            {
+                Debug.Assert(!used);
+                EmitArrayElementAddress(arrayAccess, AddressKind.Writeable);
+                EmitPopIfUnused(used);
```
AlekseyTs, line 1010: >EmitPopIfUnused(used); [](http://example.com/codeflow?start=16&length=22)

I think we cannot leave reference on the stack when a value is expected. Therefore, we should always pop and throw if ```used``` is set

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_AssignmentOperator.cs
jjonescz: > Do we have any debug/release difference in behavior without the change?

No, there is no debug/release difference neither before nor after the change. Before the change both debug and release emitted `ldelem` or similar; after the change both debug and release emit `ldelema`.

> Is there a good reason to adjust behavior for this scenario?

Using `ref array[x]` and then not using the result (either by not using the local or discarding it explicitly) feels like something that should behave equivalently.

Although now I realized that in case of a ref-readonly assignment we also emit `readonly.ldelema` which doesn't do the runtime element type check, and discards could be considered readonly as well.

So I agree, I will revert this part of the change.

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
```diff
@@ -1003,6 +1003,14 @@ parameters[i].RefKind switch
 
         private void EmitArrayElementLoad(BoundArrayAccess arrayAccess, bool used)
         {
+            if (arrayAccess.IsRef)
```
AlekseyTs, line 1006: >if (arrayAccess.IsRef) [](http://example.com/codeflow?start=12&length=22)

I think it would be better to add a special node wrapping ```BoundArrayAccess``` rather than adding a new property with unclear effect that could be set improperly. The code in the `if` would go into emit helper for the new node.

src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
```diff
@@ -661,6 +661,15 @@
     <Field Name="Indices" Type="ImmutableArray&lt;BoundExpression&gt;"/>
   </Node>
 
+  <!--
+    Represents an array access that comes from the RHS of an unused ref (non-readonly) assignment.
+
+    Needed so we can emit with correct side effects (runtime element type check).
+  -->
+  <Node Name="BoundRefArrayAccess" Base="BoundExpression">
+    <Field Name="ArrayAccess" Type="BoundArrayAccess" Null="disallow"/>
```
AlekseyTs, line 670: <Field Name="Type" Type="TypeSymbol?" Override="true" Null="always"/>

src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
```diff
@@ -661,6 +661,15 @@
     <Field Name="Indices" Type="ImmutableArray&lt;BoundExpression&gt;"/>
   </Node>
 
+  <!--
+    Represents an array access that comes from the RHS of an unused ref (non-readonly) assignment.
+
+    Needed so we can emit with correct side effects (runtime element type check).
```
AlekseyTs, line 667: >Needed so we can emit with correct side effects (runtime element type check). [](http://example.com/codeflow?start=4&length=77)

Is the empty line above necessary?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenRefLocalTests.cs
```diff
@@ -1669,6 +1669,366 @@ .maxstack  2
...
 }");
+
+            CompileAndVerify(text, options: TestOptions.ReleaseDll).VerifyIL("Program.M()", @"
+{
+  // Code size       14 (0xe)
+  .maxstack  2
+  IL_0000:  ldc.i4.1
+  IL_0001:  newarr     ""int""
+  IL_0006:  ldc.i4.0
+  IL_0007:  ldelema    ""int""
```
AlekseyTs, line 1680: >IL_0007:  ldelema    ""int"" [](http://example.com/codeflow?start=2&length=28)

Is there a change in behavior for this scenario? I think it is not necessary because when array element is a value type it is not subject for variance and type checks are not necessary.

AlekseyTs:

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
```diff
@@ -1099,6 +1103,25 @@ private void EmitArrayElementLoad(BoundArrayAccess arrayAccess, bool used)
...
 
+        private void EmitArrayElementRefLoad(BoundRefArrayAccess refArrayAccess, bool used)
+        {
+            if (used)
+            {
+                throw ExceptionUtilities.Unreachable();
+            }
+
+            // Value types do not need runtime element type checks.
+            if (refArrayAccess.ArrayAccess.Type.IsValueType)
```
AlekseyTs, line 1114: >refArrayAccess.ArrayAccess.Type.IsValueType [](http://example.com/codeflow?start=16&length=43)

Why not put this check to Optimizer and avoid special node creation?

jjonescz: Seemed like emit-layer concern about which instruction to emit, but it makes sense to avoid special node creation.

AlekseyTs: We are checking `readonly ref` case in Optimizer, but this is also about an instruction choice

AlekseyTs (approved):
LGTM (commit 6)

jjonescz:

333fred (approved):

333fred:
Rerunning CI since it's been a week, will merge when green.

PR #74564, opened 2024-07-25
Rename Suggestions: Fix getting of the context string, gracefully handle errors

### LINKED ISSUE #74545: Copilot Rename Suggestions: No suggestions for identifiers referenced in methods that touch the end of file
**Version Used**: 17.12 main

**Steps to Reproduce**:

1. Create a class whose ending bracket is the last character of the file (there is no newline at the end of file)
2. Attempt to get rename suggestions
3. Observe that nothing happens. 
image

Workaround:
Add newline at the end of file:
image

This issue was introduced with #73873 
Please assign to me (if possible), I'll create a PR with fix

AmadeusW:
There was an edge case in code that gets context (references, definitions) for copilot rename suggestions.
When a reference/definition touched the end of file (because there was no empty line at the end of file), `ArgumentOutOfRangeException` was thrown from within `GetRenameContextAsync`, and Roslyn ended up not making the rename request at all.

The solution is twofold:
1. Gracefully handle errors. If context cannot be obtained, make a request without context
2. Fix the code that gets context span. It was incorrectly attempting to get an extra trailing character, which was typically `\r` on Windows machines (and therefore not visible in logs).

Fixes #74545 and DevDiv 2173894

### DIFF
--- a/src/EditorFeatures/CSharp/InlineRename/CSharpEditorInlineRenameService.cs
+++ b/src/EditorFeatures/CSharp/InlineRename/CSharpEditorInlineRenameService.cs
@@ -148,7 +148,7 @@ internal sealed class CSharpEditorInlineRenameService([ImportMany] IEnumerable<I
             // expand to select the corresponding lines completely.
             startPosition = documentText.Lines[startLine].Start;
             endPosition = documentText.Lines[endLine].End;
-            var length = endPosition - startPosition + 1;
+            var length = endPosition - startPosition;
 
             surroundingSpanOfInterest = new TextSpan(startPosition, length);
 
--- a/src/EditorFeatures/CSharp/InlineRename/CSharpEditorInlineRenameService.cs
+++ b/src/EditorFeatures/CSharp/InlineRename/CSharpEditorInlineRenameService.cs
@@ -9,7 +9,6 @@ using System.Composition;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Editor.Implementation.InlineRename;
 using Microsoft.CodeAnalysis.GoToDefinition;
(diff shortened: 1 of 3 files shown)


src/EditorFeatures/Core.Wpf/InlineRename/UI/SmartRename/SmartRenameViewModel.cs
```diff
@@ -171,14 +172,21 @@ await Task.Delay(_smartRenameSession.AutomaticFetchDelay, cancellationToken)
...
+                var editorRenameService = document.GetRequiredLanguageService<IEditorInlineRenameService>();
+                var renameLocations = await this.BaseViewModel.Session.AllRenameLocationsTask.JoinAsync(cancellationToken)
+                    .ConfigureAwait(false);
+                var context = await editorRenameService.GetRenameContextAsync(this.BaseViewModel.Session.RenameInfo, renameLocations, cancellationToken)
+                    .ConfigureAwait(false);
+                smartRenameContext = ImmutableDictionary.CreateRange<string, string[]>(
+                    context
+                    .Select(n => new KeyValuePair<string, string[]>(n.Key, n.Value.ToArray())));
+            }
+            catch (Exception e) when (FatalError.ReportAndCatch(e, ErrorSeverity.Diagnostic))
```
AmadeusW, line 186: I tried both `ErrorSeverity.Diagnostic` and `ErrorSeverity.General` but did not see any new telemetry. What do you recommend I use here?

Cosifne: `Diagnostic` should be OK.
I think it should be classified as NonFatalWaston event

src/EditorFeatures/CSharp/InlineRename/CSharpEditorInlineRenameService.cs
```diff
@@ -148,7 +148,7 @@ void AddSpanOfInterest(SourceText documentText, TextSpan fallbackSpan, TextSpan?
             // expand to select the corresponding lines completely.
             startPosition = documentText.Lines[startLine].Start;
             endPosition = documentText.Lines[endLine].End;
-            var length = endPosition - startPosition + 1;
+            var length = endPosition - startPosition;
```
Cosifne, line 151: 🤔Do you feel it needs to be covered by a test?
either integration test(is querying copilot in CI allowed?)/unit test (might be lighter, and assert `GetRenameContextAsync` directly, but we might need to hard-code the result in test)

AmadeusW: Definitely. We synced over message, @Cosifne is helping set up the test

Cosifne (approved):

PR #74590, opened 2024-07-29
Automatically tag PRs which may affect the C# extension.

JoeRobich:
It can be tedious to determine what changes in Roslyn are relevant to the C# extension. The idea is that changes under these folders may change the behavior of the Roslyn LSP and thus impact the C# extension. PR authors may remove the label if they do not believe their change is impactful.

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -146,6 +146,28 @@ configuration:
       triggerOnOwnActions: false
     - if:
       - payloadType: Pull_Request
+      - filesMatchPattern:
+          pattern: src/(Analyzers|CodeStyle|Features|LanguageServer|Workspaces)/.*\.(cs|vb)
+      - not:
+          isActivitySender:
+            user: dotnet-bot
+            issueAuthor: False
+      - or:
+        - isAction:
+            action: Opened
+        - isAction:
+            action: Synchronize
+      - and:
+        - not:
+            hasLabel:
+              label: VSCode
+      then:
+      - addLabel:
+          label: VSCode
+      description: Adds "VSCode" tag on PRs which may affect C# extension.
+      triggerOnOwnActions: false
+    - if:
+      - payloadType: Pull_Request
       - isPullRequest
       - filesMatchPattern:
           pattern: '[xX][aA][mM][lL]$'
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -8,6 +8,7 @@ where:
 configuration:
   resourceManagementConfiguration:
     scheduledSearches:
+
     - description: Close "Need More Info" Issues
       frequencies:
       - hourly:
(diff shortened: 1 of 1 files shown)


.github/policies/resourceManagement.yml
```diff
@@ -144,6 +144,28 @@ configuration:
           label: Community
       description: Label Community Pull Requests
       triggerOnOwnActions: false
+    - if:
```
333fred, line 147: Does this actually work? I've not see the equivalent API Owners tagging actually work well.

dibarbet (approved):
I'm fine with this. But definitely should get thoughts from other @dotnet/roslyn-infrastructure folks

.github/policies/resourceManagement.yml
JoeRobich: I have no idea. I based it on the other PR taggers, but am not sure there is a good way to test.

JoeRobich: Looking at [docs](https://eng.ms/docs/more/github-inside-microsoft/policies/examples/pr-triage) and I think the indentation is all wrong. Will update.

.github/policies/resourceManagement.yml
```diff
@@ -8,6 +8,7 @@ where:
 configuration:
   resourceManagementConfiguration:
     scheduledSearches:
+
```
JoeRobich, line 11: Added empty line between searches and tasks

.github/policies/resourceManagement.yml
```diff
@@ -23,8 +24,12 @@ configuration:
       - closeIssue
       - addReply:
           reply: "Closing this issue as we've seen no reply to the request for more information. If you are able to get the requested information, please add it to the issue and we will retriage it. "
+
     eventResponderTasks:
-    - if:
+
+    - description: Auto-approve auto-merge PRs
+      triggerOnOwnActions: false
```
JoeRobich, line 31: Moved description and trigger to the top of each task to improve readability

.github/policies/resourceManagement.yml
```diff
@@ -187,12 +222,14 @@ configuration:
...
       - addLabel:
           label: Needs UX Triage
-      description: Add "Needs UX Triage" on PRs
+
+    - description: Adds "Needs API Review" on PRs that touch public APIs
       triggerOnOwnActions: false
-    - if:
+      if:
       - payloadType: Pull_Request
+      - isPullRequest
```
JoeRobich, line 230: Added isPullRequest filter where applicable

.github/policies/resourceManagement.yml
JoeRobich: Doesn't look like the formatting was _that_ different after all. Still made changes to improve readability.

JoeRobich:
Gonna get this merged so that I can test it. If the tagging doesn't work then it is an easy revert.

PR #74597, opened 2024-07-30
Allow more of Razor to use Roslyn Lsp types

davidwengier:
Once Web Tools has moved to System.Text.Json, we on Razor can move off of Visual Studio Lsp types, which resolves a licensing mismatch in the C# extension. To do this, we just need a bit more access to the Roslyn Lsp types.

There was a couple of things missing, presumably added after forking, so I went ahead and added them, but only the minimum that Razor actually needs. Mikayla's PR probably already covers these too, and more obviously.


JoeRobich (approved):

dibarbet (approved):

JoeRobich (approved):

davidwengier:
Okay, https://github.com/dotnet/razor/pull/11504 will fix any source breaks this introduces, and the val build passed, so I'm merging this.

PR #74608, opened 2024-07-30
Reduce allocations during SourceText serialization

ToddGrun:
These allocations are showing up during the typing portion of the scrolling speedometer test as about 9.4% of allocations for that scenario.

The issue here is that in .Net framework, the BinaryWriter.Write method allocates a local array for the results of the Encoding.GetBytes call. .NET core does not allocate in this case, opting to call the GetBytes version that writes to an array given it (and uses a pooled array as such).

The change is to simulate what .NET core does, which should get rid of all these temporary array allocations when serializing source texts from .net fx.

*** relevant allocations from typing portion of scrolling speedometer test ***
image

### DIFF
--- a/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Serialization/ObjectWriter.cs
+++ b/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Serialization/ObjectWriter.cs
@@ -27,8 +27,10 @@ internal sealed partial class ObjectWriter : IDisposable
 {
     private static class BufferPool<T>
     {
+        public const int BufferSize = 32768;
+
         // Large arrays that will not go into the LOH (even with System.Char).
-        public static ObjectPool<T[]> Shared = new(() => new T[32768], 512);
+        public static ObjectPool<T[]> Shared = new(() => new T[BufferSize], 512);
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


CyrusNajmabadi (approved):

chsienki (approved):

ToddGrun:
Waiting on test insertion results:

https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/568475

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Serialization/ObjectWriter.cs
```diff
@@ -74,6 +76,8 @@ private static class BufferPool<T>
     /// </summary>
     private WriterReferenceMap _stringReferenceMap;
 
+    private static readonly Encoding s_encoding = Encoding.UTF8;
```
333fred, line 79: Why make this a static readonly? I believe it should already be cached on both Framework and modern. If the goal is just to avoid specifying `Encoding.UTF8` in multiple places, consider making this a property instead, but I'm personally fine with just using UTF8 directly in both locations.

333fred (approved):
One minor comment, otherwise LGTM.

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Serialization/ObjectWriter.cs
ToddGrun: Sure, I've gone ahead and changed

PR #74617, opened 2024-07-31
Reduce overhead for resolving SymbolKey special types

sharwell:


src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/SymbolKey/SymbolKey.SymbolKeyWriter.cs
```diff
@@ -401,7 +402,13 @@ public override void VisitModule(IModuleSymbol moduleSymbol)
 
         public override void VisitNamedType(INamedTypeSymbol namedTypeSymbol)
         {
-            if (namedTypeSymbol.TypeKind == TypeKind.Error)
+            if (namedTypeSymbol.IsSpecialType())
+            {
+                WriteType(SymbolKeyType.SpecialType);
+                WriteInteger((int)namedTypeSymbol.SpecialType);
+                throw new NotSupportedException("Forced exception to identify impacted tests.");
```
sharwell, line 409: 📝 Tests for resolving named types might be using special types. This exception will help identify those tests for review to see which ones need to be converted to or supplemented by a test that doesn't fall on the new code paths.

PR #74618, opened 2024-07-31
Reduce allocations in SymbolKey.ResolveString

sharwell:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2136177

Extracted from #74513, and only contains changes which do not affect the public APIs.

### DIFF
--- a/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/SymbolKey/SymbolKey.ModuleSymbolKey.cs
+++ b/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/SymbolKey/SymbolKey.ModuleSymbolKey.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Immutable;
+
 namespace Microsoft.CodeAnalysis;
 
 internal partial struct SymbolKey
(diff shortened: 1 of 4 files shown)


src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/SymbolKey/SymbolKey.ModuleSymbolKey.cs
```diff
@@ -24,12 +26,27 @@ public sealed override void Create(IModuleSymbol symbol, SymbolKeyWriter visitor
...
+                if (assemblyModules is ImmutableArray<IModuleSymbol> modules)
+                {
+                    // Avoid allocations if possible
+                    // https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2136177
+                    result.AddValuesIfNotNull(modules);
+                }
+                else
+                {
+                    // Visual Basic implementation of IAssemblySymbol.Modules relies on covariance of IEnumerable<T>, so
+                    // the preceding concrete type check will fail.
```
333fred, line 47: We should be able to adjust VB's explicit implementation to use `ImmutableArray.CastUp` to avoid this path.

CyrusNajmabadi (approved):

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/SymbolKey/SymbolKey.ModuleSymbolKey.cs
sharwell: This may be true, but the code path would likely stay in place (just the comment would change). I haven't see VB show up in the perf trace that led to this change so not sure how important this is.

333fred: > but the code path would likely stay in place (just the comment would change)

Sure, not disagreeing with that. I think I'd prefer if we made this change for VB.

sharwell: ➡️ Updated to use `CastUp`

333fred (approved):

sharwell:

jjonescz (approved):

PR #74620, opened 2024-07-31
Move existing RemoveWhere code into an extension method on ArrayBuilder

ToddGrun:
I have an upcoming PR that needs to update an ArrayBuilder in place and I'd prefer not to duplicate that logic. Seems like a useful enough API to warrant an extension method.

### DIFF
--- a/src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
+++ b/src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
@@ -1679,7 +1679,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
 
             if (name == null)
-                FilterNotReferenceable(results);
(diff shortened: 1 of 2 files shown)


CyrusNajmabadi (approved):

src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
```diff
@@ -1679,7 +1679,7 @@ private void CheckModelAndSyntaxNodeToSpeculate(CSharpSyntaxNode syntax)
             }
 
             if (name == null)
-                FilterNotReferenceable(results);
+                results.RemoveWhere(static (symbol, _, _) => !symbol.CanBeReferencedByName, default(VoidResult));
```
333fred, line 1682: Consider using a parameter name for the `default`.

ToddGrun: done

chsienki (approved):

333fred (approved):

PR #74627, opened 2024-08-01
Reduce allocations again in BlockStructure collection

ToddGrun:
There are basically two places where array allocations have been improved (particularly when the array exceed the default array builder pool size).

1) AbstractBlockStructureProvider.ProvideBlockStructure now calls BlockSpanCollector.CollectBlockSpans with the context.Spans, not with a new TemporaryArray. This required a change to the Sort call to pass in the start index, and to the filtering code at the end of the method to only filter out from the range that the current provider added to context.Spans

2) BlockStructureContext has it's own pool for the ArrayBuilder<BlockSpan>. As this array was very commonly exceeding the maximum reusable size allowed in the default array pool, we were previously no placing these arrays back in a pool.

Note that BlockStructureServiceWithProviders.CreateBlockStructure no longer uses FixedSizeArrayBuilder, instead modifying the entries in place. There isn't really an allocation change in this method, as both old and new allocate a single array in this method.

After the GetMultiLineRegions change earlier today, these allocations are only about 0.8% in the scrolling speedometer profile I'm looking at. Of that, I'd expect this to only remove about half of that, so only somewhat impactful.

*** relevant allocations from scrolling speedometer profile ***
image

### DIFF
--- a/src/EditorFeatures/CSharpTest/Structure/CommentStructureTests.cs
+++ b/src/EditorFeatures/CSharpTest/Structure/CommentStructureTests.cs
@@ -8,6 +8,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.CSharp.Structure;
 using Microsoft.CodeAnalysis.Editor.UnitTests.Structure;
+using Microsoft.CodeAnalysis.PooledObjects;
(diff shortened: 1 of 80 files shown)


ToddGrun:
@CyrusNajmabadi -- ptal

CyrusNajmabadi (approved):

Issue #74615, opened 2024-07-31
Collection expressions use `AddRange` instead of manual `foreach` in presence of a struct enumerator


DoctorKrolic:
**Version Used**: 
Noticed while working on other PR, so latest `main`

**Steps to Reproduce**:
```cs
using System.Collections;
using System.Collections.Generic;

class MyCollection(List<int> list) : IEnumerable<int>
{
    public Enumerator GetEnumerator() => new(list.GetEnumerator());

    IEnumerator<int> IEnumerable<int>.GetEnumerator() => GetEnumerator();

    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();

    public struct Enumerator(List<int>.Enumerator enumerator) : IEnumerator<int>
    {
        public int Current => enumerator.Current;

        object IEnumerator.Current => Current;

        public bool MoveNext() => enumerator.MoveNext();

        public void Dispose() => enumerator.Dispose();

        public void Reset() { }
    }
}

class C
{
    static List<int> M(MyCollection c) => [..c];
}
```

**Expected Behavior**:
Codegen of `C.M` is a manual `foreach` due to the fact, that `MyCollection` has a struct enumerator and doesn't implement `ICollection<T>`

**Actual Behavior**:
Codegen of `C.M` is:
```cs
private static List<int> M(MyCollection c)
{
    List<int> list = new List<int>();
    list.AddRange(c);
    return list;
}
```
[...]

PR #74630, opened 2024-08-01
Use manual `foreach` instead of `AddRange` method for certain collection expression spreads

DoctorKrolic:
Fixes: https://github.com/dotnet/roslyn/issues/74615

### DIFF
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
@@ -1069,22 +1069,22 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                         var type = rewrittenSpreadOperand.Type!;
 
-                        var useSiteInfo = GetNewCompoundUseSiteInfo();
-
                         if (spreadElement.EnumeratorInfoOpt is { } enumeratorInfo)
                         {
                             var iCollectionOfTType = _compilation.GetSpecialType(SpecialType.System_Collections_Generic_ICollection_T);
                             var iCollectionOfElementType = iCollectionOfTType.Construct(enumeratorInfo.ElementType);
+                            var discardedUseSiteInfo = CompoundUseSiteInfo<AssemblySymbol>.Discarded;
 
                             // If collection has a struct enumerator but doesn't implement ICollection<T>
                             // then manual `foreach` is always more efficient then using `AddRange` method
                             if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType &&
-                                !enumeratorInfo.CollectionType.ImplementsInterface(iCollectionOfElementType, ref useSiteInfo))
+                                !enumeratorInfo.CollectionType.ImplementsInterface(iCollectionOfElementType, ref discardedUseSiteInfo))
                             {
                                 return false;
                             }
                         }
 
+                        var useSiteInfo = GetNewCompoundUseSiteInfo();
                         var conversion = _compilation.Conversions.ClassifyConversionFromType(type, addRangeMethod.Parameters[0].Type, isChecked: false, ref useSiteInfo);
                         _diagnostics.Add(rewrittenSpreadOperand.Syntax, useSiteInfo);
                         if (conversion.IsIdentity || (conversion.IsImplicit && conversion.IsReference))
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
@@ -1067,8 +1067,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                         if (addRangeMethod is null)
                             return false;
 
-                        var type = rewrittenSpreadOperand.Type!;
-
                         if (spreadElement.EnumeratorInfoOpt is { } enumeratorInfo)
                         {
                             var iCollectionOfTType = _compilation.GetSpecialType(SpecialType.System_Collections_Generic_ICollection_T);
@@ -1083,6 +1081,8 @@ namespace Microsoft.CodeAnalysis.CSharp
                                 return false;
                             }
                         }
+
+                        var type = rewrittenSpreadOperand.Type!;
 
                         var useSiteInfo = GetNewCompoundUseSiteInfo();
                         var conversion = _compilation.Conversions.ClassifyConversionFromType(type, addRangeMethod.Parameters[0].Type, isChecked: false, ref useSiteInfo);
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
@@ -1070,6 +1070,21 @@ namespace Microsoft.CodeAnalysis.CSharp
                         var type = rewrittenSpreadOperand.Type!;
 
                         var useSiteInfo = GetNewCompoundUseSiteInfo();
+
+                        if (spreadElement.EnumeratorInfoOpt is { } enumeratorInfo)
+                        {
+                            var iCollectionOfTType = _compilation.GetSpecialType(SpecialType.System_Collections_Generic_ICollection_T);
+                            var iCollectionOfElementType = iCollectionOfTType.Construct(enumeratorInfo.ElementType);
+
+                            // If collection has a struct enumerator but doesn't implement ICollection<T>
+                            // then manual `foreach` is always more efficient then using `AddRange` method
+                            if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType &&
+                                !enumeratorInfo.CollectionType.ImplementsInterface(iCollectionOfElementType, ref useSiteInfo))
+                            {
+                                return false;
+                            }
+                        }
+
                         var conversion = _compilation.Conversions.ClassifyConversionFromType(type, addRangeMethod.Parameters[0].Type, isChecked: false, ref useSiteInfo);
                         _diagnostics.Add(rewrittenSpreadOperand.Syntax, useSiteInfo);
                         if (conversion.IsIdentity || (conversion.IsImplicit && conversion.IsReference))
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -34640,6 +34640,519 @@ .maxstack  2
...
+                    IL_0036:  endfinally
+                  }
+                  IL_0037:  ldloc.0
+                  IL_0038:  ret
+                }
+                """);
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/74615")]
+        public void List_SingleSpread_CustomCollection_NotICollectionAndStructEnumerator_MixedWithOtherAddRangeSpread()
```
DoctorKrolic, line 34955:
```suggestion
        public void List_CustomCollection_NotICollectionAndStructEnumerator_MixedWithOtherAddRangeSpread()
```

DoctorKrolic:
CC @RikkiGibson @cston

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -1070,6 +1070,21 @@ private BoundExpression CreateAndPopulateList(BoundCollectionExpression node, Ty
...
+
+                        if (spreadElement.EnumeratorInfoOpt is { } enumeratorInfo)
+                        {
+                            var iCollectionOfTType = _compilation.GetSpecialType(SpecialType.System_Collections_Generic_ICollection_T);
+                            var iCollectionOfElementType = iCollectionOfTType.Construct(enumeratorInfo.ElementType);
+
+                            // If collection has a struct enumerator but doesn't implement ICollection<T>
+                            // then manual `foreach` is always more efficient then using `AddRange` method
+                            if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType &&
+                                !enumeratorInfo.CollectionType.ImplementsInterface(iCollectionOfElementType, ref useSiteInfo))
```
DoctorKrolic, line 1082: Should we maybe use discarded use site info here? Since we only use `ICollection<T>` type for checks, but don't emit it, we don't care about its use site diagnostics, right?

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -34640,6 +34640,519 @@ .maxstack  2
...
+                  IL_000e:  stloc.1
+                  IL_000f:  ldloc.1
+                  IL_0010:  ldloc.0
+                  IL_0011:  call       "void System.Runtime.InteropServices.CollectionsMarshal.SetCount<int>(System.Collections.Generic.List<int>, int)"
+                  IL_0016:  ldloc.1
+                  IL_0017:  call       "System.Span<int> System.Runtime.InteropServices.CollectionsMarshal.AsSpan<int>(System.Collections.Generic.List<int>)"
+                  IL_001c:  stloc.2
+                  IL_001d:  ldc.i4.0
+                  IL_001e:  stloc.3
+                  IL_001f:  callvirt   "System.Collections.Generic.IEnumerator<int> MyCollection.GetEnumerator()"
```
RikkiGibson, line 34751: I am confused by this codegen. Because it implements `ICollection<int>`, I expected `List<int>.AddRange(IEnumerable<int>)` to be used here.

DoctorKrolic: For this case we don't even get to `Add` vs `AddRange` method choice, because we fall into [this Span-based optimization branch](https://github.com/dotnet/roslyn/blob/3328d049728aa83db1af20cf5a1630ad83d946e6/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs#L984).

I suggest to leave it as is because a) it seems out of scope of the precise issue we try to fix (maybe we need to additionally clarify when to use span-based vs `AddRange`-based optimization since the suggestion itself is quite reasonable, but that is another workitem IMO) and b) the next optimization on my list after this one is merged is a [ToList optimization for a single spread](https://github.com/dotnet/roslyn/issues/71217) (no pun intended), so this will turn into a single `ToList` call anyway

DoctorKrolic:
@RikkiGibson @cston PTAL

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -1069,6 +1069,21 @@ private BoundExpression CreateAndPopulateList(BoundCollectionExpression node, Ty
 
                         var type = rewrittenSpreadOperand.Type!;
```
cston, line 1070: >var type = rewrittenSpreadOperand.Type!; [](http://example.com/codeflow?start=24&length=40)

Consider moving declaration after `if` block.

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -34640,6 +34640,519 @@ .maxstack  2
...
+                    static void Main()
+                    {
+                        M(new([1, 2, 3])).Report();
+                    }
+
+                    static List<int> M(MyCollection c) => [..c];
+                }
+                """;
+
+            var verifier = CompileAndVerify([source, s_collectionExtensionsWithSpan], expectedOutput: IncludeExpectedOutput("[1, 2, 3],"), targetFramework: TargetFramework.Net80, verify: Verification.Skipped);
```
cston, line 34723: >var verifier = CompileAndVerify([source, s_collectionExtensionsWithSpan], expectedOutput: IncludeExpectedOutput("[1, 2, 3],"), targetFramework: TargetFramework.Net80, verify: Verification.Skipped); [](http://example.com/codeflow?start=12&length=197)

Consider also compiling and verifying IL for the default target framework. That should avoid the `CollectionsMarshal` optimization and use `AddRange()` instead. And it should be possible to test `expectedOutput` regardless of platform for that case. Please consider adding to this test rather than creating a separate test with duplicated source.
```
verifier = CompileAndVerify([source, s_collectionExtensions], expectedOutput: "[1, 2, 3],", verify: Verification.Skipped);
verifier.VerifyIL("C.M", ...);
```

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -34640,6 +34640,519 @@ .maxstack  2
...
```
cston, line 35093: >var verifier = CompileAndVerify([source, s_collectionExtensionsWithSpan], expectedOutput: IncludeExpectedOutput("[1, 2, 3],"), targetFramework: TargetFramework.Net80, verify: Verification.Skipped); [](http://example.com/codeflow?start=12&length=197)

Consider also compiling and verifying IL for the default target framework. That should avoid the `CollectionsMarshal` optimization and use `AddRange()` instead. And it should be possible to test `expectedOutput` regardless of platform for that case. Please consider adding to this test rather than creating a separate test with duplicated source.
```
verifier = CompileAndVerify([source, s_collectionExtensions], expectedOutput: "[1, 2, 3],", verify: Verification.Skipped);
verifier.VerifyIL("C.M", ...);
```

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -34640,6 +34640,519 @@ .maxstack  2
...
```
cston, line 34668: >s_collectionExtensionsWithSpan [](http://example.com/codeflow?start=53&length=30)

It looks like all new tests can use `s_collectionExtensions` rather than `s_collectionExtensionsWithSpan`.

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -34640,6 +34640,569 @@ .maxstack  2
...
+                    static void Main()
+                    {
+                        M(new([1, 2, 3])).Report();
+                    }
+
+                    static List<int> M(MyCollection c) => [..c];
+                }
+                """;
+
+            var verifier = CompileAndVerify([source, s_collectionExtensions], expectedOutput: "[1, 2, 3],", targetFramework: targetFramework, verify: Verification.Skipped);
```
cston, line 34725: >"[1, 2, 3]," [](http://example.com/codeflow?start=94&length=12)

I think you'll need to use `IncludeExpectedOutput()` for the `TargetFramework.Net80` case.

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
DoctorKrolic: > And it should be possible to test `expectedOutput` regardless of platform for that case

Unfortunatelly, it is not. When tests are run on .NET Framework, it cannot load and execute code, generated for .NET 8

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -34722,7 +34722,18 @@ static void Main()
                 }
                 """;
 
-            var verifier = CompileAndVerify([source, s_collectionExtensions], expectedOutput: "[1, 2, 3],", targetFramework: targetFramework, verify: Verification.Skipped);
+            var expectedOutputString = "[1, 2, 3],";
+            var verifier = CompileAndVerify(
+                [source, s_collectionExtensions],
+                expectedOutput: targetFramework switch
```
cston, line 34728: Perhaps simplify the `targetFramework` checks for `expectedOutput` and `VerifyIL()` to use `targetFramework == TargetFramework.Standard` rather than `targetFramework switch ...`.

DoctorKrolic: My original motivation for `switch` was to hard-fail if test is modified and not all cases are handled, but checking for `TargetFramework.Standard` seems to fulfill this purpose as well

cston (approved):

RikkiGibson (approved):

cston:
Thanks @DoctorKrolic.

PR #74641, opened 2024-08-02
Handle various cases for field-backed properties

cston:
Changes:
- Generate implicit accessor bodies
- Support `[field:]` attributes targeting synthesized backing field
- Diagnostics for properties with ref-like type


src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -326,6 +324,198 @@ class C
...
+        [InlineData("struct")]
+        [InlineData("ref struct")]
+        [InlineData("record")]
+        [InlineData("record struct")]
+        public void ImplicitAccessorBody_01(string typeKind)
+        {
+            string source = $$"""
+                {{typeKind}} A
+                {
+                    public static object P1 { get; set { _ = field; } }
```
cston, line 338: >{ get; set { _ = field; } } [](http://example.com/codeflow?start=44&length=27)

Allowing mix of auto- and explicitly-implemented accessors should be tied to language version.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -326,6 +324,198 @@ class C
...
+
+        [Fact]
+        public void ImplicitAccessorBody_02()
+        {
+            string source = """
+                interface I
+                {
+                    static object P1 { get; set { _ = field; } }
+                    static object P2 { get { return field; } set; }
+                    object Q1 { get; set { _ = field; } }
```
cston, line 455: >{ get; set { _ = field; } } [](http://example.com/codeflow?start=30&length=27)

Should this be reported as an error, with `ErrorCode.ERR_InterfacesCantContainFields` perhaps?

RikkiGibson: Will review the tests in depth later, but it feels reasonable for the synthesized field's "location" to be the first usage of the field keyword, and to report the exact error you are saying there.

RikkiGibson (commented):
Had a few comments, and it looks like some assertions are failing in tests

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
```diff
@@ -260,21 +270,22 @@ public override OneOrMany<SyntaxList<AttributeListSyntax>> GetAttributeDeclarati
...
             }
             else
             {
                 var body = GetArrowExpression(syntax);
-                accessorsHaveImplementation = body is object;
+                hasGetAccessorImplementation = body is object;
+                hasSetAccessorImplementation = false;
                 usesFieldKeyword = body is { } && containsFieldKeyword(body);
-                Debug.Assert(accessorsHaveImplementation); // it's not clear how this even parsed as a property if it has no accessor list and no arrow expression.
+                Debug.Assert(hasGetAccessorImplementation); // it's not clear how this even parsed as a property if it has no accessor list and no arrow expression.
```
333fred, line 282: Since you're modifying this line anyway, consider making this comment an actual assert message.

333fred (approved):

333fred (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
```diff
@@ -140,7 +142,9 @@ internal static SourcePropertySymbol Create(SourceMemberContainerTypeSymbol cont
             {
                 Binder.CheckFeatureAvailability(
                     syntax,
-                    (hasGetAccessor && !hasSetAccessor) ? MessageID.IDS_FeatureReadonlyAutoImplementedProperties : MessageID.IDS_FeatureAutoImplementedProperties,
+                    hasGetAccessor && hasSetAccessor ?
+                        (hasAutoPropertyGet && hasAutoPropertySet ? MessageID.IDS_FeatureAutoImplementedProperties : MessageID.IDS_FeatureFieldKeyword) :
+                        MessageID.IDS_FeatureReadonlyAutoImplementedProperties,
```
RikkiGibson, line 147: nit: It seems like the new factoring is going to diagnose a setter-only auto property (error case) as requiring the readonly auto-properties feature.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -485,35 +549,335 @@ .maxstack  1
...
-                  IL_0000:  ldarg.0
-                  IL_0001:  ldarg.1
-                  IL_0002:  stfld      "object I.<Q2>k__BackingField"
-                  IL_0007:  ret
+                    public static int P1 { get; set { } }
+                    public static int P2 { get { return -2; } set; }
+                    public int P3 { get; set { } }
+                    public int P4 { get { return -4; } set; }
+                    public int P5 { get; init { } }
+                    public int P6 { get { return -6; } init; }
```
RikkiGibson, line 685: for the future, I would def consider a "half-auto" prop, where the manually implemented accessor never reads `field`, to possibly be a warning scenario.

For example, if you have `get => notField; set;` or `get; set => notField = value`, it practically means you want to ignore the value that was set. There are better ways to do that than introducing a backing field. It very likely means you are making a mistake.

RikkiGibson (approved):

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
cston: There is an [open question](https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#scenarios-similar-to--set-) in the proposal.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
cston: Good catch, thanks.

RikkiGibson (approved):

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
jnm2: To obtain these warnings, I added <https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#field-usage-warnings> since the proposal hadn't spoken to this aspect at all yet.

PR #74646, opened 2024-08-05
Add span type inference

jjonescz:
Corresponding speclet update: https://github.com/dotnet/csharplang/pull/8333

### DIFF
--- a/src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
+++ b/src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
@@ -105,14 +105,10 @@ namespace Microsoft.CodeAnalysis.CSharp.UnitTests
                 }
             }
             """;
-        internal const string s_collectionExtensionsWithSpan = s_collectionExtensions +
+        internal const string s_collectionExtensionsWithReadOnlySpan = s_collectionExtensions +
             """
             static partial class CollectionExtensions
             {
-                internal static void Report<T>(this in Span<T> s)
-                {
-                    Report((ReadOnlySpan<T>)s);
-                }
                 internal static void Report<T>(this in ReadOnlySpan<T> s)
                 {
                     var builder = new StringBuilder();
@@ -126,6 +122,16 @@ namespace Microsoft.CodeAnalysis.CSharp.UnitTests
                     builder.Append("]");
                     Console.Write(builder.ToString());
                     Console.Write(", ");
+                }
+            }
+            """;
+        internal const string s_collectionExtensionsWithSpan = s_collectionExtensionsWithReadOnlySpan +
+            """
+            static partial class CollectionExtensions
+            {
+                internal static void Report<T>(this in Span<T> s)
+                {
+                    Report((ReadOnlySpan<T>)s);
                 }
             }
             """;
@@ -8155,7 +8161,7 @@ static class Program
                 """;
 
             var verifier = CompileAndVerify(
-                new[] { source, s_collectionExtensionsWithSpan },
+                new[] { source, s_collectionExtensionsWithReadOnlySpan },
                 options: TestOptions.ReleaseExe,
                 targetFramework: TargetFramework.Net70,
                 verify: Verification.Skipped,
@@ -8182,32 +8188,37 @@ static class Program
                 ("IEnumerable<int>", "int[]") =>
                     """
                     {
-                      // Code size       24 (0x18)
+                      // Code size       31 (0x1f)
                       .maxstack  3
+                      .locals init (System.ReadOnlySpan<int> V_0)
                       IL_0000:  newobj     "System.Collections.Generic.List<int>..ctor()"
                       IL_0005:  dup
                       IL_0006:  ldarg.0
                       IL_0007:  callvirt   "void System.Collections.Generic.List<int>.AddRange(System.Collections.Generic.IEnumerable<int>)"
                       IL_000c:  callvirt   "int[] System.Collections.Generic.List<int>.ToArray()"
-                      IL_0011:  ldc.i4.0
-                      IL_0012:  call       "void CollectionExtensions.Report(object, bool)"
-                      IL_0017:  ret
+                      IL_0011:  call       "System.ReadOnlySpan<int> System.ReadOnlySpan<int>.op_Implicit(int[])"
+                      IL_0016:  stloc.0
+                      IL_0017:  ldloca.s   V_0
+                      IL_0019:  call       "void CollectionExtensions.Report<int>(in System.ReadOnlySpan<int>)"
+                      IL_001e:  ret
                     }
                     """,
                 ("int[]", "int[]") =>
                     """
                     {
-                      // Code size       21 (0x15)
-                      .maxstack  2
+                      // Code size       28 (0x1c)
+                      .maxstack  1
                       .locals init (System.ReadOnlySpan<int> V_0)
                       IL_0000:  ldarg.0
                       IL_0001:  newobj     "System.ReadOnlySpan<int>..ctor(int[])"
                       IL_0006:  stloc.0
                       IL_0007:  ldloca.s   V_0
                       IL_0009:  call       "int[] System.ReadOnlySpan<int>.ToArray()"
-                      IL_000e:  ldc.i4.0
-                      IL_000f:  call       "void CollectionExtensions.Report(object, bool)"
-                      IL_0014:  ret
+                      IL_000e:  call       "System.ReadOnlySpan<int> System.ReadOnlySpan<int>.op_Implicit(int[])"
+                      IL_0013:  stloc.0
+                      IL_0014:  ldloca.s   V_0
+                      IL_0016:  call       "void CollectionExtensions.Report<int>(in System.ReadOnlySpan<int>)"
+                      IL_001b:  ret
                     }
                     """,
                 ("ReadOnlySpan<int>", "ReadOnlySpan<int>") =>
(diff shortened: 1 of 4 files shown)


src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -4361,26 +4359,31 @@ static class C
...
+            Diagnostic(ErrorCode.ERR_NoSuchMemberOrExtension, "M").WithArguments("int[]", "M").WithLocation(11, 28));
 
-        // PROTOTYPE: Some of these need type inference to work.
         var expectedDiagnostics = new[]
         {
-            // (4,10): error CS8917: The delegate type could not be inferred.
+            // (4,12): error CS0123: No overload for 'M' matches delegate 'Func<int, int>'
             // var d1 = a.M;
-            Diagnostic(ErrorCode.ERR_CannotInferDelegateType, "a.M").WithLocation(4, 10),
+            Diagnostic(ErrorCode.ERR_MethDelegateMismatch, "M").WithArguments("M", "System.Func<int, int>").WithLocation(4, 12),
```
jjonescz, line 4374: Note that type inference successfully infers the delegate type to be `Func<int, int>` but then binding fails (because the span conversion doesn't exist for extension method receiver). I'm not sure if this is a problem, I think this could happen for some existing scenarios as well.

jjonescz:

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -7408,4 +7543,282 @@ static class C
...
+    }
+
+    [Fact]
+    public void TypeInference_ReadOnlySpan_ReadOnlySpan_02()
+    {
+        var source = """
+            using System;
+            class C
+            {
+                void M(ReadOnlySpan<string> a, object b) => M1(a, b);
```
333fred, line 7753: Let's test a scenario like this with a non-variant non-array conversion, like `ROS<int>` and `object`/vice versa; I expect the former to fail, and the latter to succeed.

333fred (commented):
Mostly LGTM. Couple of minor comments.

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -279,6 +279,125 @@ static class E
...
+            static class C
+            {
+                public static void R<T>(IEnumerable<T> e) => Console.Write(1);
+                public static void R<T>(Span<T> s) => Console.Write(2);
+            }
+            """;
+        var comp = CreateCompilationWithSpanAndMemoryExtensions(source, parseOptions: TestOptions.Regular13);
+        CompileAndVerify(comp, expectedOutput: "1").VerifyDiagnostics();
+
+        var expectedOutput = ExecutionConditionUtil.IsCoreClr ? "3" : "2";
```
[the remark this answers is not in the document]
cston: Consider reporting the exception type rather than "3".

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -279,6 +279,125 @@ static class E
...
+
+        comp = CreateCompilationWithSpanAndMemoryExtensions(source, parseOptions: TestOptions.RegularNext);
+        CompileAndVerify(comp, expectedOutput: expectedOutput).VerifyDiagnostics();
+
+        comp = CreateCompilationWithSpanAndMemoryExtensions(source);
+        CompileAndVerify(comp, expectedOutput: expectedOutput).VerifyDiagnostics();
+    }
```
cston, line 399: >} [](http://example.com/codeflow?start=4&length=1)

Are we testing similar cases to these new tests, but with `IEnumerable<T>` and `ReadOnlySpan<T>`?

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -7408,4 +7543,282 @@ static class C
...
+            {
+                void M(Span<string[]> a, object b) => M1(a, b);
+                void M1<T>(Span<T[]> x, T y) { }
+            }
+            """;
+        CreateCompilationWithSpanAndMemoryExtensions(source).VerifyDiagnostics(
+            // (4,43): error CS0411: The type arguments for method 'C.M1<T>(Span<T[]>, T)' cannot be inferred from the usage. Try specifying the type arguments explicitly.
+            //     void M(Span<string[]> a, object b) => M1(a, b);
+            Diagnostic(ErrorCode.ERR_CantInferMethTypeArgs, "M1").WithArguments("C.M1<T>(System.Span<T[]>, T)").WithLocation(4, 43));
+    }
```
cston, line 7620: >} [](http://example.com/codeflow?start=4&length=1)

Consider testing with `M(Span<object[]> a, string b) => M1(a, b)` as well.

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -7408,4 +7543,282 @@ static class C
...
+            using System;
+            class C
+            {
+                void M(object[] a, string b) => M1(a, b);
+                void M1<T>(ReadOnlySpan<T> x, T y) { }
+            }
+            """;
+        var comp = CreateCompilationWithSpanAndMemoryExtensions(source).VerifyDiagnostics();
+        AssertEx.Equal("C.M1<System.Object>", DisplayInvokedMethodTypeArguments(comp));
+    }
```
cston, line 7729: >} [](http://example.com/codeflow?start=4&length=1)

Please test:
```
void M2(int[] a, object b) => M1(a, b);
void M3(object[] a, int b) => M1(a, b);
```

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -7408,4 +7543,282 @@ static class C
...
+            {
+                void M(ReadOnlySpan<int> a) => M1(a);
+                void M1<T>(Span<T> x) { }
+            }
+            """;
+        CreateCompilationWithSpanAndMemoryExtensions(source).VerifyDiagnostics(
+            // (4,36): error CS0411: The type arguments for method 'C.M1<T>(Span<T>)' cannot be inferred from the usage. Try specifying the type arguments explicitly.
+            //     void M(ReadOnlySpan<int> a) => M1(a);
+            Diagnostic(ErrorCode.ERR_CantInferMethTypeArgs, "M1").WithArguments("C.M1<T>(System.Span<T>)").WithLocation(4, 36));
+    }
```
cston, line 7823: >} [](http://example.com/codeflow?start=4&length=1)

Are we testing inferring `ReadOnlySpan<T>` from call with `Span<U>`?
```csharp
void M(Span<string> a, object b) => M1(a, b);
void M1<T>(ReadOnlySpan<T> a, T b) { }
```

src/Features/CSharp/Portable/CodeRefactorings/UseRecursivePatterns/UseRecursivePatternsCodeRefactoringProvider.cs
```diff
@@ -600,7 +600,7 @@ static bool CanConvertToSubpattern(IdentifierNameSyntax name, SemanticModel mode
 
         // Process all nodes to refactor in reverse to ensure nested nodes
         // are processed before the outer nodes to refactor.
-        foreach (var originalNode in nodes.Reverse())
+        foreach (var originalNode in Enumerable.Reverse(nodes))
```
cston, line 603: >Enumerable.Reverse(nodes) [](http://example.com/codeflow?start=37&length=25)

Why was this change needed? Are we now binding `nodes.Reverse()` to `MemoryExtensions.Reverse<T>(Span<T>)`? And if so, is that an issue? (It looks like the array that is created from `var nodes = ... ToArray();` should not include array variance.)

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -105,14 +105,10 @@ private static string Concat(string container, string name)
                 }
             }
             """;
-        internal const string s_collectionExtensionsWithSpan = s_collectionExtensions +
+        internal const string s_collectionExtensionsWithReadOnlySpan = s_collectionExtensions +
```
cston, line 108: >s_collectionExtensionsWithReadOnlySpan [](http://example.com/codeflow?start=30&length=38)

Alternatively, could we leave the name unchanged, and simply remove the `Report<T>(this in Span<T>)` overload? Would that result in many additional test changes?

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
cston, line 7823: >} [](http://example.com/codeflow?start=4&length=1)

Consider testing variance when the array and span are the type arguments. For instance, with the following, for the calls with `I<object[]>` and `IOut<object[]>` I'd expect type inference to succeed but the subsequent argument conversion to fail, and for the call with `IIn<object[]>`, I'd expect type inference to fail.
```csharp
interface I<T> where T : allows ref struct { }
class A
{
    static void M(I<object[]> a) => M1(a);
    static void M1<T>(I<ReadOnlySpan<T>> x) { }
}
interface IOut<out T> where T : allows ref struct { }
class B
{
    static void M(IOut<object[]> a) => M1(a);
    static void M1<T>(IOut<ReadOnlySpan<T>> x) { }
}
interface IIn<in T> where T : allows ref struct { }
class C
{
    static void M(IIn<object[]> a) => M1(a);
    static void M1<T>(IIn<ReadOnlySpan<T>> x) { }
}
```

src/Features/CSharp/Portable/CodeRefactorings/UseRecursivePatterns/UseRecursivePatternsCodeRefactoringProvider.cs
jjonescz: > Are we now binding `nodes.Reverse()` to `MemoryExtensions.Reverse<T>(Span<T>)`?

Yes.

> And if so, is that an issue?

Yes, because `MemoryExtensions.Reverse` returns `void`.

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
jjonescz: Not many additional changes, I can do that.

cston (approved):

PR #74651, opened 2024-08-05
Remove razor source generator hack

chsienki:
A while back we added a razor specific hack to remove any documents the razor SG generated. Razor is supposed to be able to operate in a 'suppressed' mode where it doesn't produce anything, but a bug in the generator meant it wasn't respecting that flag. Rather than trying to service the SDK we took a tactical hack to just remove the docs in Roslyn.

We're now at the point in razor co-hosting where we *want* razor to produce documents, but this hack of course just removes them. The razor SG is no longer loaded out of the SDK, but comes from tooling, where the suppression has been fixed, so this hack is no longer required anyway.

### DIFF
--- a/src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.RegularCompilationTracker_Generators.cs
+++ b/src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.RegularCompilationTracker_Generators.cs
@@ -282,7 +282,7 @@ internal partial class SolutionCompilationState
             if (compilationWithStaleGeneratedTrees != null)
             {
                 var generatedTreeCount =
-                    runResult.Results.Sum(r => IsGeneratorRunResultToIgnore(r) ? 0 : r.GeneratedSources.Length);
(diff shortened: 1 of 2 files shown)


src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.RegularCompilationTracker_Generators.cs
```diff
@@ -282,7 +282,7 @@ private partial class RegularCompilationTracker : ICompilationTracker
             if (compilationWithStaleGeneratedTrees != null)
             {
                 var generatedTreeCount =
-                    runResult.Results.Sum(r => IsGeneratorRunResultToIgnore(r) ? 0 : r.GeneratedSources.Length);
+                    runResult.Results.Sum(r => r.GeneratedSources.IsDefaultOrEmpty ? 0 : r.GeneratedSources.Length);
```
chsienki, line 285: This is another case of the bug that was fixed here: https://github.com/dotnet/roslyn/pull/74507

chsienki:
@dotnet/roslyn-ide for review please :)

CyrusNajmabadi (approved):
yaay.  i always found this code impossible to understand.

PR #74653, opened 2024-08-05
Fix CG issue for Microsoft.IO.Redist

jaredpar:
Move to a version with the appropriate fix

The explicit ref of Microsoft.IO.Redist is required because of https://github.com/dotnet/sdk/issues/42608

### DIFF
--- a/src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
+++ b/src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
@@ -34,6 +34,7 @@
     <PackageReference Include="System.CommandLine" />
     <PackageReference Include="System.Collections.Immutable" />
     <PackageReference Include="Newtonsoft.Json" />
+    <PackageReference Include="Microsoft.IO.Redist" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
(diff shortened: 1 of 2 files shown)


Cosifne (approved):

333fred (approved):

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageReference Include="System.CommandLine" />
     <PackageReference Include="System.Collections.Immutable" />
     <PackageReference Include="Newtonsoft.Json" />
+    <PackageReference Include="Microsoft.IO.Redist" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
```
333fred, line 37: Womp

jaredpar: Yep.

jasonmalinowski (approved):

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
ericstj: It's busted you have to do this, you want all of MSBuild to be Exclude="Runtime"

I wonder if instead of referencing the packages for MSBuild which bring in it's entire NuGet graph, you could just reference the API it exposes.  As reference assemblies.  cc @baronfel

PR #74658, opened 2024-08-06
CodeQL fixes

jaredpar:
This takes care of a number of CodeQL violations in our code base. Virtually all of these were about uses of MD5 or SHA-1 that we have to support due to the file formats we produce and consume. As such I added suppressions for those cases. There was one real case that could be migrated that I took care of.

### DIFF
--- a/src/Compilers/Core/Portable/PEWriter/SigningUtilities.cs
+++ b/src/Compilers/Core/Portable/PEWriter/SigningUtilities.cs
@@ -19,23 +19,25 @@ namespace Microsoft.CodeAnalysis
     {
         internal static byte[] CalculateRsaSignature(IEnumerable<Blob> content, RSAParameters privateKey)
         {
-            var hash = CalculateSha1(content);
+            var hash = calculateSha1(content);
 
             using (var rsa = RSA.Create())
             {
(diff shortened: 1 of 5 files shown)


jaredpar:
@dotnet/roslyn-compiler PTAL

src/Compilers/Core/Portable/PEWriter/SigningUtilities.cs
```diff
@@ -19,23 +19,25 @@ internal static class SigningUtilities
...
+            static byte[] calculateSha1(IEnumerable<Blob> content)
             {
-                hash.AppendData(content);
-                return hash.GetHashAndReset();
+                // CodeQL [SM02196] ECMA-335 requires us to use SHA-1 and there is no alternative.
+                using (var hash = IncrementalHash.CreateHash(HashAlgorithmName.SHA1))
+                {
+                    hash.AppendData(content);
+                    return hash.GetHashAndReset();
+                }
```
jaredpar, line 40: Moved this to a local function to make it clear that the scope of the SHA1 usage was this function.

cston (approved):

chsienki (approved):

PR #74669, opened 2024-08-06
Remove test dependency from BuildValidator

tmat:
Another attempt to fix https://github.com/dotnet/sdk/pull/42443/checks?check_run_id=28431211393

Actually, not required but it would still be better to remove the test dependency imo.


tmat:
@jaredpar

jaredpar (approved):

jaredpar:
Is the basic problem here that the SDK is only building a subset of our repo?

tmat:
Yup

PR #74675, opened 2024-08-07
Ensure that `GetSymbolInfo` returns only constructor symbols for object creation expressions

DoctorKrolic:
Separated from https://github.com/dotnet/roslyn/pull/74218


src/Compilers/CSharp/Test/Semantic/Semantics/TypeParameterCreationTests.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        typeInfo = model.GetTypeInfo(typeNode);
+        Assert.Null(typeInfo.Type);
+        Assert.Null(typeInfo.ConvertedType);
+
+        symbolInfo = model.GetSymbolInfo(typeNode);
+        Assert.Null(symbolInfo.Symbol);
+        Assert.Equal(CandidateReason.NotCreatable, symbolInfo.CandidateReason);
+        Assert.Collection(symbolInfo.CandidateSymbols,
+            s => Assert.Equal("T", s.ToTestDisplayString()));
```
DoctorKrolic, line 142: @333fred I didn't change this scenario because it fully matches behavior with normal types: e.g. if we have some `new X()` expression and get symbol info from just `X` we will get symbol `X` if the expression itself is correct or `X` as a candidate symbol with candidate reason `NotCreatable` if expression has errors (e.g. if `X` is a `static` or an `abstract` class). With type parameters we get the same behavior: `T` if `new T()` expression is legal and candidate `T` with reason `NotCreatable` if parent object creation expression is an error. Does this look correct to you?

src/Compilers/CSharp/Test/Semantic/Semantics/TypeParameterCreationTests.cs
```diff
@@ -0,0 +1,272 @@
...
+            .DescendantNodes()
+            .OfType<ObjectCreationExpressionSyntax>()
+            .Single();
+
+        var typeInfo = model.GetTypeInfo(objectCreationNode);
+        Assert.Equal("T", typeInfo.Type.ToTestDisplayString());
+        Assert.Equal("T", typeInfo.ConvertedType.ToTestDisplayString());
+
+        var symbolInfo = model.GetSymbolInfo(objectCreationNode);
+        Assert.True(symbolInfo.IsEmpty);
```
DoctorKrolic, line 130: So the PR only changes this particular scenario: if we get symbol info from the whole `new T()` (or just `new()`) we get empty symbol info both in legal and error cases

333fred (approved):

DoctorKrolic:
Can I get a second review here please?

src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
```diff
@@ -3369,6 +3369,14 @@ private OneOrMany<Symbol> GetSemanticSymbols(
 
                         if (expr.Syntax.Kind() is SyntaxKind.ObjectCreationExpression or SyntaxKind.ImplicitObjectCreationExpression)
                         {
+                            // Type parameters don't have associated constructor symbols.
+                            // Therefore to make API behavior consistent with non-error scenario
+                            // we return empty symbol list here
+                            if (expr.Type?.TypeKind == TypeKind.TypeParameter)
```
cston, line 3375: >(expr.Type?.TypeKind == TypeKind.TypeParameter) [](http://example.com/codeflow?start=31&length=47)

Are there other cases where we should also return an empty set? For instance `new InterfaceType()`?

src/Compilers/CSharp/Test/Semantic/Semantics/TypeParameterCreationTests.cs
cston, line 130: >Assert.True(symbolInfo.IsEmpty); [](http://example.com/codeflow?start=8&length=32)

Let's check `symbolInfo.CandidateReason` as well.

src/Compilers/CSharp/Test/Symbol/Compilation/SemanticModelGetSemanticInfoTests.cs
```diff
@@ -12325,9 +12325,8 @@ static void f()
             Assert.Equal(ConversionKind.Boxing, semanticInfo.ImplicitConversion.Kind);
 
             Assert.Null(semanticInfo.Symbol);
-            Assert.Equal(CandidateReason.NotCreatable, semanticInfo.CandidateReason);
-            Assert.Equal(1, semanticInfo.CandidateSymbols.Length);
-            Assert.Equal("T", semanticInfo.CandidateSymbols.First().Name);
+            Assert.Equal(CandidateReason.None, semanticInfo.CandidateReason);
```
cston, line 12328: >CandidateReason.None [](http://example.com/codeflow?start=25&length=20)

Why did `CandidateReason` change?

src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
DoctorKrolic: So in the end out list is the following:
- Interface types
- Type parameters
- Static types

Added handling for all of them (they all were returning type itself as symbol candidate for `new`). I think, we should add an assert, that symbol info of `new()` or `new TypeName()` can contain only constructor symbols or just be empty if there are no constructors in the first place. Can you point me where we can add such assert?

src/Compilers/CSharp/Test/Symbol/Compilation/SemanticModelGetSemanticInfoTests.cs
DoctorKrolic: Ok, let's think what the expected behavior here is. Because 2 explanations can be taken:
1. `CandidateReason` tells why instead of single `Symbol` we get only candidates as part of `SymbolInfo`. Therefore if there are no symbols to begin with, there can be no reason for candidates, so it should be `CandidateReason.None`
2. We have an error case where we cannot create type with `new`, so `CandidateReason` should be `NotCreatable`

I lean towards the second option because it is consistent with behavior we observe if we get symbol info from `TypeName` inside `new TypeName()`: we get candidate symbol with reason `NotCreatable` instead of a single symbol if the whole object creation expression has errors even in cases when the identifier itself is bound with no ambiguities. But I want your conclusion on this question first

DoctorKrolic:
@cston PTAL

src/Compilers/CSharp/Test/Symbol/Compilation/SemanticModelGetSemanticInfoTests.cs
cston: In the previous implementation, `GetSymbolInfo()` for `new T()` returned `CandidateReason.NotCreatable` and the type symbol `T`. Returning `NotCreatable` seemed reasonable. Returning a type rather than no symbols since there was no constructor seemed somewhat surprising, but it looks like `SymbolInfoFactory.Create()` typically returns `CandidateReason.None` if there are no symbols. Returning `CandidateReason.NotCreatable` for this scenario seems more appropriate, but will callers expect symbols when the reason is something other than `CandidateReason.None`?

Is there a strong reason to make a change to `CSharpSemanticModel.GetSemanticSymbols()` for these cases?

DoctorKrolic: > Is there a strong reason to make a change to `CSharpSemanticModel.GetSemanticSymbols()` for these cases?

I originally marked current behavior as "inconsistent" in https://github.com/dotnet/roslyn/pull/74218. Fred agreed with me. You yourself consider that "surprising". So 3 people, 2 of which are from the roslyn compiler team, see that behavior as abnormal. Isn't that enough to tweak behavior a little bit to much our expectations?

> Returning `CandidateReason.NotCreatable` for this scenario seems more appropriate

> will callers expect symbols when the reason is something other than `CandidateReason.None`

There are 2 expectations here:
1. If a symbol cannot be constructed, the candidate reason should be `NotCreatable`
2. `CandidateReason` can be something other than `None` only if instead of returning a specific symbol we fallback to candidates due to semantic errors

Unfortunatelly, in our case these 2 invariants contradict each other. Picking either breaks one of that expectations. Leaving it as is breaks expectations for returned symbols. Given that in the vast majority of use cases candidate reason is not even checked (users mostly need symbols themselves), I think changing behavior in the way, which makes returned symbols more expected in an expense of making candidate reason a bit surprising in one way or another is justified. I also think we should not interfere into `SymbolInfoFactory.Create()` logic and leave `None` as a final candidate reason.

cston: I'm not familiar enough with use of `SemanticModel.GetSymbolInfo()` to know if one approach should be preferred over the other, although from the documentation on `CandidateReason`, it seems that `NotCreatable` was at least initially intended for unconstrained type parameters. I will defer to other reviewers on this.

src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
```diff
@@ -3371,6 +3371,11 @@ private OneOrMany<Symbol> GetSemanticSymbols(
                         {
                             if (resultKind == LookupResultKind.NotCreatable)
                             {
+                                if (expr.Type is { IsStatic: true } or { TypeKind: TypeKind.Interface or TypeKind.TypeParameter })
```
AlekseyTs, line 3374: >if (expr.Type is { IsStatic: true } or { TypeKind: TypeKind.Interface or TypeKind.TypeParameter }) [](http://example.com/codeflow?start=32&length=98)

The condition doesn't look anything like a check for "only constructor symbols", which is sated as the goal of this PR. I suggest we do that instead. I.e. check for what you are looking for rather than checking for what you are not looking for.

PR #74688, opened 2024-08-08
Only report project load events for initial load in VSCode

dibarbet:
This was requested to match how VS and DevKit report project load telemetry (once when the solution is opened).  Before we would report the event every time the project loaded or reloaded.

This does mean we do not report telemetry when things like the TFM change, but that is known and expected.

Verified in VSCode that the initial event is reported and changes to the project to not trigger the event to be fired.


dibarbet:
@JoeRobich probably not as important given the usage numbers, but O# should be updated to have the same behavior here.

JoeRobich (approved):

PR #74689, opened 2024-08-08
Remove most usage of "new ArrayBuilder"

ToddGrun:
Most callers of this were intending to instead use ArrayBuilder.GetInstance (as that actually is what pulls items from the pool). Otherwise, the code would create a new ArrayBuilder (and consequently the ImmutableArray.Builder), reallocate until the array hit the appropriate size, and likely return the item to a pool that would never be pulled from.


ToddGrun:
Note, I didn't touch the compilers folder, but I'm guessing they have a fair number of mis-uses also.

CyrusNajmabadi (approved):

JoeRobich:
Can we ban the use of the ArrayBuilder constructor?

ToddGrun:
> Can we ban the use of the ArrayBuilder constructor?

I thought about that, but there are legitimate uses of this when creating pools of ArrayBuilders. Hopefully, most of these came from copy/paste mistakes, and now there shouldn't be much code using that form to copy (outside of the compilers folder)

PR #74700, opened 2024-08-09
Fix issues with VSCode LSP EA causing handlers to fail to load

dibarbet:
https://github.com/dotnet/roslyn/commit/2d311bd0fa69603143b1acfc2455865f5460309c caused both @333fred 's compiler SDK extension and some XAML handlers to break.

This was caused due to the removal of the enum member, changing the numerical value of each.  Code compiled against an older version of Roslyn that used the enum would be broken.

For the compiler SDK extension, this caused an issue because the enum was exposed in the EA library, which was being loaded from the extension (and not from roslyn).  The fix is to ship and load the EA library from Roslyn (only 21kb).

For XAML, the LSP service factory exposed the enum as a default parameter, so all XAML LSP service factories had the wrong enum value.  Xaml is updating their Roslyn version to fix, but here I also obsolete the API that exposed the enum (it wasn't necessary).

Also fixed an issue where MEF discovery errors werent being recorded.

### DIFF
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
@@ -87,7 +87,9 @@ internal sealed class ExportProviderBuilder
         //         part definition Microsoft.CodeAnalysis.ExternalAccess.Pythia.PythiaSignatureHelpProvider
         var erroredParts = configuration.CompositionErrors.FirstOrDefault()?.SelectMany(error => error.Parts).Select(part => part.Definition.Type.Name) ?? Enumerable.Empty<string>();
         var expectedErroredParts = new string[] { "PythiaSignatureHelpProvider" };
-        if (erroredParts.Count() != expectedErroredParts.Length || !erroredParts.All(part => expectedErroredParts.Contains(part)) || !catalog.DiscoveredParts.DiscoveryErrors.IsEmpty)
+        var hasUnexpectedErroredParts = erroredParts.Any(part => !expectedErroredParts.Contains(part));
+
+        if (hasUnexpectedErroredParts || !catalog.DiscoveredParts.DiscoveryErrors.IsEmpty)
         {
             try
             {
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
@@ -57,7 +57,7 @@ internal sealed class ExportProviderBuilder
         var config = CompositionConfiguration.Create(catalog);
 
         // Verify we only have expected errors.
-        ThrowOnUnexpectedErrors(config, logger);
+        ThrowOnUnexpectedErrors(config, catalog, logger);
 
         // Prepare an ExportProvider factory based on this graph.
         var exportProviderFactory = config.CreateExportProviderFactory();
@@ -75,7 +75,7 @@ internal sealed class ExportProviderBuilder
         return exportProvider;
     }
 
-    private static void ThrowOnUnexpectedErrors(CompositionConfiguration configuration, ILogger logger)
+    private static void ThrowOnUnexpectedErrors(CompositionConfiguration configuration, ComposableCatalog catalog, ILogger logger)
     {
         // Verify that we have exactly the MEF errors that we expect.  If we have less or more this needs to be updated to assert the expected behavior.
         // Currently we are expecting the following:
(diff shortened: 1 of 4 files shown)


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Microsoft.CodeAnalysis.LanguageServer.csproj
```diff
@@ -78,6 +78,7 @@
     <ProjectReference Include="..\..\Features\ExternalAccess\AspNetCore\Microsoft.CodeAnalysis.ExternalAccess.AspNetCore.csproj" />
     <ProjectReference Include="..\ExternalAccess\VisualDiagnostics\Microsoft.CodeAnalysis.ExternalAccess.VisualDiagnostics.csproj" />
     <ProjectReference Include="..\..\Tools\ExternalAccess\Xaml\Microsoft.CodeAnalysis.ExternalAccess.Xaml.csproj" />
+    <ProjectReference Include="..\ExternalAccess\CompilerDeveloperSDK\Microsoft.CodeAnalysis.ExternalAccess.CompilerDeveloperSDK.csproj" />
```
dibarbet, line 81: adding so @333fred doesn't have to ship this.  Matches how XAML works as well.  Only 21kb so not concerned about size.

src/Tools/ExternalAccess/Xaml/External/ExportXamlLspServiceFactoryAttribute.cs
```diff
@@ -14,8 +14,14 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Xaml;
 [AttributeUsage(AttributeTargets.Class, AllowMultiple = false), MetadataAttribute]
 internal sealed class ExportXamlLspServiceFactoryAttribute : ExportLspServiceFactoryAttribute
 {
-    public ExportXamlLspServiceFactoryAttribute(Type type, WellKnownLspServerKinds serverKind = WellKnownLspServerKinds.Any)
```
dibarbet, line 17: this was the issue for XAML - exposing the enum as a default value.

src/Tools/ExternalAccess/Xaml/External/ExportXamlLspServiceFactoryAttribute.cs
```diff
@@ -14,8 +14,14 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Xaml;
...
+    [Obsolete("Use the constructor that takes only a type")]
+    public ExportXamlLspServiceFactoryAttribute(Type type, WellKnownLspServerKinds serverKind)
         : base(type, ProtocolConstants.RoslynLspLanguagesContract, serverKind)
     {
     }
+
+    public ExportXamlLspServiceFactoryAttribute(Type type)
```
dibarbet, line 23: @etvorun at some point once this change merges you'll want to update xaml to use this so  Ican delete the old one.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -87,15 +87,16 @@ private static void ThrowOnUnexpectedErrors(CompositionConfiguration configurati
         //         part definition Microsoft.CodeAnalysis.ExternalAccess.Pythia.PythiaSignatureHelpProvider
         var erroredParts = configuration.CompositionErrors.FirstOrDefault()?.SelectMany(error => error.Parts).Select(part => part.Definition.Type.Name) ?? Enumerable.Empty<string>();
         var expectedErroredParts = new string[] { "PythiaSignatureHelpProvider" };
-        if (erroredParts.Count() != expectedErroredParts.Length || !erroredParts.All(part => expectedErroredParts.Contains(part)))
+        if (erroredParts.Count() != expectedErroredParts.Length || !erroredParts.All(part => expectedErroredParts.Contains(part)) || !catalog.DiscoveredParts.DiscoveryErrors.IsEmpty)
```
JoeRobich, line 90: I don't like how this enters the if when `erroredParts.Count()` is `0`.

```c#
var expectedErroredParts = new HashSet<string>(){ "PythiaSignatureHelpProvider" );
var hasUnexpectedCompositionErrors = configuration.CompositionErrors.FirstOrDefault()?
    .SelectMany(error => error.Parts)
    .Select(part => part.Definition.Type.Name)
    .Any(part => !expectedErroredParts.Contains(part));

if (hasUnexpectedCompositionErrors || !catalog.DiscoveredPArts.DiscoveryErrors.IsEmpty)
```

JoeRobich (approved):

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Microsoft.CodeAnalysis.LanguageServer.csproj
333fred: Do we know what will happen with the current version of my extension, where I'm including the EA dll with my extension? Will it just not be loaded?

dibarbet: @333fred  Both get loaded (in different ALCs) which is not ideal.  However everything currently works, and I'm otherwise unsure how we could stage this easily.  So it seems OK for a short time to load both until you're able to update your extension.

image

333fred: Sounds good, thanks.

333fred (approved):

davidwengier:
Sorry!

Issue #57012, opened 2021-10-07
Test plan for "field keyword in properties"


jcouv:
Championed proposal: https://github.com/dotnet/csharplang/issues/8635
Spec: [proposals/field-keyword.md](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-14.0/field-keyword.md)

```csharp
public string PropertyConstraint
{
    get;
    set => field = value ?? throw new ArgumentNullException();
} = "";
```

Compiler:
- [x] language version
- [x] `field` keyword
  - [x] based on language version
  - [x] as *primary_expression* only
  - [x] escaped with `@`
  - [x] `this.field`, `base.field`, `Qualifier.field`
  - [x] in simple property: `object P => expr;`
  - [x] in property accessors: expression body; block body
  - [x] in indexer accessors
  - [x] in event accessors
  - [x] in nested lambdas and local functions
  - [x] in property initializer
  - [x] in attributes on property, accessor, nested function
  - [x] in nested function signature as default value
  - [x] in `nameof()`
  - [x] in call to `[Conditional]`
- [x] warning when previously binding to other symbol
  - [x] based on language version
  - [x] member of containing type, base type
  - [x] local, in same or nested function
  - [x] parameter in nested function
  - [x] type parameter, in containing type or nested function
- [x] mixed auto- and manually-implemented accessors (see `FieldKeywordTests.ImplicitAccessorBody_*`)
  - [x] `get`, `set`, `init`
  - [x] in `interface`
- [ ] field use
  - [x] static and instance properties on `class`, `struct`, `ref struct`, `interface`, `record*`
  - [x] `get`-only, `set`-only, `init`-only (see `FieldKeywordTests.AutoPropertyMustHaveGetAccessor_*`)
  - [x] `field` references in multiple properties: each get private backing field
  - [x] auto-implemented properties (including `field`) must override all accessors (see `FieldKeywordTests.Override_*`, `New_*`)
  - [x] in nested function, including `static` field reference in `static` function (see `FieldKeywordTests.Lambda_*`, `LocalFunction_*`)
  - [x] in attribute arguments (see `FieldKeywordTests.Attribute_*`)
  - [x] `ref` returning property (see `FieldKeywordTests.RefReturning_*`)
  - [x] properties with restricted types (see `FieldKeywordTests.RestrictedTypes`)
  - [x] properties with by-ref-like types (see `FieldKeywordTests.ByRefLikeType_*`)
[...]

PR #74706, opened 2024-08-09
Handle partial properties in struct field ordering warning

### LINKED ISSUE #74679: Partial properties on struct warn about field ordering
[SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAFxFANgHwAICYCMBYAKAAcIAnBKCFAAgGcFSBXAYwWoGVqBvQ6v6kuUo1YbAJLdqAcwCmCANzUAvoRUFCgilToMWbTjwL9qGAMwcAFAEpqAXgB81KHeoAWLPN78ipKADcIBBlqUhkIABMAexgUAE8nGDYoTyNvMi0RROoJBycUtSA==)

```csharp
partial struct S {
    partial int I { get; }
}

partial struct S {
    public S() => i = 42;
    private readonly int i;
    partial int I => i;
}
```

Expected behavior: no warning. I should be able to implement a partial property on a struct with my own backing field.

Actual behavior:
```
warning CS0282: There is no defined ordering between fields in multiple declarations of partial struct 'S'. To specify an ordering, all instance fields must be in the same declaration.
```

RikkiGibson:
Closes #74679

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
@@ -2734,7 +2734,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
 
                 foreach (var m in syntax.Members)
                 {
-                    if (HasInstanceData(m))
+                    if (hasInstanceData(m))
                     {
                         if (whereFoundField != null && whereFoundField != syntaxRef)
                         {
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
RikkiGibson, line 1: whitespace off.

src/Compilers/CSharp/Test/Symbol/Symbols/PartialPropertiesTests.cs
```diff
@@ -5000,5 +5000,26 @@ partial class C
                 //     public partial int Prop2 { get; set { } }
                 Diagnostic(ErrorCode.ERR_ConcreteMissingBody, "get").WithArguments("C.Prop2.get").WithLocation(7, 32));
         }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/74679")]
+        public void WRN_SequentialOnPartialClass_NotReportedForPartialProperty()
```
333fred, line 5005: Please add a version of this test that looks like this:

```cs
                partial struct S
                {
                    partial int I { get; }
                }

                partial struct S
                {
                    public S() => i = 42;
                }

                partial struct S
                {
                    private readonly int i;
                    partial int I => i;
                }
```
Probably also worth adding a note to the field test plan, cc @cston.

RikkiGibson: #57012 says:
> Check warning WRN_SequentialOnPartialClass is produced when needed

333fred (approved):

cston (approved):

PR #74712, opened 2024-08-11
Refactor semantic snippets

DoctorKrolic:
I was working on another snippet-related PR and saw a few things that IMO can be improved. So I extracted all these refactorings into this PR. What have I done and why:
- `ISnippetProvider` had `GetSnippetDataAsync` and `GetSnippetAsync` methods and for an external person it is somewhat confusing what is the difference between them. Moreover, from the name `GetSnippetDataAsync` I expect to get eigher `Task<SnippetData>` or a `ValueTask<SnippetData>`, but not `ValueTask<SnippetData?>`. By pooling `AdditionalFilterTexts` up to the interface I can replace confusing `GetSnippetDataAsync` with `IsValidSnippetLocationAsync` and construct `SnippetData` struct when this method returns `true`
- Applied `static` to a few things
- Renamed and reordered a few members for clarity

### DIFF
--- a/src/Features/Core/Portable/Snippets/SnippetChange.cs
+++ b/src/Features/Core/Portable/Snippets/SnippetChange.cs
@@ -25,7 +25,7 @@ internal readonly struct SnippetChange
     public readonly ImmutableArray<SnippetPlaceholder> Placeholders;
 
     /// <summary>
-    /// The position that the cursor should end up on
+    /// The position that the caret should end up on
     /// </summary>
     public readonly int FinalCaretPosition;
 
--- a/src/Features/Core/Portable/Snippets/SnippetChange.cs
+++ b/src/Features/Core/Portable/Snippets/SnippetChange.cs
@@ -19,20 +19,20 @@ internal readonly struct SnippetChange
     public readonly ImmutableArray<TextChange> TextChanges;
 
     /// <summary>
-    /// The position that the cursor should end up on
-    /// </summary>
-    public readonly int CursorPosition;
-
-    /// <summary>
     /// The items that we will want to rename as well as the ordering
     /// in which to visit those items.
     /// </summary>
     public readonly ImmutableArray<SnippetPlaceholder> Placeholders;
 
+    /// <summary>
+    /// The position that the cursor should end up on
+    /// </summary>
+    public readonly int FinalCaretPosition;
+
     public SnippetChange(
         ImmutableArray<TextChange> textChanges,
-        int cursorPosition,
-        ImmutableArray<SnippetPlaceholder> placeholders)
+        ImmutableArray<SnippetPlaceholder> placeholders,
+        int finalCaretPosition)
     {
         if (textChanges.IsEmpty)
         {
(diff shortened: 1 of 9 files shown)


src/Features/Core/Portable/Snippets/SnippetChange.cs
```diff
@@ -18,29 +18,29 @@ namespace Microsoft.CodeAnalysis.Snippets;
...
     /// <summary>
     /// The items that we will want to rename as well as the ordering
     /// in which to visit those items.
     /// </summary>
     public readonly ImmutableArray<SnippetPlaceholder> Placeholders;
 
+    /// <summary>
+    /// The position that the cursor should end up on
+    /// </summary>
+    public readonly int FinalCaretPosition;
```
DoctorKrolic, line 30: This is a rename `CursorPosition` -> `FinalCaretPosition`. First, I wanted to emphasize that this position is used after text changes and placeholders, so `Final`. And second, the standard term for roslyn is 'caret', not 'cursor', so made it more consistent with the rest of the codebase

src/Features/Core/Portable/Snippets/SnippetChange.cs
```diff
@@ -18,29 +18,29 @@ namespace Microsoft.CodeAnalysis.Snippets;
...
+
     public SnippetChange(
         ImmutableArray<TextChange> textChanges,
-        int cursorPosition,
-        ImmutableArray<SnippetPlaceholder> placeholders)
+        ImmutableArray<SnippetPlaceholder> placeholders,
+        int finalCaretPosition)
```
DoctorKrolic, line 35: Reordered for clarity. We first insert _text changes_, then let user loop through _placeholders_ and only then move user's caret to its _final position_

src/Features/Core/Portable/Snippets/SnippetPlaceholder.cs
```diff
@@ -10,42 +10,42 @@ namespace Microsoft.CodeAnalysis.Snippets;
 internal readonly struct SnippetPlaceholder
 {
     /// <summary>
-    /// The identifier in the snippet that needs to be renamed.
+    /// Editable text in the snippet.
     /// </summary>
-    public readonly string Identifier;
+    public readonly string Text;
```
DoctorKrolic, line 15: This can be an actual identifier, but it also can be any other text, e.g. in `if` snippet the placeholder is a full condition expression. So `Identifier` is a confusing name for this thing, renamed it to more generic `Text`

src/Features/Core/Portable/Snippets/SnippetPlaceholder.cs
```diff
@@ -10,42 +10,42 @@ namespace Microsoft.CodeAnalysis.Snippets;
...
 
     /// <summary>
     /// The positions associated with the identifier that will need to
     /// be converted into LSP formatted strings.
     /// </summary>
-    public readonly ImmutableArray<int> PlaceHolderPositions;
+    public readonly ImmutableArray<int> StartingPositions;
```
DoctorKrolic, line 21: Wanted to emphasize that these are starting positions for clarity. Also since they are already inside `SnippetPlaceholder` type, having `PlaceHolder` in the name is an overkill

DoctorKrolic:
@akhera99 PTAL

akhera99 (approved):
I like the refactoring of the SnippetProvider. I agree, if I were someone new coming into this codebase I would be confused by the distinction of GetSnippetData and GetSnippetChange (I should write better comments).
Thanks for the contribution!

PR #74714, opened 2024-08-12
Fix allow ref like types for synthesized TypeParameterSymbols in Ad-Hoc Delegates

bernd5:
Currently synthesized TypeParameterSymbols for array params return AllowsRefLikeType=true, which is not true.
An array can not contain ref-structs.

### DIFF
--- a/src/Compilers/CSharp/Test/Semantic/Semantics/LocalFunctionTests.cs
+++ b/src/Compilers/CSharp/Test/Semantic/Semantics/LocalFunctionTests.cs
@@ -32,6 +32,32 @@ namespace Microsoft.CodeAnalysis.CSharp.UnitTests
     [CompilerTrait(CompilerFeature.LocalFunctions)]
     public class LocalFunctionTests : LocalFunctionsTestBase
     {
+        [Fact]
+        public void TestAdHocDelegateWithParams_NotAllowRefLikeTest()
+        {
+            var code = """
+            var counter = Counter;
+            counter.Invoke();
+
+            void Counter(params int[] arr)
+            {
+            }
+            """;
+            var comp = CreateCompilation(code);
+            comp.VerifyDiagnostics();
+            var tree = comp.SyntaxTrees[0];
+            var model = comp.GetSemanticModel(tree);
+
+            var decl = tree.GetRoot().DescendantNodes().OfType<LocalDeclarationStatementSyntax>().Single();
+            var declType = decl.Declaration.Type;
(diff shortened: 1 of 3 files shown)


jaredpar:
Do we have a bug for this?

bernd5:
fixes #74823

src/Compilers/CSharp/Test/Semantic/Semantics/LocalFunctionTests.cs
```diff
@@ -32,6 +32,40 @@ internal static void VerifyDiagnostics(string source, params DiagnosticDescripti
     [CompilerTrait(CompilerFeature.LocalFunctions)]
     public class LocalFunctionTests : LocalFunctionsTestBase
     {
+        [Fact]
+        public void TestAdHocDelegateWithParams_NotAllowRefLikeTest()
```
jjonescz, line 36: It feels like this should be in a different file than `LocalFunctionTests`, for example `RefStructInterfacesTests` seems better.

src/Compilers/CSharp/Portable/Symbols/AllowsRefLikeTypeSymbolVisitor.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.CodeAnalysis.CSharp.Symbols
+{
+    internal sealed class AllowsRefLikeTypeSymbolVisitor : SymbolVisitor
```
jjonescz, line 13: Are there any other cases than just `params T[]` that need to be handled specially?

- If yes, we should have tests for these cases.
- If no, we don't need a visitor, we could simply check just that one case.

src/Compilers/CSharp/Test/Semantic/Semantics/LocalFunctionTests.cs
bernd5: I didn't know where to define it - that would be a better place or something like "allowRefStructTests.cs"

src/Compilers/CSharp/Portable/Symbols/AllowsRefLikeTypeSymbolVisitor.cs
bernd5: Not as far as I know

PR #74729, opened 2024-08-13
Report field diagnostic as warning, only report when ambiguous, and only with preview language version

cston:
Report field diagnostic as warning, only report when field is ambiguous, and only when compiling with preview language version.

### DIFF
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -6869,7 +6869,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>Compiling requires binding the lambda expression many times. Consider declaring the lambda expression with explicit parameter types, or if the containing method call is generic, consider using explicit type arguments.</value>
   </data>
   <data name="WRN_FieldIsAmbiguous" xml:space="preserve">
-    <value>In language version {0}, the 'field' keyword binds to a synthesized backing field for the property. To avoid generating a synthesized backing field, and to refer to the existing member instead, use 'this.field' or '@field' instead.</value>
+    <value>In language version {0}, the 'field' keyword binds to a synthesized backing field for the property. To avoid generating a synthesized backing field, and to refer to the existing member, use 'this.field' or '@field' instead.</value>
   </data>
   <data name="WRN_FieldIsAmbiguous_Title" xml:space="preserve">
     <value>The 'field' keyword binds to a synthesized backing field for the property.</value>
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -6869,10 +6869,10 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>Compiling requires binding the lambda expression many times. Consider declaring the lambda expression with explicit parameter types, or if the containing method call is generic, consider using explicit type arguments.</value>
   </data>
   <data name="WRN_FieldIsAmbiguous" xml:space="preserve">
-    <value>'{0}' binds to the synthesized backing field for the property in language version {1}. Use '@{0}' to bind to the existing symbol instead.</value>
+    <value>In language version {0}, the 'field' keyword binds to a synthesized backing field for the property. To avoid generating a synthesized backing field, and to refer to the existing member instead, use 'this.field' or '@field' instead.</value>
   </data>
   <data name="WRN_FieldIsAmbiguous_Title" xml:space="preserve">
-    <value>'field' binds to the synthesized backing field for the property. Use '@field' to bind to the existing symbol instead.</value>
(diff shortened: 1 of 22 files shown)


src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -6868,8 +6868,8 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="INF_TooManyBoundLambdas_Title" xml:space="preserve">
     <value>Compiling requires binding the lambda expression many times. Consider declaring the lambda expression with explicit parameter types, or if the containing method call is generic, consider using explicit type arguments.</value>
   </data>
-  <data name="INF_IdentifierConflictWithContextualKeyword" xml:space="preserve">
-    <value>'{0}' is a contextual keyword in property accessors starting in language version {1}. Use '@{0}' instead.</value>
+  <data name="WRN_FieldIsAmbiguous" xml:space="preserve">
+    <value>'{0}' binds to the synthesized backing field for the property in language version {1}. Use '@{0}' to bind to the existing symbol instead.</value>
```
RikkiGibson, line 6872: It would be nice to do a "first chance" lookup to see if `field` is a member of `this` (assuming there is a `this` parameter in scope), and in that case recommend user to use `this.field` instead.

RikkiGibson:
I think this is a good direction for this warning, but there's one case I'm concerned about:

```cs
// Library.dll
public class Base
{
    protected string field;
}

// MyApp.csproj
public class Derived : Base
{
    public string Prop => field; // should we warn here? What should the user do if they want to use field?
}
```

I feel optimistic that people are not going to be hit by this in real code, that in 99.9% of real cases that they want to use field, the user's natural reaction will be to rename the thing in their code which they named `field`, in order to avoid this warning. But, it feels like it would be good to record/demonstrate this scenario anyway.

cston:
I've added the suggested test, along with possible work arounds.

To reference `Base.field`, the user could use `this.field` or `base.field`; and to reference the synthesized backing field, the user could add `pragma warning disable 9258`.

---
In reply to: [2286824916](https://github.com/dotnet/roslyn/pull/74729#issuecomment-2286824916) [](http://example.com/codeflow?ancestors=2286824916)

src/Compilers/CSharp/Portable/CSharpResources.resx
cston: Let's wait on this. I'd prefer to use one warning rather than two but the wording may get confusing if we recommend something like "... or use `this.field` if a member of the instance" in cases where `field` is not an instance member.

RikkiGibson (approved):

RikkiGibson (approved):

cston:
@333fred, @dotnet/roslyn-compiler, for a second review, thanks.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -919,46 +907,54 @@ class A : Attribute
...
-                Diagnostic(ErrorCode.INF_IdentifierConflictWithContextualKeyword, "field").WithArguments("field", "preview").WithLocation(9, 41),
                 // (9,41): error CS0182: An attribute argument must be a constant expression, typeof expression or array creation expression of an attribute parameter type
-                //     object P3 { get { return null; } [A(field)] set { } }
-                Diagnostic(ErrorCode.ERR_BadAttributeArgument, "field").WithLocation(9, 41));
+                //     object P2 { get { return null; } [A(field)] set { } }
+                Diagnostic(ErrorCode.ERR_BadAttributeArgument, "field").WithLocation(9, 41),
+                // (14,20): warning CS9258: In language version preview, the 'field' keyword binds to a synthesized backing field for the property. To avoid generating a synthesized backing field, and to refer to the existing member, use 'this.field' or '@field' instead.
+                //     object P3 { [A(field)] get { return null; } set { } }
+                Diagnostic(ErrorCode.WRN_FieldIsAmbiguous, "field").WithArguments("preview").WithLocation(14, 20),
```
333fred, line 932: Isn't our default with no options Preview? And if it isn't, isn't this baseline going to change whenever we increment the default?

cston: Yes, the default with no options is language version preview. This baseline will change after the feature ships, with the baseline including the exact language version. I believe this matches how we typically handling language version diagnostics in tests.

333fred: For those types of baselines, we usually use `RegularNext` as an explicit language version.

cston: I've updated most of the tests to use `LanguageVersionFacts.CSharpNext`. I've left this one test unchanged though because I think it should use the latest language version always.

333fred (approved):

PR #74730, opened 2024-08-13
Expose document symbols to Razor cohosting

davidwengier:
Roslyn side of https://github.com/dotnet/razor/issues/10689
Razor side: https://github.com/dotnet/razor/pull/10728

Slightly annoying having to work around the VS dependency for images, but seems to look good in VS regardless:
image

### DIFF
--- a/src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
+++ b/src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
@@ -25,7 +25,7 @@ namespace Microsoft.CodeAnalysis.LanguageServer.Handler
     /// </summary>
     [ExportCSharpVisualBasicStatelessLspService(typeof(DocumentSymbolsHandler)), Shared]
     [Method(Methods.TextDocumentDocumentSymbolName)]
-    internal sealed class DocumentSymbolsHandler : ILspServiceDocumentRequestHandler<RoslynDocumentSymbolParams, SumType<DocumentSymbol[], SymbolInformation[]>>
+    internal sealed class DocumentSymbolsHandler : ILspServiceDocumentRequestHandler<RoslynDocumentSymbolParams, SumType<RoslynDocumentSymbol[], SymbolInformation[]>>
     {
         public bool MutatesSolutionState => false;
         public bool RequiresLSPSolution => true;
@@ -38,7 +38,7 @@ namespace Microsoft.CodeAnalysis.LanguageServer.Handler
 
         public TextDocumentIdentifier GetTextDocumentIdentifier(RoslynDocumentSymbolParams request) => request.TextDocument;
 
-        public Task<SumType<DocumentSymbol[], SymbolInformation[]>> HandleRequestAsync(RoslynDocumentSymbolParams request, RequestContext context, CancellationToken cancellationToken)
+        public Task<SumType<RoslynDocumentSymbol[], SymbolInformation[]>> HandleRequestAsync(RoslynDocumentSymbolParams request, RequestContext context, CancellationToken cancellationToken)
         {
             var document = context.GetRequiredDocument();
             var clientCapabilities = context.GetRequiredClientCapabilities();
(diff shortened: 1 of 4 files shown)


src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.CodeAnalysis.LanguageServer.Handler
     /// </summary>
     [ExportCSharpVisualBasicStatelessLspService(typeof(DocumentSymbolsHandler)), Shared]
     [Method(Methods.TextDocumentDocumentSymbolName)]
-    internal sealed class DocumentSymbolsHandler : ILspServiceDocumentRequestHandler<RoslynDocumentSymbolParams, object[]>
+    internal sealed class DocumentSymbolsHandler : ILspServiceDocumentRequestHandler<RoslynDocumentSymbolParams, SumType<RoslynDocumentSymbol[], SymbolInformation[]>>
```
dibarbet, line 28: could this be `SumType<DocumentSymbol[], SymbolInformation[]>` (and return RoslynDocumentSymbol) or does that not work?

src/Tools/ExternalAccess/Razor/Cohost/Handlers/DocumentSymbols.cs
```diff
@@ -0,0 +1,47 @@
...
+using LSP = Roslyn.LanguageServer.Protocol;
+
+namespace Microsoft.CodeAnalysis.ExternalAccess.Razor.Cohost.Handlers;
+
+internal static class DocumentSymbols
+{
+    public static async Task<SumType<DocumentSymbol[], SymbolInformation[]>> GetDocumentSymbolsAsync(Document document, bool useHierarchicalSymbols, CancellationToken cancellationToken)
+    {
+        // The symbol information service in Roslyn lives in EditorFeatures and has VS dependencies. for glyph images,
+        // so isn't available in OOP. The default implementation is available in OOP, but not in the Roslyn MEF composition,
```
dibarbet, line 19: > but not in the Roslyn MEF composition

Hmm - not sure why this isn't the case.  I think its regular workspace service.  Not 100% sure but I would have thought it should be available.

dibarbet: oh is the protocol project not in the OOP mef composition?

src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
davidwengier: That's what I had at first, but the Document Outline tests failed. Guessing it is missing a converter, but it wasn't clear to me why that uses LSP at all, so didn't want to poke the bear :)

src/Tools/ExternalAccess/Razor/Cohost/Handlers/DocumentSymbols.cs
davidwengier: Yes, precisely. That might come up again in other endpoints so could be something we need to talk about in future, but document symbol didn't seem important enough to block on it

src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
```diff
@@ -38,53 +38,58 @@ public DocumentSymbolsHandler()
 
         public TextDocumentIdentifier GetTextDocumentIdentifier(RoslynDocumentSymbolParams request) => request.TextDocument;
 
-        public async Task<object[]> HandleRequestAsync(RoslynDocumentSymbolParams request, RequestContext context, CancellationToken cancellationToken)
+        public Task<SumType<RoslynDocumentSymbol[], SymbolInformation[]>> HandleRequestAsync(RoslynDocumentSymbolParams request, RequestContext context, CancellationToken cancellationToken)
         {
             var document = context.GetRequiredDocument();
             var clientCapabilities = context.GetRequiredClientCapabilities();
+            var useHierarchicalSymbols = clientCapabilities.TextDocument?.DocumentSymbol?.HierarchicalDocumentSymbolSupport == true || request.UseHierarchicalSymbols;
```
ryzngard, line 45: It seems weird that even if the client doesn't support `HierarchicalSymbols` we'd still use it if the request said to?

davidwengier: The Document Outline window in Roslyn calls the document symbols LSP endpoint, and wants hierarchical results all the time, but the VS LSP client doesn't support them. In the LSP spec the non-hierarchical result is deprecated and I should probably log an issue on the platform to move off it, and then a lot of this nonsense can go away.

src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
dibarbet: I would guess the document outline tests just need to request the base type, and then cast the result it gets back here https://github.com/dotnet/roslyn/blob/main/src/VisualStudio/CSharp/Test/DocumentOutline/DocumentOutlineTestsBase.cs#L89

src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
dibarbet: this is for document outline.  base vs client doesn't have hierarchical support IIRC, but we need it for our custom doc outline support.

ryzngard (approved):

src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
davidwengier: It's an actual serialization issue, the `Glyph` property is just lost. I could probably add a call to `AddOrReplaceConverter<DocumentSymbol, RoslynDocumentSymbol>();` [here](https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/Protocol/Protocol/Internal/Converters/VSInternalExtensionUtilities.cs#L31) though if you like?

dibarbet: if that works, fine with me.  otherwise OK with this too

dibarbet (approved):

src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
davidwengier: I added the converter, seems to work fine. It's cleaner on the Razor side too, which is nice.

dibarbet (approved):

src/LanguageServer/Protocol/Handler/Symbols/DocumentSymbolsHandler.cs
davidwengier: lol, that caused a different set of test failures 🤦‍♂️

PR #74750, opened 2024-08-13
Make host outputs public

chsienki:
Implements https://github.com/dotnet/roslyn/issues/74753

### DIFF
--- a/src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs
+++ b/src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs
@@ -10476,11 +10476,13 @@ class C
 
             var generator = new PipelineCallbackGenerator((ctx) =>
             {
+#pragma warning disable RSEXPERIMENTAL004 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
                 ctx.RegisterHostOutput(ctx.CompilationProvider, (hostCtx, value) =>
                 {
                     hostOutputRan = true;
                     hostCtx.AddOutput("output", "value");
                 });
+#pragma warning restore RSEXPERIMENTAL004 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
 
                 ctx.RegisterSourceOutput(ctx.CompilationProvider, (spc, po) =>
                 {
@@ -10592,8 +10594,6 @@ class C
 
             return output;
         }
-
-
 
         [WorkItem(899050, "http://vstfdevdiv:8080/DevDiv2/DevDiv/_workitems/edit/899050")]
         [Fact]
--- a/src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs
+++ b/src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs
@@ -10462,7 +10462,7 @@ class C
 
             void RunWithCache() => VerifyOutput(dir, src, includeCurrentAssemblyAsAnalyzerReference: false, additionalFlags: new[] { "/langversion:preview", "/features:enable-generator-cache" }, generators: new[] { generator.AsSourceGenerator() }, driverCache: cache, analyzers: null);
         }
-        
+
         [Fact]
         public void Compiler_DoesNot_RunHostOutputs()
         {
--- a/src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs
+++ b/src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs
@@ -10462,6 +10462,42 @@ class C
 
             void RunWithCache() => VerifyOutput(dir, src, includeCurrentAssemblyAsAnalyzerReference: false, additionalFlags: new[] { "/langversion:preview", "/features:enable-generator-cache" }, generators: new[] { generator.AsSourceGenerator() }, driverCache: cache, analyzers: null);
         }
+        
+        [Fact]
+        public void Compiler_DoesNot_RunHostOutputs()
+        {
+            var dir = Temp.CreateDirectory();
+            var src = dir.CreateFile("temp.cs").WriteAllText(@"
+class C
+{
+}");
+            bool hostOutputRan = false;
+            bool sourceOutputRan = false;
+
+            var generator = new PipelineCallbackGenerator((ctx) =>
+            {
+                ctx.RegisterHostOutput(ctx.CompilationProvider, (hostCtx, value) =>
+                {
+                    hostOutputRan = true;
+                    hostCtx.AddOutput("output", "value");
+                });
+
+                ctx.RegisterSourceOutput(ctx.CompilationProvider, (spc, po) =>
+                {
+                    sourceOutputRan = true;
+                    spc.AddSource("output.cs", "//value");
+                });
+            });
+
+            VerifyOutput(dir, src, includeCurrentAssemblyAsAnalyzerReference: false, additionalFlags: new[] { "/langversion:preview" }, generators: new[] { generator.AsSourceGenerator() }, analyzers: null);
+
+            Assert.False(hostOutputRan);
+            Assert.True(sourceOutputRan);
+
+            // Clean up temp files
+            CleanupAllGeneratedFiles(src.Path);
+            Directory.Delete(dir.Path, true);
+        }
 
         private static int OccurrenceCount(string source, string word)
         {
@@ -10556,6 +10592,8 @@ class C
 
             return output;
         }
+
+
 
         [WorkItem(899050, "http://vstfdevdiv:8080/DevDiv2/DevDiv/_workitems/edit/899050")]
         [Fact]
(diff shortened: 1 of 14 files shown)


chsienki:
@dotnet/roslyn-compiler for reviews please :)

333fred:
Looks like failures are real.

src/Compilers/CSharp/Test/Semantic/SourceGeneration/GeneratorDriverTests.cs
```diff
@@ -4465,5 +4465,95 @@ public void GeneratorDriver_ReparsesPostInitTrees_IfParseOptionsChange_WhileSupp
...
+        }
+
+        [Fact]
+        public void GeneratorDriver_No_HostOutputs_WhenDisabled()
+        {
+            var generator = new PipelineCallbackGenerator((ctx) => { ctx.RegisterHostOutput(ctx.CompilationProvider, (hostCtx, c) => { hostCtx.AddOutput("a", "value"); }); });
+
+            var parseOptions = CSharpParseOptions.Default;
+            var compilation = CreateCompilation("class Compilation1{}", parseOptions: parseOptions);
+            GeneratorDriver driver = CSharpGeneratorDriver.Create([generator.AsSourceGenerator()], parseOptions: parseOptions, driverOptions: new GeneratorDriverOptions(IncrementalGeneratorOutputKind.Host));
```
jjonescz, line 4495: Consider testing that host outputs are present when another `IncrementalGeneratorOutputKind` is disabled

src/Compilers/CSharp/Test/Semantic/SourceGeneration/GeneratorDriverTests.cs
```diff
@@ -4465,5 +4465,95 @@ public void GeneratorDriver_ReparsesPostInitTrees_IfParseOptionsChange_WhileSupp
...
+            Assert.Collection(result.HostOutputs,
+                (e) => { Assert.Equal("a", e.Key); Assert.Equal("value", e.Value); },
+                (e) => { Assert.Equal("b", e.Key); Assert.Equal("value2", e.Value); }
+            );
+        }
+
+        [Fact]
+        public void GeneratorDriver_Multiple_HostOutputs_SameName()
+        {
+            var generator = new PipelineCallbackGenerator((ctx) => { ctx.RegisterHostOutput(ctx.CompilationProvider, (hostCtx, c) => { hostCtx.AddOutput("a", "value"); hostCtx.AddOutput("a", "value2"); }); });
```
jjonescz, line 4525: Consider testing multiple outputs with the same name across two different source generators.

src/Compilers/CSharp/Test/Semantic/SourceGeneration/GeneratorDriverTests.cs
```diff
@@ -4465,5 +4465,95 @@ public void GeneratorDriver_ReparsesPostInitTrees_IfParseOptionsChange_WhileSupp
...
+            var parseOptions = CSharpParseOptions.Default;
+            var compilation = CreateCompilation("class Compilation1{}", parseOptions: parseOptions);
+            GeneratorDriver driver = CSharpGeneratorDriver.Create([generator.AsSourceGenerator()], parseOptions: parseOptions);
+
+            driver = driver.RunGenerators(compilation);
+            var runResult = driver.GetRunResult();
+            var result = Assert.Single(runResult.Results);
+
+            Assert.Empty(result.HostOutputs);
+            Assert.IsType<ArgumentException>(result.Exception);
```
jjonescz, line 4536: Consider also testing when the lambda passed to RegisterHostOutput throws an exception.

src/Compilers/Core/Portable/SourceGeneration/IncrementalContexts.cs
```diff
@@ -198,6 +207,40 @@ internal GeneratorFilterContext(ISourceGenerator generator, CancellationToken ca
...
+        /// <param name="name">The name of the output to be added.</param>
+        /// <param name="value">The output to be added.</param>
+        /// <remarks>
+        /// A host output has no defined use. It does not contribute to the final compilation in any way. Any outputs registered 
+        /// here are made available via the <see cref="GeneratorRunResult.HostOutputs"/> collection, and it is up the host to
+        /// decide how to use them. A host may also disable these outputs altogether if they are not needed. The generator driver
+        /// otherwise makes no guarantees about how the outputs are used, other than that they will be present if the host has
+        /// requested they be produced.
+        /// </remarks>
+        public void AddOutput(string name, object value) => Outputs.Add((name, value));
```
jjonescz, line 236: Shouldn't this also be a dictionary so we throw exceptions for duplicate names early?

src/Compilers/Core/Portable/SourceGeneration/IncrementalContexts.cs
```diff
@@ -209,19 +252,19 @@ internal readonly struct IncrementalExecutionContext
...
             Sources = sources;
-            HostOutputBuilder = ArrayBuilder<(string, string)>.GetInstance();
+            HostOutputBuilder = ImmutableDictionary.CreateBuilder<string, object>();
             Diagnostics = DiagnosticBag.GetInstance();
         }
 
-        internal (ImmutableArray<GeneratedSourceText> sources, ImmutableArray<Diagnostic> diagnostics, GeneratorRunStateTable executedSteps, ImmutableArray<(string Key, string Value)> hostOutputs) ToImmutableAndFree()
-                => (Sources.ToImmutableAndFree(), Diagnostics.ToReadOnlyAndFree(), GeneratorRunStateBuilder.ToImmutableAndFree(), HostOutputBuilder.ToImmutableAndFree());
+        internal (ImmutableArray<GeneratedSourceText> sources, ImmutableArray<Diagnostic> diagnostics, GeneratorRunStateTable executedSteps, ImmutableDictionary<string, object> hostOutputs) ToImmutableAndFree()
+                => (Sources.ToImmutableAndFree(), Diagnostics.ToReadOnlyAndFree(), GeneratorRunStateBuilder.ToImmutableAndFree(), HostOutputBuilder.ToImmutable());
```
jjonescz, line 267: Why don't we free the HostOutputBuilder here?

src/Compilers/Core/Portable/SourceGeneration/IncrementalContexts.cs
```diff
@@ -209,19 +252,19 @@ internal readonly struct IncrementalExecutionContext
...
 
         internal void Free()
```
jjonescz, line 269: Why don't we free the HostOutputBuilder and GeneratorRunStateBuilder here?

src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs
```diff
@@ -10462,6 +10462,44 @@ class C
 
             void RunWithCache() => VerifyOutput(dir, src, includeCurrentAssemblyAsAnalyzerReference: false, additionalFlags: new[] { "/langversion:preview", "/features:enable-generator-cache" }, generators: new[] { generator.AsSourceGenerator() }, driverCache: cache, analyzers: null);
         }
+        
+        [Fact]
+        public void Compiler_DoesNot_RunHostOutputs()
```
jjonescz, line 10467: Consider adding a similar test for VB.

src/Compilers/Core/Portable/SourceGeneration/IncrementalContexts.cs
chsienki: This one is tricky. Because of the way the NodeStateTables work, we would need to share the dictionary across multiple calls of the callback. It's not impossible, and in fact I actually started doing it, but it's not simple.

This is the same way `RegisterSourceOutput` works today, we effectively let you add duplicates then throw when we combine them (which was different from V1 generators where it was at the point of addition). I agree it's a little odd, but you do get the same effect, just a different call stack, and the complexity of it didn't seem worth doing.

src/Compilers/Core/Portable/SourceGeneration/IncrementalContexts.cs
chsienki: Because its not pooled, so there is nothing to free.

src/Compilers/CSharp/Test/Semantic/SourceGeneration/GeneratorDriverTests.cs
```diff
@@ -3085,7 +3085,7 @@ class C { }
             // NOTE: adding new output types will cause this test to fail. Update above as needed.
             foreach (IncrementalGeneratorOutputKind kind in Enum.GetValues(typeof(IncrementalGeneratorOutputKind)))
             {
-                if (kind == IncrementalGeneratorOutputKind.None)
+                if (kind == IncrementalGeneratorOutputKind.None || kind == IncrementalGeneratorOutputKind.Host)
```
jjonescz, line 3088: Instead of skipping this output kind should we update the test as the comment says

```cs
// NOTE: adding new output types will cause this test to fail. Update above as needed.
```

Wouldn't that be just matter of adding a `ctx.RegisterHostOutput` line?

chsienki: Yeah, I looked at doing that, but the rest of the test is checking against the `GeneratedSources` collection, because up to now they all added to that. It didn't seem worth updating it to special case it, but its simple enough I can.

src/Compilers/CSharp/Test/Semantic/SourceGeneration/GeneratorDriverTests.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections;
```
jjonescz, line 6: This added `using` looks unnecessary.

jjonescz (approved):

chsienki:
@333fred for a second review please :)

333fred (approved):

PR #74763, opened 2024-08-14
Remove net35 reference assemblies

jaredpar:
This will save us 100MB of disk I/O in CI. It was only being used in one test. Reworked that a bit to not need `net35` anymore

### DIFF
--- a/src/Compilers/Test/Core/Generated.cs
+++ b/src/Compilers/Test/Core/Generated.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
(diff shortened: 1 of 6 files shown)


jaredpar:
@dotnet/roslyn-compiler PTAL

src/Compilers/Test/Core/Generated.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
```
333fred, line 1: BOM

src/Compilers/Test/Core/Generated.targets
```diff
@@ -1,4 +1,4 @@
-﻿<Project>
+<Project>
```
333fred, line 1: BOM

333fred (approved):

src/Compilers/Test/Core/Generated.cs
jaredpar: will clean these up on the next PR in this series

PR #74766, opened 2024-08-15
Move net451 test assets to net461

jaredpar:
This migrates our tests which depended on `net451` assemblies to use `net461` assemblies instead. The reason for this is to consolidate our test assets to make them easier to manager. This also saves 127MB of disk I/O when doing a restore in CI which should reduce PR validation time. The under pinning mentality of this change is that `net451` and `net461` are not significantly different hence this update should be very straight forward.

This PR is broken into **two commits**. The first commit is a mostly mechanical migration from `net451` to `net461`. There are a number of places where I had to do minor fixups in baselines like changing `System.dll(net451)` to `System(net461)`. The only significant functional issue that popped up reading through the tests is a number of `net451` tests depended on `Array.Empty<T>` not existing. Those required functional changes to work: typically by using `MakeMemberMissing` calls to maintain the behavior. To help reviewing these are the tests impacted by this behavior. 

- `CodeGenExprLambdaTests.MethodCallWithParams`
- `CollectionExpressionTests.ArrayEmpty_01`
- `CollectionExpressionTests.ArrayEmpty_02`
- `InternalsVisibleToAndStrongNameTests.IvtVirtualCall1`
- `InternalsVisibleToAndStrongNameTests.IvtVirtualCall2`
- `MissingSpecialMember.AllSpecialTypeMembers`

Note: Had to temporarily disable [BC30645ERR_InvalidOptionalParameterUsage1b][test-issue]. Will re-enable that test once I can publish to nuget.org again, fix the ref assembly bug that will let me port this test to `net461`

The second commit is refactor rename of various APIs that had the part`WithMscorlib45` to `WithMscorlib461`. Suggest mostly spot checking that one. 

[test-issue]: https://github.com/dotnet/roslyn/issues/74779


jaredpar:
@dotnet/roslyn-compiler PTAL

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -8,6 +8,7 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using ICSharpCode.Decompiler.IL;
```
cston, line 11: Is this `using` needed?

jaredpar: why does that keep happening ...

cston (approved):

src/Compilers/CSharp/Test/Symbol/Symbols/Retargeting/NoPia.cs
```diff
@@ -12,6 +12,8 @@
 using Microsoft.CodeAnalysis.Test.Utilities;
 using Xunit;
 using Roslyn.Test.Utilities;
+using static Roslyn.Test.Utilities.TestMetadata;
```
333fred, line 15: Nit: reorder.

333fred (approved):
I spot checked a number of the files in the PR, and went through the test helper classes. Going to sign off overall.

cston (approved):

jaredpar:
Integration test failures are unrelated

PR #74767, opened 2024-08-15
Snippet refactoring 2

DoctorKrolic:
Follow up to https://github.com/dotnet/roslyn/pull/74712

2 changes here:
1. Do not pass `SnippetContext` around by `in` reference. The only actual data, that `SnippetContext` holds is 1 reference to a `SyntaxContext`. Therefore passing it around by `in` doesn't win anything, but adds unnecessary cognitive overhead when reading that code
2. Since `SnippetContext` already holds a `SyntaxContext` instance, all async work of getting syntax tree and semantic model from the document has already been done. Therefore it isn't practical to keep this path async. I remember there was a discussion, that semantic snippets engine is a potential candidate for becoming public API. And since there is no way `SyntaxContext` become public, I additionaly exposed `SemanticModel` as a public property of `SnippetContext`. That way a potential API user can get everything they need without having to deal with any async stuff:
- Syntax tree: `context.SemanticModel.SyntaxTree`
- Syntax root: `context.SemanticModel.SyntaxTree.GetRoot()`
- Semantic model: `context.SemanticModel`
- Compilation: `context.SemanticModel.Compilation`

### DIFF
--- a/src/Features/Core/Portable/Snippets/SnippetProviders/ISnippetProvider.cs
+++ b/src/Features/Core/Portable/Snippets/SnippetProviders/ISnippetProvider.cs
@@ -28,7 +28,7 @@ internal interface ISnippetProvider
     /// <summary>
     /// Determines if a snippet can exist at a particular location.
     /// </summary>
-    ValueTask<bool> IsValidSnippetLocationAsync(in SnippetContext context, CancellationToken cancellationToken);
+    ValueTask<bool> IsValidSnippetLocationAsync(SnippetContext context, CancellationToken cancellationToken);
 
     /// <summary>
     /// Gets the Snippet change from the corresponding snippet provider.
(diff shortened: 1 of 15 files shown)


src/Features/Core/Portable/Snippets/SnippetProviders/ISnippetProvider.cs
```diff
@@ -28,7 +28,7 @@ internal interface ISnippetProvider
     /// <summary>
     /// Determines if a snippet can exist at a particular location.
     /// </summary>
-    ValueTask<bool> IsValidSnippetLocationAsync(in SnippetContext context, CancellationToken cancellationToken);
+    bool IsValidSnippetLocation(SnippetContext context, CancellationToken cancellationToken);
```
sharwell, line 31: 💭 I was hesitant to make this change since this may be a public API in the future.

DoctorKrolic: As I said, by exposing `SemanticModel` in `SnippetContext` we eliminate the need for async for a potential user even if semantic snippets become public API. When you have it, all further symbol checking and stuff are normal syncronus code

sharwell (approved):

akhera99 (approved):
Looks good, thanks for the contribution!

PR #74775, opened 2024-08-15
Redirect SDK analyzers to VS

jjonescz:
Spec: https://github.com/jjonescz/dotnet-sdk/blob/tearing-redirect/documentation/general/analyzer-redirecting.md
Part of https://github.com/dotnet/sdk/issues/42087.

### DIFF
--- a/src/VisualStudio/Core/Def/Diagnostics/InsertedAnalyzerProvider.cs
+++ b/src/VisualStudio/Core/Def/Diagnostics/InsertedAnalyzerProvider.cs
@@ -1,0 +1,163 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Composition;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Editor.Shared.Utilities;
+using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.PooledObjects;
+using Microsoft.VisualStudio.Shell;
+using Roslyn.Utilities;
+
+namespace Microsoft.VisualStudio.LanguageServices.Implementation.Diagnostics;
+
+/// <summary>
+/// Discovers analyzers inserted into Visual Studio which are candidates for redirection.
+/// Analyzers loaded from SDK are redirected to these VS-inserted analyzers
+/// by <see cref="RedirectingAnalyzerAssemblyResolver"/>.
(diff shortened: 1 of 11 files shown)


src/VisualStudio/Core/Def/Diagnostics/InsertedAnalyzerProvider.cs
```diff
@@ -0,0 +1,163 @@
...
+using Roslyn.Utilities;
+
+namespace Microsoft.VisualStudio.LanguageServices.Implementation.Diagnostics;
+
+/// <summary>
+/// Discovers analyzers inserted into Visual Studio which are candidates for redirection.
+/// Analyzers loaded from SDK are redirected to these VS-inserted analyzers
+/// by <see cref="RedirectingAnalyzerAssemblyResolver"/>.
+/// </summary>
+internal sealed class InsertedAnalyzerProvider
```
jjonescz, line 25: This is for the most part a copy of `VisualStudioDiagnosticAnalyzerProvider` which does something similar - discovers analyzers from VSIXes.

src/VisualStudio/Core/Def/ProjectSystem/VisualStudioProjectFactory.cs
```diff
@@ -77,6 +80,8 @@ public async Task<ProjectSystemProject> CreateAndAddToWorkspaceAsync(
 
         var vsixAnalyzerProvider = await _vsixAnalyzerProviderFactory.GetOrCreateProviderAsync(cancellationToken).ConfigureAwait(false);
 
+        _visualStudioWorkspaceImpl.LazyInsertedAnalyzerProvider = await _insertedAnalyzerProviderFactory.GetOrCreateProviderAsync(cancellationToken).ConfigureAwait(false);
```
jjonescz, line 83: I wasn't sure where to put this "discovery of VSIX analyzers" (I'm not very familiar with the IDE part of the codebase) - I need that to be done before the `RedirectingAnalyzerAssemblyResolver` is called. I chose this place because it's where the similar `IVisualStudioDiagnosticAnalyzerProviderFactory` is called.

src/VisualStudio/Core/Def/Diagnostics/RedirectingAnalyzerAssemblyResolver.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                var redirectedPath = Path.Combine(matcher.TargetDirectory, redirectedDirectory, assemblyName.Name + ".dll");
+                return Assembly.LoadFile(redirectedPath);
+            }
+        }
+
+        return null;
+    }
+
+    private ImmutableArray<Matcher> CreateMatchers()
```
jjonescz, line 46: This could be improved (e.g., by creating regexes), but I'm not sure the spec is final yet.

PR #74781, opened 2024-08-15
Fix Copilot option bug

akhera99:
Fixes bug where default option value was not being respected

### DIFF
--- a/src/VisualStudio/CSharp/Impl/Options/CSharpVisualStudioCopilotOptionsService.cs
+++ b/src/VisualStudio/CSharp/Impl/Options/CSharpVisualStudioCopilotOptionsService.cs
@@ -77,7 +77,7 @@ internal sealed class CSharpVisualStudioCopilotOptionsService : ICopilotOptionsS
         var settingManager = await _settingsManagerTask.ConfigureAwait(false);
         // The bool setting is persisted as 0=None, 1=True, 2=False, so it needs to be retrieved as an int.
         return settingManager.TryGetValue($"{CopilotOptionNamePrefix}.{optionName}", out int isEnabled) == GetValueResult.Success
-            && isEnabled == 1;
+            && isEnabled != 2;
     }
 
     public Task<bool> IsCodeAnalysisOptionEnabledAsync()


src/VisualStudio/CSharp/Impl/Options/CSharpVisualStudioCopilotOptionsService.cs
```diff
@@ -77,7 +77,7 @@ public async Task<bool> IsCopilotOptionEnabledAsync(string optionName)
         var settingManager = await _settingsManagerTask.ConfigureAwait(false);
         // The bool setting is persisted as 0=None, 1=True, 2=False, so it needs to be retrieved as an int.
         return settingManager.TryGetValue($"{CopilotOptionNamePrefix}.{optionName}", out int isEnabled) == GetValueResult.Success
-            && isEnabled == 1;
+            && isEnabled != 2;
```
Cosifne, line 80: Any context?

Cosifne: Oh nevermind I see the comment.

genlu: yea, this is now enabled by default (the source of truth is in copilot repo, where all those options are defined)

genlu (approved):

src/VisualStudio/CSharp/Impl/Options/CSharpVisualStudioCopilotOptionsService.cs
JoeRobich: Copilot needs an option provider service so they can apply the defaults when you request values.

genlu: do you have an example for that?

JoeRobich: No, it was all me wishing something existed. They could write the defaults for their settings as `$"{CopilotOptionNamePrefix}.{option.Name}.Default"` or something similar. Then we could defer to whatever they ship.

tmat: Why are we not using our existing options infra?

PR #74790, opened 2024-08-16
Reduce allocations in CSharpSyntaxFacts.AppendMembers (again)

ToddGrun:
A recent change(https://github.com/dotnet/roslyn/pull/74596) modified this code to take in an ArrayBuilder to allow callers to easily use pooled arrays. However, the size of the data populated into those arrays commonly exceeds the ArrayBuilder reuse threshold, those reducing the allocation benefit realized in those codepaths.

This PR changes the callers to instead use a pooled list, thus avoiding the threshold issue. Current scrolling speedometer results for the typing scenario still show this as allocating around 35 MB (which was about 3.5% at the time of the earlier PR, about 4.6% of allocations currently). My hope is that this PR will get rid of nearly all those allocations this time.

*** relevant allocations from the typing scenario in the scrolling speedometer ***
image

### DIFF
--- a/src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
@@ -9,6 +9,7 @@ using System.Collections.Immutable;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
@@ -385,11 +386,15 @@ namespace Roslyn.Utilities
             if (source == null)
                 return ImmutableArray<TResult>.Empty;
 
-            var builder = ArrayBuilder<TResult>.GetInstance(source.Count);
+            var builder = new TResult[source.Count];
+            var index = 0;
             foreach (var item in source)
-                builder.Add(selector(item));
+            {
+                builder[index] = selector(item);
+                index++;
+            }
 
-            return builder.ToImmutableAndFree();
+            return ImmutableCollectionsMarshal.AsImmutableArray(builder);
         }
 
         public static ImmutableArray<TResult> SelectManyAsArray<TSource, TResult>(this IEnumerable<TSource>? source, Func<TSource, IEnumerable<TResult>> selector)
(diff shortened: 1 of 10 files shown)


333fred:
@ToddGrun were you going to run a smoke test for this one?

ToddGrun:
> e threshold issue. Current scrolling speedometer results for the typing scenario still show this as allocating around 35 MB (which was about 3.5% at the time of the earlier PR, about 4.6% of allocations currently). My hope is that thi

Yeah, I was going to post once the PR was created, but here's the pipeline run that will be generating the PR shortly: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10060803&view=results

src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
```diff
@@ -385,11 +386,15 @@ public static ImmutableArray<TResult> SelectAsArray<TSource, TResult>(this IRead
             if (source == null)
                 return ImmutableArray<TResult>.Empty;
 
-            var builder = ArrayBuilder<TResult>.GetInstance(source.Count);
+            var builder = new TResult[source.Count];
```
333fred, line 389: There are a _lot_ of `SelectAsArray` overloads, and I don't feel particularly comfortable with having one use a different pattern than the rest. If we think that this is a beneficial pattern, we should use it uniformly.

ToddGrun: The other ones all take in an IEnumerable, so there is no way to no the array size to allocate up front.

333fred: I don't necessarily just mean the ones in this file. For example, all of the ones that operate on `ImmutableArray` itself use the `ArrayBuilder` pattern. I would rather see a dedicated commit that switches us to the pattern across the board.

ToddGrun: Went ahead and changed what I could find in the ImmutableArrayExtensions class

CyrusNajmabadi (approved):

RikkiGibson (approved):
Compilers changes LGTM

ToddGrun:
> Numbers look a bit wonky to me. Going to make a tweak and re-run the process.

Did multiple insertion PRs after the last commit, all looks good now. The full 3.5% of SyntaxNode[] allocations look to be gone.

https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/573228
https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/573250

PR #74791, opened 2024-08-16
Fix issue where output path set in construction never updated the project update state

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/7402

https://github.com/dotnet/roslyn/commit/863962d8dd2e0151b1387259eeef0cbf66dfba75 added support to set the backing field for `CompilationOutputAssemblyFilePath` when creating the `ProjectSystemProject`.  This had an issue where the project update state ever observed this value as it did not go through appropriate setter on `CompilationOutputAssemblyFilePath` which is responsible for updating that.

This caused us to later throw exceptions when attempting to remove that path as there was nothing to remove in the project update state.

I updated the code to instead update the path immediately after the `ProjectSystemProject` (and the corresponding solution) was created, which resolves the issue on the test project linked in the issue.

### DIFF
--- a/src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProjectFactory.cs
+++ b/src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProjectFactory.cs
@@ -102,8 +102,7 @@ internal sealed partial class ProjectSystemProjectFactory
             assemblyName,
             creationInfo.CompilationOptions,
             creationInfo.FilePath,
-            creationInfo.ParseOptions,
-            creationInfo.CompilationOutputAssemblyFilePath);
+            creationInfo.ParseOptions);
 
         var versionStamp = creationInfo.FilePath != null
             ? VersionStamp.Create(File.GetLastWriteTimeUtc(creationInfo.FilePath))
@@ -163,6 +162,10 @@ internal sealed partial class ProjectSystemProjectFactory
                 onBeforeUpdate: null,
                 onAfterUpdate: null);
         }).ConfigureAwait(false);
+
+        // Set this value early after solution is created so it is available to Razor.  This will get updated
+        // when the command line is set, but we want a non-null value to be available as soon as possible.
+        project.CompilationOutputAssemblyFilePath = creationInfo.CompilationOutputAssemblyFilePath;
 
         return project;
     }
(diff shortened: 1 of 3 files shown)


ryzngard (approved):
Approved assuming testing with razor with and without cs devkit

dibarbet:
> Approved assuming testing with razor with and without cs devkit

yup, seems to work in both cases

CyrusNajmabadi (approved):

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProjectFactory.cs
```diff
@@ -164,6 +163,10 @@ await ApplyChangeToWorkspaceAsync(w =>
                 onAfterUpdate: null);
         }).ConfigureAwait(false);
 
+        // Set this value early after solution is created so it is available to Razor.  This will get updated
```
jasonmalinowski, line 166: Not sure why Razor is being commented here -- this is just doing the thing the API is asking us to do....

jasonmalinowski, line 166: You probably want to acquire a batch scope around this -- since setting this is going to require going to the global state, we're going to acquire a lock. If there's a batch scope then we'll do it asynchronously, otherwise this will take a blocking lock and block the thread.

jasonmalinowski (commented):
Can we add a test?

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProjectFactory.cs
dibarbet: the early set was initially added for Razor here - https://github.com/dotnet/roslyn/commit/863962d8dd2e0151b1387259eeef0cbf66dfba75

It does get set anyway later on when the cmdline is set.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProjectFactory.cs
```diff
@@ -164,6 +163,10 @@ await ApplyChangeToWorkspaceAsync(w =>
...
+        // when the command line is set, but we want a non-null value to be available as soon as possible.
+        project.CompilationOutputAssemblyFilePath = creationInfo.CompilationOutputAssemblyFilePath;
```
JoeRobich, line 168: Could this have essentially been the old change needed here?

```
_projectUpdateState = _projectUpdateState.WithProjectOutputPath(creationInfo.CompilationOutputAssemblyFilePath, creationInfo.ProjectId);
```

dibarbet: Possibly - but going through the same property setter as everything else feels more safer.

jasonmalinowski (approved):

PR #74794, opened 2024-08-17
Async rename commit 

Cosifne:
(In progress, not ready to be review)


Cosifne:
A few open questions:
1. While this change would make most commit operations async, sync commit is still needed in many command handlers (example: `TypeCharCommand`, when you type a character in the code, it will cause the commit session commit, this is a synchronous commit. The Same concept applies to many other command handlers causing code changes, like `Backspace` `Delete` and etc.)
The problem is if we async commit when the user types a char, likely the rename operation would override the user's typing (or maybe rename would unable to rename the workspace, because Workspace has changed during rename). 
Like this:
AsyncRenameProblem1
I rename the class and add `1234` suffix, then press `X` after `static`.
If the commit operation is async `X` will disappear. 
It's a regression from today's experience, which renames could be committed and `X` would be added after `static`
AsyncRenameProblem1Now

2. We handle the typing command handler so the code **looks like** can't be edited when commit is in-progress
But if workspace changed for other reason, we rely on `Workspace.TryApplyChanges()` to do all the checks here.

CyrusNajmabadi:
ic an review tomorrow :)

src/EditorFeatures/Core.Wpf/InlineRename/UI/Adornment/RenameFlyoutViewModel.cs
```diff
@@ -206,20 +222,64 @@ public bool IsRenameOverloadsEditable
...
             return true;
         }
 
+        private async Task CommitAsync()
+        {
+            try
+            {
+                await Session.CommitAsync(previewChanges: false).ReportNonFatalErrorAsync().ConfigureAwait(false);
+            }
+            catch (Exception ex) when (FatalError.ReportAndCatch(ex, ErrorSeverity.Critical))
```
CyrusNajmabadi, line 263: why do we need this cache if we already hve ReportNonFatalErrorAsync above?

CyrusNajmabadi: need to at least doc to explain.

src/EditorFeatures/Core.Wpf/InlineRename/UI/Adornment/RenameFlyoutViewModel.cs
```diff
@@ -206,20 +222,64 @@ public bool IsRenameOverloadsEditable
...
+            {
+                // Show a nice error to the user via an info bar
+                var errorReportingService = Session.Workspace.Services.GetService<IErrorReportingService>();
```
CyrusNajmabadi, line 266: there was a CA(false) above, but i think IErrorService is thread affinitized.  Likely needs to be CA(true), or an explicit SwithcToMainThread here.

src/EditorFeatures/Core.Wpf/InlineRename/UI/Adornment/RenameFlyoutViewModel.cs
```diff
@@ -206,20 +222,64 @@ public bool IsRenameOverloadsEditable
...
+                if (errorReportingService is null)
+                {
+                    return;
+                }
+
+                errorReportingService.ShowGlobalErrorInfo(
```
CyrusNajmabadi, line 272: have you tested this codepath (manually?)

src/EditorFeatures/Core.Wpf/InlineRename/UI/Dashboard/RenameDashboard.xaml.cs
```diff
@@ -328,10 +329,15 @@ private void Apply_Click(object sender, RoutedEventArgs e)
...
+        {
+            _ = CommitAsync();
+        }
+
+        private async Task CommitAsync()
         {
             try
             {
-                _model.Session.Commit();
+                await _model.Session.CommitAsync(previewChanges: false).ConfigureAwait(true);
```
CyrusNajmabadi, line 340: comment why CA(true).

CyrusNajmabadi: there is error reportin gto hte user below as well.  it's unclear to me which codepaths will show errors to the user, and if we'll do that multiple times.

src/EditorFeatures/Core.Wpf/InlineRename/UI/Dashboard/RenameDashboardViewModel.cs
```diff
@@ -140,11 +143,30 @@ private void UpdateSeverity()
             }
         }
 
+        private void CommitStateChange(object sender, bool commitStart)
+            => CommitNotStart = !commitStart;
+
+        public bool CommitNotStart
+        {
+            get => _commitNotStart;
```
CyrusNajmabadi, line 151: would love threading asserts on these.

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
```diff
@@ -126,6 +126,13 @@ private set
     /// </summary>
     private CancellationTokenSource _conflictResolutionTaskCancellationSource = new();
 
+    /// <summary>
+    /// Task to track the commit operation of the session. Null if commit operation has never started.
+    /// </summary>
+    private Task<bool> _commitTask;
+
+    public bool IsCommitInProgress => !_dismissed && _commitTask is not null && _commitTask is not { Status: TaskStatus.RanToCompletion or TaskStatus.Faulted or TaskStatus.Canceled };
```
CyrusNajmabadi, line 134: !dismissed && _commitTask is not { IsCompleted: true }

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
```diff
@@ -742,14 +754,34 @@ public async Task CommitAsync(bool previewChanges)
...
+    {
+        await _threadingContext.JoinableTaskFactory.SwitchToMainThreadAsync();
+        // We are going to commit in async manner.
+        // 1. If the session is dismissed, stop and do not start another commit task
+        if (_dismissed)
+        {
+            return false;
+        }
+
+        if (_commitTask is null || !IsCommitInProgress)
```
CyrusNajmabadi, line 775: how can we have a commit in progress if we're calling into StartCommit?  doc?

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
```diff
@@ -817,6 +849,7 @@ await DismissUIAndRollbackEditsAndEndRenameSessionAsync(
 
     private async Task CommitCoreAsync(IUIThreadOperationContext operationContext, bool previewChanges)
     {
+        CommitStateChange?.Invoke(this, true);
```
CyrusNajmabadi, line 852: what is this event for?

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
```diff
@@ -842,6 +875,7 @@ private async Task CommitCoreAsync(IUIThreadOperationContext operationContext, b
                 if (newSolution == null)
                 {
                     // User clicked cancel.
+                    CommitStateChange?.Invoke(this, false);
```
CyrusNajmabadi, line 878: seems error prine to have to issue these.  can we have this be in a finally somewhere?

src/EditorFeatures/Core.Wpf/InlineRename/UI/Dashboard/RenameDashboard.xaml.cs
Cosifne: Will test & clean error reporting part.

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
Cosifne: It's possible and I will add this to the comment.

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
Cosifne: In legacy UI, when commit starts, we want to disable these buttons, because you can't edit these options once commit starts.
image
In new UI, when commit starts this event will make the UI collapse. Because it hides the background indicator tooltip
image

CyrusNajmabadi:
> In legacy UI

Serious question: why do we have hte legacy UI.  We've had the new UI for at least a year now.  I would ask that we have a PR before thsi one that just rips it out and simplifies things greatly.

src/EditorFeatures/Core.Wpf/InlineRename/UI/Dashboard/RenameDashboard.xaml.cs
Cosifne: OK I doc why `ConfigureAwait(true);` and for the exception reporting, I just leave this logic unchanged for the legacy UI

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
Cosifne: done. Doc added.

src/EditorFeatures/Core.Wpf/InlineRename/UI/Adornment/RenameFlyoutViewModel.cs
Cosifne: : ) Removed. It will be reported via `ReportNonFatalErrorAsync`

src/EditorFeatures/Core.Wpf/InlineRename/UI/Dashboard/RenameDashboard.xaml.cs
Cosifne: So I just leave this part unchanged. e.g. This is the legacy UI and we used to report the stack trace to user.

src/EditorFeatures/Core.Wpf/InlineRename/UI/Dashboard/RenameDashboardViewModel.cs
Cosifne: Added

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
Cosifne: Added.

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
Cosifne: Yeah I have put it into finally cause>

Cosifne:
> > In legacy UI
> 
> Serious question: why do we have hte legacy UI. We've had the new UI for at least a year now. I would ask that we have a PR before thsi one that just rips it out and simplifies things greatly.

@ryzngard  knows more here, but for me it's a fallback safe plan when things broken in new UI.

src/EditorFeatures/Core.Wpf/InlineRename/UI/Adornment/RenameFlyoutViewModel.cs
Cosifne: Removed : )

src/EditorFeatures/Core/InlineRename/InlineRenameSession.cs
Cosifne: Oh a unexpected case just comes to my mind. Mark this as unresovled.

PR #74796, opened 2024-08-17
Move to basic ref assemblies for net40

jaredpar:
This moves us to use `Basic.Reference.Assemblies` for our `net40` test
assets. This saves ~200MB of I/O on CI restore.

This is a purely mechanical change. The bulk of the changes is

- Rename `Net40.mscorlib` to `Net40.Resources.mscorlib`
- Add an `Import` or `using` for `Basic.Reference.Assemblies`

Functionally there should be no change here.

**Note** there is one test change in `RefStructInterfacesTests` (see second commit). Adding `Basic.Reference.Assembiles.Net40` also updated to the newer version of `Basic.Reference.Assemblies.Net90`. That effectively moved us from Preview3 to Preview6 and that has the updated version of `Activator.CreateInstance`. Needed to update a baseline to account for that.

### DIFF
--- a/scripts/vscode-build.ps1
+++ b/scripts/vscode-build.ps1
@@ -14,7 +14,7 @@ $projectFileInfo = Get-ProjectFile $fileInfo
 if ($projectFileInfo) {
   $buildTool = InitializeBuildTool
   $frameworkArg = if ($framework -ne "") { " -p:TargetFramework=$framework" } else { "" }
-  $buildArgs = "$($buildTool.Command) -v:m -m -p:RunAnalyzersDuringBuild=false -p:GenerateFullPaths=true$frameworkArg $($projectFileInfo.FullName)"
+  $buildArgs = "$($buildTool.Command) -v:m -tl:off -m -p:RunAnalyzersDuringBuild=false -p:GenerateFullPaths=true$frameworkArg $($projectFileInfo.FullName)"
 
   Exec-Command $buildTool.Path $buildArgs
   exit 0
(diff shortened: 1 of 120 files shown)


jaredpar:
@dotnet/roslyn-compiler PTAL

scripts/vscode-build.ps1
```diff
@@ -14,7 +14,7 @@ $projectFileInfo = Get-ProjectFile $fileInfo
 if ($projectFileInfo) {
   $buildTool = InitializeBuildTool
   $frameworkArg = if ($framework -ne "") { " -p:TargetFramework=$framework" } else { "" }
-  $buildArgs = "$($buildTool.Command) -v:m -m -p:RunAnalyzersDuringBuild=false -p:GenerateFullPaths=true$frameworkArg $($projectFileInfo.FullName)"
+  $buildArgs = "$($buildTool.Command) -v:m -tl:off -m -p:RunAnalyzersDuringBuild=false -p:GenerateFullPaths=true$frameworkArg $($projectFileInfo.FullName)"
```
jaredpar, line 17: The terminal logger isn't compatible with the VS Code problem matcher.

jjonescz (approved):

scripts/vscode-build.ps1
JoeRobich: If we preferred the terminal logger, we could write a matcher for it. https://code.visualstudio.com/Docs/editor/tasks#_defining-a-problem-matcher

PR #74798, opened 2024-08-18
Field-backed properties: update uses of IsAutoProperty

cston:
Relates to test plan https://github.com/dotnet/roslyn/issues/57012

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
@@ -1765,7 +1765,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             var propertyIsStatic = propertySymbol.IsStatic;
 
             return (object)sourceProperty != null &&
-                    sourceProperty.HasSynthesizedBackingField &&
+                    sourceProperty.IsAutoPropertyOrUsesFieldKeyword &&
                     TypeSymbol.Equals(sourceProperty.ContainingType, fromMember.ContainingType, TypeCompareKind.AllIgnoreOptions) &&
                     IsConstructorOrField(fromMember, isStatic: propertyIsStatic) &&
                     (propertyIsStatic || receiver.Kind == BoundKind.ThisReference);
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
@@ -1754,7 +1754,8 @@ namespace Microsoft.CodeAnalysis.CSharp
             return AccessingAutoPropertyFromConstructor(propertyAccess.ReceiverOpt, propertyAccess.PropertySymbol, fromMember);
         }
 
-        private static bool AccessingAutoPropertyFromConstructor(BoundExpression receiver, PropertySymbol propertySymbol, Symbol fromMember)
+        // PROTOTYPE: Review all callers for allowFieldKeyword.
+        private static bool AccessingAutoPropertyFromConstructor(BoundExpression receiver, PropertySymbol propertySymbol, Symbol fromMember, bool allowFieldKeyword = false)
         {
             if (!propertySymbol.IsDefinition && propertySymbol.ContainingType.Equals(propertySymbol.ContainingType.OriginalDefinition, TypeCompareKind.IgnoreNullableModifiersForReferenceTypes))
             {
@@ -1765,7 +1766,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             var propertyIsStatic = propertySymbol.IsStatic;
 
             return (object)sourceProperty != null &&
-                    sourceProperty.IsAutoPropertyOrUsesFieldKeyword &&
+                    (allowFieldKeyword ? sourceProperty.IsAutoPropertyOrUsesFieldKeyword : sourceProperty.IsAutoProperty) &&
                     TypeSymbol.Equals(sourceProperty.ContainingType, fromMember.ContainingType, TypeCompareKind.AllIgnoreOptions) &&
                     IsConstructorOrField(fromMember, isStatic: propertyIsStatic) &&
                     (propertyIsStatic || receiver.Kind == BoundKind.ThisReference);
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
@@ -1765,7 +1765,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             var propertyIsStatic = propertySymbol.IsStatic;
 
             return (object)sourceProperty != null &&
-                    sourceProperty.IsAutoProperty &&
+                    sourceProperty.HasSynthesizedBackingField &&
                     TypeSymbol.Equals(sourceProperty.ContainingType, fromMember.ContainingType, TypeCompareKind.AllIgnoreOptions) &&
                     IsConstructorOrField(fromMember, isStatic: propertyIsStatic) &&
                     (propertyIsStatic || receiver.Kind == BoundKind.ThisReference);
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -4891,7 +4891,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>Expected identifier or numeric literal</value>
   </data>
   <data name="ERR_InitializerOnNonAutoProperty" xml:space="preserve">
-    <value>Only properties with backing fields can have initializers.</value>
+    <value>Only a property with a backing field and no setter, or a property with an auto-implemented setter, can have an initializer.</value>
   </data>
   <data name="ERR_InstancePropertyInitializerInInterface" xml:space="preserve">
     <value>Instance properties in interfaces cannot have initializers.</value>
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -4891,7 +4891,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>Expected identifier or numeric literal</value>
   </data>
   <data name="ERR_InitializerOnNonAutoProperty" xml:space="preserve">
-    <value>Only auto-implemented properties can have initializers.</value>
+    <value>Only properties with backing fields can have initializers.</value>
   </data>
   <data name="ERR_InstancePropertyInitializerInInterface" xml:space="preserve">
     <value>Instance properties in interfaces cannot have initializers.</value>
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -4891,7 +4891,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>Expected identifier or numeric literal</value>
   </data>
   <data name="ERR_InitializerOnNonAutoProperty" xml:space="preserve">
-    <value>Only a property with a backing field and no setter, or a property with an auto-implemented setter, can have an initializer.</value>
+    <value>Only auto-implemented properties, or properties that use the 'field' keyword, can have initializers.</value>
   </data>
   <data name="ERR_InstancePropertyInitializerInInterface" xml:space="preserve">
     <value>Instance properties in interfaces cannot have initializers.</value>
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
@@ -302,8 +302,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             {
                 diagnostics.Add(ErrorCode.ERR_InstancePropertyInitializerInInterface, Location);
             }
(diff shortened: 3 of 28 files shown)


src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -1031,6 +1028,1438 @@ static void ReportField(FieldInfo field)
...
+                }
+                """;
+            var comp = CreateCompilation(source, targetFramework: TargetFramework.Net80);
+            comp.VerifyEmitDiagnostics(
+                // (3,25): error CS8145: Auto-implemented properties cannot return by reference
+                //     ref          object P1 { get; }
+                Diagnostic(ErrorCode.ERR_AutoPropertyCannotBeRefReturning, "P1").WithLocation(3, 25),
+                // (4,25): error CS8145: Auto-implemented properties cannot return by reference
+                //     ref          object P2 { get => ref field; }
+                Diagnostic(ErrorCode.ERR_AutoPropertyCannotBeRefReturning, "P2").WithLocation(4, 25),
```
RikkiGibson, line 1812: we spec'd that the `field` keyword can't be used on a ref-returning property right?

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -1031,6 +1028,1438 @@ static void ReportField(FieldInfo field)
...
+                // (12,32): error CS8146: Properties which return by reference must have a get accessor
+                //     static ref          object PI { set { _ = field; } }
+                Diagnostic(ErrorCode.ERR_RefPropertyMustHaveGetAccessor, "PI").WithLocation(12, 32));
+        }
+
+        [Theory]
+        [InlineData(false, false)]
+        [InlineData(false, true)]
+        [InlineData(true, false)]
+        public void SetOnly(bool useStatic, bool useInit)
```
RikkiGibson, line 1976: I found the test name a little confusing as most cases seemed to have getters

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -1031,6 +1028,1438 @@ static void ReportField(FieldInfo field)
...
+                    public override object P1 { get => field; }
+                    public override object P2 { get => field; }
+                    public override object P3 { get => field; }
+                }
+                """;
+            comp = CreateCompilation([sourceA, sourceB3], targetFramework: TargetFramework.Net80);
+            comp.VerifyEmitDiagnostics(
+                // (3,28): error CS8080: Auto-implemented properties must override all accessors of the overridden property.
+                //     public override object P1 { get => field; }
+                Diagnostic(ErrorCode.ERR_AutoPropertyMustOverrideSet, "P1").WithLocation(3, 28),
```
RikkiGibson, line 2103: I always forget about this. Yes I feel fine with saying that if you use the field keyword then you must override all the accessors. It feels like if you forgot to override one and you are using that keyword then something is going wrong, a value is being dropped on the floor somewhere.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -1031,6 +1028,1438 @@ static void ReportField(FieldInfo field)
...
+            {
+                parseOptions = parseOptions.WithPreprocessorSymbols("DEBUG");
+            }
+            var verifier = CompileAndVerify(
+                [sourceA, sourceB],
+                parseOptions: parseOptions,
+                targetFramework: TargetFramework.Net80,
+                verify: Verification.Skipped,
+                expectedOutput: IncludeExpectedOutput("""
+                    C.<P3>k__BackingField: System.Runtime.CompilerServices.CompilerGeneratedAttribute,
```
RikkiGibson, line 2414: nit: feels like it would be good to verify presence of the field symbols in NamedTypeSymbol.GetMembers() as well.

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -665,9 +662,13 @@ internal bool IsAutoProperty
         protected bool AccessorsHaveImplementation
             => (_propertyFlags & Flags.AccessorsHaveImplementation) != 0;
 
+        internal bool HasSynthesizedBackingField
```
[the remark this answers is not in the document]
cston: I've reverted the name to `IsAutoPropertyOrUsesFieldKeyword` for now. Ideally, this property should be renamed to `IsAutoProperty` and replace the current `IsAutoProperty`. But the name `HasSynthesizedBackingField` or `HasBackingField` is misleading because it's not equivalent to `BackingField is { }` since `BackingField` is set in the error case where a non-auto-property includes an initializer.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -1053,6 +1053,7 @@ static void Main()
                 }
                 """;
             var verifier = CompileAndVerify(source, expectedOutput: "(1, 2, 3, 5, 0, 9, 11)");
+            verifier.Compilation.VerifyDiagnostics();
```
RikkiGibson, line 1056: verifier itself has a VerifyDiagnostics method that will do a little less work, I think, and also include any emit diagnostics.

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -1765,7 +1765,7 @@ private static bool AccessingAutoPropertyFromConstructor(BoundExpression receive
             var propertyIsStatic = propertySymbol.IsStatic;
 
             return (object)sourceProperty != null &&
-                    sourceProperty.IsAutoProperty &&
+                    sourceProperty.IsAutoPropertyOrUsesFieldKeyword &&
```
333fred, line 1768: I'm not so sure that this is 100% correct. There's a usage in `AbstractFlowPass` that visits the backing field when this returns true; are we sure we want that behavior for all flow passes?

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -715,9 +717,9 @@ internal override void AfterAddingTypeMembersChecks(ConversionsBase conversions,
                 diagnostics.Add(ErrorCode.ERR_RefReturningPropertiesCannotBeRequired, Location);
             }
 
-            if (IsAutoProperty)
+            if (IsAutoPropertyOrUsesFieldKeyword)
             {
-                if (!IsStatic && SetMethod is { IsInitOnly: false })
+                if (!IsStatic && ((_propertyFlags & Flags.HasAutoPropertySet) != 0) && SetMethod is { IsInitOnly: false })
```
333fred, line 722: I'm not sure why we'd condition this on having `set;`, as opposed to any setter.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -302,7 +302,8 @@ protected void CheckInitializerIfNeeded(BindingDiagnosticBag diagnostics)
             {
                 diagnostics.Add(ErrorCode.ERR_InstancePropertyInitializerInInterface, Location);
             }
-            else if (!IsAutoProperty)
+            else if (!IsAutoPropertyOrUsesFieldKeyword ||
+                ((_propertyFlags & Flags.HasAutoPropertySet) == 0) && SetMethod is { })
```
333fred, line 306: Why can't we just check `IsAutoPropertyOrUsesFieldKeyword` here?

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -738,10 +740,15 @@ internal override void AfterAddingTypeMembersChecks(ConversionsBase conversions,
...
 
-                // get-only auto property should not override settable properties
-                if (this.IsOverride && SetMethod is null && !this.IsReadOnly)
+                // Auto property should override both accessors.
+                if (this.IsOverride)
                 {
-                    diagnostics.Add(ErrorCode.ERR_AutoPropertyMustOverrideSet, Location);
+                    var overriddenProperty = (PropertySymbol)this.GetLeastOverriddenMember(this.ContainingType);
+                    if ((overriddenProperty.GetMethod is { } && GetMethod is null) ||
+                        (overriddenProperty.SetMethod is { } && SetMethod is null))
```
333fred, line 748: Correct me if I'm wrong, but this appears to be a fundamentally different check than was here earlier. What was reporting the error on a missing `get` override earlier?

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
333fred: This actually doesn't appear to follow what we've specified: https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#property-initializers.

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -4891,7 +4891,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>Expected identifier or numeric literal</value>
   </data>
   <data name="ERR_InitializerOnNonAutoProperty" xml:space="preserve">
-    <value>Only auto-implemented properties can have initializers.</value>
+    <value>Only a property with a backing field and no setter, or a property with an auto-implemented setter, can have an initializer.</value>
```
333fred, line 4894:
```suggestion
    <value>Only auto-implemented properties, or properties that use the 'field' keyword, can have initializers.</value>
```

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -1031,6 +1028,1433 @@ static void ReportField(FieldInfo field)
...
+        [Fact]
+        public void Initializer_03()
+        {
+            string source = """
+                class C
+                {
+                    public static int P4 { get => field; set { } } = 4;
+                    public static int P7 { get; set { } } = 7;
+                    public static int P8 { set { field = value; } } = 8;
+                    public static int P9 { get { return field; } set { field = value; } } = 9;
```
333fred, line 1134: I expect these 4 to compile.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -1031,6 +1028,1433 @@ static void ReportField(FieldInfo field)
...
+        public void Initializer_04(bool useInit)
+        {
+            string setter = useInit ? "init" : "set";
+            string source = $$"""
+                class C
+                {
+                    public int P4 { get => field; {{setter}} { } } = 4;
+                    public int P7 { get; {{setter}} { } } = 7;
+                    public int P8 { {{setter}} { field = value; } } = 8;
+                    public int P9 { get { return field; } {{setter}} { field = value; } } = 9;
```
333fred, line 1172: I expect these 4 to compile.

333fred (commented):
Done review pass. Feels like there's a fundamental difference between the implementation and the spec here, so we need to clarify what the expected behavior is.

cston:
> Feels like there's a fundamental difference between the implementation and the spec here, so we need to clarify what the expected behavior is.

The proposal text [for initializers](https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#properties) appears to have been based on [LDM-2022-01-12](https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-01-12.md#conclusion) but there was a more recent decision in [LDM-2022-03-02](https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-03-02.md#conclusion), as you found.

I'll update the proposal and implementation.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
cston: The proposal was specifying this inconsistently. Updated in [https://github.com/dotnet/csharplang/pull/8386](https://github.com/dotnet/csharplang/pull/8386).

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
cston: Previously, for a set-only auto-property that overrides a virtual property with a get, the compiler would report "CS8051: Auto-implemented properties must have get accessors" but not "CS8080: Auto-implemented properties must override all accessors of the overridden property", see sharplab.io.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
cston: Added an open question for language design: [https://github.com/dotnet/csharplang/pull/8389](https://github.com/dotnet/csharplang/pull/8389)

333fred (commented):
No further questions in this commit, but I believe there's still an unanswered question or two from previous commits.

333fred (approved):

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
cston: Good catch, thanks. For now, I've limited this change to the one caller that expects this method to include `field` cases.

RikkiGibson (approved):

Issue #74793, opened 2024-08-17
CS8344 error even with C# 13 (ref in iterators/async)


ufcpp:
**Version Used**: 

VS 17.12.0 Preview 1.0
C# Tools   4.12.0-1.24379.11+cf82d399c36008e7936d545cde24141f8d3790fa

**Steps to Reproduce**:

```cs
static async Task M1()
{
    // CS8344.
    foreach (var r in new E1()) ;

    await Task.Yield();

    // No error with C# 13.
    E1 e1 = new();
    while (e1.MoveNext()) { var x = e1.Current; }

    await Task.Yield();

    // No error with C# 13.
    foreach (ref var x in new E2()) ;

    await Task.Yield();

    // No error with C# 13.
    E2 e2 = new();
    while (e2.MoveNext()) { ref var x = ref e2.Current; }
}

static IEnumerable<object?> M2()
{
    // CS8344.
    foreach (var r in new E1()) ;

    yield return null;

    // No error.
    E1 e1 = new();
    while (e1.MoveNext()) { var x = e1.Current; }

    yield return null;

    // No error with C# 13.
    foreach (ref var x in new E2()) ;

    yield return null;

    // No error with C# 13.
    E2 e2 = new();
    while (e2.MoveNext()) { ref var x = ref e2.Current; }
}

ref struct E1
{
    public E1 GetEnumerator() => this;
    public bool MoveNext() => false;
    public int Current => 0;
}

struct E2
{
    public E2 GetEnumerator() => this;
    public bool MoveNext() => false;
    public ref int Current => ref System.Runtime.CompilerServices.Unsafe.NullRef<int>();
}
```

**Expected Behavior**:

No error with the ref/unsafe in iterators/async feature.

**Actual Behavior**:

CS8344 error on the foreach.

I am not sure if this is a bug as it is not on [the proposal document](https://github.com/dotnet/csharplang/blob/main/proposals/ref-unsafe-in-iterators-async.md), but I would consider this to be a specification omission.

jjonescz:
Thanks for reporting this. The feature did not change anything about the CS8344 error which disallows `ref struct` enumerators inside `async`/iterator methods (although the enumerable itself can be a `ref struct`). I think there is no reason for keeping this error; we could simply remove it in C# 13 and let the ref-in-async feature disallow unsafe usages (i.e., it would still disallow the `foreach` on `ref struct` enumerator if there are any `await`s or `yield return`s in its body).

[...]

PR #74800, opened 2024-08-19
Allow ref enumerators in async/iterator methods

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/74793.
Related to test plan https://github.com/dotnet/roslyn/issues/72662.


jjonescz:

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAwaitForeachTests.cs
```diff
@@ -2083,22 +2086,34 @@ public ref struct Enumerator
...
+                Diagnostic(ErrorCode.ERR_ByRefTypeAndAwait, @"await foreach (var s in new C())
+        {
+            " + body + @"
+        }").WithArguments("C.Enumerator").WithLocation(6, 9)
             };
 
-            CreateCompilation(source, parseOptions: TestOptions.Regular12).VerifyDiagnostics(expectedDiagnostics);
-            CreateCompilation(source, parseOptions: TestOptions.Regular13).VerifyDiagnostics(expectedDiagnostics);
-            CreateCompilation(source).VerifyDiagnostics(expectedDiagnostics);
+            CreateCompilation(source, parseOptions: TestOptions.Regular13).VerifyEmitDiagnostics(expectedDiagnostics);
```
jcouv, line 2104: I didn't understand why this is allowed.
The rough lowered version of `await foreach` (without disposal) is:
```
        var enumerator = new C().GetAsyncEnumerator();
        while (await enumerator.MoveNextAsync())
        {
            enumerator.Current.ToString();
        }
```

The enumerator is a ref struct and it gets used *across* suspension points (there's an `await` in the middle).

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
```diff
@@ -17429,33 +17469,33 @@ static async System.Threading.Tasks.Task Main()
...
+                Diagnostic(ErrorCode.ERR_FeatureNotAvailableInVersion12, "foreach").WithArguments("ref and unsafe in async and iterator methods", "13.0").WithLocation(6, 15)
                 );
 
             comp2 = CreateCompilation(src1 + src2, targetFramework: s_targetFrameworkSupportingByRefLikeGenerics, parseOptions: TestOptions.Regular12);
             comp2.VerifyEmitDiagnostics(
                 // (14,24): error CS9202: Feature 'ref struct interfaces' is not available in C# 12.0. Please use language version 13.0 or greater.
                 // public ref struct S2 : IAsyncDisposable
                 Diagnostic(ErrorCode.ERR_FeatureNotAvailableInVersion12, "IAsyncDisposable").WithArguments("ref struct interfaces", "13.0").WithLocation(14, 24),
-                // (40,15): error CS8344: foreach statement cannot operate on enumerators of type 'S2' in async or iterator methods because 'S2' is a ref struct or a type parameter that allows ref struct.
+                // (40,15): error CS9202: Feature 'ref and unsafe in async and iterator methods' is not available in C# 12.0. Please use language version 13.0 or greater.
```
jcouv, line 17482: Diagnostic clarity and not directly related to this PR: is it intentional that when we report the LangVer error we no longer report CS4007?
This seems rather unfriendly (user gets a bad surprise after following our advice to upgrade LangVer).

jcouv (approved):
LGTM Thanks (iteration 3)

jcouv:
@cston @dotnet/roslyn-compiler for second review. Thanks

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
jjonescz: Yes, that's a pre-existing issue. The CS4007 is reported during emit, so we never get to report it when we report other binding-time errors. It's also a pre-existing error but perhaps we could detect just the new scenarios and report a LangVersion error for them in the emit phase instead (I'm not sure how feasible that would be).

jcouv:
@cston @dotnet/roslyn-compiler for second review. Thanks

chsienki (approved):

Issue #73909, opened 2024-06-09
Semantic snippets test flaw


DoctorKrolic:
Semantic snippets tests, which verify, that a snippet is absent in the given context, don't actually verify anything. As shown in https://github.com/dotnet/roslyn/pull/73908, a test, where `foreach` is clearly valid (a test above verifies that), doesn't fail in CI

PR #74808, opened 2024-08-19
Semantic snippets tests

DoctorKrolic:
Fixes: https://github.com/dotnet/roslyn/issues/73909

Remade tests for semantic snippets from scratch:
- Now semantic snippets are tested independently from completions. There were a few completion-related tests, moved them to `SemanticSnippetCompletionProviderTests`. If we need to test that integration point more in the future, we can add new tests to this class
- As a result of decoupling new tests moved from `EditorFeatures` to `Features` layer
- We can now also verify placeholders. They are specified in the markup as annotated spans `{|0:placeholder|}`, `{|1:placeholder2|}` etc.
- Instead of workspace xml, which is hard to read, we now pass editorconfig and reference assemblies separately (I cannot think of a scenario where semantic snippets can need more advanced xml features, such as multiple documents and so on)
- Some tests got simplified (e.g. `for`/`forr` test instead of testing `Length` and `Count` property separately now use combinatorial data), some got renamed for consistency accross all semantic snippet tests
- Reused some common test data instead of duplicating `[InlineData]` blocks
- Due to #73909 there was a bug, which sneaked in unnoticed. Was a trivial fix, so included it into this PR


DoctorKrolic:
@akhera99 PTAL

akhera99 (approved):
Looks really good, thank you for your continued contributions in this space!

DoctorKrolic:
CI failed again...

DoctorKrolic:
Someone restart failed pipelines pls

Issue #74801, opened 2024-08-19
Microsoft.VisualStudio.LanguageServices.DevKit.csproj attempted to copy files to output directory during design time build


sharwell:
**Version Used**: ee863ed

**Steps to Reproduce**:

Open Roslyn.sln in Visual Studio.

**Expected Behavior**:

Design time build does not attempt to copy files to output (bin) directory.

**Actual Behavior**:

![image](https://github.com/user-attachments/assets/d5c5842c-68f1-46a3-9f9c-97c8d0bb2d53)

The caller appears to be target `_CopyDevKitExtensionFiles` in Microsoft.CodeAnalysis.LanguageServer.UnitTests.csproj. Design time build is broken, leading to lack of IDE features during development.

tmeschter:
There are two issues here:
1. Since this is an error reported through the normal MSBuild mechanisms (as opposed to an uncaught exception or a missing target or something like that) it is reasonable to expect that it will be shown in the VS Error List.
2. Regardless of the error reporting you probably don't want `_CopyDevKitExtensionFiles` to run during a design-time build. The definition of the target should be updated to include `Condition="'$(DesignTimeBuild)' != 'true'`.

PR #74812, opened 2024-08-19
Only copy devkit assemblies for unit tests when doing a real build

dibarbet:
I noticed that the design time build was actually failing when it tried to copy these.  We only need to copy them when running tests (whcih requires a real build) so skip it for design time builds.  We do this in other places too (e.g. semantic search, build host, etc).

This doesn't seem to be related to the current missing project references failures however.

Resolves https://github.com/dotnet/roslyn/issues/74801


JoeRobich (approved):

dibarbet:
@sharwell took your approach of putting it on the target

PR #74813, opened 2024-08-19
Optimize usage of `Stack<T>` in SyntaxNavigator

steveberdy:
I was looking through the control flow of some methods in SyntaxNavigator and I came across some inefficient pops and pushes on the allocated stacks. This PR fixes a couple of those incidents.

The problem here is that we are popping the enumerator off the stack, only to push it back if we're not done with the enumerator.
To fix this, we simply peek the value off the stack, operate on it, and pop it off if we're done with the enumerator.

The fixed method is faster. Below are the results of basic benchmarks that follow the above control flows:

| Method               | Mean     | Error     | StdDev    | Median   |
|--------------------- |---------:|----------:|----------:|---------:|
| PeekAndPopIfSuccess | 4.970 ns | 0.1276 ns | 0.1193 ns | 4.939 ns |
| PopAndPushIfFail    | 8.919 ns | 1.0096 ns | 2.8968 ns | 7.272 ns |

### DIFF
--- a/src/Compilers/Core/Portable/Syntax/SyntaxNavigator.cs
+++ b/src/Compilers/Core/Portable/Syntax/SyntaxNavigator.cs
@@ -101,7 +101,9 @@ namespace Microsoft.CodeAnalysis
 
                 while (stack.Count > 0)
                 {
-                    var en = stack.Pop();
+                    // only pop enumerator off the stack if it is actually done
+                    var en = stack.Peek();
+
                     if (en.MoveNext())
                     {
                         var child = en.Current;
@@ -111,12 +113,11 @@ namespace Microsoft.CodeAnalysis
                             var token = GetFirstToken(child.AsToken(), predicate, stepInto);
                             if (token.RawKind != None)
                             {
+                                // done, pop enumerator off stack
+                                stack.Pop();
                                 return token;
                             }
                         }
-
-                        // push this enumerator back, not done yet
-                        stack.Push(en);
 
                         if (child.IsNode)
                         {
@@ -147,7 +148,8 @@ namespace Microsoft.CodeAnalysis
 
                 while (stack.Count > 0)
                 {
-                    var en = stack.Pop();
+                    // only pop enumerator off the stack if it is actually done
+                    var en = stack.Peek();
 
                     if (en.MoveNext())
                     {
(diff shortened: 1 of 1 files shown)


src/Compilers/Core/Portable/Syntax/SyntaxNavigator.cs
```diff
@@ -101,7 +101,9 @@ internal SyntaxToken GetFirstToken(SyntaxNode current, Func<SyntaxToken, bool>?
 
                 while (stack.Count > 0)
                 {
-                    var en = stack.Pop();
+                    // only pop enumerator off the stack if it is actually done
+                    var en = stack.Peek();
+
                     if (en.MoveNext())
```
cston, line 107: >if (en.MoveNext()) [](http://example.com/codeflow?start=20&length=18)

If `en.MoveNext()` fails, does the enumerator remain on the stack?

jaredpar:
> The fixed method is faster. Below are the results of basic benchmarks that follow the above control flows:

Can you post the benchmark that you wrote?

steveberdy:
@jaredpar sure. It's just a bare-minimum replication of the general control flow
```csharp
public class Benchmarks
{
    private static readonly Stack<string> s = new();

    static Benchmarks()
    {
        for (int i = 0; i < 1000; i++)
        {
            s.Push(Guid.NewGuid().ToString("N"));
        }
    }

    [Benchmark]
    public string Peek_AndPopIfSuccess()
    {
        string t = s.Peek();
        if (t.Contains("11"))
        {
            s.Pop();
            return t;
        }
        return t;
    }

    [Benchmark]
    public string Pop_AndPushIfFail()
    {
        string t = s.Pop();
        if (t.Contains("11"))
        {
            return t;
        }
        s.Push(t);
        return t;
    }
}
```

jaredpar:
@steveberdy that's a benchmark for `Stack<T>` directly, where is the benchmark about how this change improves the C# code base?

CyrusNajmabadi:
There are test failures as well.

steveberdy:
Trying to trace the issue in the tests. It's not very clear from the helix logs

CyrusNajmabadi:
possibly a stack overflow.

give that this simple test project fails:

image

you can just run those tests in the IDE:

image

and see what test it fails on, and just debug that test.

jaredpar:
Closing due to lack of recent activity.

Issue #28827, opened 2018-07-25
Go to implementation (Ctrl+F12) is broken for  interface async methods


vsfeedback:
I defined an interface with async methods in a .Net Standard 2.0 project. And two implementations in two .Net Core 2.1 projects.  
When I press Ctrl+F12 over the interface name, the  Visua Studio 'implementations' window displays the two implementations.
But when I press Ctrl+F12 over a method whose implementation has async keyword method, it doesn't find any implementation, presenting a dialog into saying &quot;This symbol has no implementations&quot;.
The expected result is that the 'implementations' window should present the two implementations of the method.

_This issue has been moved from https://developercommunity.visualstudio.com/content/problem/295266/go-to-implementation-ctrlf12-is-broken-for-interfa.html
VSTS ticketId: 649802_
_These are the original issue comments:_
(no comments)
_These are the original issue solutions:_
(no solutions)

paulovila:
Hi @jinujoseph is there any progress with this bug?

jinujoseph:
@paulovila not yet , it's in our list todo

paulovila:
Any time this year?

CyrusNajmabadi:
@paulovila Would you be interested in contributing a fix here?

paulovila:
Definitely, I'd be glad to help, my email is paulovila@live.com

CyrusNajmabadi:
Great!  I'd recommend starting with: https://github.com/dotnet/roslyn/wiki/Building-Testing-and-Debugging

When you've gotten to a buildable state, let me know.  Thanks!

ryzngard:
The problem comes in because the types and namespaces are equivalent up through the global namespace, but the `ContainingSymbol` for each global namespace is different. One is `NetModule System.Runtime.dll` and one is `NetModule netstandard.dll`  when calling [AreEquivalentWorker](http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Shared/Utilities/SymbolEquivalenceComparer.EquivalenceVisitor.cs,158).

PR #74814, opened 2024-08-19
Remove dependency on deprecated ref package

jaredpar:
This removes our dependency on the deprecated Microsoft.NET.Build.Extensions package. This is used under the hood when building a `net461` (or earlier) desktop TFM against `netstandard2.0` references. It contains a number of facade assemblies that bridge the gaps in a few places.

There is only one test that dependend on this. I migrated it to a .NET core scenario that should cover the same area.

### DIFF
--- a/src/Compilers/Test/Core/Generate.ps1
+++ b/src/Compilers/Test/Core/Generate.ps1
@@ -130,10 +130,6 @@ Add-TargetFramework "SystemThreadingTasksExtensions" '$(PkgSystem_Threading_Task
   'NetStandard20Lib#\lib\netstandard2.0\System.Threading.Tasks.Extensions.dll'
 )
 
-Add-TargetFramework "BuildExtensions" '$(PkgMicrosoft_NET_Build_Extensions)\msbuildExtensions\Microsoft\Microsoft.NET.Build.Extensions' @(
-  'NetStandardToNet461#net461\lib\netstandard.dll'
-)
-
 $targetsContent += @"
   </ItemGroup>
 </Project>
--- a/src/Workspaces/CoreTest/FindReferencesTests.cs
+++ b/src/Workspaces/CoreTest/FindReferencesTests.cs
@@ -8,6 +8,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Basic.Reference.Assemblies;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.FindSymbols;
 using Microsoft.CodeAnalysis.Test.Utilities;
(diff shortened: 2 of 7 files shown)


src/Workspaces/CoreTest/FindReferencesTests.cs
```diff
@@ -352,9 +369,9 @@ public interface I
     {
         System.Uri Get();
     }
-}", NetStandard20Ref);
+}", NetStandard20.References.All);
 
-            solution = AddProjectWithMetadataReferences(solution, "DesktopProject", LanguageNames.CSharp, @"
+            solution = AddProjectWithMetadataReferences(solution, "NetCoreProject", LanguageNames.CSharp, @"
```
jaredpar, line 374: This changes the test here from being a `net461` project referencing a `netstandard2.0` to `net8.0` referencing `netstandard2.0`. That actually matches the [original issue](https://github.com/dotnet/roslyn/issues/28827) closer by my reading. It is still under the hood though verifying that we can map symbols between the two projects when the underlying references substantially change and we have to deal with mapped symbols. 

Waiting for @sharwell, @jasonmalinowski or @CyrusNajmabadi to sign off on this or let me know if I missed a subtle issue about this test.

src/Compilers/Test/Core/Generate.ps1
```diff
@@ -130,10 +130,6 @@ Add-TargetFramework "SystemThreadingTasksExtensions" '$(PkgSystem_Threading_Task
   'NetStandard20Lib#\lib\netstandard2.0\System.Threading.Tasks.Extensions.dll'
 )
```
333fred, line 131: The grim reaper advances another door.

333fred (approved):

src/Compilers/Test/Core/Generate.ps1
jaredpar: Only three doors left.

src/Workspaces/CoreTest/FindReferencesTests.cs
jasonmalinowski: We probably went with testing netstandard and .NET Framework since they'd be most "different" in terms of types being in different assemblies. Is there a reason we can't keep this a .NET Framework ref?

jaredpar: > Is there a reason we can't keep this a .NET Framework ref?

It's difficult to keep the `net461` reference because of how the mapping between `netstandard2.0` to `net461` works under the hood. It effectively downloads / loads a deprecated NuPkg that adds a couple dozen facade assemblies to the build. This is needed because `netstandard2.0` was somewhat ret-conned into `net461`. If there were a lot of tests that did the `netstandard -> net461` mapping I'd be inclined to get that all mapped out with basic ref assemblies and use it. But this the _only_ test that does this. 

> We probably went with testing netstandard and .NET Framework since they'd be most "different" in terms of types being in different assemblies

I assure you they are quite different in `netstandard2.0` and any modern .NET core TFM. For example any core primitive type goes from being defined in `netstandard.dll` to `System.Runtime.dll`. They are, quite unfortunately, substantially different.

jasonmalinowski: > or example any core primitive type goes from being defined in netstandard.dll to System.Runtime.dll

OK that would worry me less here; as long as we have some test that is asserting that we can handle the cases where System.Object comes from different assemblies, we're probably good.

PR #74815, opened 2024-08-19
Added handler for Razor Endpoint

marcarro:
# Summary of changes
Created handler corresponding to a new LSP request endpoint in razor, to be used in Extract To Component project.

### DIFF
--- a/src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
+++ b/src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
@@ -1,0 +1,157 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Composition;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using ICSharpCode.Decompiler.CSharp.Syntax;
+using Microsoft.CodeAnalysis.CodeCleanup;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Formatting;
+using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.LanguageServer;
+using Microsoft.CodeAnalysis.LanguageServer.Handler;
+using Microsoft.CodeAnalysis.LanguageServerIndexFormat;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.CodeAnalysis.Options;
+using Microsoft.CodeAnalysis.RemoveUnnecessaryImports;
+using Microsoft.CodeAnalysis.Shared.Extensions;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.CodeAnalysis.LanguageServer.ExternalAccess.Razor;
+
+[ExportCSharpVisualBasicStatelessLspService(typeof(RazorComponentInfoHandler)), Shared]
+[Method(RazorComponentInfoName)]
+internal sealed class RazorComponentInfoHandler : ILspServiceRequestHandler<RazorComponentInfoParams, RazorComponentInfo?>
+{
+    public const string RazorComponentInfoName = "roslyn/razorComponentInfo";
+    private readonly IGlobalOptionService _globalOptions;
+
+    [ImportingConstructor]
+    [Obsolete(MefConstruction.ImportingConstructorMessage, error: true)]
+    public RazorComponentInfoHandler(IGlobalOptionService globalOptions)
+    {
+        _globalOptions = globalOptions;
+    }
+
+    public bool MutatesSolutionState => false;
+
+    public bool RequiresLSPSolution => true;
+
+    public async Task<RazorComponentInfo?> HandleRequestAsync(RazorComponentInfoParams request, RequestContext context, CancellationToken cancellationToken)
+    {
+        var project = context.Solution?.GetProject(request.Project);
+
+        if (project is null)
+        {
+            return null;
+        }
+
+        // Create a document in-memory to represent the file Razor wants to add
+        //var newFilePath = ProtocolConversions.GetDocumentFilePathFromUri(request.NewDocument.Uri);
+        //var newSource = SourceText.From(request.NewContents);
+        //var newFileLoader = new SourceTextLoader(newSource, newFilePath);
+        //var newDocumentId = DocumentId.CreateNewId(project.Id);
+        //var newSolution = project.Solution.AddDocument(
+        //    DocumentInfo.Create(
+        //        newDocumentId,
+        //        name: newFilePath,
+        //        loader: newFileLoader,
+        //        filePath: newFilePath));
+
+        //var newDocument = newSolution.GetRequiredDocument(newDocumentId);
+
+        //var formattingService = newDocument.GetLanguageService<INewDocumentFormattingService>();
+        //if (formattingService is not null)
+        //{
+        //    var hintDocument = project.Documents.FirstOrDefault();
+        //    var cleanupOptions = await newDocument.GetCodeCleanupOptionsAsync(cancellationToken).ConfigureAwait(false);
+        //    newDocument = await formattingService.FormatNewDocumentAsync(newDocument, hintDocument, cleanupOptions, cancellationToken).ConfigureAwait(false);
+        //}
+
+        var document = context.Solution.GetDocument(request.Document);
+        var semanticModel = await document.GetRequiredSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+        var syntaxTree = semanticModel.SyntaxTree;
+        var root = (CompilationUnitSyntax)syntaxTree.GetRoot(cancellationToken);
+
+
+        var (methods, fields) = GetInfoInsideRazorDocument(root, semanticModel, cancellationToken);
+
+        var result = new RazorComponentInfo
+        {
+            Methods = methods,
+            Fields = fields
+        };
+
+        return result;
+    }
+
+    private static (List<MethodInsideRazorElementInfo>, List<SymbolInsideRazorElementInfo>) GetInfoInsideRazorDocument(CompilationUnitSyntax root, SemanticModel semanticModel, CancellationToken cancellationToken)
+    {
+        var invocationExpressions = root.DescendantNodes().OfType<InvocationExpressionSyntax>().ToList();
+        var identifierNames = root.DescendantNodes().OfType<IdentifierNameSyntax>().ToList();
+        
+        List<MethodInsideRazorElementInfo> methods = [];
+        List<SymbolInsideRazorElementInfo> fields = [];
+        
+        foreach (var invocation in invocationExpressions)
+        {
+            var invocationOperation = semanticModel.GetOperation(invocation, cancellationToken) as IInvocationOperation;
+            var invocationDataFlow = semanticModel.AnalyzeDataFlow(invocation);
+            if (invocationOperation is null)
+            {
+                continue;
+            }
+
+
+            var targetMethod = invocationOperation.TargetMethod;
+            if (targetMethod is null)
+            {
+                continue;
+            }
+
+            var operationReturnType = invocationOperation.Type;
+            if (operationReturnType is null)
+            {
+                continue;
+            }
+
+            var parameterTypes = targetMethod.GetParameters().Select(parameter => parameter.Type.ToDisplayString()).ToList();
+            if (parameterTypes is null)
+            {
+                continue;
+            }
+
+            var methodInfo = new MethodInsideRazorElementInfo
+            {
+                Name = targetMethod.Name,
+                ReturnType = operationReturnType.ToNameDisplayString(),
+                ParameterTypes = parameterTypes
+            };
+
+            methods.Add(methodInfo);
+        }
+
+        foreach (var identifier in identifierNames)
+        {
+            var symbolInfo = semanticModel.GetSymbolInfo(identifier, cancellationToken);
+            if (symbolInfo.Symbol is IFieldSymbol or IPropertySymbol)
+            {
+                var field = new SymbolInsideRazorElementInfo
+                {
+                    Name = symbolInfo.Symbol.Name,
+                    Type = symbolInfo.Symbol.GetType().ToString()
+                };
+
+                fields.Add(field);
+            }
+        }
+
+        return (methods, fields);
+    }
+}
--- a/src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoParams.cs
+++ b/src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoParams.cs
@@ -1,0 +1,51 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+using Microsoft.CodeAnalysis;
+using Roslyn.LanguageServer.Protocol;
+
+internal sealed record RazorComponentInfoParams
+
+{
+    [JsonPropertyName("document")]
+    public required TextDocumentIdentifier Document { get; init; }
+
+    [JsonPropertyName("newDocument")]
+    public required TextDocumentIdentifier NewDocument { get; init; }
+
+    [JsonPropertyName("project")]
+    public required TextDocumentIdentifier Project { get; init; }
+
+    [JsonPropertyName("hostDocumentVersion")]
+    public required int HostDocumentVersion { get; init; }
+
+    [JsonPropertyName("newContents")]
+    public required string NewContents { get; init; }
+}
+
+// Not sure where to put these two records
+internal sealed record RazorComponentInfo
+{
+    public required List<MethodInsideRazorElementInfo> Methods { get; set; }
+    public required List<SymbolInsideRazorElementInfo> Fields { get; set; }
+}
+
+
+internal sealed record MethodInsideRazorElementInfo
+{
+    public required string Name { get; set; }
+
+    public required string ReturnType { get; set; }
+
+    public required List<string> ParameterTypes { get; set; }
+}
+
+internal sealed record SymbolInsideRazorElementInfo
+{
+    public required string Name { get; set; }
+    public required string Type { get; set; }
+}


src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoParams.cs
```diff
@@ -0,0 +1,51 @@
...
+    public required TextDocumentIdentifier Project { get; init; }
+
+    [JsonPropertyName("hostDocumentVersion")]
+    public required int HostDocumentVersion { get; init; }
+
+    [JsonPropertyName("newContents")]
+    public required string NewContents { get; init; }
+}
+
+// Not sure where to put these two records
```
marcarro, line 30: three *

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+    public async Task<RazorComponentInfo?> HandleRequestAsync(RazorComponentInfoParams request, RequestContext context, CancellationToken cancellationToken)
+    {
+        var project = context.Solution?.GetProject(request.Project);
+
+        if (project is null)
+        {
+            return null;
+        }
+
+        var document = context.Solution.GetDocument(request.Document);
```
ryzngard, line 55: If you're getting a document you can get the project it is in directly

```suggestion
        var document = context.Solution.GetDocument(request.Document);
        var project = document.Project;
```

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+        var semanticModel = await document.GetRequiredSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+        var syntaxTree = semanticModel.SyntaxTree;
+        var root = (CompilationUnitSyntax)syntaxTree.GetRoot(cancellationToken);
```
ryzngard, line 58: I don't think the cast is needed here

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+            Fields = fields
+        };
+
+        return result;
+    }
+
+    private static (List<MethodInsideRazorElementInfo>, List<SymbolInsideRazorElementInfo>) GetInfoInsideRazorDocument(CompilationUnitSyntax root, SemanticModel semanticModel, CancellationToken cancellationToken)
+    {
+        var invocationExpressions = root.DescendantNodes().OfType<InvocationExpressionSyntax>().ToList();
+        var identifierNames = root.DescendantNodes().OfType<IdentifierNameSyntax>().ToList();
```
ryzngard, line 74: This looks like you're getting all the invocations and identifiers in the whole file, not just ones used by razor.

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+        
+        List<MethodInsideRazorElementInfo> methods = [];
```
ryzngard, line 76: Likely these should be pooled types

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+        List<SymbolInsideRazorElementInfo> fields = [];
+        
+        foreach (var invocation in invocationExpressions)
+        {
+            var invocationOperation = semanticModel.GetOperation(invocation, cancellationToken) as IInvocationOperation;
+            var invocationDataFlow = semanticModel.AnalyzeDataFlow(invocation);
```
ryzngard, line 82: This appears to be unused

ryzngard (commented):
Some initial feedback. It's hard to follow how this information will be used. Do you have the razor side ready? It would be helpful to see that side to better comment on this code.

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+        _globalOptions = globalOptions;
+    }
+
+    public bool MutatesSolutionState => false;
+
+    public bool RequiresLSPSolution => true;
+
+    public async Task<RazorComponentInfo?> HandleRequestAsync(RazorComponentInfoParams request, RequestContext context, CancellationToken cancellationToken)
+    {
+        var project = context.Solution?.GetProject(request.Project);
```
davidwengier, line 48: I don't know what is in `request.Project`, but generally speaking the way things work in LSP would be to send the Uri of the document in the request, and then access everything via the `context` parameter. So `var document = context.GetRequiredDocument();`, and from there if you need the project, it would just be `document.Project`.

Seems like this project isn't needed though.

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
davidwengier, line 74: Rather that descending through the whole tree twice, is it better to do it once and look for both of the node types you care about?

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
davidwengier, line 82: Can you include a comment on what this is analyzing? Is it the parameters to the method, or the method itself?

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        foreach (var identifier in identifierNames)
+        {
+            var symbolInfo = semanticModel.GetSymbolInfo(identifier, cancellationToken);
+            if (symbolInfo.Symbol is IFieldSymbol or IPropertySymbol)
+            {
+                var field = new SymbolInsideRazorElementInfo
+                {
+                    Name = symbolInfo.Symbol.Name,
+                    Type = symbolInfo.Symbol.GetType().ToString()
```
davidwengier, line 124: Pretty sure this is going to end up being `Microsoft.CodeAnalysis.SourceFieldSymbol` or something. Unlikely to be what you intended.

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+                };
+
+                fields.Add(field);
```
davidwengier, line 127: Seems like there should be some filtering of duplicates here, if a field is referred to more than once

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            return null;
+        }
+
+        var document = context.Solution.GetDocument(request.Document);
+        var semanticModel = await document.GetRequiredSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+        var syntaxTree = semanticModel.SyntaxTree;
+        var root = (CompilationUnitSyntax)syntaxTree.GetRoot(cancellationToken);
+
+        var (methods, fields) = GetInfoInsideRazorDocument(root, semanticModel, cancellationToken);
```
davidwengier, line 60: It would be good to have a comment somewhere about what this endpoint actually does and is used for? I don't know why Razor wants to know about all of the method calls and fields used in a C# file.

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+            var parameterTypes = targetMethod.GetParameters().Select(parameter => parameter.Type.ToDisplayString()).ToList();
+            if (parameterTypes is null)
+            {
+                continue;
+            }
+
+            var methodInfo = new MethodInsideRazorElementInfo
+            {
+                Name = targetMethod.Name,
+                ReturnType = operationReturnType.ToNameDisplayString(),
```
davidwengier, line 109: What happens with this string? I don't think `ToNameDisplayString()`, or `ToDisplayString()` used above, guarantee that the result will be valid C# code.

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+using Microsoft.CodeAnalysis.Options;
+using Microsoft.CodeAnalysis.RemoveUnnecessaryImports;
+using Microsoft.CodeAnalysis.Shared.Extensions;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.CodeAnalysis.LanguageServer.ExternalAccess.Razor;
+
+[ExportCSharpVisualBasicStatelessLspService(typeof(RazorComponentInfoHandler)), Shared]
+[Method(RazorComponentInfoName)]
+internal sealed class RazorComponentInfoHandler : ILspServiceRequestHandler<RazorComponentInfoParams, RazorComponentInfo?>
```
davidwengier, line 30: Given that this handler seems to deal with a specific document, it should probably be a `ILspServiceDocumentRequestHandler`, to get the value out of the `context` parameter I mentioned below.

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoParams.cs
```diff
@@ -0,0 +1,51 @@
...
+    public required TextDocumentIdentifier NewDocument { get; init; }
+
+    [JsonPropertyName("project")]
+    public required TextDocumentIdentifier Project { get; init; }
+
+    [JsonPropertyName("hostDocumentVersion")]
+    public required int HostDocumentVersion { get; init; }
+
+    [JsonPropertyName("newContents")]
+    public required string NewContents { get; init; }
```
davidwengier, line 27: What goes in here?

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoParams.cs
```diff
@@ -0,0 +1,51 @@
...
+using Roslyn.LanguageServer.Protocol;
+
+internal sealed record RazorComponentInfoParams
+
+{
+    [JsonPropertyName("document")]
+    public required TextDocumentIdentifier Document { get; init; }
+
+    [JsonPropertyName("newDocument")]
+    public required TextDocumentIdentifier NewDocument { get; init; }
```
davidwengier, line 18: What goes in here?

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoParams.cs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+
+internal sealed record MethodInsideRazorElementInfo
+{
+    public required string Name { get; set; }
+
+    public required string ReturnType { get; set; }
+
+    public required List<string> ParameterTypes { get; set; }
```
davidwengier, line 44: Since these types are all for serialization, the `List` here, and the two in `RazorComponentInfo` should probably just be arrays.

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoParams.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    [JsonPropertyName("hostDocumentVersion")]
+    public required int HostDocumentVersion { get; init; }
+
+    [JsonPropertyName("newContents")]
+    public required string NewContents { get; init; }
+}
+
+// Not sure where to put these two records
+internal sealed record RazorComponentInfo
```
davidwengier, line 31: This doesn't seem to have anything to do with Razor components. Doesn't this contain C# info, not Razor info?

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoParams.cs
davidwengier: I think here is fine

davidwengier (commented):
I'm not sure what the aim of this endpoint is, so some of my comments might be irrelevant. Feel free to ignore :)

src/LanguageServer/Protocol/ExternalAccess/Razor/RazorComponentInfoHandler.cs
marcarro: Added some more information in the initial post!

davidwengier (commented):
I think there are two areas to focus on here: Firstly, generally speaking this is pretty clean code, and I can see what it's doing. What is not clear is _why_ it is doing it, and that is where code comments would be helpful.

Secondly, C# is very complicated, and it's usually a worrying sign to take shortcuts and make assumptions. eg, assuming that the first token of something will be important is unlikely to hold in all cases. Assuming that a matching on strings, names, `ToDisplayString()` etc. will work is similarly unlikely to hold in all cases. Roslyn is very powerful, and when things get reduced to comparing strings, it usually means that power is being thrown away.

src/LanguageServer/Protocol/ExternalAccess/Razor/GetSymbolicInfoHandler.cs
```diff
@@ -0,0 +1,209 @@
...
+        var isWrittenTo = writtenInsideBlock.Any(symbol => symbol == declarationInfo.Name);
+
+        // Handle special case: attribute is string type or value type.
+        // Attributes of these types are not added to the 'WrittenInside' property of a data flow analysis when written to or mutated.
+
+        // Erring on the side of caution, assume they are written to if they are involved in some type of expression.
+
+        // The 'isWrittenTo' property is not critical to functionality in current usage; it's only used in ExtractToComponent
+        // to determine if a code attribute that has been promoted to a parameter in a component should include a comment warning.
+        if (typeInfo.Type.ToDisplayString() == "string" || typeInfo.Type.IsValueType)
```
davidwengier, line 196: https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQhGAzAATZkDKALlAJYB2A5oQN6FlcXm0MsAKDDgAMZBAEoybMgF9C8ogVIUsZAMLTO3FRhRkAsgKkcC3cxRwA6Pk2biyAXjIAiAJIByALZkAhpZEXAG5tCzJmABsIYF8IkBBhGzo7MgAvJ1dPH1iAZwg/AODQi1sWMgBPDMYYAHdqZMF3bzJGCBoAQhcJELNuRVkgA==

src/LanguageServer/Protocol/ExternalAccess/Razor/GetSymbolicInfoHandler.cs
```diff
@@ -0,0 +1,312 @@
...
+        return type.ToString();
+    }
+
+    private static string FormatType(ITypeSymbol typeSymbol)
+    {
+        // Check if the symbol is a named type symbol (e.g., List<T>)
+        if (typeSymbol is INamedTypeSymbol namedTypeSymbol)
+        {
+            // Get the base name of the type, e.g., "List"
+            var typeName = namedTypeSymbol.Name;
```
marcarro, line 290: The case with integer bothers me a bit: "int" will become "Int32" (etc.) after accessing the type symbol name. This is technically correct and causes no errors, but I wonder if there's a way to just keep the type name as the user defined it.

src/LanguageServer/Protocol/ExternalAccess/Razor/GetSymbolicInfoHandler.cs
```diff
@@ -0,0 +1,312 @@
...
+        var generatedSpans = request.GeneratedDocumentRanges.Select(r => ProtocolConversions.RangeToTextSpan(r, sourceText));
+
+        // First, get the class declaration for the component (implements Microsoft.AspNetCore.Components.ComponentBase). There might be a better way to get the type.
+        var componentBaseSymbol = semanticModel.Compilation.GetTypeByMetadataName("Microsoft.AspNetCore.Components.ComponentBase");
+        if (componentBaseSymbol is null)
+        {
+            return null;
+        }
+
+        var classDeclarationNode = root.DescendantNodes().OfType<ClassDeclarationSyntax>().FirstOrDefault(classSymbol => InheritsFromComponentBase(componentBaseSymbol, classSymbol, semanticModel));
```
davidwengier, line 83: Nit: The parameter is not a `classSymbol` it is a `classNode`. "Symbol" has a specific meaning in Roslyn, in that it comes from the semantic model not the syntax tree.

src/LanguageServer/Protocol/ExternalAccess/Razor/GetSymbolicInfoHandler.cs
```diff
@@ -0,0 +1,312 @@
...
+        var fieldsInClass = fieldsBuilder.ToImmutableAndFree();
+        var propertiesInClass = propertiesBuilder.ToImmutableAndFree();
+
+        var identifiersInRange = identifiersInClass.Where(identifier => generatedSpans.Any(span => span.Contains(identifier.Span)))
+                    .Select(identifier => new IdentifierAndSymbol
+                    {
+                        Identifier = identifier,
+                        Symbol = semanticModel.GetSymbolInfo(identifier).Symbol
+                    })
+                    .Where(x => x.Symbol is not null);
```
davidwengier, line 137: Consider moving this into the above loop, and simply building `identifiersInRange` inside that loop, rather than collecting all identifiers and filtering later.

src/LanguageServer/Protocol/ExternalAccess/Razor/GetSymbolicInfoHandler.cs
```diff
@@ -0,0 +1,312 @@
...
+
+        var methodsInRange = methodsInClass.Where(method => identifiersInRange
+                                    .Any(identifier => SymbolEqualityComparer.Default.Equals(identifier.Symbol, semanticModel.GetDeclaredSymbol(method))));
+
+        var fieldsInRange = fieldsInClass.Where(field => field.Declaration.Variables
+                                            .Any(variable => identifiersInRange
+                                                .Any(identifier => SymbolEqualityComparer.Default.Equals(identifier.Symbol, semanticModel.GetDeclaredSymbol(variable)))));
```
davidwengier, line 144: This comment applies to the `methodsInRange` and `propertyInRange` calculation too, but this happens to be the worst offender: Consider how many times `GetDeclaredSymbol` is called in this calculation, for each specific value of `variable`.

src/LanguageServer/Protocol/ExternalAccess/Razor/GetSymbolicInfoHandler.cs
```diff
@@ -0,0 +1,312 @@
...
+
+            pooledMethods.Add(new MethodSymbolicInfo
+            {
+                Name = method.Identifier.Text,
+                ReturnType = GetFullTypeName(method.ReturnType!, semanticModel),
+                ParameterTypes = parameterTypes
+            });
+        }
+
+        var expressionIdentifiersInRange = identifiersInRange.Where(i => i.Identifier.Ancestors().OfType<ExpressionStatementSyntax>().Any());
```
davidwengier, line 169: Comment what this is finding? Why isn't this needed for methods? etc.

src/LanguageServer/Protocol/ExternalAccess/Razor/GetSymbolicInfoHandler.cs
```diff
@@ -0,0 +1,312 @@
...
+            return false;
+        }
+
+        var baseTypes = classDeclaration.BaseList?.Types;
+        if (baseTypes is null)
+        {
+            return false;
+        }
+
+        foreach (var baseTypeSyntax in baseTypes)
```
davidwengier, line 208: This loop seems unnecessary. I think it should be possible to get the symbol info for the class declaration itself (GetDeclaredSymbol I think) and then `InheritsFrom` could be called directly with that type symbol.

src/LanguageServer/Protocol/ExternalAccess/Razor/GetSymbolicInfoHandler.cs
```diff
@@ -0,0 +1,312 @@
...
+        if (typeSymbol is INamedTypeSymbol namedTypeSymbol)
+        {
+            // Get the base name of the type, e.g., "List"
+            var typeName = namedTypeSymbol.Name;
+
+            // If there are type arguments, format them recursively
+            if (namedTypeSymbol.TypeArguments.Length > 0)
+            {
+                var typeArguments = string.Join(", ", namedTypeSymbol.TypeArguments.Select(FormatType));
+                return $"{typeName}<{typeArguments}>"; // Returning a formatted string seems hacky so might need to be revisited.
```
davidwengier, line 296: Best not to do this manually, but rather use a call to `ToDisplayString()` that specifies an appropriate set of `SymbolDisplayFormat` values to create the desired result.

PR #74820, opened 2024-08-20
Add analyzer redirecting API

jjonescz:
- Part of https://github.com/dotnet/sdk/issues/42087
- SDK counterpart: https://github.com/dotnet/sdk/pull/42861
- Design: https://github.com/dotnet/sdk/pull/42437 ([markdown](https://github.com/jjonescz/dotnet-sdk/blob/tearing-redirect/documentation/general/analyzer-redirecting.md))


src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.cs
```diff
@@ -424,5 +427,41 @@ internal string GetRealAnalyzerLoadPath(string originalFullPath)
...
+                {
+                    try
+                    {
+                        if (redirector.RedirectPath(fullPath) is { } currentlyRedirectedPath)
+                        {
+                            if (redirectedPath == null)
+                            {
+                                redirectedPath = currentlyRedirectedPath;
+                            }
+                            else if (redirectedPath != currentlyRedirectedPath)
```
jaredpar, line 449: Given this is a path lets use `string.Equals` with an explicit comparer.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.cs
```diff
@@ -424,5 +427,41 @@ internal string GetRealAnalyzerLoadPath(string originalFullPath)
...
+                            {
+                                throw new InvalidOperationException($"Multiple redirectors disagree on the path to redirect '{fullPath}' to ('{redirectedPath}' vs '{currentlyRedirectedPath}').");
```
jaredpar, line 451: The `catch` below means this exception never manifests correct? It's just a Watson report mechanism?

src/Compilers/Core/Portable/DiagnosticAnalyzer/IAnalyzerAssemblyRedirector.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.CodeAnalysis.Diagnostics.Redirecting;
```
jaredpar, line 5: Why did we put this into a separate namespace?

src/Compilers/Core/Portable/DiagnosticAnalyzer/IAnalyzerAssemblyRedirector.cs
```diff
@@ -0,0 +1,26 @@
...
+
+/// <summary>
+/// Any MEF component implementing this interface will be used to redirect analyzer assemblies.
```
jaredpar, line 8: It feels a bit off to be mentioning MEF at the compiler layer. True this is shared code but still feels a bit off. Maybe we should move this to the remarks section? 

Also .... are there any other cases in the compiler layer where interfaces are MEF imported in workspaces? Basically an analogous example to this? I'm considering if we should go this way or possibly make workspaces define a new interface that is specifically for MEF importing then uses an adapter to bridge to this.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs
```diff
@@ -54,15 +54,32 @@ public AnalyzerFileReference(string fullPath, IAnalyzerAssemblyLoader assemblyLo
...
             _diagnosticAnalyzers = new(this, typeof(DiagnosticAnalyzerAttribute), GetDiagnosticsAnalyzerSupportedLanguages, allowNetFramework: true);
             _generators = new(this, typeof(GeneratorAttribute), GetGeneratorSupportedLanguages, allowNetFramework: false, coerceFunction: CoerceGeneratorType);
 
             // Note this analyzer full path as a dependency location, so that the analyzer loader
             // can correctly load analyzer dependencies.
             assemblyLoader.AddDependencyLocation(fullPath);
+
+            static string getAssemblyLocation(string fullPath, IAnalyzerAssemblyLoader assemblyLoader)
+            {
+                // We use reflection because AnalyzerAssemblyLoader can come from a different assembly (it's source-shared).
```
jaredpar, line 69: What is the case where `AnalyzerFileReference` doesn't have access to `IAnalyzerAssemblyLoader` members that are `internal`?

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.cs
jjonescz: Yes, we don't want to continue if there's a mismatch, but we probably shouldn't crash either.

src/Compilers/Core/Portable/DiagnosticAnalyzer/IAnalyzerAssemblyRedirector.cs
jjonescz: So we can use RestrictedInternalsVisibleTo to make this (and only this) namespace visible to the SDK counterpart.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs
jjonescz: `AnalyzerFileReference` doesn't have access to `AnalyzerAssemblyLoader.RedirectAssemblyPathExternally` because there are two `AnalyzerAssemblyLoader`s:
- one in Microsoft.CodeAnalysis, that one is fine and AnalyzerFileReference can access it,
- one in Microsoft.CodeAnalysis.Workspaces (it's a different type, only the source code is shared), this is the one actually passed here, but AnalyzerFileReference cannot access it

src/Compilers/Core/Portable/DiagnosticAnalyzer/IAnalyzerAssemblyRedirector.cs
jjonescz: > are there any other cases in the compiler layer where interfaces are MEF imported in workspaces?

IAnalyzerAssemblyResolver is exactly like that (it's what I based IAnalyzerAssemblyRedirector on). I don't know about any others.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.cs
jjonescz: `==` is equivalent to ordinal comparison; so did you mean to use ignore-case ordinal comparison, or just be more explicit?

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.cs
```diff
@@ -77,6 +78,8 @@ internal abstract partial class AnalyzerAssemblyLoader : IAnalyzerAssemblyLoader
         /// <see cref="Assembly"/> winning.</remarks>
         private readonly ImmutableArray<IAnalyzerAssemblyResolver> _externalResolvers;
 
+        private readonly ImmutableArray<IAnalyzerAssemblyRedirector> _externalRedirectors;
```
jaredpar, line 81: Have a layering question. The high level view of this change is that when Visual Studio is loading analyzers / generators it is able to redirect from paths in the .NET SDK to paths inside of Visual Studio. Basically just change the path that we are loading from. 

Why are we doing this work in the compiler vs. say the workspaces layer? That seems like the more natural fit to my eye. Is there a reason that it needs to be done here instead? 

Note: appologies cause I think we talked about this before but it's been paged out of my brain.

jjonescz: Note that this file is actually source-shared between Compiler and Workspaces.

But yes, we could implement this somewhere else in the IDE layers, outside these shared components. I've tried that at first, but it was difficult (more so as I'm unfamiliar with the IDE codebase).

The current implementation changes AnalyzerFileReference.FullPath to the redirected one. And AnalyzerFileReference.FullPath is then used by everything else, so we are basically done.

If this was done in IDE, we would keep the FullPath pointing to the original path and then we would need to change the IDE to know about redirecting, e.g.,
- the assembly loaders,
- Solution Explorer analyzer UI,
- some data serialization / checksums also needed changing IIRC.

jaredpar: I think it's reasonable for the IDE to know about redirecting as they are the ones that actually do it. It also seems like it should be reflected in the IDE that they are loading the VS copy of analyzers vs. it being hidden by the compiler.

jjonescz: Hm, I can move this into ProjectSystemProject.GetMappedAnalyzerPaths. That's what [CodeStyle analyzer redirecting](https://github.com/dotnet/roslyn/pull/75250) uses as well. I thought @chsienki said we don't want to use that API, but now I think the reason was only that Razor [needs to do more complex assembly resolution](https://github.com/dotnet/roslyn/pull/73185) which we don't need here.

I'm not sure if moving there will change anything functionally except layering, but I will try to do that to see what happens.

> I think it's reasonable for the IDE to know about redirecting as they are the ones that actually do it. It also seems like it should be reflected in the IDE that they are loading the VS copy of analyzers vs. it being hidden by the compiler.

You mean it should be reflected in the UI somehow? Technically I think that's orthogonal to where this lives - even if it's in the compiler layer, it can expose some APIs which the IDE could read to see which analyzers were redirected.

Btw, [IAnalyzerAssemblyResolver](https://github.com/dotnet/roslyn/pull/73185) also lives in the compiler layer even though it's also meant to be used just by IDEs I think.

jjonescz: Tried locally moving the functionality into GetMappedAnalyzerPaths and it seems to work fine. I can push the changes tomorrow when I polish them a bit.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs
```diff
@@ -106,6 +106,9 @@ public bool Equals(AnalyzerReference? other)
         public override int GetHashCode()
             => Hash.Combine(RuntimeHelpers.GetHashCode(_assemblyLoader), FullPath.GetHashCode());
 
+        public override string ToString()
```
333fred, line 109: Just to be sure; this was only to make your life easier when you were debugging something, right? There's no actual dependency on this format?

333fred (approved):

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.cs
JoeRobich: @jjonescz Would we be against updating the Resolver interface to support Razor's use case? I think moving CodeStyle and Razor redirection to this interface would be preferable.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs
jjonescz: Yes, this just makes debugging easier.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.cs
jjonescz: > Would we be against updating the Resolver interface to support Razor's use case?

IAnalyzerAssemblyResolver supports Razor's use case since it was added for Razor. Although it's true that Razor redirection currently lives in both the resolver interface and GetMappedAnalyzerPaths, not sure why, perhaps backcompat.

> I think moving CodeStyle and Razor redirection to this interface would be preferable.

I guess CodeStyle could be moved to use the Resolver interface, not sure, but also seems unrelated to this PR.

Note that this PR adds another interface - IAnalyzerAssemblyRedirector - which is a bit different than the Resolver one - it comes from external VSIX via MEF, that's why I made it a separate interface.

The logic of both interfaces (Redirector and Resolver) could be similar - and indeed it was before this thread where Jared asked me to move the new one (Redirector) out of compiler layer. The Resolver still lives in the compiler layer - and it needs to because it intercepts assembly loading inside AssemblyLoadContext which currently is shared between compiler and IDE so technically lives in the compiler layer. Whereas Redirector only redirects paths, so it can live in IDE layer.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs
333fred: Consider, rather than exposing a public override of `ToString`, using `DebuggerDisplayAttribute`. Not a big deal either way.

jjonescz: I would prefer exposing `ToString` as it also improved some details of error messages that might appear in VS gold bars IIRC when I was debugging this a few months back.

333fred:
@RikkiGibson @jaredpar for reviews

jaredpar (approved):

jaredpar:
@dotnet/roslyn-ide, @jasonmalinowski PTAL

jjonescz:
@jasonmalinowski for a review, thanks

jaredpar:
@jasonmalinowski PTAL

jaredpar (approved):

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
```diff
@@ -1126,9 +1127,48 @@ private OneOrMany<string> GetMappedAnalyzerPaths(string fullPath)
             return GetMappedRazorSourceGenerator(fullPath);
         }
 
+        if (TryRedirectAnalyzerAssembly(fullPath) is { } redirectedPath)
```
jasonmalinowski, line 1130: If @jaredpar's work doesn't come first, we should then refactor this method away by converting the IsSdkRazorSourceGenerator/GetMappedRazorSourceGenerator methods into a redirector.

src/Workspaces/Core/Portable/Microsoft.CodeAnalysis.Workspaces.csproj
```diff
@@ -146,6 +146,7 @@
     <RestrictedInternalsVisibleTo Include="Microsoft.VisualStudio.IntelliCode" Partner="Pythia" Key="$(IntelliCodeKey)" />
     <RestrictedInternalsVisibleTo Include="Microsoft.VisualStudio.IntelliCode.CSharp" Partner="Pythia" Key="$(IntelliCodeCSharpKey)" />
     <RestrictedInternalsVisibleTo Include="Microsoft.VisualStudio.IntelliCode.CSharp.Extraction" Partner="Pythia" Key="$(IntelliCodeCSharpKey)" />
+    <RestrictedInternalsVisibleTo Include="Microsoft.Net.Sdk.AnalyzerRedirecting" Namespace="Microsoft.CodeAnalysis.Workspaces.AnalyzerRedirecting" />
```
jasonmalinowski, line 149: Just to be 100% clear here, this Microsoft.Net.Sdk.AnalyzerRedirecting assembly is something that ships with VS, and is versioned with VS, and does not ship inside the .NET SDK?

src/Workspaces/Core/Portable/Workspace/ProjectSystem/IAnalyzerAssemblyRedirector.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// Original full path of the analyzer assembly.
+    /// </param>
+    /// <returns>
+    /// The redirected full path of the analyzer assembly
+    /// or <see langword="null"/> if this instance cannot redirect the given assembly.
+    /// </returns>
+    /// <remarks>
+    /// If two redirects return different paths for the same assembly, no redirection will be performed.
+    /// </remarks>
+    string? RedirectPath(string fullPath);
```
jasonmalinowski, line 27: Would this ever need to be async? Or is the assumption that any async would be to create the redirector in the first place?

src/Workspaces/Core/Portable/Workspace/ProjectSystem/IAnalyzerAssemblyRedirector.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.CodeAnalysis.Workspaces.AnalyzerRedirecting;
+
+/// <summary>
+/// Any MEF component implementing this interface will be used to redirect analyzer assemblies.
```
jasonmalinowski, line 8: This should have a comment explicitly stating what threading expectations exist here. Since it's not async that means no thread switching would be allowed. (And if it was async, it might still require the stronger guarantee that no thread switching is allowed.)

Per the conversation I had with @jaredpar earlier today, I'm not entirely convinced this interface could actually be implemented safely, especially if to actually implement it you need to fetch any VS service, which would require asynchrony.

jasonmalinowski (commented):
Some comments. The code itself is fine, but after the discussion I had with @jaredpar I'm not entirely convinced this can be implemented as the interface is currently structured given some of the async requirements that exist in VS and the stronger requirements that exist in Roslyn. Do we already have an implementation of this that I could look at?

jasonmalinowski (commented):
Digging further I found https://github.com/dotnet/sdk/pull/42861 as the implementation, which answered a few of my questions. So that works fine because the implementation isn't needing to grab any VS services. (And it's doing so by looking at AppDomain.CurrentDomain.BasePath which seems...highly suspect long term) So although this works fine _for that implementation_ I'm not sure any other implementations would be able to use this, and I don't think we could migrate the Razor code over to use this if we wanted to without having to revise the API shape.

@jaredpar given you had some other refactorings coming, thoughts on this?

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
jjonescz: Yeah, I think Jared's work is going to refactor all this and we just want to get this in so we can continue on the SDK side etc.

src/Workspaces/Core/Portable/Microsoft.CodeAnalysis.Workspaces.csproj
jjonescz: It is built and versioned from the .NET SDK repo, then inserted into VS and shipped with VS.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/IAnalyzerAssemblyRedirector.cs
jjonescz: Haven't thought about that. Currently, no async is needed. When we need async during creation or redirecting, we can refactor, I think.

jjonescz:
> And it's doing so by looking at AppDomain.CurrentDomain.BasePath which seems...highly suspect long term

Hm, I just need to find path to VS (because that's where the runtime analyzers are deployed and we want to redirect to them). Is there a better way?

> So although this works fine _for that implementation_ I'm not sure any other implementations would be able to use this

I don't think it's the plan to have more implementations, this is merely a hook for the SDK.

jasonmalinowski (approved):

Issue #74827, opened 2024-08-20
Flaky test: Microsoft.CodeAnalysis.MSBuild.UnitTests.NewlyCreatedProjectsFromDotNetNew.ValidateVisualBasicTemplateProjects


dibarbet:
https://helix.dot.net/BuildAnalysis/SearchTests?error=&testName=Microsoft.CodeAnalysis.MSBuild.UnitTests.NewlyCreatedProjectsFromDotNetNew.ValidateVisualBasicTemplateProjects&definition=roslyn-integration-CI&branch=main&dateValue=1&dateType=Day%28s%29

e.g.
```
[...]

PR #74831, opened 2024-08-20
Skip ValidateVisualBasicTemplateProjects failing on CI due to AVE

dibarbet:
https://github.com/dotnet/roslyn/issues/74827


JoeRobich (approved):

JoeRobich:
We definitely need to work out how to avoid this and the NuGet issue which causes us to skip the C# tests. These are the only thing that validates our project system loads all the various project types.

jaredpar:
Is anyone able to repro this? I'm surprised that we aren't seeing dumps uploaded for these crashes.

PR #74844, opened 2024-08-22
Report nullable ctor warnings for struct primary constructors

### LINKED ISSUE #74726: Nullable field warnings are missing when using struct primary constructors
[SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEDMACAzgFwCcBXAY32wGUiz8AVAUwIAoBKAWACgBvL7f7FkEBGAAzZGADwrdsAcwb4A3HkUqAvl02cuQgiXJUa5RgQBMXXpwGCc6MRIbTsshctXvtNodQP0m+GZsLtja6kA===)

```cs
public struct StructTest() // no warning
{
    public string Text { get; set; }
}

public struct StructTest2
{
    public string Text { get; set; }
    public StructTest2() { } // warning CS8618: Non-nullable property 'Text' must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring the property as nullable.
}
```

RikkiGibson:
Closes #74726

### DIFF
--- a/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
+++ b/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
@@ -1735,12 +1735,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             initializersBody ??= GetSynthesizedEmptyBody(method);
 
-            if (method is SynthesizedPrimaryConstructor primaryCtor && method.ContainingType.IsStructType())
-            {
-                body = BoundBlock.SynthesizedNoLocals(primaryCtor.GetSyntax());
-                nullableInitialState = getInitializerState(body);
-            }
-            else if (method is SourceMemberMethodSymbol sourceMethod)
+            if (method is SourceMemberMethodSymbol sourceMethod)
             {
                 CSharpSyntaxNode syntaxNode = sourceMethod.SyntaxNode;
 
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -639,8 +639,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                             Debug.Assert(thisParameter is object);
                             thisSlot = GetOrCreateSlot(thisParameter);
                         }
-                        // https://github.com/dotnet/roslyn/issues/46718: give diagnostics on return points, not constructor signature
-                        var exitLocation = method.DeclaringSyntaxReferences.IsEmpty ? null : method.TryGetFirstLocation();
+                        var exitLocation = method is SynthesizedPrimaryConstructor || method.DeclaringSyntaxReferences.IsEmpty ? null : method.TryGetFirstLocation();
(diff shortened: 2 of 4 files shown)


src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -1735,12 +1735,7 @@ private static void GetStateMachineSlotDebugInfo(
 
             initializersBody ??= GetSynthesizedEmptyBody(method);
 
-            if (method is SynthesizedPrimaryConstructor primaryCtor && method.ContainingType.IsStructType())
-            {
-                body = BoundBlock.SynthesizedNoLocals(primaryCtor.GetSyntax());
-                nullableInitialState = getInitializerState(body);
-            }
-            else if (method is SourceMemberMethodSymbol sourceMethod)
```
RikkiGibson, line 1743: This path was causing us to miss nullable analyzing the struct primary constructor.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -639,8 +639,7 @@ void enforceMemberNotNull(SyntaxNode? syntaxOpt, LocalState state)
                             Debug.Assert(thisParameter is object);
                             thisSlot = GetOrCreateSlot(thisParameter);
                         }
-                        // https://github.com/dotnet/roslyn/issues/46718: give diagnostics on return points, not constructor signature
-                        var exitLocation = method.DeclaringSyntaxReferences.IsEmpty ? null : method.TryGetFirstLocation();
+                        var exitLocation = method is SynthesizedPrimaryConstructor || method.DeclaringSyntaxReferences.IsEmpty ? null : method.TryGetFirstLocation();
```
RikkiGibson, line 642: If a type has a primary constructor, all other instance constructors will be required to chain to it (directly or indirectly), so it will be the only constructor which will get warnings for uninitialized non-nullable instance fields. That means the field/property declaration is really the best place to report diagnostics in such cases.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -1735,12 +1735,7 @@ private static void GetStateMachineSlotDebugInfo(
 
             initializersBody ??= GetSynthesizedEmptyBody(method);
 
-            if (method is SynthesizedPrimaryConstructor primaryCtor && method.ContainingType.IsStructType())
-            {
-                body = BoundBlock.SynthesizedNoLocals(primaryCtor.GetSyntax());
```
333fred, line 1740: What replicates this logic now? What body is created for the primary constructor?

RikkiGibson: The same logic which was used for class primary constructors is now also used for struct primary constructors. You can see that where the assertions were adjusted.

I reviewed the code around this area and didn't spot anything which would be problematic for structs, but, it's quite possible I missed something.

cston (approved):

333fred (approved):

PR #74861, opened 2024-08-22
Update Code Clean up Fixer to Support Async Code Cleanup on Save

cdblake1:
To enable asynchronous code cleanup, we have introduced a check to verify whether modifications can still be applied to the file after asynchronous operations have taken place. This ensures that the code cleanup process respects any changes that might have occurred concurrently, avoiding conflicts or invalid state in the solution.

Experience: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1828769/
VS-Platform: https://devdiv.visualstudio.com/DevDiv/_git/VS-Platform/pullrequest/564633

### DIFF
--- a/src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
+++ b/src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
@@ -9,12 +9,10 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.CodeActions;
 using Microsoft.CodeAnalysis.CodeCleanup;
 using Microsoft.CodeAnalysis.Editor.Shared.Extensions;
 using Microsoft.CodeAnalysis.Editor.Shared.Utilities;
 using Microsoft.CodeAnalysis.Host;
-using Microsoft.CodeAnalysis.Options;
 using Microsoft.CodeAnalysis.Progress;
 using Microsoft.CodeAnalysis.Shared.Extensions;
 using Microsoft.CodeAnalysis.Shared.Utilities;
@@ -183,6 +181,7 @@ internal abstract partial class AbstractCodeCleanUpFixer : ICodeCleanUpFixer
     {
         using (var scope = context.OperationContext.AddScope(allowCancellation: true, EditorFeaturesResources.Waiting_for_background_work_to_finish))
         {
+
             var workspaceStatusService = workspace.Services.GetService<IWorkspaceStatusService>();
             if (workspaceStatusService != null)
                 await workspaceStatusService.WaitUntilFullyLoadedAsync(context.OperationContext.UserCancellationToken).ConfigureAwait(true);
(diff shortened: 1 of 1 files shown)


src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
 
+            if (!IsSolutionValidForChanges(initialSolution, workspace.CurrentSolution, solution, out solution))
+            {
+                // Solution has changed. We can't apply this change.
+                return false;
+            }
+
             return workspace.TryApplyChanges(solution, progress);
+
+            bool IsSolutionValidForChanges(Solution initialSolution, Solution currentSolution, Solution changedSolution, out Solution updatedSolution)
```
JoeRobich, line 208: Should be static to avoid capturing `solution`.

```suggestion
            static bool IsSolutionValidForChanges(Solution initialSolution, Solution currentSolution, Solution changedSolution, out Solution updatedSolution)
```

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
+            {
+                updatedSolution = changedSolution;
+                if (initialSolution.WorkspaceVersion != currentSolution.WorkspaceVersion)
```
JoeRobich, line 211: Reverse the comparison here, set `updatedSolution` to `changedSolution`, and return early if the workspace is unchanged.

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
+            {
+                // Solution has changed. We can't apply this change.
+                return false;
+            }
+
             return workspace.TryApplyChanges(solution, progress);
+
+            bool IsSolutionValidForChanges(Solution initialSolution, Solution currentSolution, Solution changedSolution, out Solution updatedSolution)
+            {
+                updatedSolution = changedSolution;
```
JoeRobich, line 210: I think it makes sense for `updatedSolution` to start out as `currentSolution` then to eventually be modified with the changes from `changedSolution`.

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
+                            if (originalDocument.HasTextChanged(currentDocument, ignoreUnchangeableDocument: false))
+                            {
+                                // Document has changed, we can't apply this change.
+                                return false;
+                            }
+                        }
+                    }
+
+                    // If we got here, all documents in this project are the same as they were when we started.
+                    var codeCleanUpChanges = solution.GetChanges(currentSolution);
```
JoeRobich, line 276: To get the changes that are in `changedSolution` we will be starting from `initialSolution`.

```suggestion
                    var codeCleanUpChanges = changedSolution.GetChanges(initialSolution);
```

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
+                            var changedDocument = projectChanges.NewProject.GetRequiredDocument(documentId);
+                            var currentDocument = solution.GetRequiredDocument(documentId);
+
+                            if (currentDocument is null)
+                            {
+                                // Document is gone, we can't apply this change.
+                                return false;
+                            }
+
+                            solution = currentSolution.WithTextDocumentText(documentId, changedDocument.GetTextSynchronously(cancellationToken));
```
JoeRobich, line 301: We should use `updatedSolution` here. We will keep modifying it so that it accumulates all the changes from `changedSolution`.

```suggestion
                            updatedSolution = updatedSolution.WithTextDocumentText(documentId, changedDocument.GetTextSynchronously(cancellationToken));
```

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
+                        if (codeCleanupDocChanges.Length == 0)
+                        {
+                            //nothing has changed in this project. We can't apply this change.
+                            return false;
+                        }
+
+                        foreach (var documentId in codeCleanupDocChanges)
+                        {
+                            var changedDocument = projectChanges.NewProject.GetRequiredDocument(documentId);
+                            var currentDocument = solution.GetRequiredDocument(documentId);
```
JoeRobich, line 293: Since we are updating `updatedSolution`, we would need to locate the document from it in order to make changes.

```suggestion
                            var currentDocument = updatedSolution.GetRequiredDocument(documentId);
```

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
+                                // Document is gone, we can't apply this change.
+                                return false;
+                            }
+
+                            solution = currentSolution.WithTextDocumentText(documentId, changedDocument.GetTextSynchronously(cancellationToken));
+                        }
+                    }
+                }
+
+                updatedSolution = solution;
```
JoeRobich, line 306: No need to set `updatedSolution` here. It would have already been updated in the code above.
```suggestion
```

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
+
+                    foreach (var projectChanges in codeCleanUpChanges.GetProjectChanges())
+                    {
+                        var codeCleanupDocChanges = projectChanges.GetChangedDocuments()
+                            .Concat(projectChanges.GetChangedAdditionalDocuments())
+                            .Concat(projectChanges.GetChangedAnalyzerConfigDocuments()).ToImmutableArray();
+
+                        if (codeCleanupDocChanges.Length == 0)
+                        {
+                            //nothing has changed in this project. We can't apply this change.
```
JoeRobich, line 286: Is this possible? I thought we checked above that the files were the same between `initialSolution` and `currentSolution`.

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +192,120 @@ private async Task<bool> FixAsync(
...
             return workspace.TryApplyChanges(solution, progress);
+
+            bool IsSolutionValidForChanges(Solution initialSolution, Solution currentSolution, Solution changedSolution, out Solution updatedSolution)
+            {
+                updatedSolution = changedSolution;
+                if (initialSolution.WorkspaceVersion != currentSolution.WorkspaceVersion)
+                {
+                    // something has changed. Lets make sure we can still apply changes.
+
+                    var solutionChanges = initialSolution.GetChanges(currentSolution);
```
JoeRobich, line 215: To get changes from `currentSolution` we will be starting from `initialSolution`. See https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.solution.getchanges?view=roslyn-dotnet-4.9.0

```suggestion
                    var solutionChanges = currentSolution.GetChanges(initialSolution);
```

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -183,6 +181,7 @@ private async Task<bool> FixAsync(
     {
         using (var scope = context.OperationContext.AddScope(allowCancellation: true, EditorFeaturesResources.Waiting_for_background_work_to_finish))
         {
+
```
JoeRobich, line 184: nit: remove empty line at start of block
```suggestion
```

JoeRobich (commented):
I think trying to modify `solution` from the local function was confusing things.

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +191,120 @@ private async Task<bool> FixAsync(
...
 
+            if (!IsSolutionValidForChanges(initialSolution, workspace.CurrentSolution, solution, out solution))
+            {
+                // Solution has changed. We can't apply this change.
+                return false;
+            }
+
             return workspace.TryApplyChanges(solution, progress);
+
+            bool IsSolutionValidForChanges(Solution initialSolution, Solution currentSolution, Solution changedSolution, out Solution updatedSolution)
```
CyrusNajmabadi, line 207: i feel like we have this logic elsewhere. (maybe in lightbulbs)?  can we share it?  i can help you look for teh existing helper function if needed.

src/VisualStudio/Core/Def/CodeCleanup/AbstractCodeCleanUpFixer.cs
```diff
@@ -193,11 +191,120 @@ private async Task<bool> FixAsync(
             var cancellationToken = context.OperationContext.UserCancellationToken;
             var progress = scope.GetCodeAnalysisProgress();
 
+            var initialSolution = workspace.CurrentSolution;
             var solution = await applyFixAsync(progress, cancellationToken).ConfigureAwait(true);
 
             await _threadingContext.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
 
+            if (!IsSolutionValidForChanges(initialSolution, workspace.CurrentSolution, solution, out solution))
```
CyrusNajmabadi, line 199: i don't think this logic actually works.  THe reason for thsi is that TryApplyChanges already checks for this case:

```
            // If the workspace has already accepted an update, then fail
            if (newSolution.WorkspaceVersion != oldSolution.WorkspaceVersion)
            {
```

image

So if an intermediary edit happened, TryApplyChanges is a no-op.

PR #74865, opened 2024-08-22
Use InterceptorsNamespaces feature name instead of InterceptorsPreviewNamespaces

### LINKED ISSUE #74511: Interceptors still require `InterceptorsPreviewNamespaces` property
The compiler will emit a warning for source generators that use interceptors about the need for an `InterceptorsPreviewNamespaces` property in the target project.

```
/Users/captainsafia/git/aspnetcore/openapi-fixes/artifacts/obj/Microsoft.AspNetCore.OpenApi/Debug/net9.0/Microsoft.AspNetCore.Http.RequestDelegateGenerator/Microsoft.AspNetCore.Http.RequestDelegateGenerator.RequestDelegateGenerator/GeneratedRouteBuilderExtensions.g.cs(61,10): error CS9137: The 'interceptors' experimental feature is not enabled in this namespace. Add '<InterceptorsPreviewNamespaces>$(InterceptorsPreviewNamespaces);Microsoft.AspNetCore.Http.Generated</InterceptorsPreviewNamespaces>' to your project.
```

cc: @jaredpar @RikkiGibson

RikkiGibson:
Closes #74511 

Here I am making an assumption that the build tasks and the associated compiler will be in sync. If that assumption is wrong, then this handling will probably have to be pushed into the command line layer (i.e. concat'ing `/features:InterceptorsNamespaces=...` and `/features:InterceptorsPreviewNamespaces=...`.

### DIFF
--- a/docs/features/interceptors.md
+++ b/docs/features/interceptors.md
@@ -218,12 +218,14 @@ There is an experimental public API `GetInterceptorMethod(this SemanticModel, In
 
 ### User opt-in
 
-To use interceptors, the user project must specify the property `<InterceptorsPreviewNamespaces>`. This is a list of namespaces which are allowed to contain interceptors.
+To use interceptors, the user project must specify the property `<InterceptorsNamespaces>`. This is a list of namespaces which are allowed to contain interceptors.
 ```xml
-<InterceptorsPreviewNamespaces>$(InterceptorsPreviewNamespaces);Microsoft.AspNetCore.Http.Generated;MyLibrary.Generated</InterceptorsPreviewNamespaces>
+<InterceptorsNamespaces>$(InterceptorsNamespaces);Microsoft.AspNetCore.Http.Generated;MyLibrary.Generated</InterceptorsNamespaces>
 ```
 
-It's expected that each entry in the `InterceptorsPreviewNamespaces` list roughly corresponds to one source generator. Well-behaved components are expected to not insert interceptors into namespaces they do not own.
+It's expected that each entry in the `InterceptorsNamespaces` list roughly corresponds to one source generator. Well-behaved components are expected to not insert interceptors into namespaces they do not own.
+
+For compatibility, the property `<InterceptorsPreviewNamespaces>` can be used as an alias for `<InterceptorsNamespaces>`. If both properties have non-empty values, they will be concatenated together in the order `$(InterceptorsNamespaces);$(InterceptorsPreviewNamespaces)` when passed to the compiler.
 
 ### Implementation strategy
 
--- a/src/Compilers/CSharp/Portable/CSharpParseOptions.cs
+++ b/src/Compilers/CSharp/Portable/CSharpParseOptions.cs
@@ -22,7 +22,7 @@ namespace Microsoft.CodeAnalysis.CSharp
         public static CSharpParseOptions Default { get; } = new CSharpParseOptions();
 
         private ImmutableDictionary<string, string> _features;
-        private ImmutableArray<ImmutableArray<string>> _interceptorsPreviewNamespaces;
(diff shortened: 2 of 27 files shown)


docs/features/interceptors.md
```diff
@@ -218,12 +218,14 @@ There is an experimental public API `GetInterceptorMethod(this SemanticModel, In
 
 ### User opt-in
 
-To use interceptors, the user project must specify the property `<InterceptorsPreviewNamespaces>`. This is a list of namespaces which are allowed to contain interceptors.
+To use interceptors, the user project must specify the property `<InterceptorsNamespaces>`. This is a list of namespaces which are allowed to contain interceptors.
```
cston, line 221: >InterceptorsNamespaces [](http://example.com/codeflow?start=66&length=22)

Should the name be "InterceptorNamespaces" rather than "Interceptors..."?

docs/features/interceptors.md
```diff
@@ -218,12 +218,14 @@ There is an experimental public API `GetInterceptorMethod(this SemanticModel, In
...
 ```xml
-<InterceptorsPreviewNamespaces>$(InterceptorsPreviewNamespaces);Microsoft.AspNetCore.Http.Generated;MyLibrary.Generated</InterceptorsPreviewNamespaces>
+<InterceptorsNamespaces>$(InterceptorsNamespaces);Microsoft.AspNetCore.Http.Generated;MyLibrary.Generated</InterceptorsNamespaces>
 ```
 
-It's expected that each entry in the `InterceptorsPreviewNamespaces` list roughly corresponds to one source generator. Well-behaved components are expected to not insert interceptors into namespaces they do not own.
+It's expected that each entry in the `InterceptorsNamespaces` list roughly corresponds to one source generator. Well-behaved components are expected to not insert interceptors into namespaces they do not own.
+
+For compatibility, the property `<InterceptorsPreviewNamespaces>` can be used as an alias for `<InterceptorsNamespaces>`. If both properties have non-empty values, they will be concatenated together in the order `$(InterceptorsNamespaces);$(InterceptorsPreviewNamespaces)` when passed to the compiler.
```
cston, line 228: >concatenated together [](http://example.com/codeflow?start=177&length=21)

Will we avoid duplicate interceptors if the two properties have the same namespaces?

cston (approved):

docs/features/interceptors.md
RikkiGibson: In the issue there seemed to be agreement on the name `InterceptorsNamespaces`, i.e. the namespaces which contain interceptors. So I prefer to keep this name.

docs/features/interceptors.md
RikkiGibson: I think that test `GetInterceptorMethod_10` shows that we don't behave badly if the same namespace is included multiple times.

RikkiGibson:
@jaredpar for review. Mainly I'm unsure about my assumption that compiler and build tasks will be in sync here. Is it possible for example that an older version of the build task would end up getting used to pass arguments to a newer compiler?

jaredpar:
> Is it possible for example that an older version of the build task would end up getting used to pass arguments to a newer compiler?

Generally speaking no. The compiler and build task are always in sync. 

The one exception is if the build parameters `CscToolExe` and `CscToolPath` are set. That allows customers to direct to a specific binary on disk that isn't necessarily tied to our deployment. That is very much a _buyer beware_ territory and we don't concern ourselves with sync issues like this.

RikkiGibson:

docs/features/interceptors.md
jjonescz, line 221: Nit: it feels a bit inconsistent that the "Interceptors" part is plural, I would expect this to be named like `InterceptorNamespaces`.

PR #74866, opened 2024-08-22
Fix handling default copilot option values

genlu:
Fix a regression introduced in https://github.com/dotnet/roslyn/pull/74781
which causes all Roslyn's copilot options to be enabled by default


akhera99 (approved):

JoeRobich:
We should ask them to write the defaults for their settings as $"{CopilotOptionNamePrefix}.{option.Name}.Default" or something similar. Then we could defer to whatever they ship.

JoeRobich:
Tested locally with IntPreview. 

**Without this change:**

I am shown `Refine using Copilot` from the code action preview popup. 
image
Despite it not being enabled in my options.
image


**With this change installed:**
image

I am not shown 'Refine using Copilot' in the code action preview popup.
image

Issue #74734, opened 2024-08-13
Params collection in collection expression's .Add method causes crush


Spolutrean:
**Version Used**: 
VS 17.11.0 Preview 7.0

[Sharplab repro](https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoBAGABAlAOgGEB7AG3IFMBjAFylJgGcBuTHfFAFnazwKEASgFcYDMFRKkwAByjUATgGUqigG5QaVNpg4BmfElwAhCMyq4QXfQB5SwAFa06APkwBvTLnyGAkgCiMCKSihB0pIr2Ti6uuADiVHRBIWrhkQAUAJTeuF4YPj50ABaKpADuuDBUlQBypHR+ctSS4lQAJgEAHtqyDEzZfD4AvnoFuIHBoemKNoSJyVNpEYrZufmF+ADsCUkp0yuDuaMYJwZGuACyAJ5klC6MMChWpuZUnrk+CIYI3LgAgu12hlZBAwmBmFdbhRqPRHsYaEgsnkTmcsD9jDc7rD+jBjNYzBYPuNvvg/oDgaDwZCsTCHkxnojkR5cKixqTkLgAAoiSgAFWKVAAMlR1Go8p8odj6U9cAARUjERQiZjFbK4AC8cQA2igALp8YZAA===)

**Expected Behavior**:
Not crushing

**Actual Behavior**:
Crushing

PR #74899, opened 2024-08-26
Handle infinite cycles through params collection initializers

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/74734.


src/Compilers/CSharp/Test/Emit2/Semantics/ParamsCollectionTests.cs
```diff
@@ -14319,6 +14319,154 @@ static void Test(params CollectionBuilderAttribute a)
                 );
         }
 
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/74734")]
+        public void Cycle_16()
+        {
+            if (CompilationExtensions.EnableVerifyIOperation &&
+                ExecutionConditionUtil.Configuration == ExecutionConfiguration.Debug)
+            {
+                // A debug assert fails but an invalid operation is correctly created otherwise.
```
333fred, line 14328: Do we have a work item tracking this?

jjonescz: No, I will create one, thanks.

src/Compilers/CSharp/Test/Emit2/Semantics/ParamsCollectionTests.cs
```diff
@@ -14319,6 +14319,154 @@ static void Test(params CollectionBuilderAttribute a)
...
+                return;
+            }
+
+            var source = """
+                using System.Collections;
+                using System.Collections.Generic;
+
+                public class Base : IEnumerable<object>
```
cston, line 14336: >Base [](http://example.com/codeflow?start=29&length=4)

Minor: All of the tests use a base and derived type. Is that necessary to repro the issue? If not, consider merging the base and derived in at least one test where there is currently only one derived type.

cston: Let's also test a generic collection type. Perhaps this test could be modified to:
```csharp
public class MyCollection1<T> : IEnumerable<T>
{
    public IEnumerator<T> GetEnumerator() => throw null;
    IEnumerator IEnumerable.GetEnumerator() => throw null;
    public void Add(params MyCollection1<T> c) {}
}

public class C
{
    MyCollection1<object> M() => [1];
}
```

cston: Consider testing:
```csharp
public class MyCollection1<T> : IEnumerable<T>
{
    public IEnumerator<T> GetEnumerator() => throw null;
    IEnumerator IEnumerable.GetEnumerator() => throw null;
    public void Add(params MyCollection1<object> c) {}
}

public class C
{
    MyCollection1<int> M() => [1];
}
```

cston: Should we test overloads, such as the following, which looks like it should succeed:
```csharp
public class MyCollection1<T> : IEnumerable<T>
{
    public IEnumerator<T> GetEnumerator() => throw null;
    IEnumerator IEnumerable.GetEnumerator() => throw null;
    public void Add(T x) {}
    public void Add(params MyCollection1<T> y) {}
}

public class C
{
    MyCollection1<object> M() => [1];
}
```

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -751,6 +757,24 @@ BoundNode bindSpreadElement(BoundCollectionExpressionSpreadElement element, Type
             }
         }
 
+        private bool HasCollectionInitializerTypeInProgress(SyntaxNode syntax, TypeSymbol targetType)
+        {
+            Binder? current = this;
+            while (current?.Flags.Includes(BinderFlags.CollectionInitializerAddMethod) == true)
+            {
+                if (current is CollectionInitializerAddMethodBinder binder &&
+                    binder.Syntax.FullSpan.Equals(syntax.FullSpan) &&
```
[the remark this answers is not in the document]
jjonescz: I'm not sure if the syntax nodes are guaranteed to be the same reference.

src/Compilers/CSharp/Test/Emit2/Semantics/ParamsCollectionTests.cs
jjonescz: On a second thought, maybe it's better to just remove the assert.

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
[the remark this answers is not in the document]
jjonescz: But it seems it should, and I see other places in roslyn using that, so I will switch to reference equality as well, thanks.

Spolutrean:
@jjonescz Is it a good idea to also test different call-sites? It is leading to VS crash as well.
```
public class MyCollection1<T> : IEnumerable<T>
{
    public IEnumerator<T> GetEnumerator() => throw null;
    IEnumerator IEnumerable.GetEnumerator() => throw null;
    public void Add(params MyCollection1<object> c) {}
}

public class C
{
    public void M() 
    {
        MyCollection1<int> c = new() { 1 };
        c.Add(2);
    }
}
```

src/Compilers/Core/Portable/Operations/ControlFlowGraphBuilder.cs
```diff
@@ -6585,7 +6585,6 @@ public override IOperation VisitInstanceReference(IInstanceReferenceOperation op
                     }
                     else
                     {
```
333fred, line 6587: I'd still like to understand how we're getting here, as well as having explicit ioperation tests for it.

jjonescz: I think we are getting here because of the cycle - the `_currentImplicitInstance` is unset because the receiver is the same collection expression as the one being visited. I will add tests.

333fred: I'm not comfortable with getting rid of this assert entirely. Let's assert the specific case we're talking about, so we still catch any other violations.

333fred (approved):

cston (approved):

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -648,7 +648,13 @@ private BoundCollectionExpression ConvertCollectionExpression(
                 Debug.Assert((collectionCreation is BoundNewT && !isExpanded && constructor is null) ||
                              (collectionCreation is BoundObjectCreationExpression creation && creation.Expanded == isExpanded && creation.Constructor == constructor));
 
-                var collectionInitializerAddMethodBinder = this.WithAdditionalFlags(BinderFlags.CollectionInitializerAddMethod);
+                if (!elements.IsDefaultOrEmpty && HasCollectionInitializerTypeInProgress(syntax, targetType))
+                {
+                    diagnostics.Add(ErrorCode.ERR_CollectionInitializerInfiniteChainOfAddCalls, syntax, targetType);
```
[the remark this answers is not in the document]
jjonescz: > Also, we should confirm that there is no way to get into a cycle when constructors and `Add` methods are involved at the same time, otherwise we might be missing a guard.

I didn't think of a way that would be possible without hitting the existing guards. I can add some more tests though.

> It feels like we should unify and combine both checks, i.e. do them in the same place and use similar approaches for both.

The constructor check is separate and happens before binding the conversion - I considered implementing the Add check at the same place but that would require duplicating the complicated Add binding logic. The other way around (moving the existing constructor check) could be probably done - but it seems orthogonal to this bug fix, perhaps we can just file an issue. Also I think we would need another intermediate binder so it wouldn't share code with the Add check, i.e., maybe there are no advantages.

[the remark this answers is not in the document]
AlekseyTs: > perhaps we can just file an issue

I think this should  be sufficient for now.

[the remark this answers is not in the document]
jjonescz: https://github.com/dotnet/roslyn/issues/75024

PR #74903, opened 2024-08-26
Support for per-file content exclusion in Rename Suggestions

olegtk:
To support intelligent Copilot content exclusion we need to associate each piece of context we provide to Rename suggestions with file path this context is coming from.

Switch to the new Editor API (RenameContext has file path property):

`Task<IReadOnlyList<string>> GetSuggestionsAsync(ImmutableArray<RenameContext> context, CancellationToken cancellationToken);`

from (this one has no file path): 

`Task<IReadOnlyList<string>> GetSuggestionsAsync(ImmutableDictionary<string, string[]> context, CancellationToken cancellationToken);`

image


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "9.0.100-preview.5.24307.3",
+    "version": "8.0.300",
```
Cosifne, line 3: I don't think we can downgrade this number. (even there might be bugs in the current sdk)
Roslyn (main) change would flows to sdk (main) so it needs to be on 9.x

src/EditorFeatures/CSharp/InlineRename/CSharpEditorInlineRenameService.cs
```diff
@@ -149,10 +153,13 @@ void AddSpanOfInterest(SourceText documentText, TextSpan fallbackSpan, TextSpan?
             var length = endPosition - startPosition;
 
             surroundingSpanOfInterest = new TextSpan(startPosition, length);
-
             if (seen.Add(surroundingSpanOfInterest.Value))
             {
-                resultBuilder.Add(documentText.GetSubText(surroundingSpanOfInterest.Value).ToString());
+                var filePath = documentText.GetDocumentWithFrozenPartialSemantics(cancellationToken)?.FilePath;
```
Cosifne, line 158: Will it work if you direclty call https://sourceroslyn.io/#Microsoft.CodeAnalysis.Workspaces/Workspace/TextExtensions.cs,8844857d387b5686,references?

AmadeusW:
LGTM. Thank you, Oleg!

src/EditorFeatures/Core.Wpf/Lightup/LightupHelpers.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using Microsoft.CodeAnalysis.Contracts.EditAndContinue;
```
Cosifne, line 10: Is this using needed?

Cosifne:
Could you try to use the latest main and build this branch, and verify rename experience are working? (With Smart Rename on/off)
Asking this because we used to break the rename function because Lightup helper in public preview. So it would be worthy to make validate before merge

Cosifne (approved):
Note: Should only merge once we validate the change works in VS main

olegtk:
Validated on latest main, with and without smart rename, automatic and non-automatic, in various C# files and various symbol types, looks good.

PR #74910, opened 2024-08-27
Add required `--locale` option to BuildHost

### LINKED ISSUE #74902: Build host should localize message output
See https://github.com/dotnet/roslyn/discussions/74896

See https://github.com/dotnet/roslyn/blob/main/src/Workspaces/Core/MSBuild/MSBuild/BuildHostProcessManager.cs#L158

When starting the build host we do not pass the user value for `DOTNET_CLI_UI_LANGUAGE` to the build host, so the output is not localized.  Likely we should just be passing this over to the build host.

There is an open question of what should happen if VSCode configures one language, and `DOTNET_CLI_UI_LANGUAGE` specifies a different one.

JoeRobich:
When invoking the BuildHost pass along the `CurrentUICulture.Name`. This will allow reported exceptions to match the language of the process which is using the MSBuildWorkspace.

Resolves #74902

### DIFF
--- a/src/Workspaces/Core/MSBuild.BuildHost/Program.cs
+++ b/src/Workspaces/Core/MSBuild.BuildHost/Program.cs
@@ -41,7 +41,7 @@ internal static class Program
         catch (CultureNotFoundException)
         {
             // We couldn't find the culture, log a warning and fallback to the OS configured value.
-            logger.LogInformation($"Culture {locale} was not found, falling back to OS culture");
+            logger.LogWarning($"Culture {locale} was not found, falling back to OS culture");
         }
 
         logger.LogInformation($"BuildHost Runtime Version: {System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription}");
--- a/src/Workspaces/Core/MSBuild.BuildHost/Program.cs
+++ b/src/Workspaces/Core/MSBuild.BuildHost/Program.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Immutable;
 using System.CommandLine;
+using System.Globalization;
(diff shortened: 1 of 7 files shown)


src/Workspaces/Core/MSBuild.BuildHost/Program.cs
```diff
@@ -31,6 +34,16 @@ internal static async Task Main(string[] args)
...
         var logger = new BuildHostLogger(Console.Error);
 
+        try
+        {
+            CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo(locale);
+        }
+        catch (CultureNotFoundException)
+        {
+            // We couldn't find the culture, log a warning and fallback to the OS configured value.
+            logger.LogInformation($"Culture {locale} was not found, falling back to OS culture");
```
jasonmalinowski, line 44: Not sure if it matters really, should this be more a warning?

src/Workspaces/Core/MSBuild.BuildHost/Program.cs
```diff
@@ -31,6 +34,16 @@ internal static async Task Main(string[] args)
 
         var logger = new BuildHostLogger(Console.Error);
 
+        try
+        {
+            CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo(locale);
+        }
+        catch (CultureNotFoundException)
```
jasonmalinowski, line 41: I absolutely expect this check is needed, since the locale here might cross between .NET Core and .NET Framework (and in both directions), so best to be careful here.

jasonmalinowski (approved):
Code looks good; can we add a test somehow?

PR #74922, opened 2024-08-27
Field-backed properties: readonly backing field

cston:
From [proposal](https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#readonly-field):
> The synthesized backing field is *read-only* when the containing type is a `struct` and the property or containing type is declared `readonly`.

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
@@ -193,6 +193,9 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                 }
                 else
                 {
+                    // PROTOTYPE: We could treat the field as readonly if all manually implemented get and set
+                    // accessors are declared readonly. Although, to do so, we might need to bind the accessor
+                    // declarations before creating the backing field. See FieldKeywordTests.ReadOnly_05().
                     isReadOnly = false;
                 }
 
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
@@ -172,15 +172,31 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             {
                 Debug.Assert(!IsIndexer);
                 string fieldName = GeneratedNames.MakeBackingFieldName(_name);
-                BackingField = new SynthesizedBackingFieldSymbol(this,
-                                                                      fieldName,
-                                                                      // Synthesized backing field for 'field' should not be marked 'initonly'
-                                                                      // since the field might be modified in the get accessor.
-                                                                      // PROTOTYPE: Should the backing field be 'initonly' when the containing
-                                                                      // type, property, or accessor is declared 'readonly'?
-                                                                      isReadOnly: !usesFieldKeyword && ((hasGetAccessor && !hasSetAccessor) || isInitOnly),
-                                                                      this.IsStatic,
-                                                                      hasInitializer);
+
+                // The backing field is readonly if any of the following holds:
+                // - The containing type is declared readonly and the property is an instance property.
+                bool isReadOnly;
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -172,15 +172,31 @@ protected SourcePropertySymbolBase(
...
+                    isReadOnly = true;
+                }
+                // - The property is declared readonly.
+                else if (HasReadOnlyModifier)
+                {
+                    isReadOnly = true;
+                }
+                // - The property has no set accessor (but may have an init accessor) and
+                // the get accessor, if any, is automatically implemented.
+                else if ((!hasSetAccessor || isInitOnly) && (!hasGetAccessor || hasAutoPropertyGet))
```
cston, line 190: Should the field also be readonly if all manually implemented `get` and `set` accessors are declared `readonly`?

RikkiGibson: It's not valid to declare both accessors as readonly, you'd have to declare the containing property as readonly instead, which would make the field readonly per these rules.

cston: Declaring both as `readonly` is an error, but we may want to treat the backing field as `readonly` in the following case:
```csharp
struct S
{
    object P { get; readonly set { } }
}
```

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -2386,106 +2402,241 @@ public void ReadOnly_02(bool useReadOnlyType, bool useReadOnlyOnGet)
...
+                <P5>k__BackingField: True
+                <P7>k__BackingField: True
+                <Q1>k__BackingField: {{useReadOnlyType || useReadOnlyMember}}
+                <Q2>k__BackingField: {{useReadOnlyType || useReadOnlyMember}}
+                <Q3>k__BackingField: True
+                <Q4>k__BackingField: {{useReadOnlyType || useReadOnlyMember}}
+                <Q5>k__BackingField: True
+                <Q6>k__BackingField: {{useReadOnlyType}}
+                <Q7>k__BackingField: True
+                <Q8>k__BackingField: {{useReadOnlyType}}
```
RikkiGibson, line 2470: It feels like in the case `useReadOnlyType: false, useReadOnlyMember: true`, all accessors are readonly, yet the backing field is not readonly. That seems a little strange. I guess it doesn't really matter, though.

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
333fred: I think a more compelling example is:
```cs
struct S
{
    object P { readonly get => field; init; }
}
```

RikkiGibson: FWIW, I can't think of a case where things are made worse by the backing field not being readonly, in cases involving the accessors being marked readonly but not the containing type or property.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
cston: Yes, this is an example of the case I was expecting in [https://github.com/dotnet/roslyn/pull/74922#discussion_r1735340269](https://github.com/dotnet/roslyn/pull/74922#discussion_r1735340269).

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
cston: I've added a PROTOTYPE comment and a test for now.

333fred (approved):

PR #74936, opened 2024-08-28
Move to the real System.ValueTuple ref

jaredpar:


jaredpar:
@dotnet/roslyn-compiler PTAL

src/Compilers/Test/Core/TargetFrameworkUtil.cs
```diff
@@ -180,25 +179,28 @@ .. TestBase.WinRtRefs
         [
             TestBase.MinAsyncCorlibRef
         ];
-        public static ImmutableArray<MetadataReference> Mscorlib45ExtendedReferences =>
+
+        /*
+         * ⚠ Dev note ⚠: TestBase properties end here.
+         */
```
jaredpar, line 185: This comment was moved up because the properties below no longer depend on `TestBase`

333fred (approved):

jjonescz (approved):

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenTupleTest.cs
```diff
@@ -23146,16 +23146,7 @@ public void ValueTupleBaseError_NoSystemRuntime()
...
-                // (3,18): error CS0012: The type 'ValueType' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
-                //     ((int, int), (int, int)) F();
-                Diagnostic(ErrorCode.ERR_NoTypeDef, "(int, int)").WithArguments("System.ValueType", "System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a").WithLocation(3, 18),
-                // (3,5): error CS0012: The type 'ValueType' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
-                //     ((int, int), (int, int)) F();
-                Diagnostic(ErrorCode.ERR_NoTypeDef, "((int, int), (int, int))").WithArguments("System.ValueType", "System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a").WithLocation(3, 5));
+            comp.VerifyEmitDiagnostics();
```
333fred, line 23149: Is this not a big change in test behavior?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenTupleTest.cs
```diff
@@ -23926,13 +23917,12 @@ static void M()
             var compilation = CreateCompilationWithMscorlib40(source,
                 references: new[] { ValueTupleRef });
 
-            Assert.Equal(TypeKind.Class, compilation.GetWellKnownType(WellKnownType.System_ValueTuple_T2).TypeKind);
+            Assert.Equal(TypeKind.Struct, compilation.GetWellKnownType(WellKnownType.System_ValueTuple_T2).TypeKind);
```
333fred, line 23920: Same here.

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenTupleTest.cs
jaredpar: Been trying to chat with @cston and @jcouv about this. It's a situation where we're testing impossible scenarios and I'm not sure if we want the test before / after.

jaredpar: Got these back in sync.

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenTupleTest.cs
jaredpar: Back in sync

src/Compilers/Test/Core/TestBase.cs
```diff
@@ -235,14 +235,27 @@ public virtual void Dispose()
...
+        /// binary on nuget.org (possible a pre-release copy). This does have a few properties that were interesting
+        /// for a particular style of bug in VS that cannot be reproduced with modern TFMs. Specifically that it 
+        /// dependes on System.Runtime for parts of the impl and can't complie with only a reference to mscorlib. As
+        /// such this is kept around for those tests.
+        ///
+        /// Related issues
+        ///   - https://github.com/dotnet/roslyn/issues/14888
+        ///   - https://github.com/dotnet/roslyn/issues/14267
+        /// </summary>
+        public static MetadataReference ValueTupleLefacyRef => TestReferences.NetFx.ValueTuple.tuplelib;
```
cston, line 257: >Lefacy [](http://example.com/codeflow?start=50&length=6)

"Legacy"?

src/Compilers/Test/Core/TestBase.cs
```diff
@@ -235,14 +235,27 @@ public virtual void Dispose()
...
 
         public static MetadataReference MsvbRef_v4_0_30319_17929 => NetFramework.MicrosoftVisualBasic;
 
         public static MetadataReference CSharpRef => CSharpDesktopRef;
 
+        /// <summary>
+        /// This is a legacy copy of System.ValueTuple. The origin is unclear as this does not appear to be a released
+        /// binary on nuget.org (possible a pre-release copy). This does have a few properties that were interesting
+        /// for a particular style of bug in VS that cannot be reproduced with modern TFMs. Specifically that it 
+        /// dependes on System.Runtime for parts of the impl and can't complie with only a reference to mscorlib. As
```
cston, line 250: >dependes [](http://example.com/codeflow?start=12&length=8)

Typo.

cston, line 250: >dependes on System.Runtime for parts of the impl and can't complie [](http://example.com/codeflow?start=12&length=66)

"depends ... compile"

src/Compilers/VisualBasic/Test/Symbol/SymbolDisplay/SymbolDisplayTests.vb
```diff
@@ -5014,13 +5014,13 @@ End Class
...
                 SymbolDisplayPartKind.Space,
-                SymbolDisplayPartKind.ClassName,
+                SymbolDisplayPartKind.StructName,
                 SymbolDisplayPartKind.Punctuation,
                 SymbolDisplayPartKind.Keyword,
                 SymbolDisplayPartKind.Space,
                 SymbolDisplayPartKind.StructName,
                 SymbolDisplayPartKind.Punctuation},
-                references:={MetadataReference.CreateFromImage(TestResources.NetFX.ValueTuple.tuplelib)})
+                references:={Net461.ExtraReferences.SystemValueTuple})
```
cston, line 5023: >Net461.ExtraReferences.SystemValueTuple [](http://example.com/codeflow?start=29&length=39)

Should we just use the legacy reference here, to avoid changing the expected result?

cston (approved):

src/Compilers/Test/Core/TestBase.cs
jaredpar: I'm so proud of this typo that I thought about keeping it.

src/Compilers/VisualBasic/Test/Symbol/SymbolDisplay/SymbolDisplayTests.vb
jaredpar: I thought about that but symbol display is a user visible artifact. Decided that it was better to have our tests track down Tuple displays in the 99.999% case vs. the .001% case that I'm not sure you can reaally achieve.

cston (approved):

Issue #74894, opened 2024-08-24
Collection expression spread optimization using `List<T>.AddRange(IEnumerable<T>)` skipped for `ICollection<U>` when `T` and `U` are distinct types at compile-time only


cston:
The collection expression spread optimization for `[.. x]` where the target type is `List<T>` is skipped unnecessarily when:
- `x` has a value type enumerator, and
- `x` implements `ICollection<U>`, and
- `T` and `U` are distinct types at compile-time but not at runtime (for instance, `object` vs. `object?`, or `(int X, int Y)` vs. `(int, int)`).

See https://github.com/dotnet/roslyn/pull/74630
See https://github.com/dotnet/roslyn/pull/74769#issuecomment-2308489340

cston:
Lowering is currently using `TypeSymbolExtensions.ImplementsInterface()` to check for `ICollection<T>` (see [LocalRewriter_CollectionExpression.cs](https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs#L1079)), but we should probably use a conversion method such as `ConversionsBase.ClassifyBuiltInConversion()` instead.

DoctorKrolic:
[...]

PR #74949, opened 2024-08-29
Collection expressions: check conversion to ICollection<T> for spread optimization

cston:
Use `Conversions` method to check for `ICollection<T>` for spread optimization.

Fixes https://github.com/dotnet/roslyn/issues/74894

See https://github.com/dotnet/roslyn/pull/74630
See https://github.com/dotnet/roslyn/pull/74769#issuecomment-2308489340

### DIFF
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
@@ -1092,7 +1092,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                         var useSiteInfo = GetNewCompoundUseSiteInfo();
                         conversion = _compilation.Conversions.ClassifyConversionFromType(type, addRangeMethod.Parameters[0].Type, isChecked: false, ref useSiteInfo);
                         _diagnostics.Add(rewrittenSpreadOperand.Syntax, useSiteInfo);
-                        if (conversion.IsIdentity || (conversion.IsImplicit && conversion.IsReference))
+                        if (conversion.Kind is ConversionKind.Identity or ConversionKind.ImplicitReference)
                         {
                             conversion.MarkUnderlyingConversionsCheckedRecursive();
                             sideEffects.Add(_factory.Call(listTemp, addRangeMethod, rewrittenSpreadOperand));
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
@@ -1077,7 +1077,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                             // If collection has a struct enumerator but doesn't implement ICollection<T>
                             // then manual `foreach` is always more efficient then using `AddRange` method
-                            if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType )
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -1075,17 +1077,20 @@ private BoundExpression CreateAndPopulateList(BoundCollectionExpression node, Ty
 
                             // If collection has a struct enumerator but doesn't implement ICollection<T>
                             // then manual `foreach` is always more efficient then using `AddRange` method
-                            if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType &&
-                                !enumeratorInfo.CollectionType.ImplementsInterface(iCollectionOfElementType, ref discardedUseSiteInfo))
+                            if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType )
```
RikkiGibson, line 1080:
```suggestion
                            if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType)
```

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -1075,17 +1077,20 @@ private BoundExpression CreateAndPopulateList(BoundCollectionExpression node, Ty
 
                             // If collection has a struct enumerator but doesn't implement ICollection<T>
                             // then manual `foreach` is always more efficient then using `AddRange` method
-                            if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType &&
-                                !enumeratorInfo.CollectionType.ImplementsInterface(iCollectionOfElementType, ref discardedUseSiteInfo))
+                            if (enumeratorInfo.GetEnumeratorInfo.Method.ReturnType.IsValueType)
                             {
-                                return false;
+                                conversion = _compilation.Conversions.ClassifyBuiltInConversion(enumeratorInfo.CollectionType, iCollectionOfElementType, isChecked: false, ref discardedUseSiteInfo);
+                                if (!(conversion.Kind is ConversionKind.Identity or ConversionKind.ImplicitReference))
```
RikkiGibson, line 1083: nit: since the below conversion.Kind check for the AddRangeMethod parameter type is very similar in purpose, I would prefer if they were written in the same way, revising one or both as desired.

RikkiGibson (approved):

cston:

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -1075,19 +1077,22 @@ private BoundExpression CreateAndPopulateList(BoundCollectionExpression node, Ty
 
                             // If collection has a struct enumerator but doesn't implement ICollection<T>
```
chsienki, line 1078: Nit: 
```suggestion
                            // If collection has a struct enumerator but doesn't have a conversion to ICollection<T>
```

chsienki (approved):

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
cston: Thanks. I'll hold off changing the comment for now because changes here will conflict with another PR in progress, and since I believe the existing comment captures the intent. We can update later if needed.

PR #74962, opened 2024-08-30
Improve support for TypeLibFuncFlags.FHidden attributed properties.

ToddGrun:
Fixes feedback ticket: https://developercommunity.visualstudio.com/t/MIDL-hidden-attribute-not-respected-in-n/10652633

The issue here is that the TypeLibFuncAttribute is currently checked on the property symbol directly, not the backing getter/setter methods. I verified that midl compilation puts the property on the getter/setter.


ToddGrun:
@dotnet/roslyn-ide PTAL

PR #74978, opened 2024-09-03
Expose Go To Impl and Spell Check to Razor

davidwengier:
Part of https://github.com/dotnet/razor/issues/10746 and https://github.com/dotnet/razor/issues/10695

### DIFF
--- a/src/LanguageServer/Protocol/Handler/References/FindImplementationsHandler.cs
+++ b/src/LanguageServer/Protocol/Handler/References/FindImplementationsHandler.cs
@@ -11,6 +11,7 @@ using Microsoft.CodeAnalysis.Host.Mef;
 using Microsoft.CodeAnalysis.Options;
 using Microsoft.CodeAnalysis.PooledObjects;
 using Microsoft.CodeAnalysis.Shared.Extensions;
+using Microsoft.CodeAnalysis.Text;
 using LSP = Roslyn.LanguageServer.Protocol;
 
 namespace Microsoft.CodeAnalysis.LanguageServer.Handler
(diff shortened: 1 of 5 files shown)


ryzngard (approved):

src/LanguageServer/Protocol/Handler/References/FindImplementationsHandler.cs
```diff
@@ -33,26 +34,32 @@ public FindImplementationsHandler(IGlobalOptionService globalOptions)
 
         public LSP.TextDocumentIdentifier GetTextDocumentIdentifier(LSP.TextDocumentPositionParams request) => request.TextDocument;
 
-        public async Task<LSP.Location[]> HandleRequestAsync(LSP.TextDocumentPositionParams request, RequestContext context, CancellationToken cancellationToken)
+        public Task<LSP.Location[]> HandleRequestAsync(LSP.TextDocumentPositionParams request, RequestContext context, CancellationToken cancellationToken)
```
maryamariyan, line 37: assuming that you dropped `async` is intentional here

src/Tools/ExternalAccess/Razor/Cohost/Handlers/DocumentSpellCheck.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.CodeAnalysis.SpellCheck;
+using Roslyn.LanguageServer.Protocol;
+
+namespace Microsoft.CodeAnalysis.ExternalAccess.Razor.Cohost.Handlers;
+
+internal static class SpellCheck
+{
+    public readonly record struct SpellCheckSpan(int StartIndex, int Length, VSInternalSpellCheckableRangeKind Kind);
+
+    public static async Task<ImmutableArray<SpellCheckSpan>> GetSpellCheckSpansAsync(Document document, CancellationToken cancellationToken)
```
maryamariyan, line 20: would it be helpful to include try catch block in the `GetSpellCheckSpansAsync` method here?

maryamariyan (approved):

src/LanguageServer/Protocol/Handler/References/FindImplementationsHandler.cs
davidwengier: Indeed it would be a compiler warning if I left it in place.

If an async method only has a single `await` at the end, it is usually better to remove it, and the `async` modifier, and just return the `Task` directly. This avoids the creation of a state machine and so gives a slight perf bump. It does however change the call stack for exceptions that might be thrown so I wouldn't say its a hard and fast rule.

src/Tools/ExternalAccess/Razor/Cohost/Handlers/DocumentSpellCheck.cs
davidwengier: I think for our purposes we're better off letting the exception bubble out of the ExternalAccess layer here, and if necessary then we can handle it in Razor. If we handled the exceptions here it would go through Roslyn's exception reporting and telemetry, which might make it harder to get in front of the right team.

davidwengier:
@dibarbet @JoeRobich strictly speaking with Andrew's approval I could merge this, but that feels a little hollow. Would love a second sign-off :)

JoeRobich (approved):

Issue #74026, opened 2024-06-17
Visual Basic double to integer conversion tests stopped working in .NET 9 Preview 5


jjonescz:
**Version Used**: 62f2d479162cfb2353478d9dbd72a597b3b9fd1f (https://github.com/dotnet/roslyn/pull/73656)

Need to have `TargetFramework=net9.0` to reproduce locally.

Errors are like:

```
Failed Microsoft.CodeAnalysis.VisualBasic.UnitTests.CodeGenExpression.CIntTruncate_Implicit_01 [28 ms]
  Error Message:
   System.Exception : Error on 2147483648 -2147483648
  Stack Trace:
     at C1.CheckDouble(Double s, Int32 expected) in a.vb:line 51
   at C1.Main() in a.vb:line 42
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

This issue is mentioned in tests.

jjonescz:
Looks like this is an intentional breaking change: https://github.com/dotnet/runtime/issues/43895.

Simple repro:

```cs
var actual = floatToInt(int.MaxValue);
var expected = int.MinValue;
Console.WriteLine($"{actual == expected} {actual} {expected}");

static int floatToInt(float f) => (int)f;
```

with .NET SDK 9.0.100-preview.5.24307.3 and

```ps1
dotnet run -f net8.0 && echo '---' && dotnet run -f net9.0
```

prints

```
True -2147483648 -2147483648
---
False 2147483647 -2147483648
```

jaredpar:
> Likely cannot simply change the affected tests to observe the new behavior if they are being executed with different runtimes.

One approach to adjust the baseline based on the current runtime. Basically have the one value for `net472` and the other for `net9.0`.

PR #74980, opened 2024-09-03
Re-enable double-to-integer conversion tests

jjonescz:
Resolves https://github.com/dotnet/roslyn/issues/74026.

### DIFF
--- a/src/Compilers/VisualBasic/Test/Emit/CodeGen/CodeGenExpression.vb
+++ b/src/Compilers/VisualBasic/Test/Emit/CodeGen/CodeGenExpression.vb
@@ -75,8 +75,8 @@ Class C1
         CheckSingle(1.99F, 1)
         CheckSingle(Integer.MaxValue - 65F, 2147483520)
         ' https://github.com/dotnet/roslyn/issues/74026
-        ' CheckSingle(Integer.MaxValue - 64F, Integer.MinValue)  ' overflow
-        ' CheckSingle(Integer.MaxValue -  0F, Integer.MinValue)  ' overflow
+        CheckSingle(Integer.MaxValue - 64F, ]]><%= If(ExecutionConditionUtil.IsMonoOrCoreClr, "Integer.MaxValue", "Integer.MinValue") %><![CDATA[)  ' overflow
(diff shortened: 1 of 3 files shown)


src/Compilers/VisualBasic/Test/Emit/CodeGen/CodeGenExpression.vb
```diff
@@ -3214,9 +3214,9 @@ Class C1
         CheckDouble(Integer.MaxValue + 0D, Integer.MaxValue)
         CheckDouble(Integer.MaxValue + 0.01D, Integer.MaxValue)
         ' https://github.com/dotnet/roslyn/issues/74026
-        ' CheckDouble(Integer.MaxValue + 0.99D, Integer.MinValue)  ' overflow
-        ' CheckDouble(Integer.MaxValue + 1D, Integer.MinValue)  ' overflow
-        ' CheckDouble(Integer.MaxValue + 2D, Integer.MinValue)  ' overflow
+        CheckDouble(Integer.MaxValue + 0.99D, ]]><%= If(ExecutionConditionUtil.IsMonoOrCoreClr, "Integer.MaxValue", "Integer.MinValue") %><![CDATA[)  ' overflow
+        CheckDouble(Integer.MaxValue + 1D, ]]><%= If(ExecutionConditionUtil.IsMonoOrCoreClr, "Integer.MaxValue", "Integer.MinValue") %><![CDATA[)  ' overflow
+        CheckDouble(Integer.MaxValue + 2D, ]]><%= If(ExecutionConditionUtil.IsMonoOrCoreClr, "Integer.MaxValue", "Integer.MinValue") %><![CDATA[)  ' overflow
```
jaredpar, line 3219: Did you consider changing `CheckDouble` so that it has an overload that takes two arguments: expected framework and expected core? That may make the callsite code a bit cleaner.

jjonescz: That could be done but at the price of having to do many more code changes - the CheckDouble/CheckSingle methods are defined in each test separately (so either we would have to change the tests to share them or update each of these tests), I'm not sure that's worth it.

jaredpar (approved):

Issue #74875, opened 2024-08-23
Collection expressions may result in unnecessary tail increment


DoctorKrolic:
**Version Used**:
https://github.com/dotnet/roslyn/commit/b4dc2a111838c3fe1a494bbd3d94d82031e5353b (current sharplab `main`)

**Steps to Reproduce**:
```cs
using System.Collections.Generic;

public class C
{
    public List<int> M(ICollection<int> c) => [.. c, 4];
}
```
Expecte [codegen for `C.M`](https://sharplab.io/#v2:EYLgtghglgdgPgAQAwAIEEYAsBuAsAKAIQGY0AmFAYQIG8CUG1SAZKAZwBcAeWDgPhQBZABQBJSgHsANlICmAYw5QJMHjH4p5AShQBeAQG0AdEc0AaFJgC6efAF8gA==)

**Expected Behavior**:
No unnecessary operations

**Actual Behavior**:
Decompiled lowered code looks like this:
```cs
public List<int> M(ICollection<int> c)
{
    int num = 1 + c.Count;
    List<int> list = new List<int>(num);
    CollectionsMarshal.SetCount(list, num);
    Span<int> span = CollectionsMarshal.AsSpan(list);
    int num2 = 0;
    IEnumerator<int> enumerator = c.GetEnumerator();
    try
    {
        while (enumerator.MoveNext())
        {
            int current = enumerator.Current;
            span[num2] = current;
            num2++;
        }
    }
    finally
    {
        if (enumerator != null)
        {
            enumerator.Dispose();
        }
    }
    span[num2] = 4;
    num2++; // This increment in unnecessary
    return list;
}
```

PR #74982, opened 2024-09-03
Avoid unnecessary tail increment in collection expressions codegen

DoctorKrolic:
Fixes: https://github.com/dotnet/roslyn/issues/74875


jaredpar (dismissed):
This doesn't meet the bar for 17.12. Putting a hold on this until main is targeting 17.13

DoctorKrolic:
What kind of bar are we talking about here? Because I was thinking the opposite way: since the logic change is pretty much trivial this is a low risk change & it will get less review feedback to address, thus, I thought, it is gonna be a more suitable PR for a near-the-release time point

333fred:
> What kind of bar are we talking about here? Because I was thinking the opposite way: since the logic change is pretty much trivial this is a low risk change & it will get less review feedback to address, thus, I thought, it is gonna be a more suitable PR for a near-the-release time point

A bugfix only bar.

jaredpar:
> since the logic change is pretty much trivial this is a low risk change

Unfortunately optimization changes, even trivially ones, are not considered low risk. I understand that it can seem that way but in reality optimization changes are our most frequent source of regressions in the compiler. We've had a lot more luck with optimizations in collection expressions vs. other parts of the compiler but even so the bar for optimizations in general goes up faster than other bug fixes.

hez2010:
> in reality optimization changes

Actually the JIT is smart enough to get rid of the unnecessary increment here: https://godbolt.org/z/63ahWzrar
So the optimization in this PR actually didn't optimize anything.

DoctorKrolic:
> So the optimization in this PR actually didn't optimize anything

IL size also matters. Though a lot less of course, but still

DoctorKrolic:
@jaredpar I've seen [this PR](https://github.com/dotnet/roslyn/pull/75263). Does it mean, that `main` is `17.13` now?

jaredpar:
Yes main is 17.13 now.

DoctorKrolic:
@RikkiGibson @cston PTAL

DoctorKrolic:
@RikkiGibson @cston PTAL

PR #74993, opened 2024-09-04
Better Collection Expression Conversion From Expression

333fred:
Champion issue: https://github.com/dotnet/csharplang/issues/8374.
Spec: https://github.com/dotnet/csharplang/blob/main/proposals/collection-expressions-better-conversion.md.


333fred:
@cston, @AlekseyTs for review. I've been able to successfully compile the BCL with no regressions; still working on VS, I've had a few infra issues during previous attempts.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2914,12 +3039,26 @@ bool hasImplicitConversion(TypeSymbol source, TypeSymbol destination, ref Compou
...
-        private bool IsBetterParamsCollectionType(TypeSymbol t1, TypeSymbol t2, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+        private BetterResult IsBetterParamsCollectionType(TypeSymbol t1, TypeSymbol t2, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
         {
             CollectionExpressionTypeKind kind1 = ConversionsBase.GetCollectionExpressionTypeKind(Compilation, t1, out TypeWithAnnotations elementType1);
             CollectionExpressionTypeKind kind2 = ConversionsBase.GetCollectionExpressionTypeKind(Compilation, t2, out TypeWithAnnotations elementType2);
 
-            return IsBetterCollectionExpressionConversion(t1, kind1, elementType1.Type, t2, kind2, elementType2.Type, ref useSiteInfo);
+            if (kind1 is CollectionExpressionTypeKind.CollectionBuilder or CollectionExpressionTypeKind.ImplementsIEnumerable)
+            {
+                _binder.TryGetCollectionIterationType(CSharpSyntaxTree.Dummy.GetRoot(), t1, out elementType1);
```
333fred, line 3049: Note: I could use some ideas on getting a better syntax node here. We don't currently pass a reference to the syntax for the entire method to overload resolution. The syntax node is _mostly_ used for diagnostics, but there is at least one call to see whether calls are omitted in this syntax tree. This _could_ have an impact on binding in particularly interesting cases, where a pattern-based `GetEnumerator` call ends up being seen in a scenario where it should have been omitted, but all that will cause it to do is give us an element type here when it otherwise wouldn't have one, and we likely wouldn't even get here if that were the case, but it's interesting nonetheless.

cston: It's surprising we don't have a syntax node representing the containing expression in `MemberResolutionResult` for instance. And to add that, it looks like we'd need to pass the `SyntaxNode` or `BoundExpression` through a number of layers in the call stack for overload resolution.

333fred: Yup. It's possible to do, but it's quite a bit of work :). I was hoping there'd be something obvious I was missing.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2914,12 +3039,26 @@ bool hasImplicitConversion(TypeSymbol source, TypeSymbol destination, ref Compou
                 Conversions.ClassifyImplicitConversionFromType(source, destination, ref useSiteInfo).IsImplicit;
         }
 
-        private bool IsBetterParamsCollectionType(TypeSymbol t1, TypeSymbol t2, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+        private BetterResult IsBetterParamsCollectionType(TypeSymbol t1, TypeSymbol t2, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
```
cston, line 3042: >IsBetterParamsCollectionType [](http://example.com/codeflow?start=29&length=28)

Perhaps rename to `BetterParamsCollectionType` since this no longer returns a `bool`.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2850,35 +2847,163 @@ private BetterResult BetterConversionFromExpression(BoundExpression node, TypeSy
...
+                    case (false, true):
+                        return BetterResult.Right;
+                }
+            }
+
+            // - `E₁` does not have an identity conversion to `E₂`, and the element conversions to `E₁` are better than the element conversions to `E₂`, or
+            // - `E₁` has an identity conversion to `E₂`, and one of the following holds:
+
+            // `E₁` is compared to `E₂` as follows:
+            // If there is an identity conversion from `E₁` to `E₂`, then the element conversions are as good as each other. Otherwise, the element conversions to `E₁` are better than the element conversions to `E₂` if:
```
cston, line 2929: > Otherwise, [](http://example.com/codeflow?start=124&length=11)

Minor: Consider splitting long lines in this comment (this line and the last line).

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2850,35 +2847,163 @@ private BetterResult BetterConversionFromExpression(BoundExpression node, TypeSy
...
+            // Conversion comparisons are made using better conversion from expression if `ELᵢ` is not a spread element. If `ELᵢ` is a spread element, we use better conversion from the element type of the spread collection to `E₁` or `E₂`, respectively.
+
+            if (!Conversions.HasIdentityConversion(elementType1, elementType2))
+            {
+                var betterResult = BetterResult.Neither;
+                Debug.Assert(underlyingElementConversions1.Length == underlyingElementConversions2.Length && underlyingElementConversions1.Length == collectionExpressionElements.Length);
+
+                for (int i = 0; i < underlyingElementConversions1.Length; i++)
+                {
+                    // Conversion comparisons are made using better conversion from expression if `ELᵢ` is not a spread element. If `ELᵢ` is a spread element, we use better conversion from the element type of the spread collection to `E₁` or `E₂`, respectively.
```
cston, line 2942: > If `ELᵢ` [](http://example.com/codeflow?start=128&length=9)

Consider splitting long line.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2850,35 +2847,163 @@ private BetterResult BetterConversionFromExpression(BoundExpression node, TypeSy
...
+                    {
+                        if (spreadConversion is null)
+                        {
+                            // There were errors, don't attempt to compare for better error recovery
+                            Debug.Assert(element.HasErrors);
+                            continue;
+                        }
+
+                        var conversionToE1 = Conversions.GetCollectionExpressionSpreadElementConversion(spread, elementType1, ref useSiteInfo);
+                        var conversionToE2 = Conversions.GetCollectionExpressionSpreadElementConversion(spread, elementType2, ref useSiteInfo);
```
cston, line 2956: Are these conversions different from `underlyingElementConversions1[i]` and `underlyingElementConversions2[i]`, which were calculated at [https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs#L222](https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs#L222)?

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2850,35 +2847,163 @@ private BetterResult BetterConversionFromExpression(BoundExpression node, TypeSy
...
+
+                    BetterResult elementBetterResult;
+                    if (element is BoundCollectionExpressionSpreadElement { Conversion: var spreadConversion } spread)
+                    {
+                        if (spreadConversion is null)
+                        {
+                            // There were errors, don't attempt to compare for better error recovery
+                            Debug.Assert(element.HasErrors);
+                            continue;
+                        }
```
cston, line 2953: It doesn't look like we're hitting this case in `CollectionExpressionTests`. But is this check necessary?

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -1021,18 +1024,21 @@ class Program
                     static void Main()
                     {
                         var x = F([1, null]);
+                        x.Report(includeType: true);
                         int?[] y = [null, 2];
                         var z = F([..y]);
+                        z.Report(includeType: true);
                     }
```
cston, line 1031: >} [](http://example.com/codeflow?start=20&length=1)

This looks like the only test that hits the spread element case in `OverloadResolution.BetterCollectionExpressionConversion()`. Let's consider also testing cases where there are more than element. Perhaps:
```
F([3, ..y]).Report(includeType: true);
F([..y, 4]).Report(includeType: true);
int[] w = [5, 6, 7];
F([..y, ..w]).Report(includeType: true);
F([..w, ..y]).Report(includeType: true);
```

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -1372,10 +1380,129 @@ static void Main()
...
+                        // 0.cs(12,17): error CS0121: The call is ambiguous between the following methods or properties: 'Program.F2(ReadOnlySpan<object>)' and 'Program.F2(ReadOnlySpan<long>)'
+                        //         var y = F2([4, 5]);
+                        Diagnostic(ErrorCode.ERR_AmbigCall, "F2").WithArguments(generateMethodSignature("F2", type2), generateMethodSignature("F2", type1)).WithLocation(12, 17));
+                }
+            }
+        }
+
+        [Theory]
+        [InlineData("System.ReadOnlySpan<int>", "System.Span<int>", "System.ReadOnlySpan<System.Int32>")]
+        [InlineData("System.ReadOnlySpan<int>", "System.Span<object>", null)]
```
cston, line 1434: >"System.ReadOnlySpan<int>", "System.Span<object>", null [](http://example.com/codeflow?start=20&length=55)

These cases where the collections are `ReadOnlySpan<T1>` and `Span<T2>`, or span of `T1` and array or array interface of `T2`, and where `T1` is implicitly convertible to `T2` are the breaking change cases. Should we include an `expectedType12` parameter here as well, where the test compiles against both language versions?

cston: Please add an item to the breaking change doc.

cston: > Should we include an `expectedType12` parameter here as well, where the test compiles against both language versions?

It looks like the C#12 cases are tested in `BetterConversionFromExpression_01B_CSharp12`.

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -1562,28 +1887,48 @@ static void Main()
...
                     {
                         SpanDerived(new[] { string.Empty }); // ambiguous
-                        ArrayDerived([string.Empty]); // ambiguous
                     }
                 }
                 """;
+
             comp = CreateCompilation(
                 new[] { sourceA, sourceB2 },
+                parseOptions: TestOptions.Regular13,
```
cston, line 1917: >Regular13 [](http://example.com/codeflow?start=42&length=9)

`Regular12`?

src/Compilers/CSharp/Test/Emit2/Semantics/ParamsCollectionTests.cs
```diff
@@ -5198,24 +5208,14 @@ static string generateMethodSignature(string methodName, string parameterType) =
         [InlineData("System.ReadOnlySpan<int>", "System.Span<int>", "System.ReadOnlySpan<System.Int32>")]
         [InlineData("System.ReadOnlySpan<int>", "System.Span<object>", "System.ReadOnlySpan<System.Int32>")]
         [InlineData("System.ReadOnlySpan<int>", "System.Span<int?>", "System.ReadOnlySpan<System.Int32>")]
-
-        // Ambiguous for inline collection expression, but 'int' is a better conversion target than 'object' in params case
         [InlineData("System.ReadOnlySpan<object>", "System.Span<int>", "System.Span<System.Int32>")] // cannot convert object to int
```
cston, line 5211: >// cannot convert object to int [](http://example.com/codeflow?start=101&length=31)

Do we still need these "// cannot convert ..." comments?

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
```diff
@@ -1835,6 +2206,107 @@ static void Main()
...
+                targetFramework: TargetFramework.Net80,
+                verify: Verification.Skipped,
+                expectedOutput: IncludeExpectedOutput("ReadOnlySpan<string>"));
+
+            CreateCompilation(source, parseOptions: TestOptions.Regular12, targetFramework: TargetFramework.Net80).VerifyDiagnostics(
+                // (8,9): error CS0121: The call is ambiguous between the following methods or properties: 'Program.F2(ReadOnlySpan<string>)' and 'Program.F2(ReadOnlySpan<IFormattable>)'
+                //         F2([$"{1}"]);
+                Diagnostic(ErrorCode.ERR_AmbigCall, "F2").WithArguments("Program.F2(System.ReadOnlySpan<string>)", $"Program.F2(System.ReadOnlySpan<{formatType}>)").WithLocation(8, 9)
+                );
+        }
```
cston, line 2308: >} [](http://example.com/codeflow?start=8&length=1)

Are we testing the following, where the collection types are distinct, one or both are span types, and there is an implicit numeric conversion from one element type to the other?
```csharp
using System;

class Program
{
    static void Main()
    {
        F1([1, 2, 3]); // C#12: F1(ReadOnlySpan<byte>); C#13: ambiguous
        F2([4, 5, 6]); // C#12: ambiguous; C#13: F2(ReadOnlySpan<int>)
    }
    
    static void F1(ReadOnlySpan<byte> arg) { }
    static void F1(Span<int> arg) { }
    
    static void F2(ReadOnlySpan<int> arg) { }
    static void F2(Span<byte> arg) { }
}
```

cston: Please add an item to the breaking change doc.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
333fred: https://github.com/333fred/roslyn/blob/bd29c603437913a8609bba279e294fbb4166c4a2/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs#L5188-L5201 will create a spread element without a conversion, if that's what you're asking?

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
333fred: > It looks like the C#12 cases are tested in BetterConversionFromExpression_01B_CSharp12.

Yeah, I thought about trying to include it all in one big matrix, but the differences got too large and I didn't think the result was actually comprehensible. Will definitely add the whole change here to the breaking changes doc.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
cston: It's not clear we can get here with a spread element that has errors, but even if we do, we should have `underlyingElementConversions*` that were calculated for each element (including spreads) that we can use.

src/Compilers/CSharp/Test/Emit2/Semantics/ParamsCollectionTests.cs
333fred: No, I meant to remove them, thanks.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2850,35 +2847,158 @@ private BetterResult BetterConversionFromExpression(BoundExpression node, TypeSy
...
+                Debug.Assert(underlyingElementConversions1.Length == underlyingElementConversions2.Length && underlyingElementConversions1.Length == collectionExpressionElements.Length);
+
+                for (int i = 0; i < underlyingElementConversions1.Length; i++)
+                {
+                    // Conversion comparisons are made using better conversion from expression if `ELᵢ` is not a spread element. If `ELᵢ` is a spread element,
+                    // we use better conversion from the element type of the spread collection to `E₁` or `E₂`, respectively.
+                    var element = collectionExpressionElements[i];
+
+                    BetterResult elementBetterResult;
+                    if (element is BoundCollectionExpressionSpreadElement { Conversion: var spreadConversion } spread)
```
cston, line 2948: >Conversion: var spreadConversion [](http://example.com/codeflow?start=76&length=32)

The check for `Conversion:` is not needed.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2850,35 +2847,158 @@ private BetterResult BetterConversionFromExpression(BoundExpression node, TypeSy
...
+                    {
+                        var conversionToE1 = underlyingElementConversions1[i];
+                        var conversionToE2 = underlyingElementConversions2[i];
```
cston, line 2951: Consider inlining in the call to `BetterConversionTarget()`, or move the declarations outside the `if` and share with the `else`.

src/Compilers/CSharp/Test/Emit2/Semantics/CollectionExpressionTests.cs
cston: As you mentioned offline, none of the cases above are ambiguous in C#13, and we chose a different overload for `F1()` and `F3()` in C#12 vs. C#13. Updated the example.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 9.md
```diff
@@ -44,3 +44,47 @@ unsafe class C // unsafe context
...
+***Introduced in Visual Studio 2022 Version 17.12 and newer when using C# 13+***
+
+There are a few changes in collection expression binding in C# 13. Most of these are turning ambiguities into successful compilations,
+but a couple are breaking changes that either result in a new compilation error, or are a behavior breaking change. They are detailed
+below.
+
+### Empty collection expressions no longer use whether an API is a `Span` to tiebreak on overloads
+
+When an empty collection expression is provided to an overloaded method, and there isn't a clear element type, we no longer use whether
+an API takes a `ReadOnlySpan` or a `Span` to decide whether to prefer that API. For example:
```
cston, line 59: >`ReadOnlySpan` or a `Span` [](http://example.com/codeflow?start=15&length=26)

Consider including type parameters: "a `ReadOnlySpan<T>` or a `Span<T>`"

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 9.md
```diff
@@ -44,3 +44,47 @@ unsafe class C // unsafe context
 ```
 
 You can work around the break simply by adding the `unsafe` modifier to the local function.
+
+## Collection expression breaking changes in C# 13 and newer
```
cston, line 48: >breaking changes in [](http://example.com/codeflow?start=25&length=19)

Consider clarifying these are specific to overload resolution: "breaking changes with overload resolution in ..."

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 9.md
```diff
@@ -44,3 +44,47 @@ unsafe class C // unsafe context
...
+
+***Introduced in Visual Studio 2022 Version 17.12 and newer when using C# 13+***
+
+There are a few changes in collection expression binding in C# 13. Most of these are turning ambiguities into successful compilations,
+but a couple are breaking changes that either result in a new compilation error, or are a behavior breaking change. They are detailed
+below.
+
+### Empty collection expressions no longer use whether an API is a `Span` to tiebreak on overloads
```
cston, line 56: >`Span` [](http://example.com/codeflow?start=67&length=6)

Perhaps either "span" or "`Span<T>`".

cston (approved):

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 9.md
```diff
@@ -44,3 +44,47 @@ unsafe class C // unsafe context
...
+    {
+        M([]); // C.M(ReadOnlySpan<int>) in C# 12, error in C# 13.
+    }
+}
+```
+
+### Exact element type is preferred over all else
+
+In C# 13, we prefer an exact element type match, looking at conversions from expressions. This can result in a behavior change when involving
+numeric constants:
```
cston, line 77: Should we be more specific about the cases affected? For instance:
> This can result in a behavior change when involving numeric constants, specifically, when the overload parameters are `ReadOnlySpan<E1>` and `Span<E2>`, or a span type of `E1` and an array or interface implemented by an array of `E2`, and there is a numeric conversion from `E1` to `E2`.

333fred: I'm not certain it's limited to numerics. There's probably other conversions from expressions that can change. For example, things involving interpolated strings will probably also change.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
cston: I don't think we need to use the syntax node for the call site here, if the only use is determine whether the `params` type is a valid `params` type. That determination is made at the declaration site of the method, and if the type is valid there, it should be valid at call sites where the method is available.

cston (approved):

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2850,35 +2847,157 @@ private BetterResult BetterConversionFromExpression(BoundExpression node, TypeSy
...
+            // - `E₂` is the element type of `T₂`
+            // - `CE₁ᵢ` are the series of conversions from `ELᵢ` to `E₁`
+            // - `CE₂ᵢ` are the series of conversions from `ELᵢ` to `E₂`
+
+            var t1IsSpanType = kind1 is CollectionExpressionTypeKind.ReadOnlySpan or CollectionExpressionTypeKind.Span;
+            var t2IsSpanType = kind2 is CollectionExpressionTypeKind.ReadOnlySpan or CollectionExpressionTypeKind.Span;
+
+            // `C₁` is a ***better collection conversion from expression*** than `C₂` if:
+            // - `T₁` or `T₂` is not a *span type*, and `T₁` is implicitly convertible to `T₂`, and `T₂` is not implicitly convertible to `T₁`, or
+            if (!t1IsSpanType && !t2IsSpanType)
```
AlekseyTs, line 2911: >!t1IsSpanType && !t2IsSpanType [](http://example.com/codeflow?start=16&length=30)

It looks like the comment "`T₁` or `T₂` is not a *span type*" and the condition do not match. I interpret the comment as ```!t1IsSpanType || !t2IsSpanType```

333fred: I need to update this comment, Chuck had the same suggestion on the spec issue. Thanks for poking.

docs/Language Feature Status.md
```diff
@@ -29,6 +29,7 @@ efforts behind them.
...
 | [Ref Struct Interfaces](https://github.com/dotnet/csharplang/issues/7608) | [RefStructInterfaces](https://github.com/dotnet/roslyn/tree/features/RefStructInterfaces) | [Merged into 17.11p2](https://github.com/dotnet/roslyn/issues/72124) | [AlekseyTs](https://github.com/AlekseyTs) | [cston](https://github.com/cston), [jjonescz](https://github.com/jjonescz) | [ToddGrun](https://github.com/ToddGrun) |  [agocke](https://github.com/agocke), [jaredpar](https://github.com/jaredpar) |
+| [Collection expression better conversion from expression](https://github.com/dotnet/csharplang/issues/8374) | main | PROTOTYPE | [333fred](https://github.com/333fred) | [cston](https://github.com/cston), [AlekseyTs](https://github.com/AlekseyTs) | (no IDE impact) | [333fred](https://github.com/333fred), [CyrusNajmabadi](https://github.com/CyrusNajmabadi) |
```
jcouv, line 32: Since this is going straight to `main` branch, don't forget to update before merging. We should have a link to test plan (if available) and record the version this was merged into.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
333fred: For reference, the updated wording is "Both `T₁` and `T₂` are not *span types*"

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -2850,35 +2847,157 @@ private BetterResult BetterConversionFromExpression(BoundExpression node, TypeSy
...
+                    // we use better conversion from the element type of the spread collection to `E₁` or `E₂`, respectively.
+                    var element = collectionExpressionElements[i];
+                    var conversionToE1 = underlyingElementConversions1[i];
+                    var conversionToE2 = underlyingElementConversions2[i];
 
-            return IsBetterCollectionExpressionConversion(t1, kind1, elementType1, t2, kind2, elementType2, ref useSiteInfo);
+                    BetterResult elementBetterResult;
+                    if (element is BoundCollectionExpressionSpreadElement spread)
+                    {
+                        elementBetterResult = BetterConversionTarget(spread, elementType1, conversionToE1, elementType2, conversionToE2, ref useSiteInfo, okToDowngradeToNeither: out _);
```
AlekseyTs, line 2952: >BetterConversionTarget [](http://example.com/codeflow?start=46&length=22)

Is there a test that would fail if we were to call ```BetterConversionFromExpression``` here instead?

333fred: It's not entirely clear to me what we'd be checking for better expression on; spread elements are not expressions, so we couldn't pass them to `BetterConversionFromExpression`, and the type of the expression under the spread isn't convertible to either `elementType1` or `elementType2` (probably. I suppose there are likely some interesting scenarios you could construct for that like `StringValues`). Can you show an example of what you were thinking?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1021,18 +1024,40 @@ class Program
...
+                        F([3, ..y]).Report(includeType: true);
+                        F([..y, 4]).Report(includeType: true);
+                        int[] w = [5, 6, 7];
+                        F([..y, ..w]).Report(includeType: true);
+                        F([..w, ..y]).Report(includeType: true);
                     }
                 }
                 """;
-            CreateCompilation(source).VerifyEmitDiagnostics(
+            CompileAndVerify([source, s_collectionExtensions], parseOptions: TestOptions.Regular13,
```
AlekseyTs, line 1039: >TestOptions.Regular13 [](http://example.com/codeflow?start=77&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1372,10 +1399,130 @@ static void Main()
                     }
                 }
                 """;
+
+            verify(TestOptions.Regular12, expectedType12);
+            verify(TestOptions.Regular13, expectedType13);
```
AlekseyTs, line 1404: >TestOptions.Regular13 [](http://example.com/codeflow?start=19&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1548,12 +1893,14 @@ static void Main()
...
                         Identical([string.Empty]); // Span<string>
                         SpanDerived([string.Empty]); // Span<string>
+                        ArrayDerived([string.Empty]); // string[]
                     }
                 }
                 """;
             var comp = CreateCompilation(
                 new[] { sourceA, sourceB1 },
                 targetFramework: TargetFramework.Net80,
+                parseOptions: TestOptions.Regular13,
```
AlekseyTs, line 1903: >TestOptions.Regular13 [](http://example.com/codeflow?start=30&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1562,28 +1909,48 @@ static void Main()
...
                     {
                         SpanDerived(new[] { string.Empty }); // ambiguous
-                        ArrayDerived([string.Empty]); // ambiguous
                     }
                 }
                 """;
+
+            comp = CreateCompilation(
+                new[] { sourceA, sourceB2 },
+                parseOptions: TestOptions.Regular13,
```
AlekseyTs, line 1939: >TestOptions.Regular13 [](http://example.com/codeflow?start=30&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1668,9 +2035,26 @@ static void Main()
                     }
                 }
                 """;
+            var comp = CreateCompilation(new[] { source, s_collectionExtensionsWithSpan }, parseOptions: TestOptions.Regular13, targetFramework: TargetFramework.Net80);
```
AlekseyTs, line 2038: >TestOptions.Regular13 [](http://example.com/codeflow?start=105&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1773,21 +2157,28 @@ public void BetterConversionFromExpression_07()
...
                     }
                 }
                 """;
-            var comp = CreateCompilation(
+            CompileAndVerify(
                 source,
-                targetFramework: TargetFramework.Net80);
-            comp.VerifyEmitDiagnostics(
+                targetFramework: TargetFramework.Net80,
+                parseOptions: TestOptions.Regular13,
```
AlekseyTs, line 2174: >TestOptions.Regular13 [](http://example.com/codeflow?start=30&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1800,21 +2191,23 @@ static void Main()
...
                     static void Main()
                     {
                         F1([1, 2, 3]);
                     }
                 }
                 """;
-            var comp = CreateCompilation(source);
-            comp.VerifyEmitDiagnostics(
-                // (7,9): error CS0121: The call is ambiguous between the following methods or properties: 'Program.F1(int[])' and 'Program.F1(object[])'
+            CompileAndVerify(source, parseOptions: TestOptions.Regular13, expectedOutput: "int[]").VerifyDiagnostics();
```
AlekseyTs, line 2205: >TestOptions.Regular13 [](http://example.com/codeflow?start=51&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1835,6 +2228,144 @@ static void Main()
...
+                    {
+                        F2([1, (byte)2]);
+                    }
+                }
+                """;
+
+            // (9,9): error CS0121: The call is ambiguous between the following methods or properties: 'Program.F2(List<int>)' and 'Program.F2(List<byte>)'
+            //         F2([1, (byte)2]);
+            var expected = Diagnostic(ErrorCode.ERR_AmbigCall, "F2").WithArguments("Program.F2(System.Collections.Generic.List<int>)", "Program.F2(System.Collections.Generic.List<byte>)").WithLocation(9, 9);
+            CreateCompilation(source, parseOptions: TestOptions.Regular13).VerifyDiagnostics(expected);
```
AlekseyTs, line 2272: >TestOptions.Regular13 [](http://example.com/codeflow?start=52&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1835,6 +2228,144 @@ static void Main()
...
+                    static void F2(List<int> value) { Console.WriteLine("List<int>"); }
+                    static void F2(List<byte> value) { Console.WriteLine("List<byte>"); }
+                    static void Main()
+                    {
+                        F2([(byte)1, (byte)2]);
+                    }
+                }
+                """;
+
+            CompileAndVerify(source, parseOptions: TestOptions.Regular13, expectedOutput: "List<byte>");
```
AlekseyTs, line 2293: >TestOptions.Regular13 [](http://example.com/codeflow?start=51&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1835,6 +2228,144 @@ static void Main()
...
+                    static void F2(ReadOnlySpan<{{formatType}}> value) { Console.WriteLine("ReadOnlySpan<{{formatType}}>"); }
+                    static void Main()
+                    {
+                        F2([$"{1}"]);
+                    }
+                }
+                """;
+
+            CompileAndVerify(source,
+                parseOptions: TestOptions.Regular13,
```
AlekseyTs, line 2320: >TestOptions.Regular13 [](http://example.com/codeflow?start=30&length=21)

Cover Preview version as well?

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1835,6 +2228,144 @@ static void Main()
...
+                    public static implicit operator CustomHandler(string s) => new CustomHandler(0, 0);
+                }
+                """;
+
+            var handler = GetInterpolatedStringCustomHandlerType("CustomHandler", "partial class", useBoolReturns: false, includeOneTimeHelpers: false);
+
+            CompileAndVerify(
+                new[] { source, handler },
+                targetFramework: TargetFramework.Net80,
+                parseOptions: TestOptions.Regular13,
```
AlekseyTs, line 2357: >TestOptions.Regular13 [](http://example.com/codeflow?start=30&length=21)

Cover Preview version as well?

AlekseyTs:
Done with review pass (commit 19)

333fred:
@AlekseyTs addressed feedback.

AlekseyTs (approved):
LGTM (commit 20)

cston (approved):

docs/Language Feature Status.md
333fred: I hadn't planned on actualy making a test plan, given the size of this issue, but I can make one if we'd prefer to do so.

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -1835,6 +2264,535 @@ static void Main()
...
+                dynamic d = null;
+
+                M1([o, o]);
+                M1([d, d]);
+                M1([o, d]);
+                M1([d, o]);
+                M2([o, o]);
+                M2([d, d]);
+                M2([o, d]);
+                M2([d, o]);
```
cston, line 2446: >M2([d, o]); [](http://example.com/codeflow?start=16&length=11)

Let's test spread elements as well:
```csharp
IEnumerable<object> x = [];
IEnumerable<dynamic> y = [];
M1([..x]);
M1([..y]);
M1([o, ..x]);
M1([d, ..x]);
M1([..y, o]);
M1([..y, d]);
M1([..x, ..y]);
M1([..y, ..x]);
```

cston (approved):

PR #74994, opened 2024-09-04
Avoid BuildHost crash in Mono due to missing types

jaredpar:
The LoadProjectFileAsync routine calls EnsureMSBuildLoaded to make sure the Microsoft.Build.dll is accessible, but requires types from that assembly within the routine itself.  This may not always work with Mono, as the JIT may look up the types during compilation.

Fixed by splitting out a LoadProjectFileCoreAsync routine, similarly to what is already done for GetProjectsInSolution.

Fixes https://github.com/dotnet/runtime/issues/101121.


uweigand:
Thanks, @jaredpar !   Can this still make it into .NET 9?

jaredpar:
Yes it will make .NET 9. Our `main` branch is still flowing there.

PR #74996, opened 2024-09-04
Remove debug-only assert method

333fred:
This method is only used in one place, only wraps one method call, and can cause people to accidentally select the wrong method in RELEASE mode. I've just inlined the content and removed the method.


333fred:
@dotnet/roslyn-compiler for a small review.

RikkiGibson (approved):

333fred:
@dotnet/roslyn-compiler for a second small review.

PR #75000, opened 2024-09-05
Fix exception case in compiler resolver

jaredpar:
This fixes how we deal with exceptions in the compiler assembly resolver.

### DIFF
--- a/src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.Core.cs
+++ b/src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.Core.cs
@@ -227,7 +227,19 @@ namespace Microsoft.CodeAnalysis
         {
             private readonly AssemblyLoadContext _compilerAlc = compilerContext;
 
-            public Assembly? ResolveAssembly(AssemblyName assemblyName) => _compilerAlc.LoadFromAssemblyName(assemblyName);
+            public Assembly? ResolveAssembly(AssemblyName assemblyName)
+            {
+                try
+                {
+                    return _compilerAlc.LoadFromAssemblyName(assemblyName);
+                }
+                catch
+                {
+                    // The LoadFromAssemblyName method will throw if the assembly cannot be found. Need
+                    // to catch this exception and return null to satisfy the interface contract.
+                    return null;
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


jaredpar:
@dotnet/roslyn-compiler PTAL

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.Core.cs
```diff
@@ -227,7 +227,19 @@ internal sealed class CompilerAnalyzerAssemblyResolver(AssemblyLoadContext compi
         {
             private readonly AssemblyLoadContext _compilerAlc = compilerContext;
 
-            public Assembly? ResolveAssembly(AssemblyName assemblyName) => _compilerAlc.LoadFromAssemblyName(assemblyName);
+            public Assembly? ResolveAssembly(AssemblyName assemblyName)
+            {
+                try
+                {
+                    return _compilerAlc.LoadFromAssemblyName(assemblyName);
```
jjonescz, line 234: I'm wondering what was happening previously, where was the exception propagated (did it crash the whole process)?

jjonescz (approved):

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.Core.cs
jaredpar: The code which consumes `IAnalyzerAssemblyResolver` guards against unexpected exceptions. Basically, that is an extension point and the compiler generally guards against unexpected exceptions coming from extensions. So this scenario was worknig overall but this specific component wasn't meeting its contract.

src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.cs
```diff
@@ -415,7 +416,7 @@ internal string GetRealAnalyzerLoadPath(string originalFullPath)
                             return resolvedAssembly;
                         }
                     }
-                    catch
+                    catch (Exception ex) when (FatalError.ReportAndCatch(ex, ErrorSeverity.Diagnostic))
```
jaredpar, line 419: @jjonescz your comment reminded me that we're essentially silently swallowing the exceptions from plugins. Changed that to be non-silent. It's not a fatal error but definitely something that should at least raise an alert if it happens. This would've caught the issue that I fixed with this PR.

chsienki (approved):

PR #75011, opened 2024-09-06
Workaround for InvalidOperationException thrown from EmitSolutionUpdateResults.GetAllDiagnosticsAsync

tmat:
We were missing call to `WithUpToDateSourceGeneratorDocuments` in code paths that post-processed rude edits reported by `EditAndContinueService`. The service itself did call `WithUpToDateSourceGeneratorDocuments` on the solution snapshot it analyzed but the rude edits were post-processed on the snapshot that has not been updated with the latest source generator outputs.  When the analysis found a rude edit in a document that has just been generated we did not find that document in the solution snapshot.

Following up with a better fix that removes the post-processing step, but for servicing this workaround is more appropriate: https://github.com/dotnet/roslyn/pull/75013

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2169491

### DIFF
--- a/eng/targets/Settings.props
+++ b/eng/targets/Settings.props
@@ -60,6 +60,12 @@
 
     <!-- TODO --> 
     <_SkipUpgradeNetAnalyzersNuGetWarning>true</_SkipUpgradeNetAnalyzersNuGetWarning>
+
+    <!--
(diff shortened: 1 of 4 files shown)


tmat:
@davidwengier PTAL
@phil-allen-msft for servicing

davidwengier (approved):

arkalyanms (approved):

arkalyanms:
@tmat merge when ready.

PR #75027, opened 2024-09-09
Preserve trailing trivia when replacing method with property

### LINKED ISSUE #61161: `Replace method with property` refactoring sometimes eats new line
Ported from https://github.com/dotnet/roslyn/issues/40758#issuecomment-1086711511

**Steps to Reproduce**:
```cs
class C
{
    public int Test1() { return 1; }
    public int Test2() { return 0; }
}
```
Execute `Replace method with property` on `Test1`

**Expected Behavior**:
```cs
class C
{
    public int Test1 => 1;
    public int Test2() { return 0; }
}
```

**Actual Behavior**:
```cs
class C
{
    public int Test1 => 1; public int Test2() { return 0; }
}
```

Orachor:
Fixes `CSharpReplaceMethodWithPropertyService` in some cases not preserving end of line trivia.

Closes #61161

### DIFF
--- a/src/Features/CSharp/Portable/ReplaceMethodWithProperty/CSharpReplaceMethodWithPropertyService.cs
+++ b/src/Features/CSharp/Portable/ReplaceMethodWithProperty/CSharpReplaceMethodWithPropertyService.cs
@@ -6,7 +6,6 @@
 
 using System;
 using System.Composition;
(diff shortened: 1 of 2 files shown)


CyrusNajmabadi (approved):
Looks great. Thanks!!

PR #75037, opened 2024-09-10
Remove MS.CA.Test.Resources.Proprietary PackageReference

jaredpar:
This package is the last remnants of the roslyn-internal repository. It was still holding onto a few test assets that never got ported into the open but were available through the NuPkg reference.

This package is becoming a problem with component governance as it is a `netstandard1.3` package and hence brings in a lot of vulnerable components.

Considered upgrading this package to `netstandard2.0` but that would mean un-archiving the roslyn-internal repo, finding a place to put the source code and setting up a new official pipeline for the package. That is a pretty high price. Instead I decided to finish the work that I should've completed a number of years ago.

Yes this does add a few more .dll into our git tree. These are files that haven't changed in 10+ years so versioning won't be an issue. They were all brought down already for restore hence it's not new files.


src/Compilers/Test/Core/Mocks/Silverlight.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Roslyn.Test.Utilities;
+
+/// <summary>
+/// The assemblies produced here are designed to mimic the public key token structure of 
+/// silverlight references. This often presents challenges to the compiler because it has
+/// to know that two mscorlib with different public key tokens need to be treated as the 
+/// identicial. The assemblies produced here have the same identity of those that come
+/// from silverlight but without necessarily the same type contents.
+/// </summary>
```
jaredpar, line 25: I went back and forth on this decision several times. The Silverlight references are _only_ used right now to test assembly unification policies (that I could see). Rather than the overhead of bringing in more .dll files decided to just dynamically build the same assemblies that would let us test unification policy.

src/Compilers/Test/Resources/Core/Microsoft.CodeAnalysis.Compiler.Test.Resources.csproj
```diff
@@ -85,11 +86,16 @@
     <Content Include="NetFX\aacorlib\Key.snk" />
     <EmbeddedResource Include="NetFX\Minimal\minasync.dll" />
     <EmbeddedResource Include="NetFX\Minimal\mincorlib.dll" />
+    <EmbeddedResource Include="NetFX\PortableProfile7\mscorlib.dll" />
+    <EmbeddedResource Include="NetFX\PortableProfile7\System.Runtime.dll" />
```
jaredpar, line 90: Attempted to add a portable7 variant to Basic.Reference.Assemblies but these aren't available through NuPkg anywhere. Decided to just use them directly as resources.

src/Compilers/Test/Resources/Core/Microsoft.CodeAnalysis.Compiler.Test.Resources.csproj
```diff
@@ -85,11 +86,16 @@
...
+    <EmbeddedResource Include="NetFX\WinRt\System.Runtime.WindowsRuntime.dll" />
+    <EmbeddedResource Include="NetFX\WinRt\System.Runtime.WindowsRuntime.UI.Xaml.dll" />
```
jaredpar, line 92: These two references are available as NuPkg and in theory could be added to the extra sections of Basic.Reference.Assemblies.Net461. Unfortunately, there is a substantial difference between how idiomatic `net461` WinRt projects build using these references and how our tests build. The tests do not use the standard reference versions of these DLLs and using the standard references causes the tests to break. Decided to keep the existing DLLs to maintain test behavior.

src/Compilers/Test/Resources/Core/Microsoft.CodeAnalysis.Compiler.Test.Resources.csproj
```diff
@@ -15,6 +15,7 @@
   </PropertyGroup>
   <ItemGroup Label="Project References">
     <ProjectReference Include="..\..\..\Core\Portable\Microsoft.CodeAnalysis.csproj" />
+    <PackageReference Include="stdole" IncludeAssets="none" PrivateAssets="all" GeneratePathProperty="true" />
```
jaredpar, line 18: This is not a .NET framework NuPkg hence redist directly within our repo is a bit suspect. Decided on the embed the NuPkg resource approach.

src/Compilers/Test/Core/TestBase.cs
```diff
@@ -186,11 +185,6 @@ public virtual void Dispose()
             LazyThreadSafetyMode.PublicationOnly);
         public static MetadataReference MscorlibRef => s_mscorlibRef.Value;
 
-        private static readonly Lazy<MetadataReference> s_mscorlibRefPortable = new Lazy<MetadataReference>(
-            () => AssemblyMetadata.CreateFromImage(ProprietaryTestResources.v4_0_30319.mscorlib_portable).GetReference(display: "mscorlib.v4_0_30319.portable.dll"),
-            LazyThreadSafetyMode.PublicationOnly);
-        public static MetadataReference MscorlibRefPortable => s_mscorlibRefPortable.Value;
```
jaredpar, line 192: This reference was primarily used for situations in the compiler where we needed a `System.String` type that lacked the `GetEnumerator` method. This setup for `String` is also present in the PortableProfile7 API (though corelib is `System.Runtime.dll` there). Moved compiler tests over to use that.

This was used in the symbol finder tests in the IDE but that doesn't appear to be anything specific to the portable library. Instead it was used because the symbol finder tests need to deal with retargeting between different core libraries. Flipped those to mostly use `net40` and `net461` mscorlib. 
 
https://vstfdevdiv.corp.microsoft.com/DevDiv2/DevDiv/_workitems/edit/667616

jaredpar:
@dotnet/roslyn-compiler PTAL

src/Compilers/Test/Core/TestHelpers.cs
```diff
@@ -146,5 +147,30 @@ public static string NormalizeNewLines(XCData data)
...
+
+            var bytes = new byte[input.Length >> 1];
+            for (var i = 0; i < bytes.Length; i++)
+            {
+                bytes[i] = parseByte(input.Slice(i << 1, 2), NumberStyles.HexNumber);
+            }
+
+            return ImmutableCollectionsMarshal.AsImmutableArray(bytes);
+
+            byte parseByte(ReadOnlySpan<char> input, NumberStyles numberStyle)
```
jjonescz, line 166: static?

src/Compilers/Test/Core/Mocks/Silverlight.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    public static byte[] Mscorlib => s_tuple.Value.Item1;
+
+    public static byte[] System => s_tuple.Value.Item2;
+
+    private static (byte[], byte[]) BuildImages()
+    {
+        const string corlibExtraCode = """
+            using System;
+            using System.Reflection;
```
jjonescz, line 41: Nit: these usings look unnecessary. Maybe adding VerifyDiagnostics to the compilations we create here would catch that?

jjonescz (approved):

PR #75049, opened 2024-09-11
Enable NuGet audit

jaredpar:
This enables NuGet audit using the audit sources from nuget.org. This is the standard that is being pushed into all of our repositories. The end result is that we will get early warning on CVE and won't need to wait until CG scans on our scheduled builds reveal them.

I've set this up so that it warns not errors. That is deliberate because the audit warnings are asynchronous. Setting them up as errors would mean that builds, which were passing one minute, could start failing the next t. A warning here feels like the right balance but we can experiment and see how it goes.

Related https://github.com/dotnet/arcade/issues/15019

### DIFF
--- a/eng/targets/Settings.props
+++ b/eng/targets/Settings.props
@@ -16,6 +16,14 @@
     <!-- TODO: https://github.com/dotnet/roslyn/issues/71667 -->
     <NoWarn>$(NoWarn);NU1507</NoWarn>
 
+    <!--
+      Do not error on NuGet audit warnings. These are warnings that are introduced asynchronously as
+      new vulnerabilities are discovered. Having these as errors means that CI would pass on on job,
+      and then potentially fail on the next. This is meant to be an early warning system, not a 
+      random CI failure generator.
+    -->
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
+
     <CommonExtensionInstallationRoot>CommonExtensions</CommonExtensionInstallationRoot>
     <LanguageServicesExtensionInstallationFolder>Microsoft\VBCSharp\LanguageServices</LanguageServicesExtensionInstallationFolder>
 
--- a/src/Compilers/Test/Core/TestHelpers.cs
+++ b/src/Compilers/Test/Core/TestHelpers.cs
@@ -163,7 +163,7 @@ namespace Roslyn.Test.Utilities
 
             return ImmutableCollectionsMarshal.AsImmutableArray(bytes);
 
-            byte parseByte(ReadOnlySpan<char> input, NumberStyles numberStyle)
+            static byte parseByte(ReadOnlySpan<char> input, NumberStyles numberStyle)
             {
 #if NET
                 return byte.Parse(input, numberStyle);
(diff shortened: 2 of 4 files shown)


eng/targets/Settings.props
```diff
@@ -16,6 +16,14 @@
     <!-- TODO: https://github.com/dotnet/roslyn/issues/71667 -->
     <NoWarn>$(NoWarn);NU1507</NoWarn>
 
+    <!--
+      Do not error on NuGet audit warnings. These are warnings that are introduced asynchronously as
+      new vulnerabilities are discovered. Having these as errors means that CI would pass on on job,
+      and then potentially fail on the next. This is meant to be an early warning system, not a 
+      random CI failure generator.
+    -->
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
```
jjonescz, line 25: Looks like this doesn't work, CI is failing with NU1903 :)

jjonescz (dismissed):

eng/targets/Settings.props
```diff
@@ -16,6 +16,14 @@
     <!-- TODO: https://github.com/dotnet/roslyn/issues/71667 -->
     <NoWarn>$(NoWarn);NU1507</NoWarn>
 
+    <!--
+      Do not error on NuGet audit warnings. These are warnings that are introduced asynchronously as
+      new vulnerabilities are discovered. Having these as errors means that CI would pass on on job,
```
jcouv, line 21: >on on [](http://example.com/codeflow?start=90&length=5)

typo: on on

src/Compilers/Test/Core/TestHelpers.cs
```diff
@@ -163,7 +163,7 @@ public static ImmutableArray<byte> HexToByte(ReadOnlySpan<char> input)
 
             return ImmutableCollectionsMarshal.AsImmutableArray(bytes);
 
-            byte parseByte(ReadOnlySpan<char> input, NumberStyles numberStyle)
+            static byte parseByte(ReadOnlySpan<char> input, NumberStyles numberStyle)
```
jcouv, line 166: I don't have any object, but this change (and the one in Silverlight.cs) are unexpected in this PR. Just want to confirm these were intentional

jcouv (commented):
Done with review pass (iteration 1)

eng/targets/Settings.props
jaredpar: Yes I know. It's quite frustrating. Trying to understand why that is happening.

src/Compilers/Test/Core/TestHelpers.cs
jaredpar: It was intentional. It's a follow up comment from a previous PR that I forgot to merge.

jaredpar:
Reverting to draft as there were a number of audit warnings. There is an up to date check [issue][issue] in NuGet / MSBuild that caused me to see no errors locally and believe we were clean. This PR showed me there was still work left. Dealing with that locally and will update this PR In a bit. 

[issue]: https://github.com/NuGet/Home/issues/13780

eng/targets/Settings.props
```diff
@@ -16,6 +16,14 @@
...
+      and then potentially fail on the next. This is meant to be an early warning system, not a 
+      random CI failure generator.
```
jaredpar, line 23:
```suggestion
      and then potentially fail on the next. This is meant to be an early warning system and not
      a way to cause friction in PRs.
```

jaredpar:
Closing stale draft PR.

Issue #73563, opened 2024-05-17
Assert fires in `EnsureMetadataVirtual` 


AlekseyTs:
Assert fires in `EnsureMetadataVirtual` when the tests below are executed with `CompilationExtensions.EnableVerifyUsedAssemblies` set to true. 
```
            public void EnsureMetadataVirtual()
            {
                // ACASEY: This assert is here to check that we're not mutating the value of IsMetadataVirtual after
                // someone has consumed it.  The best practice is to not access IsMetadataVirtual before ForceComplete
                // has been called on all SourceNamedTypeSymbols.  If it is necessary to do so, then you can pass
                // ignoreInterfaceImplementationChanges: true, but you must be conscious that seeing "false" may not
                // reflect the final, emitted modifier.
                Debug.Assert(!IsMetadataVirtualLocked);
                if ((_flags & IsMetadataVirtualBit) == 0)
                {
                    ThreadSafeFlagOperations.Set(ref _flags, IsMetadataVirtualBit);
                }
            }
```

Call stack:
```
>	Microsoft.CodeAnalysis.CSharp.dll!Microsoft.CodeAnalysis.CSharp.Symbols.SourceMemberMethodSymbol.Flags.EnsureMetadataVirtual() Line 272	C#
 	Microsoft.CodeAnalysis.CSharp.dll!Microsoft.CodeAnalysis.CSharp.Symbols.SourceMemberMethodSymbol.EnsureMetadataVirtual() Line 569	C#
 	Microsoft.CodeAnalysis.CSharp.dll!Microsoft.CodeAnalysis.CSharp.Symbols.SourceMemberContainerTypeSymbol.SynthesizeInterfaceMemberImplementation(Microsoft.CodeAnalysis.CSharp.Symbols.TypeSymbol.SymbolAndDiagnostics implementingMemberAndDiagnostics, Microsoft.CodeAnalysis.CSharp.Symbol interfaceMember) Line 1852	C#
 	Microsoft.CodeAnalysis.CSharp.dll!Microsoft.CodeAnalysis.CSharp.Symbols.SourceMemberContainerTypeSymbol.ComputeInterfaceImplementations(Microsoft.CodeAnalysis.CSharp.BindingDiagnosticBag diagnostics, System.Threading.CancellationToken cancellationToken) Line 179	C#
[...]

PR #75066, opened 2024-09-12
Adjust IsMetadataVirtual assertion in MethodToClassRewriter

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/73563

`IsMetadataVirtual` previously could either:
- give a final answer but this required it to be called **after** `ForceComplete` (which may initialize it in `SourceMemberContainerTypeSymbolSynthesizeInterfaceMemberImplementation`)
- give a partial answer (by passing `ignoreInterfaceImplementationChanges`)

Previously, in a scenario where `comp2` references `comp1` as a compilation reference and we emit `comp2` and then emit `comp1`:
1. emitting `comp2` called `MethodToClassRewriter.VisitCall` which called `IsMetadataVirtual` on a not-yet-completed symbol from `comp1`. The read from `IsMetadataVirtual` set the "locked" flag.
2. emitting `comp1` completed that symbol from `comp1` which would call `EnsureMetadataVirtual`. This asserted since the "locked" flag was already set

Now, the assertion in `MethodToClassRewriter.VisitCall` eagerly forces the completion of the symbol from `comp1`.


src/Compilers/CSharp/Portable/Emitter/Model/MethodSymbolAdapter.cs
```diff
@@ -658,12 +658,24 @@ internal virtual bool IsMetadataFinal
         /// signature).
         /// WARN WARN WARN: We won't have a final value for this until declaration
         /// diagnostics have been computed for all <see cref="SourceMemberContainerTypeSymbol"/>s, so pass
-        /// ignoringInterfaceImplementationChanges: true if you need a value sooner
+        /// option: <see cref="IsMetadataVirtualOption.IgnoreInterfaceImplementationChanges"/> true if you need a value sooner
```
AlekseyTs, line 661: >true [](http://example.com/codeflow?start=95&length=4)

Delete? This looks like an artifact from the old state of the comment.

AlekseyTs:
It looks like a couple of tests in `RefStructInterfacesTests.cs` were disabled because of this issue. I think we should enable them in this PR.

AlekseyTs:

AlekseyTs (approved):
LGTM (commit 2)

jcouv:

jcouv:

333fred (approved):

Issue #74362, opened 2024-07-11
Debug Assert error in MethodToStateMachineRewriter


bernd5:
**Version Used**: 

**Steps to Reproduce**:
Compile the following code with debug asserts enabled:

```csharp
using System;
using System.Collections.Generic;
using System.Threading.Tasks;

public class App
{
    public static void Main()
    {
        foreach (var x in new X().ValuesViaLocalFunction)
        {
            Console.WriteLine(x);
        }
    }
}


public class X
{
    public async IAsyncEnumerable<int> GetValues(bool extra = true)
    {
       yield return 0;
    }

    public IEnumerable<int> ValuesViaLocalFunction
    {
        get
        {
            foreach (var b in Do(true).ToBlockingEnumerable())
            {
                yield return b;
            }

            async IAsyncEnumerable<int> Do(bool extra)
            {
                await foreach (var v in GetValues(true))
                {
                    yield return v;
                }
            }
        }
    }
}
```

**Expected Behavior**:
No assert triggered.

**Actual Behavior**:
Debug assert is triggered at: [Link](https://github.com/dotnet/roslyn/blob/7241b8522990905c491dee920d32dc09c345920d/src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/MethodToStateMachineRewriter.cs#L220)

The reason is that `AsyncExceptionHandlerRewriter` does not care about `SyntheticBoundNodeFactory.Syntax`. The syntax needs to be updated to the try statement.
See: [Link](https://github.com/dotnet/roslyn/blob/7241b8522990905c491dee920d32dc09c345920d/src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncExceptionHandlerRewriter.cs#L183)

AlekseyTs:
Assigning back to Julien since he already has a PR with the fix

PR #75070, opened 2024-09-12
Keep try syntax in async exception rewriting

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/74362

The async exception handler rewriter (which is only involved for an `await` inside a `catch` or `finally`, such as awaiting the disposal in an `await foreach`) produced a bound node for the `try` whose syntax was not recognized by the `SyntaxBindingUtilities.BindsToTryStatement` assertion.
In the problematic scenario, the syntax was the block of the accessor containing the local function with `await foreach`.
The fix is to synthesize the nodes for this `try` with more specific syntax, such as the `await foreach` statement.

The `AsyncInCatchFilterLifted` test was affected and so I updated `IsClosureScope`

### DIFF
--- a/src/Compilers/CSharp/Test/Emit2/Semantics/ParamsCollectionTests.cs
+++ b/src/Compilers/CSharp/Test/Emit2/Semantics/ParamsCollectionTests.cs
@@ -58,6 +58,11 @@ namespace System.Runtime.CompilerServices
             VerifyParams(parameter.GetSymbol<ParameterSymbol>(), isParamArray: isParamArray, isParamCollection: isParamCollection);
         }
 
+        private static string ExpectedOutput(string output)
+        {
+            return ExecutionConditionUtil.IsMonoOrCoreClr ? output : null;
+        }
+
         [Fact]
         public void Span()
         {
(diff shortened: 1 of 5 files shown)


src/Compilers/CSharp/Test/Emit2/Semantics/ParamsCollectionTests.cs
```diff
@@ -58,11 +58,6 @@ private static void VerifyParams(IParameterSymbol parameter, bool isParamArray =
             VerifyParams(parameter.GetSymbol<ParameterSymbol>(), isParamArray: isParamArray, isParamCollection: isParamCollection);
         }
 
-        private static string ExpectedOutput(string output)
```
AlekseyTs, line 61: >ExpectedOutput [](http://example.com/codeflow?start=30&length=14)

I do not think moving this method to a shared location with this name an implementation is appropriate. The name is too general/unspecific and the logic is not applicable in every test.

src/Compilers/Test/Utilities/CSharp/CSharpTestBase.cs
```diff
@@ -2625,5 +2625,10 @@ public interface IAsyncDisposable
     }
 }
 ";
+
+        internal static string ExpectedOutput(string output)
```
AlekseyTs, line 2629: >ExpectedOutput [](http://example.com/codeflow?start=31&length=14)

The name doesn't reflect the purpose of the method, therefore, I do not think it should be used for a global/shared helper.

AlekseyTs:

AlekseyTs (approved):
LGTM (commit 2)

jcouv:

333fred (approved):

Issue #71570, opened 2024-01-10
Semantic snippets doesn't fallback to VS options properly


DoctorKrolic:
**Version Used**:
Latest `main`

**Description**:
Since semantic snippets are controlled by roslyn completion providers, they are not only suggested only in valid contexts, but can also use other APIs to benifit their users. One of such features is respecting user's code style options. For instance, `for` and `forr` snippets try to respect preference for using `var` vs `int` in loop header. We can easily test this by adding an `.editorconfig` document with the desired options to the test workspace and verify our results. And [we have such tests in place](https://github.com/dotnet/roslyn/blob/0f7fba61e5be3354240736997edb4a0e622e9d9d/src/EditorFeatures/CSharpTest/Completion/CompletionProviders/Snippets/CSharpForSnippetCompletionProviderTests.cs#L230-L264). However, `.editorconfig` is not the only way to control code style settings. When it is not present, snippets should correctly fallback to VS options and respect them instead. Unfortunatelly, currently this is not the case.

**Steps to Reproduce**:
1. In you VS options set code style option `'var' preference/For built-in types` to `Prefer 'var'`:
![devenv_7IKZ7HTbaJ](https://github.com/dotnet/roslyn/assets/70431552/95d379b3-9688-401d-8e3d-9b634313ee9f)
2. Make sure you don't have an `.editorconfig` file in your solution, so it cannot override VS options
3. Execute `for` snippet where it is allowed

**Expected Behavior**:
```cs
for (var i = 0; i < length; i++)
{
    $$
}
```

**Actual Behavior**:
![devenv_eeenXLtttk](https://github.com/dotnet/roslyn/assets/70431552/887cd126-b34e-413e-9663-93ff10791d8e)

**Additional info**:
I've debugged this problem and the cause seems to be here:
https://github.com/dotnet/roslyn/blob/bacd78c504f5dcb85693e2968405c51fe5ff3d3e/src/Features/Core/Portable/Snippets/SnippetProviders/AbstractSnippetProvider.cs#L163-L165
[...]

PR #75073, opened 2024-09-12
Verify that semantic snippets obey IDE preferences

DoctorKrolic:
Closes https://github.com/dotnet/roslyn/issues/71570

The bug itself is no longer reproducible and I want to ensure we are regression-proof, so added a few integration tests


src/VisualStudio/IntegrationTest/New.IntegrationTests/InProcess/EditorInProcess.cs
```diff
@@ -687,6 +687,22 @@ public async Task DismissCompletionSessionsAsync(CancellationToken cancellationT
         broker.DismissAllSessions(view);
     }
 
+    public async Task WaitForCompletionSessionsAsync(CancellationToken cancellationToken)
```
DoctorKrolic, line 690: @sharwell After `WaitForCompletionSetAsync` the `broker.IsCompletionActive(view)` returns `true`, while the completion `view` in the IDE might not shown yet and `broker.GetSessions(view)` returns an empty collection. Since I was hitting this pretty much 100% of times locally I added this helper. The behavior is either related to some async completion stuff or is a platform bug. Can you PTAL and, if needed, create an internal tracking issue for this? Also, shouldn't `WaitForCompletionSetAsync` correctly wait for the session to start in the first place?

DoctorKrolic: Also I didn't add calls to this helper in other completion-related tests, bug I suspect they are susceptible to this behavior and may be flaky due to this thing as well.

DoctorKrolic:
> Does this `foreach` snippet work for you with passing explicit types, please?

Nop, unfortunatelly, that is not implemented yet. Currently only `for` and `forr` snippets consider user's `var` preferences, since it was easy to do in the first place. I had some ideas about improving `foreach` snippet before and as a result I've proposed [this compiler API](https://github.com/dotnet/roslyn/issues/74998) that would be very helpful for `foreach` snippet and other collection-related tasks. I am waiting for the conclusion on that proposal before I touch `foreach` snippet

DoctorKrolic:
@akhera99 @sharwell PTAL

DoctorKrolic:
@akhera99 @sharwell PTAL

DoctorKrolic:
@akhera99 @sharwell PTAL

akhera99 (approved):
Signing off on the snippets related work. @sharwell for the `WaitForCompletionSetAsync` work.

src/VisualStudio/IntegrationTest/New.IntegrationTests/InProcess/EditorInProcess.cs
sharwell: The code which connects the editor behavior to `FeatureAttribute.CompletionSet` is here:

https://github.com/dotnet/roslyn/blob/55a809f91cae01bf1e6c635ee91faae6b0af23fe/src/VisualStudio/IntegrationTest/TestSetup/AsyncCompletionTracker.cs

The call to `GetSelectedItemAsync` is _intended_ to serve this purpose, but apparently is not doing so. That call will need to be updated to correctly invoke the operation that ensures the view is rendered.

src/VisualStudio/IntegrationTest/New.IntegrationTests/InProcess/EditorInProcess.cs
```diff
@@ -687,6 +687,22 @@ public async Task DismissCompletionSessionsAsync(CancellationToken cancellationT
...
+    {
+        await JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+
+        await WaitForCompletionSetAsync(cancellationToken);
+
+        var view = await GetActiveTextViewAsync(cancellationToken);
+        var broker = await GetComponentModelServiceAsync<ICompletionBroker>(cancellationToken);
+        var sessions = broker.GetSessions(view);
+        while (sessions.Count == 0)
```
sharwell, line 699: 📝 Note that it is generally not acceptable to wait in this fashion during integration tests. It is both fundamentally flaky since it doesn't account for the possibility that more than one operation complete during the wait, and slow since it fails to immediately proceed with the test code when the view rendering is complete.

src/VisualStudio/IntegrationTest/New.IntegrationTests/InProcess/EditorInProcess.cs
DoctorKrolic: Can you please clarify what changes do we need to make? Or is it a platform issue (`ISelectedItemProvider` is a VS type)?

PR #75079, opened 2024-09-12
Test and document Span breaking changes

jjonescz:

### DIFF
--- a/docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
+++ b/docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
@@ -1,0 +1,43 @@
+# This document lists known breaking changes in Roslyn after .NET 8 all the way to .NET 9.
+
+## `Span` and `ReadOnlySpan` overloads are applicable in more scenarios in C# 14 and newer
+
+***Introduced in Visual Studio 2022 version 17.13***
+
+C# 14 introduces new [built-in span conversions and type inference rules](https://github.com/dotnet/csharplang/issues/7905).
+This means that different overloads might be chosen compared to C# 13, and sometimes an ambiguity compile-time error
+might be raised because a new overload is applicable but there is no single best overload.
+
+The following example shows some ambiguities and possible workarounds.
+Note that another workaround is for API authors to use `OverloadResolutionPriorityAttribute`.
+
+```cs
+var x = new long[] { 1 };
+Assert.Equal([2], x); // previously Assert.Equal<T>(T[], T[]), now ambiguous with Assert.Equal<T>(ReadOnlySpan<T>, Span<T>)
+Assert.Equal([2], x.AsSpan()); // workaround
+
+var y = new int[] { 1, 2 };
+var s = new ArraySegment<int>(x, 1, 1);
+Assert.Equal(y, s); // previously Assert.Equal<T>(T, T), now ambiguous with Assert.Equal<T>(Span<T>, Span<T>)
+Assert.Equal(y.AsSpan(), s); // workaround
+```
+
+A `Span` overload might be chosen in C# 14 where an `IEnumerable` overload was chosen in C# 13,
+and that can lead to `ArrayTypeMismatchException`s at runtime if you are using covariant arrays:
+
+```cs
+string[] s = new[] { "a" };
+object[] o = s; // array variance
+
+C.R(o); // wrote 1 previously, now crashes in Span<T> constructor with ArrayTypeMismatchException
+C.R(o.AsEnumerable()); // workaround
+
+static class C
+{
+    public static void R<T>(IEnumerable<T> e) => Console.Write(1);
+    public static void R<T>(Span<T> s) => Console.Write(2);
+    // another workaround:
+    [OverloadResolutionPriority(1)]
+    public static void R<T>(ReadOnlySpan<T> s) => Console.Write(3);
+}
+```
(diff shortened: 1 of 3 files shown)


docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -0,0 +1,43 @@
+# This document lists known breaking changes in Roslyn after .NET 8 all the way to .NET 9.
```
cston, line 1: >.NET 8 all the way to .NET 9 [](http://example.com/codeflow?start=61&length=28)

Should this be ".NET 9 ... .NET 10"?

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -0,0 +1,43 @@
...
+
+## `Span` and `ReadOnlySpan` overloads are applicable in more scenarios in C# 14 and newer
```
cston, line 3: >`Span` and `ReadOnlySpan` [](http://example.com/codeflow?start=3&length=25)

Consider including type parameters: "`Span<T>` and `ReadOnlySpan<T>`".

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -0,0 +1,43 @@
...
+Assert.Equal([2], x); // previously Assert.Equal<T>(T[], T[]), now ambiguous with Assert.Equal<T>(ReadOnlySpan<T>, Span<T>)
+Assert.Equal([2], x.AsSpan()); // workaround
+
+var y = new int[] { 1, 2 };
+var s = new ArraySegment<int>(x, 1, 1);
+Assert.Equal(y, s); // previously Assert.Equal<T>(T, T), now ambiguous with Assert.Equal<T>(Span<T>, Span<T>)
+Assert.Equal(y.AsSpan(), s); // workaround
+```
+
+A `Span` overload might be chosen in C# 14 where an `IEnumerable` overload was chosen in C# 13,
```
cston, line 25: >`Span` [](http://example.com/codeflow?start=2&length=6)

Consider including the type parameter: `Span<T>`.

cston, line 25: >where an `IEnumerable` [](http://example.com/codeflow?start=43&length=22)

Consider changing this to `IEnumerable<T>`, or generalize this to "a interface implemented by `T[]`, such as `IEnumerable<T>`".

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -0,0 +1,43 @@
...
+## `Span` and `ReadOnlySpan` overloads are applicable in more scenarios in C# 14 and newer
+
+***Introduced in Visual Studio 2022 version 17.13***
+
+C# 14 introduces new [built-in span conversions and type inference rules](https://github.com/dotnet/csharplang/issues/7905).
+This means that different overloads might be chosen compared to C# 13, and sometimes an ambiguity compile-time error
+might be raised because a new overload is applicable but there is no single best overload.
+
+The following example shows some ambiguities and possible workarounds.
+Note that another workaround is for API authors to use `OverloadResolutionPriorityAttribute`.
```
cston, line 12: >. [](http://example.com/codeflow?start=92&length=1)

"or compile with `-langversion:13` or earlier"?

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -0,0 +1,43 @@
...
+Assert.Equal([2], x.AsSpan()); // workaround
+
+var y = new int[] { 1, 2 };
+var s = new ArraySegment<int>(x, 1, 1);
+Assert.Equal(y, s); // previously Assert.Equal<T>(T, T), now ambiguous with Assert.Equal<T>(Span<T>, Span<T>)
+Assert.Equal(y.AsSpan(), s); // workaround
+```
+
+A `Span` overload might be chosen in C# 14 where an `IEnumerable` overload was chosen in C# 13,
+and that can lead to `ArrayTypeMismatchException`s at runtime if you are using covariant arrays:
```
cston, line 26: >`ArrayTypeMismatchException`s at runtime if you are using covariant arrays [](http://example.com/codeflow?start=21&length=74)

Minor: Perhaps avoid the plural with "an `ArrayTypeMismatchException` at runtime if the array is covariant"

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
jjonescz: Hm, that's true for the whole section, not just these ambiguity examples. It should be implied by "C# 14 introduces...". It feels weird to suggest this workaround explicitly, we want people to upgrade langversions, not stay on old versions.

333fred (approved):

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -0,0 +1,43 @@
+# This document lists known breaking changes in Roslyn after .NET 9 all the way to .NET 10.
```
jcouv, line 1: I always get confused on this so feel free to correct me.
Since the breaking changes listed below are shipping as part of .NET 9, I would expect them to be in the exist document that covers the `].NET 8, .NET 9]` window.

333fred: You only see them when you turn on C# 14, which will be part of .NET 10.

jcouv: Ah, missed that. Thanks

Issue #16929, opened 2017-02-03
Incorrect System.Reflection.*Attributes checks in some parts of the code


0xd4d:
**Version Used**: Current master branch

Eg. `TypeAttributes` is a mix of flags and enums. To check if a type is eg. non-nested public, one has to first use `VisibilityMask` and then compare the result with eg. `Public`, but some of the code treat `Public` as a flag and use code like `(f & TypeAttributes.Public) == TypeAttributes.Public`.

The following attributes and sub-enums are sometimes used incorrectly:

```
TypeAttributes
	VisibilityMask (NotPublic, Public, NestedPublic, ...)
	ClassSemanticsMask (Class, Interface)

MethodImplAttributes
	CodeTypeMask (IL, ...)
	ManagedMask (Managed, Unmanaged)

MethodAttributes
	MemberAccessMask (PrivateScope, Private, ...)

ManifestResourceAttributes
	VisibilityMask (Public, Private)

FieldAttributes
	FieldAccessMask (PrivateScope, Private, ...)
```

akoeplinger:
In https://github.com/dotnet/roslyn/pull/75081 it was decided to not replace the ManifestResourceAttributes/MethodImplAttributes/TypeAttributes instances since they are not a real issue.

The rest should be good now.

PR #75081, opened 2024-09-12
Remove buggy IsPublic method from TypeAttributesExtensions

akoeplinger:
When checking visibility we need to use the VisibilityMask because TypeAttributes is a funky flags enum that uses values that aren't binary exclusive (unlike normal flags enums) so e.g. `NestedPrivate` (0x3) would be interpreted as `Public` (0x1). See https://learn.microsoft.com/en-us/dotnet/api/system.reflection.typeattributes?view=net-8.0#examples

We got lucky because the IsPublic extension method isn't actually used anywhere so this was never hit. Remove the method.

~~The same applies to checking for interface, but TypeAttributes.ClassSemanticsMask is the same value as TypeAttributes.Interface so it happens to work, but it's still better to use the canonical code for clarity.~~

~~Also found a couple other *Attributes that were described in #16929 and fixed them too. Those are essentially no-ops too.~~

Fixes https://github.com/dotnet/roslyn/issues/16929

/cc @tmat

### DIFF
--- a/src/Compilers/Core/Rebuild/CompilationOptionsReader.cs
+++ b/src/Compilers/Core/Rebuild/CompilationOptionsReader.cs
@@ -276,7 +276,7 @@ namespace Microsoft.CodeAnalysis.Rebuild
                 var content = new byte[length];
                 Marshal.Copy(new IntPtr(contentPtr), content, 0, length);
 
-                var isPublic = (resource.Attributes & ManifestResourceAttributes.Public) != 0;
+                var isPublic = (resource.Attributes & ManifestResourceAttributes.VisibilityMask) == ManifestResourceAttributes.Public;
                 var description = new ResourceDescription(name, dataProvider: () => new MemoryStream(content), isPublic);
                 return description;
             }).ToArray();
--- a/src/Compilers/Core/Rebuild/CompilationOptionsReader.cs
+++ b/src/Compilers/Core/Rebuild/CompilationOptionsReader.cs
@@ -276,7 +276,7 @@ namespace Microsoft.CodeAnalysis.Rebuild
                 var content = new byte[length];
                 Marshal.Copy(new IntPtr(contentPtr), content, 0, length);
 
-                var isPublic = (resource.Attributes & ManifestResourceAttributes.VisibilityMask) == ManifestResourceAttributes.Public;
+                var isPublic = (resource.Attributes & ManifestResourceAttributes.Public) != 0;
                 var description = new ResourceDescription(name, dataProvider: () => new MemoryStream(content), isPublic);
                 return description;
             }).ToArray();
--- a/src/Compilers/Test/Core/Platform/Custom/MetadataSignatureHelper.cs
+++ b/src/Compilers/Test/Core/Platform/Custom/MetadataSignatureHelper.cs
@@ -323,7 +323,7 @@ namespace Roslyn.Test.Utilities
 
             sb.Append(codeType);
             sb.Append(' ');
-            sb.Append((attributes & MethodImplAttributes.Unmanaged) == MethodImplAttributes.Unmanaged ? "unmanaged" : "managed");
+            sb.Append((attributes & MethodImplAttributes.ManagedMask) == MethodImplAttributes.Unmanaged ? "unmanaged" : "managed");
 
             if ((attributes & MethodImplAttributes.PreserveSig) != 0)
                 sb.Append(" preservesig");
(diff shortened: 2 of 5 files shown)


akoeplinger:
Turns out there was already an old issue about this (https://github.com/dotnet/roslyn/issues/16929) and we just had a couple instances left so I fixed them too.

tmat:
@dotnet/roslyn-compiler

333fred:
Presumably, we should be able to observe differences in these scenarios with tests; given that there are no failures here, it seems like we have a test gap and I'd like to see some tests covering it.

AlekseyTs:
Alternatively, if we are fixing internal dead code (never used), the I would prefer removing it rather than changing it.

akoeplinger:
yeah only the IsPublic change is actually a bug but the method is never used, I can remove it.

the other cases only work because the *Mask value is the same as the enum entry we're checking. in theory we could leave it as is but I think it's worthwhile just to not spread a bad pattern.

src/Compilers/Core/Rebuild/CompilationOptionsReader.cs
```diff
@@ -276,7 +276,7 @@ where PdbReader.GetGuid(cdi.Kind) == EmbeddedSourceGuid
                 var content = new byte[length];
                 Marshal.Copy(new IntPtr(contentPtr), content, 0, length);
 
-                var isPublic = (resource.Attributes & ManifestResourceAttributes.Public) != 0;
+                var isPublic = (resource.Attributes & ManifestResourceAttributes.VisibilityMask) == ManifestResourceAttributes.Public;
```
333fred, line 279: This doesn't appear to be equivalent. Yes, the only 2 flags are `Public` and `Private`, so it's arguable that `Public | Private` shouldn't be true here, but I'd prefer to keep this as is.

src/Compilers/Test/Core/Platform/Custom/MetadataSignatureHelper.cs
```diff
@@ -323,7 +323,7 @@ public static StringBuilder AppendMethodImplAttributes(StringBuilder sb, MethodI
 
             sb.Append(codeType);
             sb.Append(' ');
-            sb.Append((attributes & MethodImplAttributes.Unmanaged) == MethodImplAttributes.Unmanaged ? "unmanaged" : "managed");
+            sb.Append((attributes & MethodImplAttributes.ManagedMask) == MethodImplAttributes.Unmanaged ? "unmanaged" : "managed");
```
333fred, line 326: The current pattern doesn't actually appear to be a problem? Whether something is managed or not is just 1 bit.

src/Compilers/Core/Rebuild/CompilationOptionsReader.cs
akoeplinger: why do you think this is not equivalent? the mask nulls the other bits (which don't exist in this enum ..yet?) so we can then directly compare the entry.

I'm not saying this is nice but it's the pattern these *Mask APIs are using

src/Compilers/Test/Core/Platform/Custom/MetadataSignatureHelper.cs
akoeplinger: right, same here. I don't have a particularly strong opinion other than not spreading a bad pattern.

src/Compilers/Core/Rebuild/CompilationOptionsReader.cs
333fred: It's not equivalent exactly as I said: if `resource.Attributes` is `Public | Private`, this will now be false, where the existing code would say true.

akoeplinger: ah I misread what you said, you meant that before `Public | Private` returned true here while now it doesn't.

that's true but I'd find it kind of strange for this usage

333fred:
In talking with @AlekseyTs, we'd prefer to leave most of the code as is, rather than changing it, given that these flags aren't likely to change and they work as is. For the bad `IsPublic` check, removing it is fine.

akoeplinger:
@333fred sure, I updated the PR to just remove the IsPublic method.

333fred (approved):

333fred:
@dotnet/roslyn-compiler @AlekseyTs for a second review.

AlekseyTs (approved):
LGTM (commit 4)

Issue #73905, opened 2024-06-08
Assert failure in `BindMethodBody.assertBindIdentifierTargets()` with `[param:]` attribute on local function


cston:
**Version Used**: e867be9f6f8777b61b00bbd6ab46cd99e5da617e

**Steps to Reproduce**:

Compile the following with a `DEBUG` build:

```csharp
using System;

[AttributeUsage(AttributeTargets.All)]
class A : Attribute
{
    public A(string s) { }
}

class C
{
    event EventHandler E
    {
        add
        {
            [param: A(nameof(p))] void F(int p) { }
        }
        remove
        {
        }
    }
}
```

**Result**:

Debug.Assert() failure:
>	Microsoft.CodeAnalysis.CSharp.dll!Microsoft.CodeAnalysis.CSharp.MethodCompiler.BindMethodBody.__assertBindIdentifierTargets|42_3(Microsoft.CodeAnalysis.CSharp.InMethodBinder inMethodBinder, System.Collections.Concurrent.ConcurrentDictionary<Microsoft.CodeAnalysis.CSharp.Syntax.IdentifierNameSyntax, int> identifierMap, Microsoft.CodeAnalysis.CSharp.BoundNode methodBody, Microsoft.CodeAnalysis.CSharp.BindingDiagnosticBag diagnostics) Line 2143	C#
 	Microsoft.CodeAnalysis.CSharp.dll!Microsoft.CodeAnalysis.CSharp.MethodCompiler.BindMethodBody(Microsoft.CodeAnalysis.CSharp.Symbols.MethodSymbol method, Microsoft.CodeAnalysis.CSharp.TypeCompilationState compilationState, Microsoft.CodeAnalysis.CSharp.BindingDiagnosticBag diagnostics, bool includeInitializersInBody, Microsoft.CodeAnalysis.CSharp.BoundNode initializersBody, bool reportNullableDiagnostics, out Microsoft.CodeAnalysis.CSharp.ImportChain importChain, out bool originalBodyNested, out bool prependedDefaultValueTypeConstructorInitializer, out Microsoft.CodeAnalysis.CSharp.MethodBodySemanticModel.InitialState forSemanticModel) Line 1782	C#
[...]

PR #75100, opened 2024-09-12
Exclude identifiers in misplaced attributes from identifier map

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/73905

For context:
- `buildIdentifierMapOfBindIdentifierTargets` builds a map of identifiers that we expect to bind
- `adjustIdentifierMapIfAny` updates the map to record which identifiers we actually bound
- `assertBindIdentifierTargets` checks the resulting map (already ignored identifiers in presence of binding errors, but now also skips identifiers in presence of warning for misplaced attributes, which causes attributes not to be bound)
- those checks are meant to align with logic in `SynthesizedPrimaryConstructor.GetCapturedParameters`

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
@@ -1651,11 +1651,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                     Binder current = this;
                     while (current is not (null or InMethodBinder { IdentifierMap: not null }))
                     {
-                        if (current is ContextualAttributeBinder)
-                        {
-                            return;
-                        }
-
                         current = current.Next;
                     }
 
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
@@ -1651,6 +1651,11 @@ namespace Microsoft.CodeAnalysis.CSharp
                     Binder current = this;
                     while (current is not (null or InMethodBinder { IdentifierMap: not null }))
                     {
+                        if (current is ContextualAttributeBinder)
+                        {
+                            return;
+                        }
+
                         current = current.Next;
                     }
 
--- a/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
+++ b/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
@@ -2010,6 +2010,8 @@ namespace Microsoft.CodeAnalysis.CSharp
                 }
             }
 
+            // Logic in this lambda is based on Binder.IdentifierUsedAsValueFinder.CheckIdentifiersInNode.childrenNeedChecking.
+            // It can be more permissive (i.e. allow us to dive into more nodes), but should not be more restrictive
             static void addIdentifiers(CSharpSyntaxNode? node, ConcurrentDictionary<IdentifierNameSyntax, int> identifierMap)
             {
                 if (node is null)
@@ -2028,7 +2030,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                             case GotoStatementSyntax { RawKind: (int)SyntaxKind.GotoStatement }:
                             case TypeParameterConstraintClauseSyntax:
                             case AliasQualifiedNameSyntax:
-                            case AttributeListSyntax:
                                 // These nodes do not have anything interesting for us
                                 return false;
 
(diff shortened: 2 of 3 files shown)


src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -2028,6 +2028,7 @@ static void addIdentifiers(CSharpSyntaxNode? node, ConcurrentDictionary<Identifi
                             case GotoStatementSyntax { RawKind: (int)SyntaxKind.GotoStatement }:
                             case TypeParameterConstraintClauseSyntax:
                             case AliasQualifiedNameSyntax:
+                            case AttributeListSyntax:
```
AlekseyTs, line 2031: >case AttributeListSyntax: [](http://example.com/codeflow?start=28&length=25)

I guess I should have added a comment that logic in this lambda is based on ```Binder.IdentifierUsedAsValueFinder.CheckIdentifiersInNode.childrenNeedChecking```. It can be more permissive (i.e. allow us to dive into more nodes), but should not be more restrictive. Since ```childrenNeedChecking``` has the following case:
```
                        case AttributeListSyntax:
                            // References in attributes, if any, are errors
                            // Skip them
                            return false;
```
I am fine with adding a similar case here with the same comment/explanation. Given the sated goal "Predict all identifiers in the body that should go through Binder.BindIdentifier method", putting it into the bucket of "uninteresting" nodes is somewhat misleading because it contains identifiers that go through Binder.BindIdentifier.

At the same time, I would prefer an alternative approach to suppressing the assert. The one that will allow ```assertBindIdentifierTargets``` to be closer to achieving the stated goal (quoted above). 

```assertBindIdentifierTargets``` already disables asserts in presence of errors since there is no guaranty that we bind all the identifiers that are supposed to be bound. The scenarios that we are dealing with are actually error scenarios, we skip binding the attributes, but report a warning instead. That is why ```assertBindIdentifierTargets``` fails to disable the check. I suggest to fix this flaw instead. I'll make a specific suggestion at the relevant location.

**Comment by AlekseyTs** [comment]
>                                     }

It looks like adding the following code after this block suppresses the assert:
```
                                    if (id.Ancestors(ascendOutOfTrivia: false).OfType<AttributeListSyntax>().Any() &&
                                        diagnostics.DiagnosticBag!.AsEnumerable().Any(d => d.Code == (int)ErrorCode.WRN_AttributeLocationOnBadDeclaration))
                                    {
                                        continue;
                                    }
```

---
Refers to: src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs:2136 in a3ad201. [](commit_id = a3ad201c5ad78ae8f1229417d4d44d00688f1658, deletion_comment = False)

**Comment by AlekseyTs** [comment]
>                                 if (flags == 1)

It looks like adding the following code at the end of body of this `if` suppresses the assert:
```
if (id.Ancestors(ascendOutOfTrivia: false).OfType<AttributeListSyntax>().Any() &&
    diagnostics.DiagnosticBag!.AsEnumerable().Any(d => d.Code == (int)ErrorCode.WRN_AttributeLocationOnBadDeclaration))
{
    continue;
}
```

---
Refers to: src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs:2112 in a3ad201. [](commit_id = a3ad201c5ad78ae8f1229417d4d44d00688f1658, deletion_comment = False)

File: src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs [hunk 2]
```diff
@@ -1651,6 +1651,11 @@ void adjustIdentifierMapIfAny(SimpleNameSyntax node, bool invoked)
                     Binder current = this;
                     while (current is not (null or InMethodBinder { IdentifierMap: not null }))
                     {
+                        if (current is ContextualAttributeBinder)
```
AlekseyTs, line 1654: >if (current is ContextualAttributeBinder) [](http://example.com/codeflow?start=24&length=41)

If we end up keeping an additional logic here (which wouldn't be necessary with the alternative approach), I would prefer it to be a syntactic check based on `id`.

AlekseyTs:

jcouv:

AlekseyTs (approved):
LGTM (commit 2)

AlekseyTs:
While merging, consider adjusting commit's title and comment to align them with the latest revision.

jcouv:
> While merging, consider adjusting commit's title and comment to align them with the latest revision.

Yup. Already done :-)

jcouv:

333fred (approved):

PR #75103, opened 2024-09-13
Fix null coalescing operator emit with span conversion

jjonescz:
A bug discovered while running runtime tests (would be also discovered by building runtime with Debug version of roslyn but currently there are other unrelated asserts so I haven't tried that).

### DIFF
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_NullCoalescingOperator.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_NullCoalescingOperator.cs
@@ -87,10 +87,9 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             // if left conversion is intrinsic implicit (always succeeds) and results in a reference type
             // we can apply conversion before doing the null check that allows for a more efficient IL emit.
-            // span conversion is from a reference type (array/string) to a value type ([ReadOnly]Span) so it cannot participate here.
             Debug.Assert(rewrittenLeft.Type is { });
             if (rewrittenLeft.Type.IsReferenceType &&
-                BoundNode.GetConversion(leftConversion, leftPlaceholder) is { IsImplicit: true, IsUserDefined: false, IsSpan: false })
+                BoundNode.GetConversion(leftConversion, leftPlaceholder) is { Kind: ConversionKind.Identity or ConversionKind.ImplicitReference })
             {
                 rewrittenLeft = ApplyConversionIfNotIdentity(leftConversion, leftPlaceholder, rewrittenLeft);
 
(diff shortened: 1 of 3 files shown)


src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenOperators.cs
```diff
@@ -1479,11 +1479,9 @@ .locals init (T1 V_0)
...
   IL_001d:  call       ""void System.Console.WriteLine(object)""
   IL_0022:  ldarg.0
-  IL_0023:  stloc.0
-  IL_0024:  ldloc.0
-  IL_0025:  box        ""T1""
-  IL_002a:  pop
-  IL_002b:  ret
+  IL_0023:  box        ""T1""
+  IL_0028:  pop
+  IL_0029:  ret
```
jjonescz, line 1484: This changed because the left operand of the coalesce (`t1`) has conversion to `T2` which is not a reference type (previously the check was against `T1` which is a reference type). I'm not actually sure why that's the case since there is `where T1 : T2` I would expect both to be detected as reference types.

333fred (approved):

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_NullCoalescingOperator.cs
```diff
@@ -87,9 +87,10 @@ private BoundExpression MakeNullCoalescingOperator(
 
             // if left conversion is intrinsic implicit (always succeeds) and results in a reference type
             // we can apply conversion before doing the null check that allows for a more efficient IL emit.
+            // span conversion is from a reference type (array/string) to a value type ([ReadOnly]Span) so it cannot participate here.
             Debug.Assert(rewrittenLeft.Type is { });
             if (rewrittenLeft.Type.IsReferenceType &&
-                BoundNode.GetConversion(leftConversion, leftPlaceholder) is { IsImplicit: true, IsUserDefined: false })
+                BoundNode.GetConversion(leftConversion, leftPlaceholder) is { IsImplicit: true, IsUserDefined: false, IsSpan: false })
```
AlekseyTs, line 93: >IsSpan: false [](http://example.com/codeflow?start=118&length=13)

Given the original comment above the `if`, it feels like the condition was a somewhat "roundabout" way of checking for an implicit reference conversion. With the additional condition we are still trying to check for what we are not looking for rather than checking for what we are looking for. Consider adjusting condition accordingly in order to make it more robust in long run.

AlekseyTs:

AlekseyTs (approved):
LGTM (commit 2)

jjonescz:
@333fred for another look as there was a small product code change, thanks

333fred (approved):

PR #75104, opened 2024-09-13
Add `Enumerable.Reverse(this T[])` polyfill

jjonescz:
To avoid needing to fixup more locations when merging main into the feature branch. See also https://github.com/dotnet/roslyn/pull/75077#discussion_r1757047047.

### DIFF
--- a/src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
@@ -879,6 +879,9 @@ namespace System.Linq
             }
         }
 
+        // https://github.com/dotnet/runtime/issues/107723
+        public static IEnumerable<T> Reverse<T>(this T[] source) => Enumerable.Reverse(source);
+
 #if NETSTANDARD
 
         // Copied from https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq/src/System/Linq/Chunk.cs
--- a/src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
@@ -880,7 +880,11 @@ namespace System.Linq
         }
 
         // https://github.com/dotnet/runtime/issues/107723
+#if NET10_0_OR_GREATER
+        public static IEnumerable<T> Reverse<T>(T[] source) => Enumerable.Reverse(source);
+#else
         public static IEnumerable<T> Reverse<T>(this T[] source) => Enumerable.Reverse(source);
+#endif
 
 #if NETSTANDARD
 
--- a/src/Features/CSharp/Portable/CodeRefactorings/UseRecursivePatterns/UseRecursivePatternsCodeRefactoringProvider.cs
+++ b/src/Features/CSharp/Portable/CodeRefactorings/UseRecursivePatterns/UseRecursivePatternsCodeRefactoringProvider.cs
@@ -600,7 +600,7 @@ internal sealed class UseRecursivePatternsCodeRefactoringProvider : SyntaxEditor
 
         // Process all nodes to refactor in reverse to ensure nested nodes
         // are processed before the outer nodes to refactor.
-        foreach (var originalNode in Enumerable.Reverse(nodes))
+        foreach (var originalNode in nodes.Reverse())
         {
             // Only process nodes fully within a fixAllSpan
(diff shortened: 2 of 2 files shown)


cston (approved):

src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
```diff
@@ -879,6 +879,9 @@ public static bool SequenceEqual<T>(this IEnumerable<T>? first, IEnumerable<T>?
             }
         }
 
+        // https://github.com/dotnet/runtime/issues/107723
+        public static IEnumerable<T> Reverse<T>(this T[] source) => Enumerable.Reverse(source);
```
333fred, line 883: It's probably not correct to put this here. @jaredpar to confirm, but I'm concerned that we'll end up breaking ourselves when the bcl adds this.

jjonescz: https://github.com/dotnet/roslyn/blob/5a39d0ad691ee88dcf163808d9ed7a46b168a1c9/docs/contributing/Target%20Framework%20Strategy.md#pattern-for-extension-methods

I think this should be fine. Problems could arise if we wrap this in `#if`s incorrectly (so when bcl adds this we need to be careful making this conditional).

jaredpar: I _think_ that we should be able to mitigate this by wrapping this in 

```csharp
#if !NET10_OR_GREATER
```

By the time that we actually start targeting `net10.0` this API is going to be there so that won't cause a break on us. We generally move to the new TFM rather late in the release and think that we can expect this to be added early given it's tied to first class span support which is going in early.

333fred: Won't we potentially break in a torn dll scenario though? Or have we thoroughly removed any change of that at this point?

jaredpar: What teared scenario are you thinking of? This is just another API inside the compiler. The only impact that can happen is weirdness at compile time. The `#if` will ensure that `net9.0` or earlier we have the method. If we get deployed into `net10.0` or higher then there is just another polyfill API in our binary. In the case the `#if` is `false` then we're in `net10.0` where the API is guaranteed to be available.

333fred: Well, we certainly had a tear like this when we had the EA dll. Have we gotten rid of all of those scenarios?

jaredpar: Shouldn't have this type of tearing problem without the EA DLL

333fred (approved):

src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
jjonescz: Hm, good idea about adding the `#if` for .NET 10, I have done that. Now it matches our guidelines linked above precisely.

src/Compilers/Core/Portable/InternalUtilities/EnumerableExtensions.cs
```diff
@@ -880,7 +880,11 @@ public static bool SequenceEqual<T>(this IEnumerable<T>? first, IEnumerable<T>?
         }
 
         // https://github.com/dotnet/runtime/issues/107723
+#if NET10_0_OR_GREATER
+        public static IEnumerable<T> Reverse<T>(T[] source) => Enumerable.Reverse(source);
```
jaredpar, line 884: in `net10.0` though won't we get this method for free? Basically it would be coming from the BCL so we wouldn't need it at all.

jjonescz: Yes, but the guidelines say to define the non-extension version for binary compatibility, or does that not apply here for some reason?

jaredpar: Yes but what idiot wrote those guidelines?  ;) 

I looked back over the guidelines and I'm trying to decide why I added that particular section. I know the reason for creating the non-extension version was to avoid compile time ambiguities. I can't remember exactly why I didn't say "just remove it entirely". I'm guessing it was because there were cases where it was used by full name ... 

Anyways, wrote that for a reason so lets follow it 😄

jaredpar (approved):

PR #60601, opened 2022-04-06
[semi-auto-props]: Support assigning properties from constructor

Youssef1313:
Test plan: https://github.com/dotnet/roslyn/issues/57012

cc @AlekseyTs @333fred

Note: Code I modified is used for both allowing assigning from constructors and also for some data flow. Changes made here were more focused on the first part. Meaning that it *could* be not correct for data flow which (**if possible**) I want to leave for now (there are PROTOTYPE comments there).

This implements the following part from [LDM notes](https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-01-12.md#open-questions-for-field):

> Properties can be assigned in the constructor of a type if that property has a setter, or is a get-only property whose backing field will be emitted.

And also, from the [original proposal](https://github.com/dotnet/csharplang/blob/main/proposals/semi-auto-properties.md):

> If the auto-property does not have a set accessor, the backing field can still be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.

PR #58832, opened 2022-01-13
[semi-auto-props]: Refactor and rename 'AccessingAutoPropertyFromConstructor'

Youssef1313:
Got some time today to get a small PR :)

Test plan: https://github.com/dotnet/roslyn/issues/57012

PR #75116, opened 2024-09-13
Field-backed properties: definite assignment

cston:
See [proposal](https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#definite-assignment-in-structs).

Similar to changes started in earlier `features/semi-auto-props` branch in https://github.com/dotnet/roslyn/pull/58832, https://github.com/dotnet/roslyn/pull/60601.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
@@ -527,15 +527,14 @@ namespace Microsoft.CodeAnalysis.CSharp
                     if (!InAttributeArgument)
                     {
                         var propertyAccess = (BoundPropertyAccess)expr;
-                        var propertySymbol = propertyAccess.PropertySymbol;
-                        if (propertySymbol.RefKind == RefKind.None)
+                        if (propertyAccess.PropertySymbol.RefKind == RefKind.None)
                         {
                             bool? useAsLvalue = RequiresRValueOnly(valueKind) ? false : RequiresAssignableVariable(valueKind) ? true : null;
                             if (useAsLvalue.HasValue)
                             {
-                                if (AccessingAutoPropertyFromConstructor(propertyAccess.ReceiverOpt, propertySymbol, useAsLvalue.Value))
+                                if (CanUseBackingFieldDirectlyInConstructor(propertyAccess, useAsLvalue.Value))
                                 {
-                                    // Update expr to mark that the property access should use the backing field.
+                                    // Update the property access to indicate that the backing field should be used directly.
                                     expr = propertyAccess.Update(
                                         propertyAccess.ReceiverOpt,
                                         propertyAccess.InitialBindingReceiverIsSubjectToCloning,
@@ -1720,7 +1719,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 if (setMethod is null)
                 {
                     var containing = this.ContainingMemberOrLambda;
-                    if (!(expr is BoundPropertyAccess propertyAccess && AccessingAutoPropertyFromConstructor(propertyAccess, containing, useAsLvalue: true))
+                    if (!(expr is BoundPropertyAccess propertyAccess && AccessingAutoPropertyFromConstructor(propertyAccess, containing))
                         && !isAllowedDespiteReadonly(receiver))
                     {
                         Error(diagnostics, ErrorCode.ERR_AssgReadonlyProp, node, propertySymbol);
--- a/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
@@ -523,6 +523,31 @@ namespace Microsoft.CodeAnalysis.CSharp
                     Debug.Assert(valueKind is (BindValueKind.Assignable or BindValueKind.RefOrOut or BindValueKind.RefAssignable) || diagnostics.DiagnosticBag is null || diagnostics.HasAnyResolvedErrors());
                     return expr;
 
+                case BoundKind.PropertyAccess:
+                    {
+                        var propertyAccess = (BoundPropertyAccess)expr;
+                        var propertySymbol = propertyAccess.PropertySymbol;
+                        if (propertySymbol.RefKind == RefKind.None)
+                        {
+                            bool? useAsLvalue = RequiresRValueOnly(valueKind) ? false : RequiresAssignableVariable(valueKind) ? true : null;
+                            if (useAsLvalue.HasValue)
+                            {
+                                if (AccessingAutoPropertyFromConstructor(propertyAccess.ReceiverOpt, propertySymbol, useAsLvalue.Value))
+                                {
+                                    // Update expr to mark that the property access should use the backing field.
+                                    expr = propertyAccess.Update(
+                                        propertyAccess.ReceiverOpt,
+                                        propertyAccess.InitialBindingReceiverIsSubjectToCloning,
+                                        propertyAccess.PropertySymbol,
+                                        useAsLvalue.Value ? AccessorKind.Set : AccessorKind.Get, // PROTOTYPE: What about compound assignment, increment?
+                                        propertyAccess.ResultKind,
+                                        propertyAccess.Type);
+                                }
+                            }
+                        }
+                    }
+                    break;
+
                 case BoundKind.IndexerAccess:
                     expr = BindIndexerDefaultArgumentsAndParamsCollection((BoundIndexerAccess)expr, valueKind, diagnostics);
                     break;
@@ -1694,7 +1719,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 if (setMethod is null)
                 {
                     var containing = this.ContainingMemberOrLambda;
-                    if (!AccessingAutoPropertyFromConstructor(receiver, propertySymbol, containing, useAsLvalue: true)
+                    if (!(expr is BoundPropertyAccess propertyAccess && AccessingAutoPropertyFromConstructor(propertyAccess, containing, useAsLvalue: true))
                         && !isAllowedDespiteReadonly(receiver))
                     {
                         Error(diagnostics, ErrorCode.ERR_AssgReadonlyProp, node, propertySymbol);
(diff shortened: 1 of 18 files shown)


src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -659,6 +659,32 @@ protected bool HasExplicitAccessModifier
...
+        /// True if the property has a synthesized backing field, and
+        /// either no accessor or the accessor is auto-implemented.
+        /// </summary>
+        internal bool CanUseBackingFieldDirectlyInConstructor(bool useAsLvalue)
+        {
+            if (BackingField is null)
+            {
+                return false;
+            }
+            if (useAsLvalue)
```
333fred, line 678: Will this work for a `ref`-returning property?

333fred: Actually, I suppose it would work fine, since you can't use the backing field directly in that case... Nevermind.

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -523,6 +523,31 @@ private BoundExpression CheckValue(BoundExpression expr, BindValueKind valueKind
...
+                            if (useAsLvalue.HasValue)
+                            {
+                                if (CanUseBackingFieldDirectlyInConstructor(propertyAccess, useAsLvalue.Value))
+                                {
+                                    // Update the property access to indicate that the backing field should be used directly.
+                                    expr = propertyAccess.Update(
+                                        propertyAccess.ReceiverOpt,
+                                        propertyAccess.InitialBindingReceiverIsSubjectToCloning,
+                                        propertyAccess.PropertySymbol,
+                                        useAsLvalue.Value ? AccessorKind.Set : AccessorKind.Get, // PROTOTYPE: What about compound assignment, increment?
```
RikkiGibson, line 542: Consider extracting and reusing logic from 'GetIndexerAccessorKind' in order to determine this.

333fred (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
cston: Auto-properties cannot be `ref`-returning.

RikkiGibson (commented):
Per offline discussion I am wondering if a somewhat smaller change would be adequate to address the scenario. It looks like the existing behavior with auto properties in struct constructors is:

- property get: auto-default all unassigned fields
- property set: don't auto-default anything, if the set accessor is auto-implemented or absent.

So, shouldn't a relatively small change to logic in definite assignment be adequate here (check if setter specifically is auto-implemented rather than whole property)?

cston:
> ... It looks like the existing behavior with auto properties in struct constructors is:
> 
> * property get: auto-default all unassigned fields, even if getter is auto-implemented.
> * property set: don't auto-default anything, if the set accessor is auto-implemented or absent.

Calling `get` with an auto property today will cause that one backing field to be initialized before the call but does not affect other fields that have yet to be assigned.

For instance, with the following, `y = P1` results in initialization of `P1` but not `F1`. See [sharplab.io](https://sharplab.io/#v2:CYLg1APgzgLgTgVwMYwAQGUCMBYAUAbz1WNQAEBmVASwDs0AxTAbiJIurtQAVNV9UA5gFMYTVAF9WxdlgAUtNAA8ANKgD2CNAtQBPAJRS+hkrtQBebs2MlG51Ipa4Tk3OKA=).
```csharp
struct S1
{
    public int F1;
    public int P1 { get; }
    public S1(int x, out int y)
    {
        y = P1;
        F1 = x;
    }
}
```

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -1749,27 +1749,35 @@ private DiagnosticInfo GetBadEventUsageDiagnosticInfo(EventSymbol eventSymbol)
                 new CSDiagnosticInfo(ErrorCode.ERR_BadEventUsageNoField, leastOverridden);
         }
 
+#nullable enable
         internal static bool AccessingAutoPropertyFromConstructor(BoundPropertyAccess propertyAccess, Symbol fromMember)
         {
-            return AccessingAutoPropertyFromConstructor(propertyAccess.ReceiverOpt, propertyAccess.PropertySymbol, fromMember);
+            if (propertyAccess.UseBackingField == AccessorKind.Unknown)
+            {
+                return false;
```
RikkiGibson, line 1757: Is there anything we should assert about the property access, when we are asking this question about a property access which may not have had this calculated yet?

RikkiGibson (commented):
Sorry, I still need more time to finish review, but had a few more comments.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -2205,6 +2206,94 @@ public static bool FoundInUnboundLambda(BoundNode methodBody, IdentifierNameSynt
                 return base.Visit(node);
             }
         }
+
+        private sealed class WasPropertyBackingFieldAccessChecked : BoundTreeWalkerWithStackGuardWithoutRecursionOnTheLeftOfBinaryOperator
```
RikkiGibson, line 2210: It looks like it would also be reasonable to wrap this type declaration in `#if DEBUG`.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -2205,6 +2206,94 @@ public static bool FoundInUnboundLambda(BoundNode methodBody, IdentifierNameSynt
...
+
+            private WasPropertyBackingFieldAccessChecked()
+            {
+            }
+
+            public override BoundNode? Visit(BoundNode? node)
+            {
+                if (_found is { })
+                {
+                    return node;
```
RikkiGibson, line 2230: Is this return value used?

cston: No, the return value is not used, and in fact most of the visit methods return `null`. I'll change this to `return null;`.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -1650,150 +1650,1327 @@ .maxstack  1
...
+        {
+            string setter = useInit ? "init" : "set";
+            string source = $$"""
+                #pragma warning disable 649
+                using System;
+                struct C1
+                {
+                    public int F1;
+                    public int P1 { get; }
+                    public C1(int i) { P1 += i; }
```
RikkiGibson, line 2418: It seems like this and the other scenarios in this test would be more illustrative if `F1` were explicitly assigned after the `P1 += i;` statement. That would show whether the property access is causing auto-default of the whole instance.

Same for the `++`, `--`, `??=` tests below.

src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
```diff
@@ -2194,7 +2194,7 @@
     <!-- Whether receiver will need to be cloned during emit (only valid before lowering) -->
     <Field Name="InitialBindingReceiverIsSubjectToCloning" Type="ThreeState"/>
     <Field Name="PropertySymbol" Type="PropertySymbol"/>
-    <!-- The property accessor referenced if the property access should use the backing field directly instead. -->
+    <!-- The property accessor kind, set for properties with synthesized backing field only. -->
     <Field Name="UseBackingField" Type="AccessorKind" />
```
RikkiGibson, line 2198: I'm not 100% happy with this field name, but I don't have any suggestions for a better one. It feels like this name is saying "this property access should be replaced with a use of the backing field".

RikkiGibson (approved):
LGTM, nothing blocking.

src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
cston: Renamed to `BoundPropertyAccess.AutoPropertyAccessorKind`.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
cston: Thanks, that makes sense.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -1073,6 +1074,7 @@ static IEnumerable<Symbol> getAllMembersToBeDefaulted(Symbol requiredMember)
                             }
                         }
 
+                        // PROTOTYPE: Is checking IsAutoProperty sufficient or should we check the appropriate accessors are missing or auto-implemented?
```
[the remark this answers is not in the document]
cston: Moved comment to test plan https://github.com/dotnet/roslyn/issues/57012.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -1964,6 +1966,8 @@ protected override bool TryGetReceiverAndMember(BoundExpression expr, out BoundE
                         var propAccess = (BoundPropertyAccess)expr;
                         var propSymbol = propAccess.PropertySymbol;
                         member = propSymbol;
+                        // PROTOTYPE: Why don't we call Binder.AccessingAutoPropertyFromConstructor
```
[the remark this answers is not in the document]
cston: Moved comment to test plan https://github.com/dotnet/roslyn/issues/57012.

src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
```diff
@@ -2195,7 +2195,7 @@
     <Field Name="InitialBindingReceiverIsSubjectToCloning" Type="ThreeState"/>
     <Field Name="PropertySymbol" Type="PropertySymbol"/>
     <!-- The property accessor kind, set for properties with synthesized backing field only. -->
```
RikkiGibson, line 2197: is it also the case that this will only have a non-Unknown value when auto accessors are being used?

RikkiGibson (approved):

src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
cston: The value will be set to something other than `AccessorKind.Unknown` for classic auto properties and for properties that use `field`. Properties that use `field` may include an auto-implemented accessor.

PR #75139, opened 2024-09-16
Use MSBuild globs to determine which file changes are relevant

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/6985

The approach I took was to serialize glob information over to the host process, and use MS.Extensions.FileGlobbing to match files against the globs.

Unfortunately it is quite difficult to get full fidelity matching - ideally we would be able to call `IsMatch` on a `CompositeGlob` built from all the project globs - however we do not want to load MSBuild in the host process.

Instead, we serialize the set of `Includes`, `Excludes`, and `Removes` from the evaluated `GlobResult`.  This roughly should be the correct set of globs and matches how CPS recreates globs, see https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=/src/Microsoft.VisualStudio.ProjectSystem/Build/MsBuildGlobFactory.cs&version=GBmain&line=34&lineEnd=35&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

We then match using the project directory as the relative directory (also matches CPS).  However our matching is a bit simpler as we don't currently even get file watcher notifications for anything outside the project cone.

### DIFF
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/LoadedProject.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/LoadedProject.cs
@@ -9,6 +9,7 @@ using Microsoft.CodeAnalysis.LanguageServer.HostWorkspace.ProjectTelemetry;
 using Microsoft.CodeAnalysis.MSBuild;
 using Microsoft.CodeAnalysis.ProjectSystem;
 using Microsoft.CodeAnalysis.Workspaces.ProjectSystem;
+using Microsoft.Extensions.FileSystemGlobbing;
 using Microsoft.Extensions.Logging;
 using Roslyn.Utilities;
 
@@ -19,6 +20,9 @@ namespace Microsoft.CodeAnalysis.LanguageServer.HostWorkspace;
 /// </summary>
 internal sealed class LoadedProject : IDisposable
 {
+    private readonly string _projectFilePath;
+    private readonly string _projectDirectory;
+
     private readonly ProjectSystemProject _projectSystemProject;
     private readonly ProjectSystemProjectOptionsProcessor _optionsProcessor;
     private readonly IFileChangeContext _fileChangeContext;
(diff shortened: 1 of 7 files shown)


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/LoadedProject.cs
```diff
@@ -28,26 +32,31 @@ internal sealed class LoadedProject : IDisposable
...
         // We'll watch the directory for all source file changes
         // TODO: we only should listen for add/removals here, but we can't specify such a filter now
-        var projectDirectory = Path.GetDirectoryName(projectSystemProject.FilePath)!;
+        _projectDirectory = Path.GetDirectoryName(_projectFilePath)!;
 
         _fileChangeContext = fileWatcher.CreateContext([
-            new(projectDirectory, ".cs"),
-            new(projectDirectory, ".cshtml"),
-            new(projectDirectory, ".razor")
+            new(_projectDirectory, ".cs"),
```
tmat, line 57: Are we not watching for changes in non-Razor additional files, which might affect other source generators? How about resx files, etc.?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/LoadedProject.cs
```diff
@@ -57,7 +66,34 @@ public LoadedProject(ProjectSystemProject projectSystemProject, SolutionServices
 
     private void FileChangedContext_FileChanged(object? sender, string filePath)
     {
-        NeedsReload?.Invoke(this, EventArgs.Empty);
+        // If the project file itself changed, we almost certainly need to reload the project.
+        if (string.Equals(filePath, _projectFilePath, StringComparison.OrdinalIgnoreCase))
```
tmat, line 70: Are we gonna get .csproj file change events if we didn't include .csproj in CreateContext?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/LoadedProject.cs
dibarbet: We are watching for changes for files we already know about in the solution elsewhere.  This is attempting to look for added/removed files that we don't know about that we might need to reload the project for.

Yes - this file watcher has problems.  Additions of files outside the project cone are not watched.  Additions of other kinds of files are also not watched.   However that is pre-existing and not directly the problem I'm trying to solve at the moment.

dibarbet: Note - this code only applies to C# standalone (without Dev Kit).

JoeRobich (approved):

PR #75160, opened 2024-09-18
Skip tests on Mono

giritrivedi:
Couple of visualBasic "EmitAndContinue" test cases fail with
"Operation is not supported on this platform" error.
 Reason for this failure is that they are not supported on Mono.
 Ensured that these test cases are skipped when run on mono.


src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
```diff
@@ -159,6 +159,7 @@ public static class ExecutionConditionUtil
         public static bool IsWindowsDesktop => IsWindows && IsDesktop;
         public static bool IsMonoDesktop => Type.GetType("Mono.Runtime") != null;
         public static bool IsMono => MonoHelpers.IsRunningOnMono();
+        public static bool IsNewMono => Type.GetType("Mono.RuntimeStructs") != null;
```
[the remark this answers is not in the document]
giritrivedi: IsMono - looks for Desktop/legacy mono . 
Figured this out when I was trying to skip couple of test cases on mono environment using ClrOnly() which uses IsRunningOnMono(). 
ClrOnly() doesn't hold good for mono platforms . 

Hence to figure out if it is running on mono runtime or not implemented another check IsNewMono which uses IsUsingMonoRuntime() .

src/Compilers/Test/Core/FX/MonoHelpers.cs
```diff
@@ -11,5 +11,6 @@ namespace Microsoft.CodeAnalysis.Test.Utilities
     public static class MonoHelpers
     {
         public static bool IsRunningOnMono() => Roslyn.Test.Utilities.ExecutionConditionUtil.IsMonoDesktop;
+        public static bool IsUsingMonoRuntime() => Roslyn.Test.Utilities.ExecutionConditionUtil.IsNewMono;
```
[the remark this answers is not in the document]
giritrivedi: Please check my reply in my earlier comment. Hope that clarifies.

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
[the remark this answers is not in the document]
giritrivedi: Is there anything more information I can provide you with ?

giritrivedi:
\cc @uweigand

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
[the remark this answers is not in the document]
jaredpar: I'm still not clear though what the new mono is. I'm very familiar with desktop mono. What is the new variant of mono that you're trying to account for here?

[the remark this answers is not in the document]
JoeRobich: There is the Mono runtime which lives in dotnet/runtime and is used for WASM and other scenarios. It appears that `Mono.RuntimeStructs` lives in both mono/mono and dotnet/runtime. `Mono.Runtime` seems to only be in mono/mono.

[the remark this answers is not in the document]
uweigand: Yes, this test is about checking for running on .NET with the Mono runtime (as opposed to the CoreCLR runtime).  There is also a similar check already present in roslyn here: https://github.com/dotnet/roslyn/blob/54fd790788b03c256f736c917fc368daf0a643db/src/Compilers/Core/Portable/InternalUtilities/PlatformInformation.cs#L44

Maybe this can be re-used.  Or at least, the same terminology can be used - I think something like `IsUsingMonoRuntime` (or `IsMonoRuntime` like here https://github.com/dotnet/runtime/blob/b5833d2cd22f474485f74696609f6a6c3617cd15/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.cs#L30) would be preferable over `IsNewMono`.

giritrivedi:
@jaredpar , The suggested changes have been incorporated and all tests are passing. 
Would you mind reviewing it one more time ?

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
```diff
@@ -159,6 +159,7 @@ public static class ExecutionConditionUtil
         public static bool IsWindowsDesktop => IsWindows && IsDesktop;
         public static bool IsMonoDesktop => Type.GetType("Mono.Runtime") != null;
         public static bool IsMono => MonoHelpers.IsRunningOnMono();
+        public static bool IsMonoRuntime => Type.GetType("Mono.RuntimeStructs") != null;
```
jaredpar, line 162: Now that we're making a distinction between Mono Core and Mono Framework I would like it if the names we use here are more descriptive. It seems that we essentially want three properties here: 

- `IsMonoDesktop`: true when we are running under the Mono equivalent of .NET Framework
- `IsMonoCore`: true when we are running under the Mono equivalent of .Net Core
- `IsAnyMono`: basically `IsMonoDesktop || IsMonoCore`
- p.

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
```diff
@@ -332,6 +333,12 @@ public class ClrOnly : ExecutionCondition
         public override string SkipReason => "Test not supported on Mono";
     }
 
+    public class NotMono : ExecutionCondition
```
jaredpar, line 336: Please make the naming a more crisp here to denote if it's any, desktop or core mono.

src/Compilers/Test/Core/FX/MonoHelpers.cs
```diff
@@ -11,5 +11,6 @@ namespace Microsoft.CodeAnalysis.Test.Utilities
     public static class MonoHelpers
     {
         public static bool IsRunningOnMono() => Roslyn.Test.Utilities.ExecutionConditionUtil.IsMonoDesktop;
+        public static bool IsUsingMonoRuntime() => Roslyn.Test.Utilities.ExecutionConditionUtil.IsMonoRuntime;
```
jaredpar, line 14: Please make the naming a more crisp here to denote if it's any, desktop or core mono.

jaredpar (changes requested):

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
giritrivedi: hi @jaredpar , Thanks for taking a look. So you are saying to use "IsMonoCore" instead of "IsMonoRuntime" ?

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
giritrivedi: How about using "NotOnMonoCore" ?

src/Compilers/Test/Core/FX/MonoHelpers.cs
giritrivedi: How about using "IsRUnningOnMonoCore" ? 
Let me know if these names looks good to you ? I would like to commit my next change once we are in agreement with names to be used .

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
jaredpar: Correct. `MonoRuntime` is a bit ambiguous cause it could be the framework or core version.

uweigand: Just to chime in here a bit.  As I understand it, there are four environments we could be running under:
- We could run under .NET Framework (4.x)
- We could run under .NET Core (actually .NET 6, 8, 9,...) using the CoreCLR runtime (i.e. https://github.com/dotnet/runtime/tree/main/src/coreclr)
- We could run under .NET Core (actually .NET 6, 8, 9,...) using the Mono runtime (i.e. https://github.com/dotnet/runtime/tree/main/src/mono)
- We could run under stand-alone Mono itself (i.e. https://github.com/mono/mono)

The distinction @giritrivedi is trying to make here is between the two latter cases.  I'm not sure this is really a question of .NET Framework vs. .NET Core; if I understand correctly, the stand-alone Mono project sort-of supports both of these as well, but that's not what we care about here.

What the test checks for is "are we running under .NET (Core) using the Mono runtime".   [ I don't have strong opinions on any particular name, I just wanted to be sure we're all clear what exactly this is intended to express.  ]


As an aside, I noticed this somewhat strange situation:  in this location, we have:
```
        public static bool IsMonoDesktop => Type.GetType("Mono.Runtime") != null;
        public static bool IsMono => MonoHelpers.IsRunningOnMono();
```
but that helper function is itself defined as:
```
    public static class MonoHelpers
    {   
        public static bool IsRunningOnMono() => Roslyn.Test.Utilities.ExecutionConditionUtil.IsMonoDesktop;
    }
```
so it seems the `IsMono` definition is redundant (always equivalent to `IsMonoDesktop`)?

jaredpar: > I'm not sure this is really a question of .NET Framework vs. .NET Core; if I understand correctly, the stand-alone Mono project sort-of supports both of these as well, but that's not what we care about here.

We generally think of `mono/mono` as the framework / desktop mono. Possible this is just our vernacular though. Happy to consider other suffix here. 

> so it seems the IsMono definition is redundant (always equivalent to IsMonoDesktop)?

That is very possible. It's been a while since we had Mono in our core CI loop. I wouldn't be surprised if we had some issues like this that popped up over time.

jaredpar (approved):

jaredpar:
@dotnet/roslyn-compiler PTAL

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
```diff
@@ -332,6 +333,12 @@ public class ClrOnly : ExecutionCondition
         public override string SkipReason => "Test not supported on Mono";
     }
 
+    public class NotOnMonoCore : ExecutionCondition
+    {
+        public override bool ShouldSkip => MonoHelpers.IsRunningOnMonoCore();
+        public override string SkipReason => "Test not supported on Mono";
```
jjonescz, line 339:
```suggestion
        public override string SkipReason => "Test not supported on Mono Core";
```

src/Compilers/Test/Core/FX/MonoHelpers.cs
```diff
@@ -11,5 +11,6 @@ namespace Microsoft.CodeAnalysis.Test.Utilities
     public static class MonoHelpers
     {
         public static bool IsRunningOnMono() => Roslyn.Test.Utilities.ExecutionConditionUtil.IsMonoDesktop;
```
jjonescz, line 13: Should we rename this to `IsRunningOnMonoDesktop`?

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
```diff
@@ -159,6 +159,7 @@ public static class ExecutionConditionUtil
         public static bool IsWindowsDesktop => IsWindows && IsDesktop;
         public static bool IsMonoDesktop => Type.GetType("Mono.Runtime") != null;
         public static bool IsMono => MonoHelpers.IsRunningOnMono();
```
jjonescz, line 161: Should we remove this property or change the logic to be `IsMonoDesktop || IsMonoCore`?

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
```diff
@@ -159,6 +159,7 @@ public static class ExecutionConditionUtil
...
+        public static bool IsMonoCore => Type.GetType("Mono.RuntimeStructs") != null;
```
jjonescz, line 162: Consider adding comments saying `IsMonoCore` means https://github.com/dotnet/runtime/tree/main/src/mono and `IsMonoDesktop` means https://github.com/mono/mono or something like that.

src/Compilers/Test/Core/FX/MonoHelpers.cs
giritrivedi: Looks to be a good idea. However I would do it in a separate PR. 
Please let me know if you insist to be done as part of this PR itself.

jjonescz: Separate PR or just filing an issue is fine, thanks.

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
jjonescz: (Filing an issue here is fine as well.)

jjonescz (approved):

jjonescz:
Thanks @giritrivedi

PR #75163, opened 2024-09-19
Fix the OOM error in scripting tests

jaredpar:
Was able to reproduce the Scripting `OutOfMemoryException` issues on my local machine. After looking at the test run in PerfView it was clear that we were allocating a lot more memory than necessary and that was, likey, the cause of the OOM that we saw. Given that it was running under x86 this seems fairly likely.

The allocations were coming more from the ResultsProvider and IOperation test assemlbies. But given that scripting is small our scheduling algorithm often puts scripting into a slice that includes both these test assemblies. Their allocations are what is pushing us to OOM. That is also likely why we couldn't reproduce this by just running the scripting tests locally.

This PR attempts to address the cases that I saw. There are a few cases where I moved eager operations to lazy. The bigger change though is moving a to using an interpolated string handler for `RoslynDebug.Assert`. There are a number of cases where had roughly the following code:

```csharp
Debug.Assert(condition, $"... {syntaxNode} ...");
```

On net472 that is going to eagerly allocate the message string which realizes the entire syntax node. One assert alone was contributing 9GB of allocations all by itself. Did a simple regex search for `Debug.Assert` that had an interpolation and moved them in mass to `RoslynDebug.Assert`. Now the strings are only realized if the assert fails. Will discuss with the team if we want to think about an analyzer here

PS: if the scripting tests OOM on this PR I may take tomorrow off.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversion.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversion.cs
@@ -252,7 +252,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                     break;
             }
 
-            Debug.Assert(isTrivial, "this conversion needs additional data: " + kind);
+            RoslynDebug.Assert(isTrivial, $"this conversion needs additional data: {kind}");
         }
 
         internal static Conversion GetTrivialConversion(ConversionKind kind)
--- a/src/Compilers/Core/Portable/InternalUtilities/Debug.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/Debug.cs
@@ -5,6 +5,8 @@
 using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Text;
 
 namespace Roslyn.Utilities
 {
@@ -12,12 +14,22 @@ namespace Roslyn.Utilities
     {
         /// <inheritdoc cref="Debug.Assert(bool)"/>
         [Conditional("DEBUG")]
-        public static void Assert([DoesNotReturnIf(false)] bool b) => Debug.Assert(b);
+        public static void Assert([DoesNotReturnIf(false)] bool condition) => Debug.Assert(condition);
 
         /// <inheritdoc cref="Debug.Assert(bool, string)"/>
         [Conditional("DEBUG")]
-        public static void Assert([DoesNotReturnIf(false)] bool b, string message)
-            => Debug.Assert(b, message);
+        public static void Assert([DoesNotReturnIf(false)] bool condition, string message)
+            => Debug.Assert(condition, message);
+
+        /// <inheritdoc cref="Debug.Assert(bool, string)"/>
+        [Conditional("DEBUG")]
+        public static void Assert([DoesNotReturnIf(false)] bool condition, [InterpolatedStringHandlerArgument(nameof(condition))] ref AssertInterpolatedStringHandler message)
+        {
+            if (!condition)
+            {
+                Debug.Fail(message.ToStringAndClear());
+            }
+        }
 
         [Conditional("DEBUG")]
         public static void AssertNotNull<T>([NotNull] T value)
@@ -61,6 +73,31 @@ namespace Roslyn.Utilities
                 }
             }
 #endif
+        }
+
+        [InterpolatedStringHandler]
+        public struct AssertInterpolatedStringHandler
+        {
+            private readonly StringBuilder? _builder;
+
+            public AssertInterpolatedStringHandler(int literalLength, int formattedCount, bool condition, out bool shouldAppend)
+            {
+                shouldAppend = condition;
+                if (shouldAppend)
+                {
+                    _builder = new StringBuilder(literalLength + formattedCount);
+                }
+            }
+
+            internal string ToStringAndClear() => _builder is null
+                ? ""
+                : _builder.ToString();
+
+            public void AppendLiteral(string value) => _builder!.Append(value);
+
+            public void AppendFormatted<T>(T value) => _builder!.Append(value);
+
+            public void AppendFormatted(ReadOnlySpan<char> value) => _builder!.Append(value.ToString());
         }
     }
 }
--- a/src/Compilers/Core/Portable/InternalUtilities/Debug.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/Debug.cs
@@ -82,7 +82,7 @@ namespace Roslyn.Utilities
 
             public AssertInterpolatedStringHandler(int literalLength, int formattedCount, bool condition, out bool shouldAppend)
             {
-                shouldAppend = condition;
+                shouldAppend = !condition;
                 if (shouldAppend)
                 {
                     _builder = new StringBuilder(literalLength + formattedCount);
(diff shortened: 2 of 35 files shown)


src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversion.cs
```diff
@@ -252,7 +252,7 @@ private static void AssertTrivialConversion(ConversionKind kind)
                     break;
             }
 
-            Debug.Assert(isTrivial, "this conversion needs additional data: " + kind);
+            RoslynDebug.Assert(isTrivial, $"this conversion needs additional data: {kind}");
```
jaredpar, line 255: To emphasize how much these changes matter: this `Debug.Assert` alone was causing 50MB of allocations in the slice I was debugging.

src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmClrType.cs
```diff
@@ -44,7 +44,9 @@ internal DkmClrType(DkmClrModuleInstance module, DkmClrAppDomain appDomain, Type
             ModuleInstance = module;
             AppDomain = appDomain;
             _lmrType = lmrType;
-            _evalAttributes = GetEvalAttributes(lmrType);
+            _lazyEvalAttributes = new System.Lazy<ReadOnlyCollection<DkmClrEvalAttribute>>(
+                () => GetEvalAttributes(lmrType),
```
jaredpar, line 48: `GetEvalAttribute` realizes a huge number of `string` and arrays inside the reflection libraries. About 150MB of savings from this change.

src/Compilers/Core/Portable/Operations/OperationMapBuilder.cs
```diff
@@ -65,8 +66,9 @@ private static void RecordOperation(IOperation operation, Dictionary<SyntaxNode,
                 if (!operation.IsImplicit)
                 {
                     // IOperation invariant is that all there is at most 1 non-implicit node per syntax node.
-                    Debug.Assert(!argument.ContainsKey(operation.Syntax),
-                        $"Duplicate operation node for {operation.Syntax}. Existing node is {(argument.TryGetValue(operation.Syntax, out var original) ? original.Kind : null)}, new node is {operation.Kind}.");
+                    RoslynDebug.Assert(
+                        !argument.ContainsKey(operation.Syntax),
+                        $"Duplicate operation node for {operation.Syntax}. Existing node is {(argument.TryGetValue(operation.Syntax, out var original) ? (OperationKind?)original.Kind : null)}, new node is {operation.Kind}.");
```
jaredpar, line 71: About 9GB of allocations mostly due to the `{operation.Syntax}` portion.

tmat:
Nice!

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
```diff
@@ -5,19 +5,31 @@
...
         /// <inheritdoc cref="Debug.Assert(bool, string)"/>
         [Conditional("DEBUG")]
-        public static void Assert([DoesNotReturnIf(false)] bool b, string message)
-            => Debug.Assert(b, message);
+        public static void Assert([DoesNotReturnIf(false)] bool condition, string message)
+            => Debug.Assert(condition, message);
+
+        /// <inheritdoc cref="Debug.Assert(bool, string)"/>
+        [Conditional("DEBUG")]
+        public static void Assert([DoesNotReturnIf(false)] bool condition, [InterpolatedStringHandlerArgument(nameof(condition))] ref AssertInterpolatedStringHandler message)
```
cston, line 26: >Assert [](http://example.com/codeflow?start=27&length=6)

Should we polyfill this as a method in `System.Diagnostics.Debug` rather than `Roslyn.Utilities.RoslynDebug`, to avoid changing call sites and avoid overlooking call sites?

jaredpar: It's a `static` method so can't polyfill until we get extension everything 😦

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
```diff
@@ -62,5 +74,30 @@ internal static void AssertOrFailFast([DoesNotReturnIf(false)] bool condition, s
...
+            {
+                shouldAppend = condition;
+                if (shouldAppend)
+                {
+                    _builder = new StringBuilder(literalLength + formattedCount);
+                }
+            }
+
+            internal string ToStringAndClear() => _builder is null
+                ? ""
```
333fred, line 93: Nit: seems like this should never be hit. Consider just throwing.

333fred (approved):

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
333fred: Don't think we could anyways, the non-extension static method is applicable. We'd never look for a polyfill.

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
```diff
@@ -5,19 +5,31 @@
...
         /// <inheritdoc cref="Debug.Assert(bool)"/>
         [Conditional("DEBUG")]
-        public static void Assert([DoesNotReturnIf(false)] bool b) => Debug.Assert(b);
+        public static void Assert([DoesNotReturnIf(false)] bool condition) => Debug.Assert(condition);
 
         /// <inheritdoc cref="Debug.Assert(bool, string)"/>
         [Conditional("DEBUG")]
-        public static void Assert([DoesNotReturnIf(false)] bool b, string message)
-            => Debug.Assert(b, message);
+        public static void Assert([DoesNotReturnIf(false)] bool condition, string message)
```
jjonescz, line 21: Can this overload be removed?

jjonescz (approved):

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
jaredpar: @333fred as I'm not sure about the implications here of doing this.

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
```diff
@@ -62,5 +74,28 @@ internal static void AssertOrFailFast([DoesNotReturnIf(false)] bool condition, s
...
+            public AssertInterpolatedStringHandler(int literalLength, int formattedCount, bool condition, out bool shouldAppend)
+            {
+                shouldAppend = condition;
+                if (shouldAppend)
+                {
+                    _builder = new StringBuilder(literalLength + formattedCount);
+                }
+            }
+
+            internal string ToStringAndClear() => _builder!.ToString();
```
cston, line 92: >_builder!.ToString(); [](http://example.com/codeflow?start=50&length=21)

Should this call `_builder.Clear()` as well?

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
333fred: No, a regular string won't work with the handler.

AlekseyTs (approved):
LGTM (commit 3)

jaredpar:
image


Sigh .... Still some more work to do here.

PR #75165, opened 2024-09-19
Fix NETCOREAPP3_0_OR_NEWER checks

JoeRobich:
According to https://learn.microsoft.com/en-us/dotnet/standard/frameworks#preprocessor-symbols it should be _OR_GREATER


sharwell:
I merged an earlier version of this change instead (#74687)

PR #75171, opened 2024-09-19
Test  allocations

jaredpar:
Continuing to work on unnecessary allocations in our unit tests

### DIFF
--- a/src/Compilers/CSharp/Test/Semantic/Semantics/NativeIntegerTests.cs
+++ b/src/Compilers/CSharp/Test/Semantic/Semantics/NativeIntegerTests.cs
@@ -12384,7 +12384,7 @@ $@"class Program
 
                 if (expectedDiagnostics.Length == 0)
                 {
-                    CompileAndVerify(comp);
+                    CompileAndVerify(comp, emitOptions: EmitOptions.Default);
                 }
 
                 static bool useUnsafe(string type) => type == "void*";
--- a/src/Compilers/Test/Core/Platform/Desktop/Extensions.cs
+++ b/src/Compilers/Test/Core/Platform/Desktop/Extensions.cs
@@ -28,9 +28,9 @@ namespace Roslyn.Test.Utilities.Desktop
     {
         public static void AddArray<T>(this SerializationInfo info, string name, ImmutableArray<T> value) where T : class
         {
-            // we will copy the content into an array and serialize the copy
-            // we could serialize element-wise, but that would require serializing
-            // name and type for every serialized element which seems worse than creating a copy.
+            // This will store the underlying T[] directly into the SerializationInfo. That is safe because it
+            // only ever reads from the array. This is done instead of creating a copy because it is a 
+            // significant source of allocations in our unit tests.
             info.AddValue(name, value.IsDefault ? null : ImmutableCollectionsMarshal.AsArray(value), typeof(T[]));
         }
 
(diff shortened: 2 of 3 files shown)


src/Compilers/CSharp/Test/Semantic/Semantics/NativeIntegerTests.cs
```diff
@@ -12384,7 +12384,7 @@ void binaryOperator(string op, string leftType, string rightType, string expecte
 
                 if (expectedDiagnostics.Length == 0)
                 {
-                    CompileAndVerify(comp);
+                    CompileAndVerify(comp, emitOptions: EmitOptions.Default);
```
jaredpar, line 12387: By default `CompileAndVerify` embeds portable PDBs and the sheer number of tests here was causing that allocate 100MB of unnecessary `byte[]`.

src/Compilers/Test/Core/Platform/Desktop/Extensions.cs
```diff
@@ -13,6 +13,7 @@
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.InteropServices;
```
jaredpar, line 16: This is a significant source of redundant `byte[]` allocations.

src/Compilers/Test/Core/TargetFrameworkUtil.cs
```diff
@@ -113,13 +113,26 @@ public static class NetCoreApp
...
+        private static ImmutableArray<MetadataReference> s_references;
+
         /// <summary>
         /// This is the full set of references provided by default on the .NET Framework TFM
         /// </summary>
         /// <remarks>
         /// Need to special case tuples until we move to net472
         /// </remarks>
-        public static ImmutableArray<MetadataReference> References { get; } = [.. Net461.References.All, Net461.ExtraReferences.SystemValueTuple];
+        public static ImmutableArray<MetadataReference> References
```
jaredpar, line 124: There are a number of tests that call say `NetFramework.mscorlib`. When this was a static initializer that realized all these values when they didn't really need to be. This is only a savings in certain silces of our tests but it's a significant one in them.

src/Compilers/Test/Core/Platform/Desktop/Extensions.cs
```diff
@@ -30,27 +31,27 @@ public static void AddArray<T>(this SerializationInfo info, string name, Immutab
             // we will copy the content into an array and serialize the copy
             // we could serialize element-wise, but that would require serializing
             // name and type for every serialized element which seems worse than creating a copy.
-            info.AddValue(name, value.IsDefault ? null : value.ToArray(), typeof(T[]));
+            info.AddValue(name, value.IsDefault ? null : ImmutableCollectionsMarshal.AsArray(value), typeof(T[]));
```
AlekseyTs, line 34: Is the comment above still accurate?

src/Compilers/Test/Core/Platform/Desktop/Extensions.cs
```diff
@@ -30,27 +31,27 @@ public static void AddArray<T>(this SerializationInfo info, string name, Immutab
...
+            return ImmutableCollectionsMarshal.AsImmutableArray(array);
         }
 
         public static void AddByteArray(this SerializationInfo info, string name, ImmutableArray<byte> value)
         {
             // we will copy the content into an array and serialize the copy
             // we could serialize element-wise, but that would require serializing
             // name and type for every serialized element which seems worse than creating a copy.
-            info.AddValue(name, value.IsDefault ? null : value.ToArray(), typeof(byte[]));
+            info.AddValue(name, value.IsDefault ? null : ImmutableCollectionsMarshal.AsArray(value), typeof(byte[]));
```
AlekseyTs, line 48: Here as well

jaredpar: Think those comments are somewhat wrong even in the before case. Let me clean that up.

333fred (approved):

AlekseyTs (commented):
LGTM (commit 2)

PR #75172, opened 2024-09-19
Updates content exclusion for on-the-fly-docs

akhera99:
Looks to see if any of the file paths of the syntax nodes that declared the symbol are excluded.
If a user is looking at docs for an excluded symbol, it now returns a new message indicating that content is excluded.

image


src/Features/Core/Portable/QuickInfo/OnTheFlyDocsElement.cs
```diff
@@ -13,11 +13,12 @@ namespace Microsoft.CodeAnalysis.QuickInfo;
 /// <param name="declarationCode">the symbol's declaration code</param>
 /// <param name="language">the language of the symbol</param>
 /// <param name="hasComments">whether the symbol has comments</param>
-internal sealed class OnTheFlyDocsElement(string symbolSignature, ImmutableArray<string> declarationCode, string language, bool hasComments = false)
+internal sealed class OnTheFlyDocsElement(string symbolSignature, ImmutableArray<string> declarationCode, string language, bool hasContentExcluded, bool hasComments = false)
 {
     public string SymbolSignature { get; } = symbolSignature;
     public ImmutableArray<string> DeclarationCode { get; } = declarationCode;
     public string Language { get; } = language;
+    public bool IsContentExcluded { get; set; } = hasContentExcluded;
```
genlu, line 21: nit: change it to `isContentExcluded` for consistency?

genlu:
>         }, LogLevel.Information));

this reminds me, we haven't added some of those telemetry to the dashboard

---
Refers to: src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs:220 in f65aa66. [](commit_id = f65aa667986cbc97edccfa1e02e3db1f2d67b424, deletion_comment = False)

src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs
```diff
@@ -137,6 +137,19 @@ private async Task SetResultTextAsync(ICopilotCodeAnalysisService copilotService
 
         try
         {
+            if (_onTheFlyDocsElement.IsContentExcluded)
+            {
+                SetResultText(EditorFeaturesResources.Your_organization_excluded_a_referenced_file_from_Copilot);
+                CurrentState = OnTheFlyDocsState.Finished;
+                Logger.Log(FunctionId.Copilot_On_The_Fly_Docs_Content_Excluded, KeyValueLogMessage.Create(m =>
+                {
+                    m["SymbolHeaderText"] = _onTheFlyDocsElement.SymbolSignature;
```
genlu, line 146: Not sure we should log details like this, knowing we couldn't display the link due to content exclusion is sufficient

genlu:
>             m["SymbolHeaderText"] = _onTheFlyDocsElement.SymbolSignature;

same here, do we need this in the log?

---
Refers to: src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs:218 in f65aa66. [](commit_id = f65aa667986cbc97edccfa1e02e3db1f2d67b424, deletion_comment = False)

src/Features/CSharp/Portable/QuickInfo/CSharpSemanticQuickInfoProvider.cs
```diff
@@ -182,6 +175,17 @@ protected override NullableFlowState GetNullabilityAnalysis(SemanticModel semant
...
         }
 
+        // Checks to see if any of the files containing the symbol are excluded.
+        var hasContentExcluded = false;
+        var symbolFilePaths = symbol.DeclaringSyntaxReferences.Select(reference => reference.SyntaxTree.FilePath);
+        foreach (var symbolFilePath in symbolFilePaths)
+        {
+            if (await copilotService.IsFileExcludedAsync(symbolFilePath, cancellationToken).ConfigureAwait(false))
+            {
+                hasContentExcluded = true;
```
genlu, line 185: we could bail once hasContentExcluded is flipped to true

genlu:
>                     ClassifiedTextElement.CreateHyperlink(EditorFeaturesResources.Describe_with_Copilot, EditorFeaturesResources.Generate_summary_with_Copilot, () =>

sorry I missed this earlier, but if I understand this correctly, we are still displaying "describe with copilot" link even for symbols we already know are being excluded, and only show the "is excluded" text after user click the link. Is that correct? If so, I think that's not a good experience, I'd suggest we show a message immediately, like "describe with copilot is unavailable since the referenced document is excluded by your org"

---
Refers to: src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs:79 in f900324. [](commit_id = f9003247da6d480e59fc7d1249c7afec7a60f720, deletion_comment = False)

src/EditorFeatures/Core.Wpf/QuickInfo/OnTheFlyDocsView.xaml.cs
```diff
@@ -137,6 +139,14 @@ private async Task SetResultTextAsync(ICopilotCodeAnalysisService copilotService
 
         try
         {
+            //if (_onTheFlyDocsElement.IsContentExcluded)
```
genlu: Remove the comment. Also we are no longer logging  telemetry when symbol is excluded, just  want to make sure its intended

src/EditorFeatures/Core/EditorFeaturesResources.resx
```diff
@@ -953,4 +953,7 @@ Do you want to proceed?</value>
   <data name="An_error_occurred_while_generating_documentation_for_this_code" xml:space="preserve">
     <value>An error occurred while generating documentation for this code.</value>
   </data>
+  <data name="Describe_with_Copilot_is_unavailable_since_the_referenced_document_is_excluded_by_your_organization" xml:space="preserve">
+    <value>Describe with Copilot is unavailable since the referenced document is excluded by your organization</value>
```
genlu, line 957: Do we need quotes here?

genlu:
@akhera99 Could you pls add a screenshot for this change? Thanks

src/EditorFeatures/Core/EditorFeaturesResources.resx
akhera99: Quotes surrounding the value? I don't know why we would?

genlu: No, I meant something like

```suggestion
    <value>'Describe with Copilot' is unavailable since the referenced document is excluded by your organization</value>
```

Because there's this a few lines above

```xml
<data name="Computing_Encapsulate_Field_information" xml:space="preserve">
    <value>Computing 'Encapsulate Field' information</value>
  </data>
```

akhera99: screenshot added to original post

genlu (approved):
:shipit:

PR #75175, opened 2024-09-19
Adding back code to handle LinePosition values correctly

matteo-prosperi:


matteo-prosperi:
@JoeRobich, for visibility

JoeRobich (approved):

PR #75176, opened 2024-09-19
Remove ResultId from RelatedDocumentsHandler

genlu:
Because cursor position would impact the response from IRelatedDocumentsService


dibarbet (approved):
lgtm, probably some tests need to be updated

PR #75177, opened 2024-09-19
More test allocation improvements

jaredpar:


jaredpar:
@dotnet/roslyn-compiler PTAL

src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmClrType.cs
```diff
@@ -188,9 +189,14 @@ private static System.Type GetProxyType(System.Type type)
             return assembly.GetType(proxyName);
         }
 
-        private static ReadOnlyCollection<DkmClrEvalAttribute> GetEvalAttributes(Type type)
+        private static ReadOnlyCollection<DkmClrEvalAttribute> GetEvalAttributes(DkmClrAppDomain appDomain, Type type)
         {
             var reflectionType = ((TypeImpl)type).Type;
+            if (appDomain.TypeToEvalAttributesMap.TryGetValue(reflectionType, out var cachedAttributes))
```
jaredpar, line 195: Actually executing `GetEvalAttributes` is allocation wise very expensive. The DKM layer creates `DkmClrType` instances for the same `Type` many many times. That means we're doing so much unnecessary work. This simple cache saves 200MB allocations in the case the unit test OOMs. Not sure how much it saves in the case where it actually completes

src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmClrAppDomain.cs
```diff
@@ -10,14 +10,20 @@
...
 using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Microsoft.VisualStudio.Debugger.Evaluation.ClrCompilation;
 
 namespace Microsoft.VisualStudio.Debugger.Clr
 {
     public class DkmClrAppDomain
     {
+        public Dictionary<Type, ReadOnlyCollection<DkmClrEvalAttribute>> TypeToEvalAttributesMap { get; }
```
cston, line 21: >public [](http://example.com/codeflow?start=8&length=6)

`internal`

src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmClrType.cs
```diff
@@ -12,6 +12,7 @@
 using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Linq;
+using System.Runtime.InteropServices.WindowsRuntime;
```
cston, line 15: >System.Runtime.InteropServices.WindowsRuntime [](http://example.com/codeflow?start=6&length=45)

Is this `using` needed?

src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmClrAppDomain.cs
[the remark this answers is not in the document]
jaredpar: Is this type used concurrently? My looking through seemed to indicate it was not designed for that but I don't have a lot of context into this code abes.

src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmClrType.cs
jaredpar: why does VS keep adding that???

src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmClrAppDomain.cs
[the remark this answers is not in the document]
cston: I'm not sure if the tests use `Dkm` instances concurrently, although we are using `Interlocked.CompareExchange()` to initialize a couple of lazy members in this assembly. But it seemed safer to allow for concurrent access.

cston (approved):

333fred (approved):

Issue #75182, opened 2024-09-20
MSBuildWorkspace reports warnings as errors/failures


svick:
**Version Used**: Roslyn 4.11.0

**Steps to Reproduce**:

Run this code:

```c#
using Microsoft.CodeAnalysis.MSBuild;

var workspace = MSBuildWorkspace.Create();

var project = await workspace.OpenProjectAsync(@"..\..\..\..\project\project.csproj");

foreach (var diagnostic in workspace.Diagnostics)
{
    Console.WriteLine($"{diagnostic.Kind}: {diagnostic.Message}");
}
```

Where project.csproj is:

```csproj
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
  </ItemGroup>

</Project>
```

**Expected Behavior**:

The workspace reports the `NU1903` warning about a vulnerability as a warning, i.e. the output should be

```
Warning: Msbuild failed when processing the file 'C:\src\tmp\roslynworkspaceapp\project\project.csproj' with message: Package 'Newtonsoft.Json' 12.0.3 has a known high severity vulnerability, https://github.com/advisories/GHSA-5crp-9r3c-p9vr
```

**Actual Behavior**:

The warning is reported as a failure:

```
Failure: Msbuild failed when processing the file 'C:\src\tmp\roslynworkspaceapp\project\project.csproj' with message: Package 'Newtonsoft.Json' 12.0.3 has a known high severity vulnerability, https://github.com/advisories/GHSA-5crp-9r3c-p9vr
```

The problem is in `Microsoft.CodeAnalysis.MSBuild.DiagnosticReporter`, whose `Report(DiagnosticLog log)` method ignores `DiagnosticLogItem.Kind`.

tmat:
Related: https://github.com/dotnet/roslyn/issues/75170

Issue #75170, opened 2024-09-19
Improve MSBuildWorkspace diagnostic reporting


tmat:
`MSBuildWorkspace.WorkspaceFailed` reports `WorkspaceDiagnosticKind.Failure` even for issues that can be safely ignored, such as failure to load project of unregistered extension. 

Internally, `DiagnosticReporter.Report` conflates logging with exception handling. It's unclear why `DiagnosticReportingMode` is needed.
Callers that pass `DiagnosticReportingMode.Throw` should just throw the exception directly.
Callers that pass `DiagnosticReportingMode.Ignore` should not call the method at all - it's a no op.

`MSBuildWorkspace.WorkspaceFailed` should be called with `WorkspaceDiagnosticKind.Failure` only for unrecoverable errors. Presumably, we wouldn't immediately throw in these cases so that we can report all the errors. Failure should be reported if and only if the operation fails. If the operation can succeed (e.g. by ignoring certain projects) a warning should be reported.

PR #75183, opened 2024-09-20
Fix reporting of MSBuild warnings in MSBuildWorkspace

svick:
Fixes https://github.com/dotnet/roslyn/issues/75182

### DIFF
--- a/src/Workspaces/Core/MSBuild/MSBuild/DiagnosticReporter.cs
+++ b/src/Workspaces/Core/MSBuild/MSBuild/DiagnosticReporter.cs
@@ -58,13 +58,24 @@ namespace Microsoft.CodeAnalysis.MSBuild
         {
             foreach (var logItem in log)
             {
-                Report(DiagnosticReportingMode.Log, GetMSBuildFailedMessage(logItem.ProjectFilePath, logItem.ToString()));
+                var diagnosticKind = logItem.Kind is DiagnosticLogItemKind.Warning ? WorkspaceDiagnosticKind.Warning : WorkspaceDiagnosticKind.Failure;
+
+                Report(new WorkspaceDiagnostic(diagnosticKind, GetMSBuildFailedMessage(logItem.ProjectFilePath, logItem.ToString(), diagnosticKind)));
             }
         }
 
-        private static string GetMSBuildFailedMessage(string projectFilePath, string message)
+        private static string GetMSBuildFailedMessage(string projectFilePath, string message, WorkspaceDiagnosticKind diagnosticKind)
             => RoslynString.IsNullOrWhiteSpace(message)
-                ? string.Format(WorkspaceMSBuildResources.Msbuild_failed_when_processing_the_file_0, projectFilePath)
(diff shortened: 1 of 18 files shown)


src/Workspaces/Core/MSBuild/MSBuild/DiagnosticReporter.cs
```diff
@@ -58,13 +58,24 @@ public void Report(IEnumerable<DiagnosticLogItem> log)
...
-                Report(DiagnosticReportingMode.Log, GetMSBuildFailedMessage(logItem.ProjectFilePath, logItem.ToString()));
+                var diagnosticKind = logItem.Kind is DiagnosticLogItemKind.Warning ? WorkspaceDiagnosticKind.Warning : WorkspaceDiagnosticKind.Failure;
+
+                Report(new WorkspaceDiagnostic(diagnosticKind, GetMSBuildFailedMessage(logItem.ProjectFilePath, logItem.ToString(), diagnosticKind)));
             }
         }
 
-        private static string GetMSBuildFailedMessage(string projectFilePath, string message)
+        private static string GetMSBuildFailedMessage(string projectFilePath, string message, WorkspaceDiagnosticKind diagnosticKind)
             => RoslynString.IsNullOrWhiteSpace(message)
```
jasonmalinowski, line 68: It might just be easier to fill in a blank message with "no message" or something so the log entry would be a bit clearer we're missing one (and it reduces the number of strings we have). That said, happy to keep this as is since you're just maintaining behavior.

src/Workspaces/Core/MSBuild/WorkspaceMSBuildResources.resx
```diff
@@ -133,10 +133,16 @@
...
   <data name="Msbuild_failed_when_processing_the_file_0" xml:space="preserve">
-    <value>Msbuild failed when processing the file '{0}'</value>
+    <value>MSBuild failed when processing the file '{0}'</value>
   </data>
   <data name="Msbuild_failed_when_processing_the_file_0_with_message_1" xml:space="preserve">
-    <value>Msbuild failed when processing the file '{0}' with message: {1}</value>
+    <value>MSBuild failed when processing the file '{0}' with message: {1}</value>
+  </data>
+  <data name="Msbuild_warning_when_processing_the_file_0" xml:space="preserve">
+    <value>MSBuild warning when processing the file '{0}'</value>
```
JoeRobich, line 142: I think we should change this string to `MSBuild warned ...` to keep it inline with the `MSBuild failed ...` string.

tmat:
Will this change fix https://github.com/dotnet/roslyn/issues/75170?

src/Workspaces/Core/MSBuild/WorkspaceMSBuildResources.resx
svick: Done.

svick:
@tmat I'm not sure, but probably not.

If the issues are originally reported as warnings, then this PR will make sure they will end up as warnings. Otherwise, it won't change anything.

PR #75191, opened 2024-09-21
EnC: allow changing accessibility of members

tmat:


tmat:
@davidwengier PTAL. Will only merge if the linked Mono issue is fixed.

davidwengier (approved):

PR #75202, opened 2024-09-21
Optimise allocations in `AssertEx.NormalizeWhitespace` and `OperationTreeVerifier.Verify`

hamarb123:
They should now be amortised allocation free (in the assert succeed case). I have also removed pointless allocations for indentation I noticed in `OperationTreeVerifier`.

Works by using char[] buffers from `ArrayPool<char>.Shared` (along with `int[]` ones in some parts too), and checking the test condition manually (instead of via the `Assert` type), then calling into `Assert` if it fails with a string from `.ToString`. Buffers are returned after the assertion condition does not fail.

Please let me know if there's any parts which don't follow style rules that I missed, or need more comments, etc. :)

/cc @jaredpar who wanted these optimised for allocations.


src/Compilers/Test/Core/Compilation/OperationTreeVerifier.cs
```diff
@@ -62,13 +63,186 @@ public static string GetOperationTree(Compilation compilation, IOperation operat
...
-            expectedOperationTree = expectedOperationTree.Trim(newLineChars);
-            expectedOperationTree = expectedOperationTree.Replace("\r\n", "\n").Replace(" \n", "\n").Replace("\n", Environment.NewLine);
+            // Finds spaces before newlines and places them in the specified toRemove buffer in reverse index order
+            static void FindSpacesBeforeNewlineHelper(char[] buffer, int bufferSize, ref int[] toRemove, ref int toRemoveCount)
+            {
+                int i = bufferSize;
+                while (true)
+                {
+                    // Go to next preceding \r or \n (if any)
+                    i = buffer.AsSpan().Slice(0, i).LastIndexOfAny("\r\n".AsSpan());
```
huoyaoyuan, line 73: Nit: `AsSpan(0, i)` (and other places)

src/Compilers/Test/Core/Compilation/OperationTreeVerifier.cs
```diff
@@ -62,13 +63,186 @@ public static string GetOperationTree(Compilation compilation, IOperation operat
 
         public static void Verify(string expectedOperationTree, string actualOperationTree)
         {
-            char[] newLineChars = Environment.NewLine.ToCharArray();
-            string actual = actualOperationTree.Trim(newLineChars);
-            actual = actual.Replace(" \n", "\n").Replace(" \r", "\r");
-            expectedOperationTree = expectedOperationTree.Trim(newLineChars);
-            expectedOperationTree = expectedOperationTree.Replace("\r\n", "\n").Replace(" \n", "\n").Replace("\n", Environment.NewLine);
+            // Finds spaces before newlines and places them in the specified toRemove buffer in reverse index order
+            static void FindSpacesBeforeNewlineHelper(char[] buffer, int bufferSize, ref int[] toRemove, ref int toRemoveCount)
```
huoyaoyuan, line 67: I really wonder whether the complex dance is measurably better than doing manipulation in `StringBuilder`.

hamarb123: @jaredpar does this PR measurably improve your scenario?

What would the allocations for `StringBuilder.Replace` on .NET Framework be like @huoyaoyuan (which @jaredpar was concerned about iirc)? I'm not familiar with how it performs, as I usually am just `.Append`ing on `StringBuilder`s when I use them.

It seems like you'd at least have ~2x total string length in total, to store StringBuilder buffer & result string, whereas the current one is 3x for actual, and 4x for expected - so it seems like it could be at most ~2x better in the best scenario. Whereas this should be amortised allocation free (but obviously has a bit more complexity in implementation as a result).

jaredpar: > @jaredpar does this PR measurably improve your scenario?

Going to be profiling this probably Tuesday. Monday tends to be pretty meeting heavy for me. Hoping to have time to profile this out tomorrow when I'm back working on tests.

src/Compilers/Test/Core/Assert/AssertEx.cs
```diff
@@ -594,37 +616,117 @@ public static void AssertResultsEqual(string result1, string result2)
...
+            if (!actualString.StartsWith(expectedSubString, StringComparison.Ordinal))
+            {
+                // Call through to actual API to get same error message experience
+                Assert.StartsWith(expectedSubString.ToString(), actualString.ToString(), StringComparison.Ordinal);
+            }
+
+            if (pooled1 != null)
+            {
+                ArrayPool<char>.Shared.Return(pooled1);
+            }
```
jaredpar, line 662: I understand this pattern but it's also feels a bit much to me. Too much manual work. I'm wondering if we should instead approach it with something like the following: 

```csharp
struct ArrayPoolValue<T> : IDisposable
{
  private T[] _array;
  public T[] T Array
  {
    get => _array;
    set
    {
       if (_array is not null)
        ArrayPool<T>.Shared.Return(_array);
       _array = value;
    }
  }
  public void Dispose()
  {
    if (_array is not null)
    {
        ArrayPool<T>.Shared.Return(_array);
        _array = null;
    }
  }
}
```

Then we could change the signatures  to take `ref ArrayPoolValue<T> pool`. Consumption becomes simply: 

```csharp
using var pooled1 = new ArrayPool<char>;
```

@333fred for thoughts

src/Compilers/Test/Core/Assert/AssertEx.cs
```diff
@@ -594,37 +616,117 @@ public static void AssertResultsEqual(string result1, string result2)
...
+        }
+
+        internal static ReadOnlySpan<char> NormalizeWhitespace(ReadOnlySpan<char> input, out char[] pooledAlloc)
+        {
+            return NormalizeWhitespaceROM(input, out pooledAlloc).Span;
         }
 
-        internal static string NormalizeWhitespace(string input)
+        internal static ReadOnlyMemory<char> NormalizeWhitespaceROM(ReadOnlySpan<char> input, out char[] pooledAlloc)
```
jaredpar, line 674: If we do suffix on the name we'd prefer the longer form vs. the abbreviation. Overall though I think I'd prefer we just have the one overload that returns `ReadOnlyMemory<char>` and then let the consumers do `.Span` as needed. That is a bit of ceremony but it's one that we've already done in a few other places.

src/Compilers/Test/Core/Assert/AssertEx.cs
```diff
@@ -594,37 +616,117 @@ public static void AssertResultsEqual(string result1, string result2)
...
+                // Find next section
+                prev = next + 1;
+                next = input.Slice(prev).IndexOfAny("\r\n".AsSpan());
+                if (next >= 0) next += prev;
+                else next = input.Length;
             }
 
-            return output.ToString();
+            // Return the valid part of our allocation
+            if (pooledAlloc == null) return default;
```
jaredpar, line 728: Nit: the style in roslyn requires that this be put on separate lines. There are a few instances of this in the PR that need attention.

src/Compilers/Test/Core/Assert/AssertEx.cs
```diff
@@ -594,37 +616,117 @@ public static void AssertResultsEqual(string result1, string result2)
...
-            foreach (var line in inputLines)
+            // Setup allocation (start with none)
+            pooledAlloc = null;
+            int allocSize = 0;
+
+            // Loop over each line section until we run out of lines
+            int prev = 0;
+            int next = input.IndexOfAny("\r\n".AsSpan());
+            if (next < 0) next = input.Length;
+            while (true)
```
jaredpar, line 684: Reading through this I'm wondering if it should be broken up a bit. Basically should we have one method that iterates through a `ReadOnlySpan<char>` and gives us back `ReadOnlyMemory<char>` for the lines within it?

src/Compilers/Test/Core/Assert/AssertEx.cs
```diff
@@ -594,37 +616,117 @@ public static void AssertResultsEqual(string result1, string result2)
...
-                    output.AppendLine(trimmedLine);
+                    // Ensure buffer size is big enough
+                    int newChars = spacesToAppend + trimmedLine.Length + Environment.NewLine.Length;
+                    pooledAlloc ??= ArrayPool<char>.Shared.Rent(newChars);
+                    if (pooledAlloc.Length < allocSize + newChars)
+                    {
+                        char[] newArr = ArrayPool<char>.Shared.Rent(allocSize + newChars);
+                        pooledAlloc.AsSpan().Slice(0, allocSize).CopyTo(newArr.AsSpan().Slice(0, allocSize));
+                        ArrayPool<char>.Shared.Return(pooledAlloc);
+                        pooledAlloc = newArr;
```
jaredpar, line 705: This pattern of basically increasing the size of the pooled array is repeated a few times in the PR. Think we should factor this out into a helper method.

src/Compilers/Test/Core/Assert/AssertEx.cs
```diff
@@ -594,37 +616,117 @@ public static void AssertResultsEqual(string result1, string result2)
...
+                    }
+
+                    // Copy new text into buffer
+                    pooledAlloc.AsSpan().Slice(allocSize, spacesToAppend).Fill(' ');
```
jaredpar, line 709:
```suggestion
                    pooledAlloc.AsSpan(allocSize, spacesToAppend).Fill(' ');
```

src/Compilers/Test/Core/Assert/AssertEx.cs
huoyaoyuan: `MemoryPool<T>` and `IMemoryOwner<T>` does exactly that, but with overhead of extra abstraction through `Memory<T>`.

333fred: I don't think I care that much. These helpers are presumably not touched much, and I don't think we care about ensuring it's always returned?

src/Compilers/Test/Core/Compilation/OperationTreeVerifier.cs
```diff
@@ -62,13 +63,186 @@ public static string GetOperationTree(Compilation compilation, IOperation operat
 
         public static void Verify(string expectedOperationTree, string actualOperationTree)
         {
-            char[] newLineChars = Environment.NewLine.ToCharArray();
-            string actual = actualOperationTree.Trim(newLineChars);
-            actual = actual.Replace(" \n", "\n").Replace(" \r", "\r");
-            expectedOperationTree = expectedOperationTree.Trim(newLineChars);
-            expectedOperationTree = expectedOperationTree.Replace("\r\n", "\n").Replace(" \n", "\n").Replace("\n", Environment.NewLine);
+            // Finds spaces before newlines and places them in the specified toRemove buffer in reverse index order
```
333fred, line 66: Local functions go at the end of the method for our style.

src/Compilers/Test/Core/Compilation/OperationTreeVerifier.cs
```diff
@@ -62,13 +63,186 @@ public static string GetOperationTree(Compilation compilation, IOperation operat
...
+                    // Perform removal
+                    RemoveHelper(ref buffer, ref bufferSize, toRemove, toRemoveCount);
+                }
+                else
+                {
+                    throw new NotSupportedException("Only supported line endings are lf and crlf (OperationTreeVerifier.Verify)");
+                }
+            }
+
+            // Emulate actualOperationTree.Trim(newLineChars).Replace(" \n", "\n").Replace(" \r", "\r") into actualBuffer with actualBufferSize
```
333fred, line 216: I don't think I have the brain power to review this today, but I'm going to insist that we actually spell out the general algorithm going on here in more plain english.

src/Compilers/Test/Core/Assert/AssertEx.cs
hamarb123: > and I don't think we care about ensuring it's always returned

No, it's not a correctness issue to not return it. Unless assertion failure is expected (e.g., if you have a bunch of tests like `bool didError = false; try { ... calls Assert(); } catch { didError = true; } Assert.True(didError);`), there's not really any reason to worry about ensuring it imo.

> Then we could change the signatures to take `ref ArrayPoolValue<T>` pool. Consumption becomes simply:
> `using var pooled1 = new ArrayPool<char>;`

You can't pass `using` variables by reference, we will very quickly run into this issue I think.

src/Compilers/Test/Core/Assert/AssertEx.cs
hamarb123: Do you mean another inline method, or in some helper type? If you mean in a helper type, which one, or where would I create it?

src/Compilers/Test/Core/Assert/AssertEx.cs
hamarb123: I'm not exactly sure what you mean by this?

jcouv:
It looks like there was no activity on this PR for a long while. Moving to draft to make room in our review queue. Please re-publish and ping us to drive this forward if it's ready

hamarb123:
@jcouv I was waiting for @jaredpar to confirm that it did measurably improve his scenario (plus some questions I asked about some feedback), and then I forgot about this PR after some time lol.

jcouv:
@hamarb123 Since Jared's comment in 2024, we've moved to a policy of requiring some kind of benchmark (evidence of the performance improvement) when submitting optimization PRs. I don't think that Jared or the team will have the bandwidth to gather that info. For instance, showing the impact on end-to-end test runtime.

hamarb123:
I mean, it was jared that wanted me to make the PR to start with because he thought it was impactful, so if he does not have time to measure it, then I think I'm happy to just close it.

jcouv:
FWIW, here's what I observed for `Emit3` ([script](https://gist.github.com/jcouv/f01fccd9c114838d7c2c55e269124e1f)):
Baseline 28d629d
   - Wall clock: 14:13.523
   - Test line elapsed: 14:10.785
   - Tree CPU: 606.719s
   - Peak working set: 1755.7 MB
   - Peak private bytes: 1318.1 MB

  PR eeebea4
   - Wall clock: 13:25.181
   - Test line elapsed: 13:25.084
   - Tree CPU: 582.750s
   - Peak working set: 1650.8 MB
   - Peak private bytes: 1218.5 MB

  Delta for PR
   - Wall clock: -48.342s (-5.66%)
   - CPU: -23.969s (-3.95%)
   - Peak working set: -105 MB (-5.98%)
   - Peak private bytes: -99.6 MB (-7.55%)

hamarb123:
Ah nice, so ~5% savings for that test across each of the metrics you measured - lmk if you want me to reopen the PR :)

phil-allen-msft:
@hamarb123 , these do look like good improvements. Feel free to reopen.

hamarb123:
I'm closing this, as it seems like if any push or similar were to occur it would immediately waste 40% of the monthly copilot budget I have on this account, and it does not seem like anyone is that worried about it.

Issue #73439, opened 2024-05-12
`System.InsufficientExecutionStackException` thrown in `LocalBinderFactory` compiling `ErrorFacts.IsBuildOnlyDiagnostic()` on MacOS


cston:
**Version Used**: 

**Steps to Reproduce**:

1. Build dotnet/roslyn Compilers.sln on MacOS.

**Expected Behavior**:

Compilation succeeds.

**Actual Behavior**:

`System.InsufficientExecutionStackException` thrown compiling `ErrorFacts.IsBuildOnlyDiagnostic()`

>$HOME/.nuget/packages/microsoft.net.compilers.toolset/4.10.0-1.24061.4/tasks/netcore/Microsoft.CSharp.Core.targets(84,5): error : Process terminated. System.InsufficientExecutionStackException: Insufficient stack to continue executing the program safely. This can happen from having too many functions on the call stack or function on the stack using too much stack space. [/dotnet/roslyn/src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj::TargetFramework=net7.0]
    at System.Runtime.CompilerServices.RuntimeHelpers.EnsureSufficientExecutionStack() [/dotnet/roslyn/src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj::TargetFramework=net7.0]
    at Microsoft.CodeAnalysis.CSharp.CSharpSyntaxWalker.Visit(SyntaxNode node) [/dotnet/roslyn/src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj::TargetFramework=net7.0]
    at Microsoft.CodeAnalysis.CSharp.CSharpSyntaxWalker.DefaultVisit(SyntaxNode node) [/dotnet/roslyn/src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj::TargetFramework=net7.0]
    at Microsoft.CodeAnalysis.CSharp.LocalBinderFactory.DefaultVisit(SyntaxNode node) [/dotnet/roslyn/src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj::TargetFramework=net7.0]
[...]

PR #75212, opened 2024-09-23
Use manual recursion for binary patterns

333fred:
Much like binary operators, we can have deeply-nested binary patterns. An example of this is our own IsBuildOnlyDiagnostic, which has ~2500 binary patterns in a single arm, and growing every release. To avoid stack depth issues, I took the same approach we do for binary operators; rewrite recursion to use a manual stack for these cases. Fixes https://github.com/dotnet/roslyn/issues/73439.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
@@ -1734,99 +1734,137 @@ namespace Microsoft.CodeAnalysis.CSharp
             bool hasErrors,
             BindingDiagnosticBag diagnostics)
         {
-            bool isDisjunction = node.Kind() == SyntaxKind.OrPattern;
-            if (isDisjunction)
+            // Users (such as ourselves) can have many, many nested binary patterns. To avoid crashing, do left recursion manually.
+
+            var binaryPatternStack = ArrayBuilder<BinaryPatternSyntax>.GetInstance();
+            BinaryPatternSyntax? currentNode = node;
+
+            do
             {
-                MessageID.IDS_FeatureOrPattern.CheckFeatureAvailability(diagnostics, node.OperatorToken);
+                binaryPatternStack.Push(currentNode);
+                currentNode = currentNode.Left as BinaryPatternSyntax;
+            } while (currentNode != null);
 
-                permitDesignations = false; // prevent designators under 'or'
-                var left = BindPattern(node.Left, inputType, permitDesignations, hasErrors, diagnostics);
-                var right = BindPattern(node.Right, inputType, permitDesignations, hasErrors, diagnostics);
+            Debug.Assert(binaryPatternStack.Count > 0);
 
-                // Compute the common type. This algorithm is quadratic, but disjunctive patterns are unlikely to be huge
-                var narrowedTypeCandidates = ArrayBuilder<TypeSymbol>.GetInstance(2);
-                collectCandidates(left, narrowedTypeCandidates);
-                collectCandidates(right, narrowedTypeCandidates);
-                var narrowedType = leastSpecificType(node, narrowedTypeCandidates, diagnostics) ?? inputType;
-                narrowedTypeCandidates.Free();
+            BoundPattern? result = null;
 
-                return new BoundBinaryPattern(node, disjunction: isDisjunction, left, right, inputType: inputType, narrowedType: narrowedType, hasErrors);
-
-                static void collectCandidates(BoundPattern pat, ArrayBuilder<TypeSymbol> candidates)
+            while (binaryPatternStack.TryPop(out var binaryPattern))
+            {
+                if (result == null)
                 {
-                    if (pat is BoundBinaryPattern { Disjunction: true } p)
-                    {
-                        collectCandidates(p.Left, candidates);
-                        collectCandidates(p.Right, candidates);
-                    }
-                    else
-                    {
-                        candidates.Add(pat.NarrowedType);
-                    }
+                    Debug.Assert(binaryPattern.Left is not BinaryPatternSyntax);
+                    result = BindPattern(binaryPattern.Left, inputType, permitDesignations, hasErrors, diagnostics);
                 }
 
-                TypeSymbol? leastSpecificType(SyntaxNode node, ArrayBuilder<TypeSymbol> candidates, BindingDiagnosticBag diagnostics)
+                result = bindBinaryPattern(result, this, binaryPattern, inputType, permitDesignations, hasErrors, diagnostics);
+            }
+
+            binaryPatternStack.Free();
+            Debug.Assert(result != null);
+            return result;
+
+            static BoundPattern bindBinaryPattern(
+                BoundPattern preboundLeft,
+                Binder binder,
+                BinaryPatternSyntax node,
+                TypeSymbol inputType,
+                bool permitDesignations,
+                bool hasErrors,
+                BindingDiagnosticBag diagnostics)
+            {
+                bool isDisjunction = node.Kind() == SyntaxKind.OrPattern;
+                if (isDisjunction)
                 {
-                    Debug.Assert(candidates.Count >= 2);
-                    CompoundUseSiteInfo<AssemblySymbol> useSiteInfo = GetNewCompoundUseSiteInfo(diagnostics);
-                    TypeSymbol? bestSoFar = candidates[0];
-                    // first pass: select a candidate for which no other has been shown to be an improvement.
-                    for (int i = 1, n = candidates.Count; i < n; i++)
+                    MessageID.IDS_FeatureOrPattern.CheckFeatureAvailability(diagnostics, node.OperatorToken);
+
+                    permitDesignations = false; // prevent designators under 'or'
+                    var right = binder.BindPattern(node.Right, inputType, permitDesignations, hasErrors, diagnostics);
+
+                    // Compute the common type. This algorithm is quadratic, but disjunctive patterns are unlikely to be huge
+                    var narrowedTypeCandidates = ArrayBuilder<TypeSymbol>.GetInstance(2);
+                    collectCandidates(preboundLeft, narrowedTypeCandidates);
+                    collectCandidates(right, narrowedTypeCandidates);
+                    var narrowedType = leastSpecificType(node, narrowedTypeCandidates, diagnostics) ?? inputType;
+                    narrowedTypeCandidates.Free();
+
+                    return new BoundBinaryPattern(node, disjunction: isDisjunction, preboundLeft, right, inputType: inputType, narrowedType: narrowedType, hasErrors);
+
+                    static void collectCandidates(BoundPattern pat, ArrayBuilder<TypeSymbol> candidates)
                     {
-                        TypeSymbol candidate = candidates[i];
-                        bestSoFar = lessSpecificCandidate(bestSoFar, candidate, ref useSiteInfo) ?? bestSoFar;
-                    }
-                    // second pass: check that it is no more specific than any candidate.
-                    for (int i = 0, n = candidates.Count; i < n; i++)
-                    {
-                        TypeSymbol candidate = candidates[i];
-                        TypeSymbol? spoiler = lessSpecificCandidate(candidate, bestSoFar, ref useSiteInfo);
-                        if (spoiler is null)
+                        if (pat is BoundBinaryPattern { Disjunction: true } p)
                         {
-                            bestSoFar = null;
-                            break;
+                            collectCandidates(p.Left, candidates);
+                            collectCandidates(p.Right, candidates);
+                        }
+                        else
+                        {
+                            candidates.Add(pat.NarrowedType);
+                        }
+                    }
+
+                    TypeSymbol? leastSpecificType(SyntaxNode node, ArrayBuilder<TypeSymbol> candidates, BindingDiagnosticBag diagnostics)
+                    {
+                        Debug.Assert(candidates.Count >= 2);
+                        CompoundUseSiteInfo<AssemblySymbol> useSiteInfo = binder.GetNewCompoundUseSiteInfo(diagnostics);
+                        TypeSymbol? bestSoFar = candidates[0];
+                        // first pass: select a candidate for which no other has been shown to be an improvement.
+                        for (int i = 1, n = candidates.Count; i < n; i++)
+                        {
+                            TypeSymbol candidate = candidates[i];
+                            bestSoFar = lessSpecificCandidate(bestSoFar, candidate, ref useSiteInfo) ?? bestSoFar;
+                        }
+                        // second pass: check that it is no more specific than any candidate.
+                        for (int i = 0, n = candidates.Count; i < n; i++)
+                        {
+                            TypeSymbol candidate = candidates[i];
+                            TypeSymbol? spoiler = lessSpecificCandidate(candidate, bestSoFar, ref useSiteInfo);
+                            if (spoiler is null)
+                            {
+                                bestSoFar = null;
+                                break;
+                            }
+
+                            // Our specificity criteria are transitive
+                            Debug.Assert(spoiler.Equals(bestSoFar, TypeCompareKind.ConsiderEverything));
                         }
 
-                        // Our specificity criteria are transitive
-                        Debug.Assert(spoiler.Equals(bestSoFar, TypeCompareKind.ConsiderEverything));
+                        diagnostics.Add(node, useSiteInfo);
+                        return bestSoFar;
                     }
 
-                    diagnostics.Add(node, useSiteInfo);
-                    return bestSoFar;
+                    // Given a candidate least specific type so far, attempt to refine it with a possibly less specific candidate.
+                    TypeSymbol? lessSpecificCandidate(TypeSymbol bestSoFar, TypeSymbol possiblyLessSpecificCandidate, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+                    {
+                        if (bestSoFar.Equals(possiblyLessSpecificCandidate, TypeCompareKind.AllIgnoreOptions))
+                        {
+                            // When the types are equivalent, merge them.
+                            return bestSoFar.MergeEquivalentTypes(possiblyLessSpecificCandidate, VarianceKind.Out);
+                        }
+                        else if (binder.Conversions.HasImplicitReferenceConversion(bestSoFar, possiblyLessSpecificCandidate, ref useSiteInfo))
+                        {
+                            // When there is an implicit reference conversion from T to U, U is less specific
+                            return possiblyLessSpecificCandidate;
+                        }
+                        else if (binder.Conversions.HasBoxingConversion(bestSoFar, possiblyLessSpecificCandidate, ref useSiteInfo))
+                        {
+                            // when there is a boxing conversion from T to U, U is less specific.
+                            return possiblyLessSpecificCandidate;
+                        }
+                        else
+                        {
+                            // We have no improved candidate to offer.
+                            return null;
+                        }
+                    }
                 }
-
-                // Given a candidate least specific type so far, attempt to refine it with a possibly less specific candidate.
-                TypeSymbol? lessSpecificCandidate(TypeSymbol bestSoFar, TypeSymbol possiblyLessSpecificCandidate, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+                else
                 {
-                    if (bestSoFar.Equals(possiblyLessSpecificCandidate, TypeCompareKind.AllIgnoreOptions))
-                    {
-                        // When the types are equivalent, merge them.
-                        return bestSoFar.MergeEquivalentTypes(possiblyLessSpecificCandidate, VarianceKind.Out);
-                    }
-                    else if (Conversions.HasImplicitReferenceConversion(bestSoFar, possiblyLessSpecificCandidate, ref useSiteInfo))
-                    {
-                        // When there is an implicit reference conversion from T to U, U is less specific
-                        return possiblyLessSpecificCandidate;
-                    }
-                    else if (Conversions.HasBoxingConversion(bestSoFar, possiblyLessSpecificCandidate, ref useSiteInfo))
-                    {
-                        // when there is a boxing conversion from T to U, U is less specific.
-                        return possiblyLessSpecificCandidate;
-                    }
-                    else
-                    {
-                        // We have no improved candidate to offer.
-                        return null;
-                    }
-                }
-            }
-            else
-            {
-                MessageID.IDS_FeatureAndPattern.CheckFeatureAvailability(diagnostics, node.OperatorToken);
+                    MessageID.IDS_FeatureAndPattern.CheckFeatureAvailability(diagnostics, node.OperatorToken);
 
-                var left = BindPattern(node.Left, inputType, permitDesignations, hasErrors, diagnostics);
-                var right = BindPattern(node.Right, left.NarrowedType, permitDesignations, hasErrors, diagnostics);
-                return new BoundBinaryPattern(node, disjunction: isDisjunction, left, right, inputType: inputType, narrowedType: right.NarrowedType, hasErrors);
+                    var right = binder.BindPattern(node.Right, preboundLeft.NarrowedType, permitDesignations, hasErrors, diagnostics);
+                    return new BoundBinaryPattern(node, disjunction: isDisjunction, preboundLeft, right, inputType: inputType, narrowedType: right.NarrowedType, hasErrors);
+                }
             }
         }
     }
--- a/src/Compilers/CSharp/Portable/Binder/DecisionDagBuilder.cs
+++ b/src/Compilers/CSharp/Portable/Binder/DecisionDagBuilder.cs
@@ -645,32 +645,70 @@ namespace Microsoft.CodeAnalysis.CSharp
             out BoundDagTemp output,
             ArrayBuilder<BoundPatternBinding> bindings)
         {
-            var builder = ArrayBuilder<Tests>.GetInstance(2);
-            if (bin.Disjunction)
+            // Users (such as ourselves) can have many, many nested binary patterns. To avoid crashing, do left recursion manually.
+
+            var binaryPatternStack = ArrayBuilder<BoundBinaryPattern>.GetInstance();
+            var currentNode = bin;
+
+            do
             {
-                builder.Add(MakeTestsAndBindings(input, bin.Left, bindings));
-                builder.Add(MakeTestsAndBindings(input, bin.Right, bindings));
-                var result = Tests.OrSequence.Create(builder);
-                if (bin.InputType.Equals(bin.NarrowedType))
+                binaryPatternStack.Push(currentNode);
+                currentNode = currentNode.Left as BoundBinaryPattern;
+            } while (currentNode != null);
+
+            Tests? result = null;
+            BoundDagTemp? currentOutput = null;
+
+            while (binaryPatternStack.TryPop(out currentNode))
+            {
+                if (result == null)
                 {
-                    output = input;
-                    return result;
+                    Debug.Assert(currentNode.Left is not BoundBinaryPattern);
+                    Debug.Assert(currentOutput == null);
+                    result = MakeTestsAndBindings(input, currentNode.Left, out currentOutput, bindings);
+                }
+
+                Debug.Assert(currentOutput != null);
+                result = makeTestsAndBindingsForBinaryPattern(this, result, currentOutput, input, currentNode, out currentOutput, bindings);
+            }
+
+            Debug.Assert(result != null);
+            Debug.Assert(currentOutput != null);
+            output = currentOutput;
+
+            binaryPatternStack.Free();
+
+            return result;
+
+            static Tests makeTestsAndBindingsForBinaryPattern(DecisionDagBuilder @this, Tests leftTests, BoundDagTemp leftOutput, BoundDagTemp input, BoundBinaryPattern bin, out BoundDagTemp output, ArrayBuilder<BoundPatternBinding> bindings)
+            {
+                var builder = ArrayBuilder<Tests>.GetInstance(2);
+                if (bin.Disjunction)
+                {
+                    builder.Add(leftTests);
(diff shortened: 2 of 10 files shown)


333fred:
@dotnet/roslyn-compiler for review. I cannot stress strongly enough, review with whitespace off. Otherwise all the extractions to local functions will be much harder to review.

333fred:
Also note: A much simpler workaround would just be to break up IsBuildOnlyDiagnostic into a few different pattern arms. I didn't take that approach because that would just leave a maintenance task on us every so often to fix up the compilation when some builds randomly start failing; this seems much less painful long term, and I don't think the original code is unreasonable.

333fred:
Converting this back to draft while I find a few more places that binary operators are handled that we probably need to handle patterns as well.

333fred:
Alright. There are a couple of locations in the optimizer and emit where we do specially handle binary operators, but we these nodes don't survive the local rewriter, and there are existing assertions that we don't have any patterns at that point. This is ready for review.

AlekseyTs:
@333fred It looks like there are legitimate test failures

333fred:
@AlekseyTs I believe they should be fixed now.

AlekseyTs:
@333fred It looks like EndToEnd tests are crashing, likely related to the added test

src/Compilers/Core/Portable/Operations/ControlFlowGraphBuilder.cs
```diff
@@ -7559,15 +7559,42 @@ public override IOperation VisitRelationalPattern(IRelationalPatternOperation op
...
+            IBinaryPatternOperation? current = operation;
+
+            do
+            {
+                stack.Push(current);
+                current = current.LeftPattern as IBinaryPatternOperation;
+            } while (current != null);
+
+            var result = (IPatternOperation)VisitRequired(stack.Peek().LeftPattern);
+            while (stack.TryPop(out current))
```
AlekseyTs, line 7578: >while (stack.TryPop(out current)) [](http://example.com/codeflow?start=12&length=33)

It looks like the stack is never empty on entry, therefore consider using a loop with post-condition instead:
```
            do
            {
                current = stack.Pop()
                result = createOperation(this, current, result);
            }
            while (stack.Any());
```

AlekseyTs: Or perhaps:
```
current = stack.Pop();
var result = (IPatternOperation)VisitRequired(current.LeftPattern);
do
{
    result = createOperation(this, current, result);
}
while (stack.TryPop(out current));

```

src/Compilers/CSharp/Test/EndToEnd/EndToEndTests.cs
```diff
@@ -723,5 +725,57 @@ class XAttribute : System.Attribute { }
             Assert.Collection(runResult.TrackedSteps["result_ForAttributeWithMetadataName"],
                 step => Assert.True(step.Outputs.Single().Value is ClassDeclarationSyntax { Identifier.ValueText: "C1" }));
         }
+
+        [Fact]
+        public void ManyBinaryPatterns()
+        {
+            const string Preamble = $"""
```
AlekseyTs, line 732: >Preamble [](http://example.com/codeflow?start=25&length=8)

It looks like naming conventions for locals are not followed

333fred: This what our const convention is.

AlekseyTs: > This what our const convention is. 

I don't think I agree. These are locals and, therefore, shouldn't start with capital letter.

AlekseyTs: There is a bunch of existing constants like that in this file

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
```diff
@@ -1734,99 +1734,139 @@ private BoundPattern BindBinaryPattern(
...
-                collectCandidates(right, narrowedTypeCandidates);
-                var narrowedType = leastSpecificType(node, narrowedTypeCandidates, diagnostics) ?? inputType;
-                narrowedTypeCandidates.Free();
+            Debug.Assert(binaryPatternStack.Count > 0);
 
-                return new BoundBinaryPattern(node, disjunction: isDisjunction, left, right, inputType: inputType, narrowedType: narrowedType, hasErrors);
+            BoundPattern? result = null;
 
-                static void collectCandidates(BoundPattern pat, ArrayBuilder<TypeSymbol> candidates)
+            while (binaryPatternStack.TryPop(out var binaryPatternAndPermitDesignations))
```
AlekseyTs, line 1753: >while (binaryPatternStack.TryPop(out var binaryPatternAndPermitDesignations)) [](http://example.com/codeflow?start=12&length=77)

Consider using a loop with post-condition instead

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
```diff
@@ -1734,99 +1734,139 @@ private BoundPattern BindBinaryPattern(
...
-                    {
-                        collectCandidates(p.Left, candidates);
-                        collectCandidates(p.Right, candidates);
-                    }
-                    else
-                    {
-                        candidates.Add(pat.NarrowedType);
-                    }
+                    Debug.Assert(binaryPattern.Left is not BinaryPatternSyntax);
+                    result = BindPattern(binaryPattern.Left, inputType, permitDesignations, hasErrors, diagnostics);
```
AlekseyTs, line 1759: >result = BindPattern(binaryPattern.Left, inputType, permitDesignations, hasErrors, diagnostics); [](http://example.com/codeflow?start=20&length=96)

I think the logic could be simpler and much easier to follow if this conditional `result` initialization was done unconditionally before the loop. The node to bind could be "picked" from the stack, or could be stored in a local by the previous loop.

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
```diff
@@ -1734,99 +1734,139 @@ private BoundPattern BindBinaryPattern(
...
-                    // first pass: select a candidate for which no other has been shown to be an improvement.
-                    for (int i = 1, n = candidates.Count; i < n; i++)
+                    Debug.Assert(!permitDesignations);
+                    MessageID.IDS_FeatureOrPattern.CheckFeatureAvailability(diagnostics, node.OperatorToken);
+
+                    var right = binder.BindPattern(node.Right, inputType, permitDesignations, hasErrors, diagnostics);
+
+                    // Compute the common type. This algorithm is quadratic, but disjunctive patterns are unlikely to be huge
+                    var narrowedTypeCandidates = ArrayBuilder<TypeSymbol>.GetInstance(2);
+                    collectCandidates(preboundLeft, narrowedTypeCandidates);
```
AlekseyTs, line 1788: >collectCandidates(preboundLeft, narrowedTypeCandidates); [](http://example.com/codeflow?start=20&length=56)

Perhaps the assumption from the comment above that "disjunctive patterns are unlikely to be huge" has been proven wrong. Therefore, it might make sense keep the set of candidate types from the left as we bubble up, instead of keep revisiting the subtree over and over again. Also, it might be worth considering to collect only distinct types to decrease amount of memory used.

src/Compilers/CSharp/Portable/Binder/ExpressionVariableFinder.cs
```diff
@@ -339,6 +339,28 @@ public override void VisitBinaryExpression(BinaryExpressionSyntax node)
...
 
+        public override void VisitBinaryPattern(BinaryPatternSyntax node)
+        {
+            // Like binary expressions, binary patterns are left-associative, and can be deeply nested (even in our own code).
+            // Handle this with manual recursion.
+            PatternSyntax currentPattern = node;
+
+            var rightPatternStack = ArrayBuilder<PatternSyntax>.GetInstance();
+
+            while (currentPattern is BinaryPatternSyntax binaryPattern)
```
AlekseyTs, line 350: >while (currentPattern is BinaryPatternSyntax binaryPattern) [](http://example.com/codeflow?start=12&length=59)

Consider using a loop with a post-condition. We know the condition is true on entry.

src/Compilers/CSharp/Portable/Binder/LocalBinderFactory.cs
```diff
@@ -796,6 +796,27 @@ public override void VisitSwitchExpression(SwitchExpressionSyntax node)
             }
         }
 
+        public override void VisitBinaryPattern(BinaryPatternSyntax node)
+        {
+            // Users (such as ourselves) can have many, many nested binary patterns. To avoid crashing, do left recursion manually.
+            PatternSyntax currentPattern = node;
+
+            var rightPatternStack = ArrayBuilder<PatternSyntax>.GetInstance();
```
AlekseyTs, line 804: >var rightPatternStack = ArrayBuilder<PatternSyntax>.GetInstance(); [](http://example.com/codeflow?start=12&length=66)

I think the order of visiting arms is not important for the purpose of creating binders. Therefore, we don't really need to use manual stack here, we can visit leaves as we descend left in a single loop.

src/Compilers/CSharp/Portable/Binder/DecisionDagBuilder.cs
```diff
@@ -645,33 +645,71 @@ private Tests MakeTestsAndBindingsForBinaryPattern(
...
-                var result = Tests.OrSequence.Create(builder);
-                if (bin.InputType.Equals(bin.NarrowedType))
+                binaryPatternStack.Push(currentNode);
+                currentNode = currentNode.Left as BoundBinaryPattern;
+            } while (currentNode != null);
+
+            Tests? result = null;
+            BoundDagTemp? currentOutput = null;
+
+            while (binaryPatternStack.TryPop(out currentNode))
```
AlekseyTs, line 662: >while (binaryPatternStack.TryPop(out currentNode)) [](http://example.com/codeflow?start=12&length=50)

Consider using a loop with a post-condition

src/Compilers/CSharp/Portable/Binder/DecisionDagBuilder.cs
```diff
@@ -645,33 +645,71 @@ private Tests MakeTestsAndBindingsForBinaryPattern(
...
+            {
+                if (result == null)
```
AlekseyTs, line 664: >if (result == null) [](http://example.com/codeflow?start=16&length=19)

Consider moving initialization out of the loop

src/Compilers/CSharp/Portable/Binder/DecisionDagBuilder.cs
```diff
@@ -645,33 +645,71 @@ private Tests MakeTestsAndBindingsForBinaryPattern(
...
-                builder.Add(MakeTestsAndBindings(input, bin.Left, bindings));
-                builder.Add(MakeTestsAndBindings(input, bin.Right, bindings));
-                var result = Tests.OrSequence.Create(builder);
-                if (bin.InputType.Equals(bin.NarrowedType))
+                binaryPatternStack.Push(currentNode);
+                currentNode = currentNode.Left as BoundBinaryPattern;
+            } while (currentNode != null);
+
+            Tests? result = null;
+            BoundDagTemp? currentOutput = null;
```
AlekseyTs, line 660: >BoundDagTemp? currentOutput = null; [](http://example.com/codeflow?start=12&length=35)

Can we simply use ```output``` instead

src/Compilers/CSharp/Portable/BoundTree/BoundTreeWalker.cs
```diff
@@ -129,9 +127,9 @@ protected BoundTreeWalkerWithStackGuardWithoutRecursionOnTheLeftOfBinaryOperator
 
             this.Visit(current);
 
-            while (rightOperands.Count > 0)
+            while (rightOperands.TryPop(out current))
```
AlekseyTs, line 130: >while (rightOperands.TryPop(out current)) [](http://example.com/codeflow?start=12&length=41)

Since you decided to change this code, consider using a loop with a post-condition

src/Compilers/CSharp/Portable/BoundTree/BoundTreeWalker.cs
```diff
@@ -112,12 +112,10 @@ protected BoundTreeWalkerWithStackGuardWithoutRecursionOnTheLeftOfBinaryOperator
 
             rightOperands.Push(node.Right);
 
-            var binary = (BoundBinaryOperator)node.Left;
-
             BeforeVisitingSkippedBoundBinaryOperatorChildren(binary);
             rightOperands.Push(binary.Right);
 
-            BoundExpression current = binary.Left;
+            BoundExpression? current = binary.Left;
```
AlekseyTs, line 118: >? [](http://example.com/codeflow?start=27&length=1)

Is `?` necessary?

src/Compilers/CSharp/Portable/BoundTree/BoundTreeWalker.cs
```diff
@@ -142,6 +140,44 @@ protected virtual void BeforeVisitingSkippedBoundBinaryOperatorChildren(BoundBin
...
+            {
+                binary = (BoundBinaryPattern)current;
+                BeforeVisitingSkippedBoundBinaryPatternChildren(binary);
+                rightOperands.Push(binary.Right);
+                current = binary.Left;
+            }
+
+            Visit(current);
+
+            while (rightOperands.TryPop(out current))
```
AlekseyTs, line 168: >while (rightOperands.TryPop(out current)) [](http://example.com/codeflow?start=12&length=41)

Consider using a loop with a post-condition

src/Compilers/CSharp/Portable/BoundTree/BoundTreeWalker.cs
```diff
@@ -142,6 +140,44 @@ protected virtual void BeforeVisitingSkippedBoundBinaryOperatorChildren(BoundBin
...
+                return base.VisitBinaryPattern(node);
+            }
+
+            var rightOperands = ArrayBuilder<BoundPattern>.GetInstance();
+
+            rightOperands.Push(node.Right);
+            BeforeVisitingSkippedBoundBinaryPatternChildren(binary);
+            rightOperands.Push(binary.Right);
+
+            BoundPattern? current = binary.Left;
```
AlekseyTs, line 156: >? [](http://example.com/codeflow?start=24&length=1)

Is `?` necessary?

src/Compilers/CSharp/Portable/BoundTree/BoundTreeWalker.cs
```diff
@@ -142,6 +140,44 @@ protected virtual void BeforeVisitingSkippedBoundBinaryOperatorChildren(BoundBin
...
+            while (rightOperands.TryPop(out current))
+            {
+                Visit(current);
+            }
+
+            rightOperands.Free();
+            return null;
+        }
+
+        protected virtual void BeforeVisitingSkippedBoundBinaryPatternChildren(BoundBinaryPattern node)
```
AlekseyTs, line 177: >BeforeVisitingSkippedBoundBinaryPatternChildren [](http://example.com/codeflow?start=31&length=47)

It doesn't look like this method is overridden by anyone. I think we should remove it for now. It can be added later when needed.

src/Compilers/CSharp/Portable/BoundTree/BoundTreeWalker.cs
333fred: Yes, `TryPop` will assign `null` when it returns false.

src/Compilers/CSharp/Portable/BoundTree/BoundTreeWalker.cs
333fred: Same as above.

src/Compilers/CSharp/Portable/Binder/ExpressionVariableFinder.cs
333fred: I think that results in less-clear code, like we have in `VisitBinaryExpression` above. I'd prefer to keep this as is.

src/Compilers/CSharp/Portable/FlowAnalysis/AbstractFlowPass.cs
```diff
@@ -996,17 +996,27 @@ static bool patternMatchesNull(BoundPattern pattern)
                     case BoundNegatedPattern negated:
                         return !patternMatchesNull(negated.Negated);
                     case BoundBinaryPattern binary:
-                        if (binary.Disjunction)
+                        var binaryPatterns = getBinaryPatterns(binary);
```
AlekseyTs, line 999: >var binaryPatterns = getBinaryPatterns(binary); [](http://example.com/codeflow?start=24&length=47)

I would expect to see change to an override of VisitBinaryPattern in this class. And, if we don't expect it to be called, throw from it.

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
```diff
@@ -1734,99 +1734,139 @@ private BoundPattern BindBinaryPattern(
...
+            Debug.Assert(binaryPatternStack.Count > 0);
 
-                // Compute the common type. This algorithm is quadratic, but disjunctive patterns are unlikely to be huge
-                var narrowedTypeCandidates = ArrayBuilder<TypeSymbol>.GetInstance(2);
-                collectCandidates(left, narrowedTypeCandidates);
-                collectCandidates(right, narrowedTypeCandidates);
-                var narrowedType = leastSpecificType(node, narrowedTypeCandidates, diagnostics) ?? inputType;
-                narrowedTypeCandidates.Free();
+            var binaryPatternAndPermitDesignations = binaryPatternStack.Pop();
+            BoundPattern? result = BindPattern(binaryPatternAndPermitDesignations.pat.Left, inputType, binaryPatternAndPermitDesignations.permitDesignations, hasErrors, diagnostics);
```
AlekseyTs, line 1752: >? [](http://example.com/codeflow?start=24&length=1)

Is `?` still necessary?

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
```diff
@@ -1734,99 +1734,139 @@ private BoundPattern BindBinaryPattern(
...
-                        collectCandidates(p.Left, candidates);
-                        collectCandidates(p.Right, candidates);
-                    }
-                    else
-                    {
-                        candidates.Add(pat.NarrowedType);
-                    }
-                }
+            binaryPatternStack.Free();
+            Debug.Assert(result != null);
```
AlekseyTs, line 1766: >Debug.Assert(result != null) [](http://example.com/codeflow?start=12&length=28)

Is this assert still necessary?

src/Compilers/CSharp/Portable/Binder/DecisionDagBuilder.cs
```diff
@@ -645,33 +645,59 @@ private Tests MakeTestsAndBindingsForBinaryPattern(
...
+            var currentNode = bin;
+
+            do
+            {
+                binaryPatternStack.Push(currentNode);
+                currentNode = currentNode.Left as BoundBinaryPattern;
+            } while (currentNode != null);
+
+            currentNode = binaryPatternStack.Pop();
+            Tests? result = MakeTestsAndBindings(input, currentNode.Left, out output, bindings);
```
AlekseyTs, line 660: >? [](http://example.com/codeflow?start=17&length=1)

Is the `?` still necessary?

src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
```diff
@@ -2064,9 +2064,35 @@ void assignPatternVariablesAndMarkReadFields(BoundPattern pattern, bool definite
...
+                                definitely = definitely && !pat.Disjunction;
+                                stack.Push((pat, definitely));
+                                pat = pat.Left as BoundBinaryPattern;
+                            } while (pat is not null);
+
+                            Debug.Assert(stack.Count > 0);
+                            (pat, definitely) = stack.Peek();
+                            assignPatternVariablesAndMarkReadFields(pat.Left, definitely);
+
+                            while (stack.TryPop(out var patAndDef))
```
AlekseyTs, line 2090: >while (stack.TryPop(out var patAndDef)) [](http://example.com/codeflow?start=28&length=39)

Consider using a loop with a post-condition

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker_Patterns.cs
```diff
@@ -111,8 +111,26 @@ public override BoundNode VisitNegatedPattern(BoundNegatedPattern node)
 
         public override BoundNode VisitBinaryPattern(BoundBinaryPattern node)
         {
-            Visit(node.Left);
-            Visit(node.Right);
+            // Users (such as ourselves) can have many, many nested binary patterns. To avoid crashing, do left recursion manually.
```
AlekseyTs, line 114: >// Users (such as ourselves) can have many, many nested binary patterns. To avoid crashing, do left recursion manually. [](http://example.com/codeflow?start=12&length=119)

Should we add a similar override to `DebugVerifier`?

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker_Patterns.cs
```diff
@@ -111,8 +111,26 @@ public override BoundNode VisitNegatedPattern(BoundNegatedPattern node)
...
+            {
+                TakeIncrementalSnapshot(current);
+                stack.Push(current);
+                current = current.Left as BoundBinaryPattern;
+            } while (current != null);
+
+            Debug.Assert(stack.Peek().Left is not BoundBinaryPattern);
+            Visit(stack.Peek().Left);
+
+            while (stack.TryPop(out current))
```
AlekseyTs, line 128: >while (stack.TryPop(out current)) [](http://example.com/codeflow?start=12&length=33)

Consider using a node with a post condition

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker_Patterns.cs
```diff
@@ -199,8 +217,21 @@ private void LearnFromAnyNullPatterns(
                     LearnFromAnyNullPatterns(inputSlot, inputType, p.Negated);
                     break;
                 case BoundBinaryPattern p:
-                    LearnFromAnyNullPatterns(inputSlot, inputType, p.Left);
-                    LearnFromAnyNullPatterns(inputSlot, inputType, p.Right);
+                    // Do not use left recursion because we can have many nested binary patterns.
+                    var current = p;
+                    while (true)
+                    {
+                        LearnFromAnyNullPatterns(inputSlot, inputType, current.Right);
```
AlekseyTs, line 224: >LearnFromAnyNullPatterns(inputSlot, inputType, current.Right); [](http://example.com/codeflow?start=24&length=62)

Is this process not order sensitive? Consider explicitly stating this in a comment, probably with an explanation why.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker_Patterns.cs
```diff
@@ -111,8 +111,26 @@ public override BoundNode VisitNegatedPattern(BoundNegatedPattern node)
...
         {
-            Visit(node.Left);
-            Visit(node.Right);
+            // Users (such as ourselves) can have many, many nested binary patterns. To avoid crashing, do left recursion manually.
+
+            var stack = ArrayBuilder<BoundBinaryPattern>.GetInstance();
+            BoundBinaryPattern current = node;
+            do
+            {
+                TakeIncrementalSnapshot(current);
```
AlekseyTs, line 120: >TakeIncrementalSnapshot(current); [](http://example.com/codeflow?start=16&length=33)

Is state of ```_expressionIsRead``` insignificant here?

src/Compilers/CSharp/Portable/Operations/CSharpOperationFactory.cs
```diff
@@ -2607,15 +2607,42 @@ private IOperation CreateBoundNegatedPatternOperation(BoundNegatedPattern boundN
...
+            BoundBinaryPattern? current = boundBinaryPattern;
+
+            do
+            {
+                stack.Push(current);
+                current = current.Left as BoundBinaryPattern;
+            } while (current != null);
+
+            var result = (IPatternOperation)Create(stack.Peek().Left);
+            while (stack.TryPop(out current))
```
AlekseyTs, line 2626: >while (stack.TryPop(out current)) [](http://example.com/codeflow?start=12&length=33)

Consider using a loop with a post-condition

AlekseyTs:

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker_Patterns.cs
333fred: It shouldn't be to BinaryPattern itself. If any nested patterns care, they'll save and reset as appropriate, but we don't need to adjust it ourselves here.

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
333fred: I'll take the former suggestion, but I'll avoid the latter suggestion for now. We can do it in a followup if it's necessary, but it's a bigger change and I don't think we need to look at it yet.

src/Compilers/CSharp/Portable/BoundTree/BoundTreeVisitors.cs
```diff
@@ -184,22 +184,23 @@ public static Location GetTooLongOrComplexExpressionErrorLocation(BoundNode node
             {
                 SyntaxNode syntax = node.Syntax;
 
-                if (!(syntax is ExpressionSyntax))
+                if (syntax is not (ExpressionSyntax or PatternSyntax))
                 {
-                    syntax = syntax.DescendantNodes(n => !(n is ExpressionSyntax)).OfType<ExpressionSyntax>().FirstOrDefault() ?? syntax;
+                    syntax = syntax.DescendantNodes(n => n is not (ExpressionSyntax or PatternSyntax)).FirstOrDefault(n => n is ExpressionSyntax or PatternSyntax) ?? syntax;
```
AlekseyTs, line 189: >n => n is ExpressionSyntax or PatternSynta [](http://example.com/codeflow?start=118&length=42)

Why is this necessary? It looks like this condition is the opposite of the filter used for DescendantNodes.

333fred: The predicate for `DescendentNodes` controls whether we descend into children; it stops us from returning new nodes after we hit an expression or pattern. However, it will still have given us all of the parent nodes, so we need to ignore them.

AlekseyTs (approved):
LGTM (commit 10)

333fred:

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
```diff
@@ -1734,100 +1734,144 @@ private BoundPattern BindBinaryPattern(
...
-                }
+            binaryPatternStack.Free();
+            narrowedTypeCandidates.Free();
+            return result;
 
-                TypeSymbol? leastSpecificType(SyntaxNode node, ArrayBuilder<TypeSymbol> candidates, BindingDiagnosticBag diagnostics)
+            static BoundPattern bindBinaryPattern(
+                BoundPattern preboundLeft,
+                Binder binder,
+                (BinaryPatternSyntax pat, bool permitDesignations) patternAndPermitDesignations,
```
cston, line 1775: >(BinaryPatternSyntax pat, bool permitDesignations) patternAndPermitDesignations [](http://example.com/codeflow?start=16&length=79)

Consider splitting the tuple into two separate parameters.

src/Compilers/CSharp/Portable/BoundTree/BoundTreeVisitors.cs
```diff
@@ -184,22 +184,23 @@ public static Location GetTooLongOrComplexExpressionErrorLocation(BoundNode node
             {
                 SyntaxNode syntax = node.Syntax;
 
-                if (!(syntax is ExpressionSyntax))
+                if (syntax is not (ExpressionSyntax or PatternSyntax))
```
cston, line 187: >(ExpressionSyntax or PatternSyntax) [](http://example.com/codeflow?start=34&length=35)

Consider creating a static local function since this same check is used three times.

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
333fred: It was originally separate parameters; I moved it to the tuple for simplicity of the calling code.

src/Compilers/CSharp/Portable/BoundTree/BoundTreeVisitors.cs
333fred: This doesn't really feel like the type of thing that needs to be extracted to a local function; it honestly feels like that would impede readability, not help. I'd prefer to leave this as is.

src/Compilers/CSharp/Portable/Binder/Binder_Patterns.cs
cston: The method is not using the tuple as a tuple though. It feels like the caller should be responsible for deconstructing the tuple rather than the method.

src/Compilers/CSharp/Test/EndToEnd/EndToEndTests.cs
```diff
@@ -723,5 +725,57 @@ class XAttribute : System.Attribute { }
...
+        public void ManyBinaryPatterns()
+        {
+            const string preamble = $"""
+                int i = 2;
+
+                System.Console.Write(i is
+                """;
+            const string append = $"""
+
+                or
```
cston, line 739: >or [](http://example.com/codeflow?start=16&length=2)

Consider testing deep nesting of `and` patterns as well since there is some difference in `Binder.BindBinaryPattern()`.
```csharp
Console.Write(i is not 1 and not 2 and ...);
```

cston (approved):

cston (approved):

PR #35667, opened 2019-05-12
Implement Quick Info styles

sharwell:
* Support emphasis in Quick Info (`<em>` and `<i>` elements in XML documentation comments)
* Support strong in Quick Info (`<strong>` and `<b>` elements in XML documentation comments)
* Support code font in Quick Info (`<c>`, `<code>`, `<tt>` elements in XML documentation comments)
* Support bullet and numbered lists in Quick Info
* Support definition lists in Quick Info
* Support navigating to `cref` items by clicking in Quick Info
* Support navigating to `href` hyperlinks in Quick Info (`<see href` and `<a href`)

### Images

⚠️ This picture is *slightly* outdated relative to current behavior (both links to https://google.com now render correctly, and the spacing for list items has changed slightly.
image

image

PR #75215, opened 2024-09-24
Allow <code> in DocComments to render as a block in Hover response

JoeRobich:
As noticed in #75181, `<code>` from DocComment always renders inline even if it represents a block of code.

Before:
image

After:
image

In order to have highlighting, we will need to pass along the code blocks `lang` attribute with the TaggedText.

### DIFF
--- a/src/LanguageServer/ProtocolUnitTests/Hover/HoverTests.cs
+++ b/src/LanguageServer/ProtocolUnitTests/Hover/HoverTests.cs
@@ -424,6 +424,48 @@ void A.AMethod(int i)
             Assert.Equal(expectedMarkdown, results.Contents.Fourth.Value);
         }
 
+        [Theory, CombinatorialData]
+        public async Task TestGetHoverAsync_UsingMarkupContentDoesNotEscapeCodeBlock(bool mutatingLspWorkspace)
+        {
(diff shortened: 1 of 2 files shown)


sharwell:
Note that this feature was originally implemented in #35667 for Visual Studio. I believe the change here is only related to LSP clients.

Can we get tests added for the new behavior?

dibarbet (commented):
lgtm, pending requested test
https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/ProtocolUnitTests/Hover/HoverTests.cs is probably the right place

src/LanguageServer/ProtocolUnitTests/Hover/HoverTests.cs
```diff
@@ -424,6 +424,48 @@ void A.AMethod(int i)
             Assert.Equal(expectedMarkdown, results.Contents.Fourth.Value);
         }
 
+        [Theory, CombinatorialData]
```
dibarbet, line 427:
```suggestion
        [Theory, CombinatorialData, WorkItem("https://github.com/dotnet/roslyn/issues/75181")]
```

dibarbet (approved):

PR #75224, opened 2024-09-24
Add option for LSP-based auto insert

akhera99:


JoeRobich (approved):

src/LanguageServer/Protocol/LspOptionsStorage.cs
```diff
@@ -32,6 +32,8 @@ internal sealed class LspOptionsStorage
 
         private static readonly OptionGroup s_codeLensOptionGroup = new(name: "code_lens", description: "");
 
+        private static readonly OptionGroup s_onAutoInsertOptionGroup = new(name: "on_auto_insert", description: "");
```
Cosifne, line 35: `on_auto_insert` doesn't sound like a good option group? Is this related to typing?

Cosifne: Also note I believe the grouping name here would be reflected by vscode in the settings page.
Like `inlay hints`
image
So a better group name would be needed.

akhera99: "On Auto Insert" is the name of the feature, but it's not option-based in roslyn so the naming has never been an issue. We could change it to Auto Insert, @dibarbet @JoeRobich any thoughts?

JoeRobich: VSCode settings are laid out according to the grouping in the extensions's package.json (Inlay Hints options are under [Text Editor](https://github.com/dotnet/vscode-csharp/blob/main/package.json#L732)). Not really sure what function the `OptionGroup` plays here.

JoeRobich: I do like "Auto Insert" better.

dibarbet: Auto insert also seems fine to me.  Option group is used as the item after the first arrow.  For example for "Dotnet > Inlay Hints: Suppress...", the option group is "Inlay Hints".  I think having that be "Auto Insert" would be good.

Cosifne: > VSCode settings are laid out according to the grouping in the extensions's package.json (Inlay Hints options are under [Text Editor](https://github.com/dotnet/vscode-csharp/blob/main/package.json#L732)). Not really sure what function the `OptionGroup` plays here.

Yeah it's because the option group name is used in package.json

`Auto Insert` LGTM

dibarbet (approved):

Issue #75244, opened 2024-09-25
Implement nullable analysis for `field` keyword


RikkiGibson:
#57012
dotnet/csharplang#8425

Tracks implementation of short-term nullability solution for `field` keyword in properties.

Calling out a few areas to change to implement `set` enforcement (not part of the initial PR, but maybe we will do it as a stretch goal):

https://github.com/dotnet/roslyn/blob/bdebb57e8817bdfa1b43f53e905db9b4dfa95736/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs#L687-L699
```diff
                     else
                     {
+                        if (method is SourcePropertyAccessorSymbol
+                            {
+                                MethodKind: MethodKind.PropertySet,
+                                AssociatedSymbol: SourcePropertySymbolBase { UsesFieldKeyword: true, BackingField: { } backingField }
+                            })
+                        {
+                            if (memberHasBadState(member, state))
+                            {
+                                // Member '{name}' must have a non-null value when exiting.
+                                Diagnostics.Add(ErrorCode.WRN_MemberNotNull, syntaxOpt?.GetLocation() ?? methodMainNode.Syntax.GetLastToken().GetLocation(), member.Name);
+                            }
+                        }

                         do
                         {
                             foreach (var memberName in method.NotNullMembers)
                             {
                                 enforceMemberNotNullOnMember(syntaxOpt, state, method, memberName);
                             }
 
                             method = method.OverriddenMethod;
                         }
                         while (method != null);
                     }
```

https://github.com/dotnet/roslyn/blob/bdebb57e8817bdfa1b43f53e905db9b4dfa95736/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs#L949-L959

```diff
                     else
                     {
+                        if (method is SourcePropertyAccessorSymbol
+                            {
+                                MethodKind: MethodKind.PropertySet,
+                                AssociatedSymbol: SourcePropertySymbolBase { UsesFieldKeyword: true, BackingField: { } backingField }
+                            })
+                        {
+                            // https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#setter-analysis
+                            // setter assumes the backing field may be null and has to put the field into the correct state when exiting.
+                            if (getSlotForFieldOrPropertyOrEvent(backingField) is int memberSlot &&
+                                memberSlot > 0)
+                            {
+                                SetState(ref this.State, memberSlot, NullableFlowState.MaybeNull);
[...]

Issue #50244, opened 2021-01-05
Warn when auto-property setter is more tolerant than getter


RikkiGibson:
Related to #49903

[SharpLab](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOyQExAagB8BiTdAG3ImHLgAI5NraBYAKAAEAGOjgRgB0AEQCWEAOaYA9rBEBjKAIDCUnHACCTcgE8oIqAG527AG4QEdOXQC8dTHADudJQAoAlEbZyBABQRSABxs7CnJPbz9AgQAVKQBlGAQRTHF3AzoAegy7KToHc0xk8SQ6YHQYSwRoAAs4KDoICoxMGBEwOGNOAGZeACZndgBvdjpRugBtdUopBwA5UIBdEbGOHv4eSKDBunE4GHSoPfSAX3ZjoA===)
```cs
#nullable enable
using System.Diagnostics.CodeAnalysis;

var c = new C();
c.Prop = null;
c.Prop.ToString(); // no warning, but crashes at runtime

public class C
{
    [AllowNull]
    public string Prop { get; set; }
}
```

In this scenario, the setter accepts a maybe-null argument but the getter claims it will return a not-null value. A manually implemented property can fulfill this by replacing a null argument with a non-null default value, but an auto-property can't do that.

We should give a warning on the declaration of this property--perhaps one indicating that an auto-property setter should not be nullable when the getter is non-nullable.

We could also give the warning in a "roundabout" way by just making sure we
1. type the synthesized backing field based on the getter's return type and
2. make sure the nullable walker analyzes the synthesized assignment of the 'value' param to the backing field in the setter

Note that this behavior is specific to properties--fields aren't much of a concern because they don't have the behavior of assuming that more tolerant setters "sanitize" their inputs.

RikkiGibson:
While working on https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#nullability it occurred to me that this would be fairly straightforward to accomplish by simply nullable-analyzing the synthesized auto-property accessors.

This would lean on the existing analogy that the following two properties are equivalent (as much as is possible):
```cs
string Prop1 { get; set; }
string Prop2 { get => field; set => field = value; }
```

[...]

PR #75246, opened 2024-09-25
Short term field keyword nullable analysis

RikkiGibson:
Part of #75244

I found that the dotnet/csharplang#8425 did not really include enough details to attack the attribute problem.

Basically, we have a precedent with auto properties using nullable attributes, that the field is subject to those same attributes in the same way. For example, `[AllowNull] string Prop { get; set; } = null;` is valid, even though the `= null` is assigning to the backing field rather than the auto-prop which claims to accept null.

However, our goal with the field keyword nullability proposal was to allow customizing the field's nullability independently of the property. Therefore, I have arrived at the following principle:
- For classic auto properties, i.e. field-backed properties which don't use the `field` keyword, `field:` targeted nullability attributes are ignored, and the attributes from the property are used for the field instead. Essentially, the field is not a distinct thing from the property for the purposes of nullable analysis.
    - The goal here is avoid any behavior or diagnostic changes for existing auto-property scenarios in the short term. #50244 suggests making changes to existing scenarios to improve safety.
- For properties using `field` keyword, the field *is* a distinct thing. `field:` targeted nullability attributes are used on the field, and nullability attributes on the property are *not* applied to the field.
    - Therefore, `[AllowNull] string Prop { get => field; set => field = value } = null;` results in a warning--since the initializer is assigning to the field, and the AllowNull doesn't apply to the field.
    - However, we still view the property as a proxy to the field in a constructor. This is accomplished by sharing a slot between the property and the backing field in constructors. In other words, within a constructor, whatever flow state the property has, we assume the field also has that same state, and any state changes to either one are reflected in both. This means the right thing happens whether field/property assignment is used, or `if (Prop is null) throw ...;` is used, to ensure that the property is in the correct state before exiting the constructor.
    - We use the property's nullability attributes to decide what values can be assigned to the property, when the language defines that the property is the thing being assigned--such as when a property with a setter is assigned in a constructor. Similarly, the field's nullability attributes decide what values can be assigned to the field when the field is the thing being assigned, such as in a property initializer.

The short term solution reflects the need for user to specify whether the field is nullable or not independently of the property. When we introduce a solution which infers the field's nullability across members, we expect the need for the user to specify will disappear in most cases.


RikkiGibson:
Test failure appears to be just the following

```
[xUnit.net 00:00:47.12]     Microsoft.CodeAnalysis.CSharp.UnitTests.DiagnosticAnalyzerTests.TestDescriptorForConfigurableCompilerDiagnostics [FAIL]
  Failed Microsoft.CodeAnalysis.CSharp.UnitTests.DiagnosticAnalyzerTests.TestDescriptorForConfigurableCompilerDiagnostics [7 ms]
  Error Message:
   Add resource string named 'WRN_UninitializedNonNullableBackingField_Title' for Title of 'CS9264' to 'CSharpResources'
Expected: True
Actual:   False
  Stack Trace:
     at Microsoft.CodeAnalysis.CSharp.UnitTests.DiagnosticAnalyzerTests.TestDescriptorForConfigurableCompilerDiagnostics() in /_/src/Compilers/CSharp/Test/Emit3/Diagnostics/DiagnosticAnalyzerTests.cs:line 733

   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

I will try to address tomorrow.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -4565,18 +4565,18 @@ class C
             else
             {
                 comp.VerifyEmitDiagnostics(
-                    // (4,13): warning CS8618: Non-nullable property 'P1' must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring the property as nullable.
+                    // (4,13): error CS9264: Non-nullable property 'P1' must contain a non-null value when exiting constructor. Consider adding the 'required' modifier, or declaring the property as nullable, or adding '[field: MaybeNull, AllowNull]' attributes.
```
cston, line 4568: >error CS9264 [](http://example.com/codeflow?start=31&length=12)

Let's update these comments to `warning CS9264`, to avoid confusion.

src/Compilers/CSharp/Portable/Errors/ErrorFacts.cs
```diff
@@ -82,6 +82,8 @@ static ErrorFacts()
             nullableWarnings.Add(GetId(ErrorCode.WRN_NullabilityMismatchInReturnTypeOnInterceptor));
             nullableWarnings.Add(GetId(ErrorCode.WRN_NullabilityMismatchInParameterTypeOnInterceptor));
 
+            nullableWarnings.Add(GetId(ErrorCode.WRN_UninitializedNonNullableBackingField));
```
cston, line 85: >nullableWarnings [](http://example.com/codeflow?start=12&length=16)

Does this allow disabling the warning using `pragma warning disable nullable`? It doesn't look like we're testing the existing entries above, but consider adding a test for this case.
```csharp
#nullable enable
#pragma warning disable nullable
class C1
{
    object P1 => field; // 1
}
#pragma warning enable nullable
class C2
{
    object P2 => field;
}
```

RikkiGibson: I think `#pragma warning disable nullable` was ripped out, but, it should work with `#nullable disable warnings`.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -2130,6 +2134,30 @@ protected override int GetOrCreateSlot(Symbol symbol, int containingSlot = 0, bo
                 }
             }
 
+            // Share a slot between backing field and associated property/event in the context of a constructor which owns initialization of that backing field.
+            if (this._symbol is MethodSymbol constructor
+                && constructor.IsConstructor()
+                && constructor.IsStatic == symbol.IsStatic)
```
cston, line 2140: >constructor.IsStatic == symbol.IsStatic [](http://example.com/codeflow?start=19&length=39)

Are we testing this case? That is, are we testing use of static property in an instance constructor?

RikkiGibson: This check is only *really* going to be relevant in a scenario like the following:
1. static property is used in instance constructor, *and*
2. somehow, backing field of the same static property is also used in instance constructor. I think this is not possible.

I am adding some tests to poke at that area, but, even if the tests don't show the necessity of the check, I think I will feel more comfortable having it there.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -4752,6 +4752,1616 @@ class C
...
+                class C
+                {
+                    public static string Prop { get => field; set => field = value; }
+                    static C() { }
+                }
+                """;
+
+            var comp = CreateCompilation(source);
+            comp.VerifyEmitDiagnostics(
+                // (6,12): warning CS9264: Non-nullable property 'Prop' must contain a non-null value when exiting constructor. Consider adding the 'required' modifier, or declaring the property as nullable, or adding '[field: MaybeNull, AllowNull]' attributes.
```
333fred, line 4894: Hmm. Existing problem, but you can't exactly mark `Prop` as required...

333fred (approved):

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -4752,6 +4752,1617 @@
...
+                //     static C() { }
+                Diagnostic(ErrorCode.WRN_UninitializedNonNullableBackingField, "C").WithArguments("property", "Prop").WithLocation(6, 12));
+        }
+
+        [Fact]
+        public void Nullability_07()
```
cston, line 4920: >Nullability_07 [](http://example.com/codeflow?start=20&length=14)

Minor: It seems like `Nullability_08` covers this test. Similar comment for `_07_Getter_only`.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -4752,6 +4752,1617 @@
...
+            var source = """
+                #nullable enable
+                using System.Diagnostics.CodeAnalysis;
+
+                class C
+                {
+                    [MaybeNull, AllowNull]
+                    public string Prop
+                    {
+                        get => field.ToString(); // 1
```
cston, line 5410: > // 1 [](http://example.com/codeflow?start=48&length=5)

Minor: Consider moving `// 1` to setter.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -4752,6 +4752,1617 @@
...
+                    public string? Prop { get => field; set; } = "a";
+                }
+                """;
+
+            var comp = CreateCompilation([source, NotNullAttributeDefinition]);
+            comp.VerifyEmitDiagnostics();
+        }
+
+        [Fact]
+        public void Nullability_20_ManualGetter_AutoSetter_NullTest()
```
cston, line 5655: >Nullability_20_ManualGetter_AutoSetter_NullTest [](http://example.com/codeflow?start=20&length=47)

Is this distinct from the previous test? (What are we testing with the addition of `if (Prop is null) throw null!`?)

cston (approved):

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
RikkiGibson: The goal is to demonstrate that a null-test+throw on the property puts a non-nullable field into the correct state. Adding a version without initializer to demonstrate that more clearly.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
RikkiGibson: The goal of this separation was to test the constructor exit checks separately from testing field flow state being reflected when reading the property.

PR #75250, opened 2024-09-26
Do not load CodeStyle analyzers added by the SDK

JoeRobich:
To avoid a version mismatch between the version of Roslyn in the IDE and the version of Roslyn CodeStyle analyzers added by the SDK (#72811), we will not load the CodeStyle analyzers added by the SDK. The IDE ships with the same analyzers which are version matched and used during analysis.

The analyzers are still listed in the Solution Explorer but without any associated analyzers.
image

### DIFF
--- a/src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
+++ b/src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
@@ -195,13 +195,44 @@ Namespace Microsoft.VisualStudio.LanguageServices.UnitTests.ProjectSystemShim
                 Dim project = Await environment.ProjectFactory.CreateAndAddToWorkspaceAsync(
                     "Project", LanguageNames.CSharp, CancellationToken.None)
 
-                ' add Razor source generator and a couple more other analyzer filess:
+                ' add Razor source generator and a couple more other analyzer files:
                 Dim path1 = Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk.Razor", "source-generators", "Microsoft.NET.Sdk.Razor.SourceGenerators.dll")
                 Dim path2 = Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk.Razor", "source-generators", "SdkDependency1.dll")
                 project.AddAnalyzerReference(path1)
                 project.AddAnalyzerReference(path2)
 
                 AssertEx.Equal({path1, path2}, environment.Workspace.CurrentSolution.Projects.Single().AnalyzerReferences.Select(Function(r) r.FullPath))
+            End Using
+        End Function
+
+        <WpfFact>
+        Public Async Function CodeStyleAnalyzers_FromSdk() As Task
+            Using environment = New TestEnvironment()
+                Dim providerFactory = DirectCast(environment.ExportProvider.GetExportedValue(Of IVisualStudioDiagnosticAnalyzerProviderFactory), MockVisualStudioDiagnosticAnalyzerProviderFactory)
+                providerFactory.Extensions =
+                {
+                     ({
+                        Path.Combine(TempRoot.Root, "File.dll")
+                     },
+                     "AnotherExtension")
+                }
+
+                Dim project = Await environment.ProjectFactory.CreateAndAddToWorkspaceAsync(
+                    "Project", LanguageNames.CSharp, CancellationToken.None)
+
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CodeStyle.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CodeStyle.Fixes.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CSharp.CodeStyle.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CSharp.CodeStyle.Fixes.dll"))
+
+                Assert.Empty(environment.Workspace.CurrentSolution.Projects.Single().AnalyzerReferences)
+
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Dir", "File.dll"))
+
+                AssertEx.Equal(
+                {
+                    Path.Combine(TempRoot.Root, "Dir", "File.dll")
+                }, environment.Workspace.CurrentSolution.Projects.Single().AnalyzerReferences.Select(Function(r) r.FullPath))
             End Using
         End Function
     End Class
--- a/src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
+++ b/src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
@@ -206,29 +206,50 @@ Namespace Microsoft.VisualStudio.LanguageServices.UnitTests.ProjectSystemShim
         End Function
 
         <WpfFact>
-        Public Async Function CodeStyleAnalyzers_FromSdk() As Task
+        Public Async Function CodeStyleAnalyzers_CSharp_FromSdk_AreIgnored() As Task
             Using environment = New TestEnvironment()
-                Dim providerFactory = DirectCast(environment.ExportProvider.GetExportedValue(Of IVisualStudioDiagnosticAnalyzerProviderFactory), MockVisualStudioDiagnosticAnalyzerProviderFactory)
-                providerFactory.Extensions =
-                {
-                     ({
-                        Path.Combine(TempRoot.Root, "File.dll")
-                     },
-                     "AnotherExtension")
-                }
-
                 Dim project = Await environment.ProjectFactory.CreateAndAddToWorkspaceAsync(
                     "Project", LanguageNames.CSharp, CancellationToken.None)
 
+                ' These are the in-box C# codestyle analyzers that ship with the SDK
                 project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CodeStyle.dll"))
                 project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CodeStyle.Fixes.dll"))
                 project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CSharp.CodeStyle.dll"))
                 project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CSharp.CodeStyle.Fixes.dll"))
 
+                ' Ensure they are not returned when getting AnalyzerReferences
                 Assert.Empty(environment.Workspace.CurrentSolution.Projects.Single().AnalyzerReferences)
 
+                ' Add a non-codestyle analyzer to the project
                 project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Dir", "File.dll"))
 
+                ' Ensure it is returned as expected
+                AssertEx.Equal(
+                {
+                    Path.Combine(TempRoot.Root, "Dir", "File.dll")
+                }, environment.Workspace.CurrentSolution.Projects.Single().AnalyzerReferences.Select(Function(r) r.FullPath))
+            End Using
+        End Function
+
+        <WpfFact>
+        Public Async Function CodeStyleAnalyzers_VisualBasic_FromSdk_AreIgnored() As Task
+            Using environment = New TestEnvironment()
+                Dim project = Await environment.ProjectFactory.CreateAndAddToWorkspaceAsync(
+                    "Project", LanguageNames.VisualBasic, CancellationToken.None)
+
+                ' These are the in-box VB codestyle analyzers that ship with the SDK
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "vb", "Microsoft.CodeAnalysis.CodeStyle.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "vb", "Microsoft.CodeAnalysis.CodeStyle.Fixes.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "vb", "Microsoft.CodeAnalysis.VisualBasic.CodeStyle.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "vb", "Microsoft.CodeAnalysis.VisualBasic.CodeStyle.Fixes.dll"))
+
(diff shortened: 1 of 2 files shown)


src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
```diff
@@ -1030,6 +1030,49 @@ public void RemoveAnalyzerReference(string fullPath)
...
+        {
+            // Map all files in the SDK directory that contains the Razor source generator to source generator files loaded from VSIX.
+            // Include the generator and all its dependencies shipped in VSIX, discard the generator and all dependencies in the SDK
+            return GetMappedRazorSourceGenerator(fullPath);
+        }
+
+        return OneOrMany.Create(fullPath);
+    }
+
+    private static readonly string s_csharpCodeStyleAnalyzerSdkDirectory = Path.Combine("Sdks", "Microsoft.NET.Sdk", "codestyle", "cs") + PathUtilities.DirectorySeparatorStr;
```
ToddGrun, line 1054: >PathUtilities.DirectorySeparatorStr [](http://example.com/codeflow?start=138&length=35)

Should this be at the beginning of the string too?

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
```diff
@@ -1030,6 +1030,49 @@ public void RemoveAnalyzerReference(string fullPath)
...
+    private static readonly string s_visualBasicCodeStyleAnalyzerSdkDirectory = Path.Combine("Sdks", "Microsoft.NET.Sdk", "codestyle", "vb") + PathUtilities.DirectorySeparatorStr;
+
+    private bool IsSdkCodeStyleAnalyzer(string fullPath)
+    {
+        if (Language == LanguageNames.CSharp &&
+            fullPath.LastIndexOf(s_csharpCodeStyleAnalyzerSdkDirectory, StringComparison.OrdinalIgnoreCase) + s_csharpCodeStyleAnalyzerSdkDirectory.Length - 1 ==
```
ToddGrun, line 1060: >LastIndexOf [](http://example.com/codeflow?start=21&length=11)

EndsWith?

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
```diff
@@ -1030,6 +1030,49 @@ public void RemoveAnalyzerReference(string fullPath)
...
+            fullPath.LastIndexOf(Path.DirectorySeparatorChar))
+        {
+            return true;
+        }
+
+        if (Language == LanguageNames.VisualBasic &&
+            fullPath.LastIndexOf(s_visualBasicCodeStyleAnalyzerSdkDirectory, StringComparison.OrdinalIgnoreCase) + s_visualBasicCodeStyleAnalyzerSdkDirectory.Length - 1 ==
+            fullPath.LastIndexOf(Path.DirectorySeparatorChar))
```
ToddGrun, line 1068: nit: this would be a bit easier to read as a separate fn since it's done in 3 places

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
sharwell, line 1054: 💭 Should we just look at the assembly name? With the current approach, if users explicitly reference a version of the code style package instead of using the one in the SDK, they will end up in the same bad IDE state we have today.

sharwell:
Note that this will load the analyzers which ship in the IDE, but it does _not_ correct the error where the wrong options are provided to those analyzers (analyzers in the IDE use values from Tools > Options, but those same analyzers in a build do not).

arkalyanms (approved):

src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
```diff
@@ -204,5 +204,36 @@ Namespace Microsoft.VisualStudio.LanguageServices.UnitTests.ProjectSystemShim
...
+                     "AnotherExtension")
+                }
+
+                Dim project = Await environment.ProjectFactory.CreateAndAddToWorkspaceAsync(
+                    "Project", LanguageNames.CSharp, CancellationToken.None)
+
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CodeStyle.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CodeStyle.Fixes.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CSharp.CodeStyle.dll"))
+                project.AddAnalyzerReference(Path.Combine(TempRoot.Root, "Sdks", "Microsoft.NET.Sdk", "codestyle", "cs", "Microsoft.CodeAnalysis.CSharp.CodeStyle.Fixes.dll"))
```
jaredpar, line 226: Are all of these DLLs already in box that we're mapping to?

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
```diff
@@ -1041,28 +1084,27 @@ public void RemoveAnalyzerReference(string fullPath)
...
-                return OneOrMany.Create(ImmutableArray<string>.Empty);
+        if (!vsixRazorAnalyzers.IsEmpty)
+        {
+            if (s_razorSourceGeneratorAssemblyRootedFileNames.Any(
+                static (fileName, fullPath) => fullPath.EndsWith(fileName, StringComparison.OrdinalIgnoreCase), fullPath))
+            {
+                return OneOrMany.Create(vsixRazorAnalyzers);
             }
+
+            return OneOrMany<string>.Empty;
```
jaredpar, line 1107: My one fear is that this is hard to test in a unit test. Do we have an item on the queue to do manual verification of Razor / Code style once this is all merged in?

Have we thought about adding an integration test that verifies we redirected these loads?

src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
```diff
@@ -204,5 +204,36 @@ Namespace Microsoft.VisualStudio.LanguageServices.UnitTests.ProjectSystemShim
...
+        Public Async Function CodeStyleAnalyzers_FromSdk() As Task
+            Using environment = New TestEnvironment()
+                Dim providerFactory = DirectCast(environment.ExportProvider.GetExportedValue(Of IVisualStudioDiagnosticAnalyzerProviderFactory), MockVisualStudioDiagnosticAnalyzerProviderFactory)
+                providerFactory.Extensions =
+                {
+                     ({
+                        Path.Combine(TempRoot.Root, "File.dll")
+                     },
+                     "AnotherExtension")
+                }
```
jasonmalinowski, line 218: Should we just remove this bit since it's not relevant to the test at hand? I imagine this has gotten copy/pastede from prior tests and didn't matter here. It'll also make the assert an Assert.Empty( )which will be a bit clearer.

src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
jasonmalinowski, line 226: Do resource binaries get added too?

jasonmalinowski: Probably needs VB too.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
```diff
@@ -1030,6 +1030,49 @@ public void RemoveAnalyzerReference(string fullPath)
...
+
+        return OneOrMany.Create(fullPath);
+    }
+
+    private static readonly string s_csharpCodeStyleAnalyzerSdkDirectory = Path.Combine("Sdks", "Microsoft.NET.Sdk", "codestyle", "cs") + PathUtilities.DirectorySeparatorStr;
+    private static readonly string s_visualBasicCodeStyleAnalyzerSdkDirectory = Path.Combine("Sdks", "Microsoft.NET.Sdk", "codestyle", "vb") + PathUtilities.DirectorySeparatorStr;
+
+    private bool IsSdkCodeStyleAnalyzer(string fullPath)
+    {
+        if (Language == LanguageNames.CSharp &&
```
jasonmalinowski, line 1059: Any reason not to just match on Sdks\Microsoft.NET.Sdk\codestyle and work for both languages at once?

src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
```diff
@@ -204,5 +204,36 @@ Namespace Microsoft.VisualStudio.LanguageServices.UnitTests.ProjectSystemShim
                 AssertEx.Equal({path1, path2}, environment.Workspace.CurrentSolution.Projects.Single().AnalyzerReferences.Select(Function(r) r.FullPath))
             End Using
         End Function
+
+        <WpfFact>
+        Public Async Function CodeStyleAnalyzers_FromSdk() As Task
```
jasonmalinowski, line 209:
```suggestion
        Public Async Function CodeStyleAnalyzers_FromSdk_AreIgnored() As Task
```

jasonmalinowski:
> The analyzers are still listed in the Solution Explorer but without any associated analyzers.

Do we have a design-time target somewhere we can add a hack in that removes these at the MSBuild level instead? This is a perfectly fine tactical fix of course, but that'd avoid side effects like this.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
```diff
@@ -1030,6 +1030,49 @@ public void RemoveAnalyzerReference(string fullPath)
...
 
+    private OneOrMany<string> GetMappedAnalyzerPaths(string fullPath)
+    {
+        fullPath = Path.GetFullPath(fullPath);
+
+        if (IsSdkCodeStyleAnalyzer(fullPath))
+        {
+            // We discard the CodeStyle analyzers added by the SDK when the EnforceCodeStyleInBuild property is set.
+            // The same analyzers ship in-box as part of the Features layer and are version matched to the compiler.
+            return OneOrMany<string>.Empty;
```
dibarbet, line 1041: Will this work as is for VSCode?  I noticed that we don't have a codestyle dll shipped in the extension - do we need that, or do we just need the Features dll?

jasonmalinowski, line 1041: Just to call it out explicitly, doing this here will:

1. Also apply to VS Code (which might be reasonable since we'll have similar issues there I imagine)
2. Will _not_ apply to MSBuildWorkspace, which might be bad, or might not be.

The latter we can of course defer for this change (since this has huge dogfooding benefit), but wanted to call out the former.

jasonmalinowski (approved):

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
JoeRobich: This does work in VS Code. We only need the Features assemblies.

src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
JoeRobich: I don't believe resource binaries are added as AnalyzerReferences to the project. I would expect they are loaded by the AnalyzerAssemblyLoader.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
JoeRobich: Happy to make that change. Didn't have a good feel with how specific we wanted to get.

JoeRobich: Actually will keep it as is. The code currently checks for the analyzer path to end with the folder path fragment which is a stronger guarantee than just containing the fragment.

src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
JoeRobich: Removed!

src/VisualStudio/Core/Test/ProjectSystemShim/VisualStudioProjectTests/AnalyzerReferenceTests.vb
JoeRobich: These are the in box code style analyzers. We are ensuring that they do not get returned as AnalyzerReferences from the Project. Which means they will not be used for analysis.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
```diff
@@ -1041,33 +1072,33 @@ public void RemoveAnalyzerReference(string fullPath)
...
+                return OneOrMany.Create(vsixRazorAnalyzers);
             }
+
+            return OneOrMany<string>.Empty;
         }
 
         return OneOrMany.Create(fullPath);
     }
 
+    private static string CreateDirectoryPathFragment(params string[] paths) => Path.Combine([" ", .. paths, " "]).Trim();
```
ToddGrun, line 1097: >" " [](http://example.com/codeflow?start=109&length=3)

Why did this switch from the directory separator char to a space?

JoeRobich: So Path.Combine will add the separator then we trim it off the spaces. Was just trying to put the work onto Path.Combine.

JoeRobich: Maybe it is dumb. Happy to revert it.

ToddGrun: Definitely not dumb, just a little too tricky for me to spot without the explanation.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
JoeRobich: I think it should be easy enough to write an integration test for this. Will do as a follow up after the snap.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
JoeRobich: I am going to do a follow up that adds an integration test. Will add a comment here as well.

src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs
chsienki: I really don't love that we're expanding this mechanism, given the work we're doing to allow the SDK to redirect analyzers via https://github.com/dotnet/roslyn/pull/74820 

That isn't a reason to block on this, given its not done yet, but could we open an issue to investigate if we could replace this with the new API when it's introduced?

KirillOsenkov:
@JoeRobich quick question: will it work if the paths to analyzers contain `..`?

I haven't tried in VS, but some other hosts receive paths like this:
`C:\Program Files\dotnet\sdk\9.0.100-rc.1.24452.12\Sdks\Microsoft.NET.Sdk\targets\..\codestyle\cs\Microsoft.CodeAnalysis.CodeStyle.dll`, so need to first call `Path.GetFullPath()` on the reference before calling IsSdk...

I haven't checked whether the VS project system normalizes paths before this codepath is hit.

JoeRobich:
> so need to first call `Path.GetFullPath()` on the reference before calling IsSdk...

Thanks for checking! We do call `Path.GetFullPath()` just before.

https://github.com/dotnet/roslyn/blob/3bff3622487486dec7794dfd0c71e05a52c313a4/src/Workspaces/Core/Portable/Workspace/ProjectSystem/ProjectSystemProject.cs#L1033-L1037

PR #75279, opened 2024-09-27
Revert "Updates content exclusion for on-the-fly-docs"

333fred:
Reverts dotnet/roslyn#75172. Broke NGEN.


akhera99 (approved):

333fred:
I'm going to bypass to get the insertion going; we have enough test passes that I can be confident we didn't break anything.

PR #75288, opened 2024-09-28
Improve doc comments for `IConditionalOperation.IsRef`

YoshiRulz:
see dotnet/roslyn#75107

### DIFF
--- a/src/Compilers/Core/Portable/Operations/OperationInterfaces.xml
+++ b/src/Compilers/Core/Portable/Operations/OperationInterfaces.xml
@@ -1137,7 +1137,8 @@
     </Property>
     <Property Name="IsRef" Type="bool">
       <Comments>
-        <summary>Is result a managed reference</summary>
+        <value><see langword="true"/> iff the result is a reference to a struct or class. Not to be confused with <see cref="Type"/>.<see cref="ITypeSymbol.IsReferenceType"/>.</value>
+        <remarks>This occurs in C# for ternaries whose branches use <see langword="ref"/>.</remarks>
       </Comments>
     </Property>
   </Node>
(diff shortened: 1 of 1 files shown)


jaredpar:
@333fred do we want a change here? Discussion didn't mention if it thinks docs sufficient or insufficient as is.

src/Compilers/Core/Portable/Operations/OperationInterfaces.xml
```diff
@@ -1137,7 +1137,8 @@
     </Property>
     <Property Name="IsRef" Type="bool">
       <Comments>
-        <summary>Is result a managed reference</summary>
+        <value><see langword="true"/> iff the result is a reference to a struct or class. Not to be confused with <see cref="Type"/>.<see cref="ITypeSymbol.IsReferenceType"/>.</value>
+        <remarks>This occurs in C# for ternaries whose branches use <see langword="ref"/>.</remarks>
```
333fred, line 1141: I'm fine with the addition of the remark section, but I think I'd prefer a slight rewording of the original summary, rather than this change. Something like `True if the result is by-reference. I don't think we want to include the works `struct` or `class` here. I also don't think that cref will format correctly.

jaredpar: I didn't see a lot of 'iff' in our documentation.

YoshiRulz:
```suggestion
        <value><see langword="true"/> if the result is by-reference. Not to be confused with <c>op.Type.IsReferenceType</c>; by-ref can be used with either.</value>
        <remarks>This occurs in C# for ternaries whose branches use <see langword="ref"/>.</remarks>
```
?

(You were right about the links not working, at least in VS Code w/ dev kit plugin.)

333fred: I don't think we really need the second sentence at all, being perfectly honest.

333fred (approved):

333fred:

jjonescz (approved):

jcouv:
Merged/squashed. Thanks for the contribution @YoshiRulz !

Issue #75237, opened 2024-09-25
Compiler should error when it hits native PDB writer limits


jaredpar:
The native PDB writer has a size limit of 0x10000 for calls to `ISymUnmanagedWriter::SetSymAttribute`. The writer does not enforce this limit on calls to `SetSymAttribute` but rather in `CloseMethod` where it returns `E_OUTOFMEMORY`. That leads to us emitting errors like 

> error CS0041: Unexpected error writing debug information -- 'Insufficient memory to continue the execution of the program'

That is very difficult for customers and us to debug. To help with root causing these problems we should emit a better error in the compiler when we hit this limit. Specifically the error should take into account the 0x10000 limit and issue an error message that includes the name of the method that was being emitted when the limit was hit. 

This has come up two times now: #73447 and an internal feedback ticket. Having this better error would've saved us significant time in tracking down the problem.

PR #75293, opened 2024-09-30
Report error for native PDB limit

jjonescz:
Resolves https://github.com/dotnet/roslyn/issues/75237.


src/Compilers/CSharp/Test/Emit2/PDB/PDBTests.cs
```diff
@@ -367,6 +368,83 @@ public void SymWriterErrors4()
...
+        }
+
+        [ConditionalTheory(typeof(WindowsOnly), Reason = ConditionalSkipReason.NativePdbRequiresDesktop)]
+        [WorkItem("https://github.com/dotnet/roslyn/issues/75237")]
+        [InlineData(0x10_000)]
+        [InlineData(0x20_000)]
+        public void NativeWriterLimit_Over(int length)
+        {
+            CompileWithMockedCustomMetadata(length).Diagnostics.Verify(
+                // error CS0041: Unexpected error writing debug information -- 'Insufficient memory to continue the execution of the program.'
```
jcouv, line 387: I'm confused. This test is linked to [https://github.com/dotnet/roslyn/issues/75237](https://github.com/dotnet/roslyn/issues/75237) but exhibits the same/current behavior described in the issue.

Is this a different scenario than what customers had reported? Should we report a more specific error here (which limit was hit)?

src/Compilers/Core/Portable/NativePdbWriter/PdbWriter.cs
```diff
@@ -148,6 +148,16 @@ public void SerializeDebugInfo(IMethodBody methodBody, StandaloneSignatureHandle
...
             {
+                const int limit = 0x10_000;
+                if (blob.Length > limit)
+                {
+                    throw new SymUnmanagedWriterException(string.Format(
+                        CodeAnalysisResources.SymWriterMetadataOverLimit,
+                        methodBody.MethodDefinition,
+                        blob.Length,
+                        limit));
+                }
```
jaredpar, line 159: Note: the name of the attribute is part of the limit length. Is that included in the `blob`?

src/Test/PdbUtilities/Writer/SymWriterTestUtilities.cs
```diff
@@ -13,5 +13,10 @@ internal static class SymWriterTestUtilities
     {
         public static readonly Func<ISymWriterMetadataProvider, SymUnmanagedWriter> ThrowingFactory =
             _ => throw new SymUnmanagedWriterException("xxx", new NotSupportedException(), "<lib name>");
+
+        public static SymUnmanagedWriter CreateUnmanagedWriter(ISymWriterMetadataProvider metadataProvider)
```
jcouv, line 17: What's the benefit of this factory method vs. using `SymUnmanagedWriterFactory.CreateWriter` directly?

src/Test/PdbUtilities/Writer/CustomMetadataSymUnmanagedWriter.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DiaSymReader;
+
+namespace Roslyn.Test.PdbUtilities;
+
+internal sealed class CustomMetadataSymUnmanagedWriter(SymUnmanagedWriter target, byte[] customMetadata) : DelegatingSymUnmanagedWriter(target)
```
jcouv, line 9: >CustomMetadataSymUnmanagedWriter [](http://example.com/codeflow?start=22&length=32)

nit: if we don't feel this type will be used more generally, consider moving into `PDBTests.cs`

jcouv (commented):
Done with review pass (iteration 1). Only minor questions

src/Test/PdbUtilities/Writer/CustomMetadataSymUnmanagedWriter.cs
jjonescz: I think it needs to be in this separate utility assembly, because in PDBTests some of the types would conflict between Microsoft.DiaSymReader and Microsoft.CodeAnalysis which are both referenced from there

src/Test/PdbUtilities/Writer/SymWriterTestUtilities.cs
jjonescz: SymUnmanagedWriterFactory when used from PDBTests raises a compile-time error because the type is defined both in Microsoft.DiaSymReader and Microsoft.CodeAnalysis assemblies. This utility project only references Microsoft.CodeAnalysis, so this factory method was a simple way to workaround that problem (without the need for extern aliases).

src/Compilers/CSharp/Test/Emit2/PDB/PDBTests.cs
jjonescz: The purpose of these two tests (`_Over` and `_Under`) is to verify the limit is indeed 0x10000. To do that, they use a custom writer which intercepts metadata writes and passes a custom buffer with the specified size. The specific error reporting logic happens before that, so it's not hit. I don't think there's a straightforward way to test that with exact buffer sizes. But `NativeWriterLimit_EndToEnd` tests the specific error (albeit the buffer size cannot be set exactly).

src/Compilers/Core/Portable/NativePdbWriter/PdbWriter.cs
jjonescz: No, the name is not included in the `blob`. You are right the limit is wrong now, will fix.

src/Compilers/CSharp/Test/Emit2/PDB/PDBTests.cs
jcouv: Thanks for the context. Makes sense. Consider leaving comments (this is not so much about testing the compiler as testing the limit of library that we depend on)

jcouv (approved):
LGTM Thanks (iteration 2) with a minor comment suggestion

jjonescz:
@dotnet/roslyn-compiler @jaredpar for a second review, thanks

jaredpar (approved):

Issue #65176, opened 2022-11-02
Feature 'Edit and Continue' currently unavailable due to an internal error. Read out of bounds.


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Feature-Edit-and-Continue-currently-un/10172594)._

---
[severity:I'm unable to use this version]
Each time I am trying to debug unit test from Test Explorer I am getting "Feature 'Edit and Continue' is currently unavailable due to an internal error."
I had following version of Visual Studio 2022:17.3.6(when error started I had 17.3.3)

To reporuduce:
1. Start project with unit tests.
2. Open Tests Explorer
3. Right click any test and select "Debug"
4.. Observe the issue. Break point in test would not be hit.

Things I tried to remedy the situation, but failed to resolve the issue:
1. Installed VS 2022 version 17.3.6
2. Tried to repair visual studio

System info:
Microsoft Visual Studio Enterprise 2022
Version 17.3.6
VisualStudio.17.Release/17.3.6+32929.385
Microsoft .NET Framework
Version 4.8.04084

Here is stack trace:
```
StreamJsonRpc.RemoteInvocationException: Read out of bounds.
   at StreamJsonRpc.JsonRpc.d__143`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.CodeAnalysis.EditAndContinue.RemoteEditAndContinueServiceProxy.<>c__DisplayClass5_0.<b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.CodeAnalysis.Remote.BrokeredServiceConnection`1.d__20`1.MoveNext()
RPC server exception:
System.BadImageFormatException: Read out of bounds.
      at System.Reflection.Throw.OutOfBounds()
[...]

PR #75304, opened 2024-09-30
Improve error reporting when reading bad metadata during EnC

tmat:
Helps diagnosing issues like https://github.com/dotnet/roslyn/issues/65176

### DIFF
--- a/src/Compilers/Core/Portable/Emit/EditAndContinue/DefinitionMap.cs
+++ b/src/Compilers/Core/Portable/Emit/EditAndContinue/DefinitionMap.cs
@@ -291,14 +291,15 @@ namespace Microsoft.CodeAnalysis.Emit
                     debugInfo = Baseline.DebugInformationProvider(methodHandle);
                     localSignature = Baseline.LocalSignatureProvider(methodHandle);
                 }
-                catch (Exception e) when (e is InvalidDataException or IOException)
+                catch (Exception e) when (e is InvalidDataException or IOException or BadImageFormatException)
                 {
                     diagnostics.Add(MessageProvider.CreateDiagnostic(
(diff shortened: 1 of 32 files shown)


tmat:
@davidwengier @dotnet/roslyn-compiler ptal

davidwengier (approved):

src/Compilers/Core/Portable/Emit/EditAndContinue/DefinitionMap.cs
```diff
@@ -690,7 +692,7 @@ static ImmutableArray<string> getHoistedVariableNames(ImmutableArray<ISymbolInte
             {
                 provider = Baseline.DebugInformationProvider(MetadataTokens.MethodDefinitionHandle(methodRowId));
             }
-            catch (Exception e) when (e is InvalidDataException or IOException)
+            catch (Exception e) when (e is InvalidDataException or IOException or BadImageFormatException)
```
AlekseyTs, line 695: >catch (Exception e) when (e is InvalidDataException or IOException or BadImageFormatException) [](http://example.com/codeflow?start=12&length=94)

We don't want to report anything for this `catch`?

AlekseyTs (approved):
LGTM (commit 1)

src/Compilers/Core/Portable/Emit/EditAndContinue/DefinitionMap.cs
tmat: It's not a critical error. If a method is deleted and its body is not updated to throw an exception it would execute the original IL. Perhaps we could add a new warning but I don't think it's worth the effort.

tmat:
@jaredpar @cston @jjonescz for 2nd review.

cston (approved):

PR #75305, opened 2024-09-30
Fix crash with Dispose method with params

jcouv:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2187060

The issue is in `OverloadResolution.AddMemberToCandidateSet` which may or may not compute a normal result (`normalResult`) on line 1047, then decides whether to keep the normal or expanded result on line 1079 (`PreferExpandedFormOverNormalForm`).
In the failing scenario, we skip computing a normal result but end up preferring the normal result (which is `default`) which then would crash in `OverloadResolutionResult.ReportDiagnostics` (assertion in debug mode in `AssertNone(MemberResolutionKind.None)` and throwing `UnexpectedValue` at line 457 in release mode).
The fix is to prefer the expanded result since we didn't compute a normal result.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
@@ -1141,6 +1141,9 @@ outerDefault:
                     }
 
                     break;
+
+                case MemberResolutionKind.None:
+                    return true;
             }
             return false;
 
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
@@ -1044,7 +1044,8 @@ outerDefault:
 
             // Second, we need to determine if the method is applicable in its normal form or its expanded form.
             bool disallowExpandedNonArrayParams = (options & Options.DisallowExpandedNonArrayParams) != 0;
-            var normalResult = ((options & Options.IgnoreNormalFormIfHasValidParamsParameter) != 0 && IsValidParams(_binder, leastOverriddenMember, disallowExpandedNonArrayParams, out _))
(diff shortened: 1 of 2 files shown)


333fred (approved):

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1141,6 +1141,9 @@ private static bool PreferExpandedFormOverNormalForm<TMember>(MemberResolutionRe
                     }
 
                     break;
+
+                case MemberResolutionKind.None:
```
AlekseyTs, line 1145: >case MemberResolutionKind.None: [](http://example.com/codeflow?start=16&length=31)

I think it might be better to adjust the caller to never call this helper if ```IsMemberApplicableInNormalForm``` was not called. Basically, it is not normal to pass an uninitialized ```MemberResolutionResult``` around, The caller explicitly makes a choice to not initialize ```normalResult``` and, I think, it should be responsible for not passing it around. The logic could also be clearer because we will not have to rely on an indirect way to detect an uninitialized input. Also, it looks like, if ```Options.IsMethodGroupConversion``` is set, perhaps by mistake, the caller might still put the uninitialized resolution result into the output.

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -1063,6 +1064,7 @@ private void AddMemberToCandidateSet<TMember>(
                 // tricks you can pull to make overriding methods [indexers] inconsistent with overridden
                 // methods [indexers] (or implementing methods [indexers] inconsistent with interfaces). 
 
+                Debug.Assert((options & Options.IgnoreNormalFormIfHasValidParamsParameter) == 0 || (options & Options.IsMethodGroupConversion) == 0);
```
AlekseyTs, line 1067: >Debug.Assert((options & Options.IgnoreNormalFormIfHasValidParamsParameter) == 0 || (options & Options.IsMethodGroupConversion) == 0); [](http://example.com/codeflow?start=16&length=133)

Consider asserting this on entry to the function

AlekseyTs (approved):
LGTM (commit 2)

333fred (approved):

AlekseyTs (approved):
LGTM (commit 3)

PR #75325, opened 2024-10-01
Add warning when property has a backing field but a manually implemented accessor does not use it

RikkiGibson:
#57012

The goal of this change is to catch a scenario where user is in an in-between state of using the field keyword or not. See `MigratingToFieldKeyword_` tests.

This is designed to work as a basically declaration-level check. The goal was for it to be cheap and for diagnostics to appear relatively early without needing to e.g. compile the whole property and constructors, and analyze definite assignment, in order for them to appear. (i.e. a mini version of what happens at the compilation-level for ordinary fields.)

It's not meant to be airtight, but to just catch simple cases. For example, it won't complain if the keyword is used in all accessors, yet none of the uses are writing to it.

The design has a slightly odd rule as well relating to initializers. Basically, presence of an initializer removes the need for setter to use the backing field. This is because *something* is certainly setting the backing field if an initializer is present. This scenario was suggested by @jnm2, and we think would come up in a situation where an interface property is "read/write" but the implementation is readonly. See `MigratingToFieldKeyword_05`.


RikkiGibson:
@cston @333fred for review

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -31,12 +31,13 @@ private enum Flags : ushort
             IsExpressionBodied = 1 << 0,
             HasAutoPropertyGet = 1 << 1,
             HasAutoPropertySet = 1 << 2,
-            UsesFieldKeyword = 1 << 3,
-            IsExplicitInterfaceImplementation = 1 << 4,
-            HasInitializer = 1 << 5,
-            AccessorsHaveImplementation = 1 << 6,
-            HasExplicitAccessModifier = 1 << 7,
-            RequiresBackingField = 1 << 8,
+            GetterUsesFieldKeyword = 1 << 4,
```
cston, line 34: >4 [](http://example.com/codeflow?start=42&length=1)

`3`, `4`, ...?

cston:
>                     public string Prop { get; set => throw null!; } // 1

Minor: The `!` in this test shouldn't be necessary.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs:347 in b6903fd. [](commit_id = b6903fd284662b343199ed5a5b43af50dab5b4e3, deletion_comment = False)

cston:
>                     partial object P1

Consider testing declarations appearing after implementations as well, in this test and the next test. (See existing `PartialProperty_*` tests that use `bool reverseOrder`.)

---
Refers to: src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs:264 in b6903fd. [](commit_id = b6903fd284662b343199ed5a5b43af50dab5b4e3, deletion_comment = False)

cston:
>                 //     public string Prop { get; set => throw null!; } // 1

The warning for this case seems unfortunate.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs:363 in b6903fd. [](commit_id = b6903fd284662b343199ed5a5b43af50dab5b4e3, deletion_comment = False)

cston:
>             );

Please revert the formatting change.

---
Refers to: src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs:1945 in b6903fd. [](commit_id = b6903fd284662b343199ed5a5b43af50dab5b4e3, deletion_comment = False)

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests.cs
```diff
@@ -2396,6 +2396,9 @@ public class Test
                 // (6,21): warning CS0657: 'field' is not a valid attribute location for this declaration. Valid attribute locations for this declaration are 'method, return'. All attributes in this block will be ignored.
                 //     public int P { [field: A] get => throw null; set => throw null; }
                 Diagnostic(ErrorCode.WRN_AttributeLocationOnBadDeclaration, "field").WithArguments("field", "method, return").WithLocation(6, 21),
+                // (6,50): warning CS9265: The 'set' accessor of property 'Test.P' should use the backing 'field' because the other accessor is using it.
+                //     public int P { [field: A] get => throw null; set => throw null; }
```
cston, line 2400: >public int P { [field: A] get => throw null; set => throw null; } [](http://example.com/codeflow?start=23&length=65)

Are we generating a backing field for this property? That seems like a bug, even though the attribute is an error.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Locations.cs
```diff
@@ -632,24 +632,36 @@ int Goo
 }
 ";
             CreateCompilation(source).VerifyDiagnostics(
+                // (25,9): warning CS9265: The 'set' accessor of property 'C.Goo' should use the backing 'field' because the other accessor is using it.
+                //         set { }
```
cston, line 636: >set { } [](http://example.com/codeflow?start=27&length=7)

Are we generating a backing field for this property?

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests.cs
cston: Logged https://github.com/dotnet/roslyn/issues/75459.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -305,6 +312,47 @@ protected void CheckInitializerIfNeeded(BindingDiagnosticBag diagnostics)
...
 
+#nullable enable
+        private void CheckFieldKeywordUsage(BindingDiagnosticBag diagnostics)
+        {
+            if (!this.DeclaringCompilation.IsFeatureEnabled(MessageID.IDS_FeatureFieldKeyword))
+            {
+                return;
+            }
+
+            check((SourcePropertySymbolBase?)PartialImplementationPart ?? this, diagnostics);
```
333fred, line 323: Consider inlining this local function.

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -8002,4 +8002,10 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="WRN_UninitializedNonNullableBackingField_Title" xml:space="preserve">
     <value>Non-nullable property must contain a non-null value when exiting constructor. Consider adding the 'required' modifier, or declaring the property as nullable, or adding '[field: MaybeNull, AllowNull]' attributes.</value>
   </data>
+  <data name="WRN_AccessorDoesNotUseBackingField" xml:space="preserve">
+    <value>The '{0}' accessor of property '{1}' should use the backing 'field' because the other accessor is using it.</value>
```
333fred, line 8006: Honestly, `the backing 'field'` reads very awkwardly. Maybe just `should use 'field'`?

RikkiGibson:
> > ```
> >             //     public string Prop { get; set => throw null!; } // 1
> > ```
> 
> The warning for this case seems unfortunate.
> 
> Refers to: src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs:363 in [b6903fd](https://github.com/dotnet/roslyn/commit/b6903fd284662b343199ed5a5b43af50dab5b4e3). [](commit_id = [b6903fd](https://github.com/dotnet/roslyn/commit/b6903fd284662b343199ed5a5b43af50dab5b4e3), deletion_comment = False)

I think this case itself, I am fine with. We can actually see that nothing is setting that field--why are you using a field?

One I am less sure about is `string Prop { get => field ??= Compute(); set => throw null!; }`. In this case an initializer actually needs to be added to silence the warning, which the diagnostic message isn't clear about currently.

At the same time I am reluctant to abandon this syntax/declaration-based approach, and do a definite assignment pass instead a la ordinary fields, because for the getter-only case we would potentially need to analyze the constructors to know if the field is being set.

jnm2:
My personal preference would be for airtightness, and the lack of earliness (having to wait for a full compilation to update the warnings) hasn't been a noticeable problem for me with existing fields and field-like events.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests.cs
cston: This issue should be resolved by https://github.com/dotnet/roslyn/pull/75461.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Locations.cs
RikkiGibson: It looks like #75461 will fix this.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Locations.cs
```diff
@@ -684,22 +696,33 @@ int Goo
 }
 ";
             CreateCompilation(source).VerifyDiagnostics(
+                // (11,9): warning CS9265: The 'get' accessor of property 'C.Goo' should use the backing 'field' because the other accessor is using it.
+                //         get { return 0; }
```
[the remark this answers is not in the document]
RikkiGibson: It looks like #75461 will fix this.

cston:
It looks like the CI build failures are due to the fix in https://github.com/dotnet/roslyn/pull/75461.

RikkiGibson:
Have addressed feedback, merged in main, and updated more test baselines.

cston (approved):

333fred (approved):

Issue #75315, opened 2024-10-01
"Field is never assigned" warning inconsistent for ref fields


jjonescz:
**Version Used**: 4.13.0-1.24480.10 (ba7fe3580bc60e6b25729a07c7886b25bcd36c25)

**Steps to Reproduce**:

```cs
ref struct R
{
    public ref int F; // warning CS0649: Field 'R.F' is never assigned to, and will always have its default value 0

    static void M(in int x) { }

    static void Test(R r)
    {
        M(r.F);
    }
}
```

compare with:

```cs
ref struct R
{
    public ref int F; // no warning

    static void M(in int x) { }

    static void Test(R r)
    {
        M(in r.F); // `in` added here
    }
}
```

**Expected Behavior**: The two code snippets are equivalent, so I would expect the warning to be either reported for both or for none of them.

**Actual Behavior**: The warning is reported inconsistently.

**Additional Notes**:
- The warning is also not reported for other callsite modifiers like `ref` (but one cannot reproduce the "no callsite modifier" scenario there). Perhaps the warning is not supposed to be reported for ref fields at all.
- If the warning should be reported for ref fields, it shouldn't say that `R.F` will always have its default value 0 - because it's a ref field, it will have its default value `null`.

jaredpar:
I wouldn't expect `in` to remove the warning here because it's not reassigning the `ref`. I would expect only a `ref` reassignment to remove the warning.

PR #75340, opened 2024-10-02
Fix assignment analysis of ref fields

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75315.

A field cannot be assigned-to via a readonly reference, but previously we considered any reference to be a write and hence didn't emit the "field is never assigned" warning.

Similarly we didn't distinguish between ref assignments and value assignments of ref fields - now we consider only ref assignments to be writes to ref fields.

This also improves the warning to say that any never-assigned *ref* field will have its default value set to "null reference".

### DIFF
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
@@ -346,7 +346,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                         // All primary constructor parameters are definitely assigned outside of the primary constructor
                         foreach (var parameter in primaryConstructor.Parameters)
                         {
-                            NoteWrite(parameter, value: null, read: true, isRef: false);
+                            NoteWrite(parameter, value: null, read: true, isRef: parameter.RefKind != RefKind.None);
                         }
 
                         CurrentSymbol = save;
@@ -1881,7 +1881,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             {
                 // this code has no effect except in region analysis APIs such as DataFlowsOut where we unassign things
                 if (slot > 0) SetSlotState(slot, true);
-                NoteWrite(parameter, value: null, read: true, isRef: false);
+                NoteWrite(parameter, value: null, read: true, isRef: parameter.RefKind != RefKind.None);
             }
 
             if (parameter is SourceComplexParameterSymbolBase { ContainingSymbol: LocalFunctionSymbol or LambdaSymbol } sourceComplexParam)
@@ -2765,7 +2765,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 int slot = GetOrCreateSlot(iterationVariable);
                 if (slot > 0) SetSlotAssigned(slot);
                 // NOTE: do not report unused iteration variables. They are always considered used.
-                NoteWrite(iterationVariable, null, read: true, isRef: false);
+                NoteWrite(iterationVariable, null, read: true, isRef: iterationVariable.RefKind != RefKind.None);
             }
         }
 
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
@@ -951,6 +951,9 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
         }
 
+        /// <param name="isRef">
+        /// Whether this write represents a ref-assignment.
+        /// </param>
         private void NoteWrite(BoundExpression n, BoundExpression value, bool read, bool isRef)
         {
             while (n != null)
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
@@ -346,7 +346,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                         // All primary constructor parameters are definitely assigned outside of the primary constructor
                         foreach (var parameter in primaryConstructor.Parameters)
                         {
-                            NoteWrite(parameter, value: null, read: true);
+                            NoteWrite(parameter, value: null, read: true, isRef: false);
                         }
 
                         CurrentSymbol = save;
@@ -846,7 +846,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
         }
 
-        protected virtual void NoteWrite(Symbol variable, BoundExpression value, bool read)
+        protected virtual void NoteWrite(Symbol variable, BoundExpression value, bool read, bool isRef)
         {
             if ((object)variable != null)
             {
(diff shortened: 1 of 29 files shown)


src/Compilers/CSharp/Test/Emit3/FlowAnalysis/RegionAnalysisTests.cs
```diff
@@ -14109,6 +14109,9 @@ ref struct RS
                 """,
                 targetFramework: TargetFramework.NetCoreApp);
             comp.VerifyEmitDiagnostics(
+                // (3,13): warning CS0649: Field 'RS.ri' is never assigned to, and will always have its default value (null reference)
```
RikkiGibson, line 14112: The wording of the diagnostic feels a little confusing. Perhaps it would make sense to introduce a new message for ref fields which indicates the need to ref-assign.

src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
```diff
@@ -346,7 +346,7 @@ protected override ImmutableArray<PendingBranch> Scan(ref bool badRegion)
                         // All primary constructor parameters are definitely assigned outside of the primary constructor
                         foreach (var parameter in primaryConstructor.Parameters)
                         {
-                            NoteWrite(parameter, value: null, read: true);
+                            NoteWrite(parameter, value: null, read: true, isRef: false);
```
RikkiGibson, line 349: Does this behave correctly when a primary constructor parameter is by-`ref`? For example, do ref parameters in primary constructors behave consistently with by-value parameters in the dataflow APIs?

src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
```diff
@@ -2748,20 +2764,8 @@ public override void VisitForEachIterationVariables(BoundForEachStatement node)
...
-        }
-
-        public override BoundNode VisitObjectInitializerMember(BoundObjectInitializerMember node)
-        {
-            var result = base.VisitObjectInitializerMember(node);
-
-            if ((object)_sourceAssembly != null && node.MemberSymbol != null && node.MemberSymbol.Kind == SymbolKind.Field)
-            {
-                _sourceAssembly.NoteFieldAccess((FieldSymbol)node.MemberSymbol.OriginalDefinition, read: false, write: true);
+                NoteWrite(iterationVariable, null, read: true, isRef: false);
```
RikkiGibson, line 2767: Does this behave correctly (e.g. is there any behavior change in the dataflow APIs) for a case like the following?

```cs
void M(Span<int> span) {
    foreach (ref var x in span) { }
}
```

src/Compilers/CSharp/Test/Emit3/FlowAnalysis/RegionAnalysisTests.cs
jjonescz: Sounds good. Does this wording seem fine?

> Field '{0}' is never ref-assigned to, and will always have its default value (null reference)

src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
jjonescz: I don't think this should affect dataflow APIs in any way, since the `isRef` is only passed down to `_sourceAssembly.NoteFieldAccess` which isn't part of dataflow but is only used to report the "unused field" diagnostics IIUC. Moreover, it's only used if the symbol is a field but here it's a parameter, so that branch won't be reached.

That being said, in case we used the parameter elsewhere in the future, we could probably pass `isRef: parameter.RefKind != RefKind.None` here.

src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
```diff
@@ -961,7 +963,9 @@ private void NoteWrite(BoundExpression n, BoundExpression value, bool read)
                             if ((object)_sourceAssembly != null)
                             {
                                 var field = fieldAccess.FieldSymbol.OriginalDefinition;
-                                _sourceAssembly.NoteFieldAccess(field, read: value == null || WriteConsideredUse(fieldAccess.FieldSymbol.Type, value), write: true);
+                                _sourceAssembly.NoteFieldAccess(field,
+                                    read: value == null || WriteConsideredUse(fieldAccess.FieldSymbol.Type, value),
+                                    write: field.RefKind == RefKind.None || isRef);
```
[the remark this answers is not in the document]
jjonescz: Sounds good, I can add `isRef = false`. I don't think I can assert anything useful because the receiver can be a ref field in an error scenario (I will add a test).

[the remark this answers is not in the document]
jjonescz: Good catch, I think `isRef = false` should be there. Even though it would be an error scenario today to have a ref field to a ref struct, it could be allowed in the future, and the warnings can be observed even today. And if we are ref-assigning (i.e., `isRef = true`) we don't want to propagate that to the receiver (we are not ref assigning the receiver).

src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
jjonescz: Similar response as above, there should be no change, since `iterationVariable` is a `LocalSymbol`, hence `isRef` has no effect on it.

src/Compilers/CSharp/Test/Emit3/FlowAnalysis/RegionAnalysisTests.cs
jjonescz: Btw, technically that will be a new warning that can be reported for existing code - even where it wasn't previously reported due to the bug being fixed - but also where it was reported but under a different error code. So should it be part of a new warning wave?

RikkiGibson: > So should it be part of c# 14 warning wave?

I think the simplest, lowest risk way will be to use new wording and put in a warning wave.

RikkiGibson (approved):

RikkiGibson (approved):

jjonescz:

src/Compilers/CSharp/Portable/FlowAnalysis/DefiniteAssignment.cs
```diff
@@ -846,15 +846,17 @@ private void NoteRead(BoundNode fieldOrEventAccess)
             }
         }
 
-        protected virtual void NoteWrite(Symbol variable, BoundExpression value, bool read)
+        protected virtual void NoteWrite(Symbol variable, BoundExpression value, bool read, bool isRef)
```
jaredpar, line 849: Does `isRef` mean literally `ref` or any kind like `in / out`. Consider adding a comment here to make the intent clear.

jjonescz: It means "ref assignment", so there's no in/out variant. Added comment.

jaredpar (approved):

Issue #74275, opened 2024-07-05
```MakeInvocationExpression``` helper doesn't belong in ```SyntheticBoundNodeFactory```

### ISSUE THREAD

**Issue by AlekseyTs** [dotnet/roslyn#74275]
Multiple helpers in ```SyntheticBoundNodeFactory``` make use of  ```MakeInvocationExpression``` helper,
which actually uses a `Binder` to bind an invocation expression (do overload resolution, apply necessary
conversions to the arguments, etc.). Therefore, there is no guarantee that it is going to produce a node in its lowered form, which we generally expect from  ```SyntheticBoundNodeFactory``` to do. The helper is used indirectly in lowering and in the post lowering phases of the compiler. When the operation is performed against an unknown set of overloads (and that is how it is used today) the result is unpredictable in terms of what we might get back. For example, we might get back a node that is not going to pass `DiagnosticPass` checks with respect to operations allowed in an expression tree, resulting in a compiler crash (see, for example, #74163 caused by exactly that). Or we might get back a not lowered node in a post lowering phase, etc.

I propose the following action:
1. Remove the helper in the current form from ```SyntheticBoundNodeFactory```
2. If a binding against an unknown set of overload should occur, that should be done in initial binding phase, the resulting node should be stored "on a side", used by the `LocalRewriter` and be a subject to ```DiagnosticPass``` checks. We have plenty examples of how to do that, just search for use of "placeholders" in bound nodes.
3. Otherwise, we should consider binding against a known and restricted set of overloads with known argument types, so that we can guarantee that we are going to get back a node that doesn't require lowering and cannot contain "forbidden" operations.   

```
        internal BoundExpression MakeInvocationExpression(
            BinderFlags flags,
            SyntaxNode node,
            BoundExpression receiver,
            string methodName,
            ImmutableArray<BoundExpression> args,
            BindingDiagnosticBag diagnostics,
            ImmutableArray<TypeSymbol> typeArgs = default(ImmutableArray<TypeSymbol>),
            bool ignoreNormalFormIfHasValidParamsParameter = false)
        {
            if (_binder is null || _binder.Flags != flags)
            {
                _binder = new SyntheticBinderImpl(this).WithFlags(flags);
            }

            return _binder.MakeInvocationExpression(
                node,
                receiver,
                methodName,
                args,
                diagnostics,
                typeArgs: typeArgs.IsDefault ? default(ImmutableArray<TypeWithAnnotations>) : typeArgs.SelectAsArray(t => TypeWithAnnotations.Create(t)),
[...]

PR #75341, opened 2024-10-02
Do not use overload resolution while lowering interpolated strings

AlekseyTs:
Related to #74275.


src/Compilers/CSharp/Portable/Binder/Binder_InterpolatedString.cs
```diff
@@ -241,29 +243,32 @@ private BoundInterpolatedString BindUnconvertedInterpolatedStringToString(BoundU
...
-
-            if (tryBindAsHandlerType(out var result))
+            if ((unconvertedInterpolatedString.Parts.Length > 4 || !AllInterpolatedStringPartsAreStrings(unconvertedInterpolatedString.Parts)) &&
+                tryBindAsHandlerType(out var result))
             {
                 // Case 3
                 return result;
             }
 
             // The specifics of 4 vs 5 aren't necessary for this stage of binding. The only thing that matters is that every part needs to be convertible
```
333fred, line 257: Should this comment be updated, since we're now checking whether we can lower to `Concat` or not?

src/Compilers/CSharp/Portable/Binder/Binder_InterpolatedString.cs
```diff
@@ -288,6 +293,151 @@ bool tryBindAsHandlerType([NotNullWhen(true)] out BoundInterpolatedString? resul
...
+                name,
+                expressions,
+                diagnostics,
+                typeArgs: default(ImmutableArray<TypeWithAnnotations>),
+                allowFieldsAndProperties: false,
+                ignoreNormalFormIfHasValidParamsParameter: true, // if an interpolation expression is the null literal, it should not match a params parameter.
+                disallowExpandedNonArrayParams: InExpressionTree);
+
+            // We do not verify expected return type of the chosen factory method.
+            // This is technically a spec violation because there is no guaranty what
```
333fred, line 356:
```suggestion
            // This is technically a spec violation because there is no guarantee what
```

src/Compilers/CSharp/Portable/Binder/Binder_InterpolatedString.cs
```diff
@@ -288,6 +293,151 @@ bool tryBindAsHandlerType([NotNullWhen(true)] out BoundInterpolatedString? resul
...
+                    }
+                }
+
+                expressions[0] = new BoundLiteral(syntax, ConstantValue.Create(formatString.ToStringAndFree()), GetSpecialType(Microsoft.CodeAnalysis.SpecialType.System_String, diagnostics, syntax)) { WasCompilerGenerated = true };
+                return expressions.ToImmutableAndFree();
+            }
+
+            static string escapeInterpolatedStringLiteral(string value)
+            {
+                var builder = PooledStringBuilder.GetInstance();
```
333fred, line 422: I know you're just copying this over, but... this double builder seems useless to me. This method is only ever called to append to a different builder. Consider just passing in `stringBuilder` and making this return void.

333fred (approved):

AlekseyTs:

jjonescz (approved):

Issue #75353, opened 2024-09-28
crash on assigning result of a switch expression to a constant


thalber:
## Environment data
`dotnet --info` output: [info.txt](https://github.com/user-attachments/files/17175967/info.txt)

VS Code version: 1.93.1
C# Extension version: 2.45.25, 2.39.29, presumably others

## OmniSharp log

[4-C#.log](https://github.com/user-attachments/files/17175941/4-C.log)

## Steps to reproduce

1. Open any .cs file in the editor that contains following code:

```cs
const float //this is where i forgot to write the rest of the line lol
angVel = (firstChunk.vel.magnitude, angVel, firstChunk.ContactPoint.x, firstChunk.contactPoint.y) switch
{
	( <= 1f, _, _, _) => 0f,
	( >= 1f, 0f, not 0, 0) => Random.Range(5, 10f) * Mathf.Sign(firstChunk.vel.x),
	( >= 1f, _, not 0, -1) => 1f,
	_ => angVel,
};
rot = Mathf.LerpAngle(rot, rot + angVel, 1f);
```

## Expected  behavior

A compiler error is displayed for assigning a non-static expression to a const value.

## Actual behavior

`microsoft.codeanalysis.languageserver.exe` crashes with a stack overflow
![image](https://github.com/user-attachments/assets/c78fb22b-3c1e-46f4-a927-7ffb32f5bd79)

This causes the LSP to implode. the compiler exhibits similar behavior 
![image](https://github.com/user-attachments/assets/e557de15-f338-4bce-9eb6-29db7582d87a)

## Additional context

Bug is seemingly context-independent as it occurs even if i paste the snippet into a completely blank project. Doesn't show up for shorter switches.

PR #75366, opened 2024-10-03
Avoid cycles in constant initializer

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75353.

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
@@ -574,7 +574,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
 
             internal override ConstantValue GetConstantValue(SyntaxNode node, LocalSymbol inProgress, BindingDiagnosticBag diagnostics = null)
             {
-                if (this.IsConst && inProgress == this)
+                if (this.IsConst && (inProgress == this || this.ForbiddenZone?.Contains(node) == true))
                 {
                     if (diagnostics != null)
                     {
--- a/src/Compilers/CSharp/Test/Semantic/Semantics/ConstantTests.cs
+++ b/src/Compilers/CSharp/Test/Semantic/Semantics/ConstantTests.cs
@@ -3246,6 +3246,86 @@ enum E{0}
                 Diagnostic(ErrorCode.ERR_CircConstValue, "A").WithArguments("E8.A"));
         }
 
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75353")]
+        public void ConstCycle_SwitchExpression()
+        {
+            var source = """
+                const int x =
+                    x switch
+                    {
+                	    _ => 0,
+                    };
+                """;
+            CreateCompilation(source).VerifyDiagnostics(
+                // (2,5): error CS0110: The evaluation of the constant value for 'x' involves a circular definition
+                //     x switch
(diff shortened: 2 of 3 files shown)


src/Compilers/CSharp/Test/Semantic/Semantics/ConstantTests.cs
```diff
@@ -3246,6 +3246,86 @@ enum E{0}
...
+                // 	var y when y == x => 0,
+                Diagnostic(ErrorCode.ERR_CircConstValue, "x").WithArguments("x").WithLocation(3, 18));
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75353")]
+        public void ConstCycle_Checked()
+        {
+            var source = """
+                const int x =
+                    checked(x);
```
jjonescz, line 3321: Note that the issue is not just with switch expressions. I added this test to demonstrate that. The stack overflow can happen anywhere we get a binder using `this.GetBinder(node)` because then we lose the current `LocalInProgressBinder` which the cycle detection was relying on.

src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
```diff
@@ -574,7 +574,7 @@ private void MakeConstantTuple(LocalSymbol inProgress, BoundExpression boundInit
 
             internal override ConstantValue GetConstantValue(SyntaxNode node, LocalSymbol inProgress, BindingDiagnosticBag diagnostics = null)
             {
-                if (this.IsConst && inProgress == this)
+                if (this.IsConst && (inProgress == this || this.ForbiddenZone?.Contains(node) == true))
```
jcouv, line 577: This will catch a direct reference to `x` in the initializer for `const x`. Is there also a problem with an indirect reference: `const int x = y; const int y = ...x...;`?

jcouv (commented):
Done with review pass (iteration 2)

src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
jjonescz: Could not reproduce that failing even today. Added tests.

src/Compilers/CSharp/Test/Semantic/Semantics/ConstantTests.cs
```diff
@@ -3246,6 +3246,156 @@ enum E{0}
...
+    {
+        _ => 0,
+    }").WithArguments("x").WithLocation(2, 5));
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75353")]
+        public void ConstCycle_SwitchExpression_MutuallyReferencing()
+        {
+            var source = """
+                const int y = x switch { };
```
jcouv, line 3275: >{ } [](http://example.com/codeflow?start=39&length=3)

nit: consider adding `_ => 0` to simplify diagnostics

jcouv (approved):
LGTM Thanks (iteration 3)

src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
AlekseyTs, line 577: > || this.ForbiddenZone?.Contains(node) == true) [](http://example.com/codeflow?start=55&length=47)

While, perhaps, this change prevents the stack overflow for the specific scenario, I think we should be addressing the root cause of the problem, i.e. the binder hierarchy used to perform binding. 
I am fine if we proceed with this change as a quick, low risk,  temporary fix. But we should follow-up on the issue around binder hierarchy.

PR #75371, opened 2024-10-03
Fix circularity issue with binding const local

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/75353

Opened issue https://github.com/dotnet/roslyn/issues/75416 (wrong binder used in `switch` expression binding)
Opened issue https://github.com/dotnet/roslyn/issues/75437 (injected binders may get dropped by `GetBinder`)

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/LocalBinderFactory.cs
+++ b/src/Compilers/CSharp/Portable/Binder/LocalBinderFactory.cs
@@ -949,13 +949,14 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             if (node.Initializer is { } initializer)
             {
-                var oldEnclosing = _enclosing;
-                var localInProgressBinder = new LocalInProgressBinder(initializer, _enclosing);
-                AddToMap(initializer, localInProgressBinder);
+                var enclosing = _enclosing;
+                if (node.Parent is VariableDeclarationSyntax { Parent: LocalDeclarationStatementSyntax { IsConst: true } })
+                {
+                    enclosing = new LocalInProgressBinder(initializer, _enclosing);
+                    AddToMap(initializer, enclosing);
+                }
 
-                _enclosing = localInProgressBinder;
-                Visit(initializer.Value);
-                _enclosing = oldEnclosing;
+                Visit(initializer.Value, enclosing);
             }
         }
 
--- a/src/Compilers/CSharp/Portable/Binder/LocalBinderFactory.cs
+++ b/src/Compilers/CSharp/Portable/Binder/LocalBinderFactory.cs
@@ -946,7 +946,17 @@ namespace Microsoft.CodeAnalysis.CSharp
         public override void VisitVariableDeclarator(VariableDeclaratorSyntax node)
         {
             Visit(node.ArgumentList);
-            Visit(node.Initializer?.Value);
+
+            if (node.Initializer is { } initializer)
+            {
+                var oldEnclosing = _enclosing;
+                var localInProgressBinder = new LocalInProgressBinder(initializer, _enclosing);
+                AddToMap(initializer, localInProgressBinder);
+
+                _enclosing = localInProgressBinder;
+                Visit(initializer.Value);
+                _enclosing = oldEnclosing;
+            }
         }
 
         public override void VisitReturnStatement(ReturnStatementSyntax node)
--- a/src/Compilers/CSharp/Portable/Binder/LocalInProgressBinder.cs
+++ b/src/Compilers/CSharp/Portable/Binder/LocalInProgressBinder.cs
@@ -2,11 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
+using System;
+using System.Diagnostics;
+using System.Threading;
 using Microsoft.CodeAnalysis.CSharp.Symbols;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using Microsoft.CodeAnalysis.Text;
+using Roslyn.Utilities;
 
 namespace Microsoft.CodeAnalysis.CSharp
 {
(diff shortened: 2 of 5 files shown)


src/Compilers/CSharp/Portable/Binder/LocalInProgressBinder.cs
```diff
@@ -17,20 +18,26 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+            InitializerSyntax = initializerSyntax;
         }
 
-        internal override LocalSymbol LocalInProgress
+        internal override LocalSymbol? LocalInProgress
         {
             get
             {
-                return _inProgress;
+                return _localSymbol;
```
RikkiGibson, line 34: Is it sensible to access this while `_localSymbol` is still `null`? Or does that mean that we forgot to call `SetLocalSymbol` somewhere?

RikkiGibson (commented):
Just one minor question, otherwise looks good

src/Compilers/CSharp/Test/Semantic/Semantics/ConstantTests.cs
333fred, line 1: Consider adding a test for a collection expression like `[x]`.

src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
```diff
@@ -563,13 +563,27 @@ private void MakeConstantTuple(LocalSymbol inProgress, BoundExpression boundInit
                     var type = this.Type;
                     if (boundInitValue == null)
                     {
-                        var inProgressBinder = new LocalInProgressBinder(this, this._initializerBinder);
-                        boundInitValue = inProgressBinder.BindVariableOrAutoPropInitializerValue(_initializer, this.RefKind, type, diagnostics);
+                        recordLocalInBinderChain(this, _initializer, this._initializerBinder);
```
AlekseyTs, line 566: >recordLocalInBinderChain(this, _initializer, this._initializerBinder) [](http://example.com/codeflow?start=24&length=69)

It feels like this can be done in ```LocalWithInitializer```'s constructor

src/Compilers/CSharp/Portable/Binder/LocalInProgressBinder.cs
AlekseyTs, line 34: >_localSymbol [](http://example.com/codeflow?start=23&length=12)

We probably should assert that the value is not null here.

src/Compilers/CSharp/Portable/Binder/LocalBinderFactory.cs
```diff
@@ -946,7 +946,17 @@ public override void VisitLocalDeclarationStatement(LocalDeclarationStatementSyn
...
-            Visit(node.Initializer?.Value);
+
+            if (node.Initializer is { } initializer)
+            {
+                var oldEnclosing = _enclosing;
+                var localInProgressBinder = new LocalInProgressBinder(initializer, _enclosing);
+                AddToMap(initializer, localInProgressBinder);
+
+                _enclosing = localInProgressBinder;
+                Visit(initializer.Value);
```
AlekseyTs, line 957: >Visit [](http://example.com/codeflow?start=16&length=5)

We probably have an overload that takes an enclosing binder as parameter and deals with saving/setting/restoring the context.

src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
```diff
@@ -536,7 +536,7 @@ public LocalWithInitializer(
                 Debug.Assert(initializer != null);
 
                 _initializer = initializer;
-                _initializerBinder = initializerBinder;
+                _initializerBinder = initializerBinder.GetBinder(initializer);
```
AlekseyTs, line 539: >initializerBinder.GetBinder(initializer) [](http://example.com/codeflow?start=37&length=40)

It looks like this operation can return null when  ```LocateDeclaredVariableSymbol``` is doing an error recovery. In that case we might want to inject the binder here explicitly.

AlekseyTs:
Consider opening an issue to check for other binders that are created on the fly, which could be "lost" in similar scenarios.

AlekseyTs:

src/Compilers/CSharp/Portable/Binder/LocalBinderFactory.cs
```diff
@@ -946,7 +946,17 @@ public override void VisitLocalDeclarationStatement(LocalDeclarationStatementSyn
         public override void VisitVariableDeclarator(VariableDeclaratorSyntax node)
         {
             Visit(node.ArgumentList);
-            Visit(node.Initializer?.Value);
+
+            if (node.Initializer is { } initializer)
+            {
+                var oldEnclosing = _enclosing;
+                var localInProgressBinder = new LocalInProgressBinder(initializer, _enclosing);
```
AlekseyTs, line 953: >new LocalInProgressBinder(initializer, _enclosing) [](http://example.com/codeflow?start=44&length=50)

Would it make sense to limit creation of the extra binder to `const` declarations?

src/Compilers/CSharp/Portable/Binder/LocalInProgressBinder.cs
jcouv: It should no longer be `null` by the time and if someone asks. Added an assertion

src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
```diff
@@ -536,7 +536,25 @@ public LocalWithInitializer(
                 Debug.Assert(initializer != null);
 
                 _initializer = initializer;
-                _initializerBinder = initializerBinder;
+                _initializerBinder = initializerBinder.GetBinder(initializer) ?? new LocalInProgressBinder(_initializer, initializerBinder);
```
AlekseyTs, line 539: >new LocalInProgressBinder [](http://example.com/codeflow?start=81&length=25)

We probably don't want to inject this binder for non-constants.

src/Compilers/CSharp/Portable/Symbols/Source/SourceLocalSymbol.cs
```diff
@@ -536,7 +536,26 @@ public LocalWithInitializer(
                 Debug.Assert(initializer != null);
 
                 _initializer = initializer;
-                _initializerBinder = initializerBinder;
+                _initializerBinder = initializerBinder.GetBinder(initializer);
```
AlekseyTs, line 539: >initializerBinder.GetBinder(initializer) [](http://example.com/codeflow?start=37&length=40)

It is expected to get null from this call for non-constant declarations. Most likely this is the reason why CI is failing.

AlekseyTs:

AlekseyTs (approved):
LGTM (commit 5)

333fred (approved):

PR #75372, opened 2024-10-03
Fixing the OOM in scripting

jaredpar:
Dug into the Scripting OOM tests more and figured out what was going on. The scripting tests predominantly create `PortableExecutableReference` instances using `Stream` as the source. Under the hood this creates a lot of native allocations in the form of memory mapped files and `AllocHGlobal` memory. As `PortableExecutableReference` is not disposable, these allocations are cleaned up passively in finalizers.

This means that our tests rely on the finalizer running fast enough that it frees up enough native memory for the next scripting test that allocates native memory to succeed. When the scripting tests are run in isolation or in an x64 process this works out just fine. But when we run the tests on x86 and in combination with other test assemblies this can break down and lead to OOM situations.

This fix approaches the problem from three directions:

1. Scripting had an existing mechanism for intercepting files being converted to `PortableExecutableReference`. This change expands on that to make it more thoroughly plumbed through the scripting code base and to have a single place, `ScriptOptions`, where the intercept can be hooked up.
2. Creating a `ScriptTestBase` type which sets up the intercepting of file loads and then actively disposing the resulting `MetadataImageReference` on `Dispose`. This means the native memory is actively freed at the end of each test. It also provides a `ScriptOptions` value that derived tests can leverage.
3. Plumbing through `ScriptOptions` to the individual tests. 

After doing this I ran the script tests under PerfView in a number of configurations and verified that the memory was being actively freed and the heap size of the process was reduced as expected.

### DIFF
--- a/src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
+++ b/src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
@@ -11,6 +11,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Reflection.PortableExecutable;
 using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
@@ -210,7 +211,7 @@ namespace Microsoft.CodeAnalysis
             return (path, properties) =>
             {
                 var peStream = FileSystem.OpenFileWithNormalizedException(path, FileMode.Open, FileAccess.Read, FileShare.Read);
-                return MetadataReference.CreateFromFile(peStream, path, properties, documentation: null);
+                return MetadataReference.CreateFromFile(peStream, path, PEStreamOptions.PrefetchEntireImage, properties, documentation: null);
             };
         }
 
--- a/src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
+++ b/src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
@@ -210,7 +210,7 @@ namespace Microsoft.CodeAnalysis
             return (path, properties) =>
             {
                 var peStream = FileSystem.OpenFileWithNormalizedException(path, FileMode.Open, FileAccess.Read, FileShare.Read);
-                return MetadataReference.CreateFromFile(peStream, path, properties);
+                return MetadataReference.CreateFromFile(peStream, path, properties, documentation: null);
             };
         }
 
--- a/src/Compilers/Core/Portable/MetadataReference/MetadataReference.cs
+++ b/src/Compilers/Core/Portable/MetadataReference/MetadataReference.cs
@@ -252,19 +252,8 @@ namespace Microsoft.CodeAnalysis
         internal static MetadataImageReference CreateFromFile(
             string path,
             PEStreamOptions options,
-            MetadataReferenceProperties properties) =>
-            CreateFromFile(
-                StandardFileSystem.Instance.OpenFileWithNormalizedException(path, FileMode.Open, FileAccess.Read, FileShare.Read),
-                path,
-                options,
-                properties,
-                documentation: null);
-
-        internal static MetadataImageReference CreateFromFile(
-            string path,
-            PEStreamOptions options,
             MetadataReferenceProperties properties,
-            DocumentationProvider? documentation) =>
+            DocumentationProvider? documentation = null) =>
             CreateFromFile(
                 StandardFileSystem.Instance.OpenFileWithNormalizedException(path, FileMode.Open, FileAccess.Read, FileShare.Read),
                 path,
@@ -275,20 +264,8 @@ namespace Microsoft.CodeAnalysis
         internal static MetadataImageReference CreateFromFile(
             Stream peStream,
             string path,
-            MetadataReferenceProperties properties,
-            DocumentationProvider? documentation) =>
-            CreateFromFile(
-                peStream,
-                path,
-                PEStreamOptions.PrefetchEntireImage,
-                properties,
-                documentation);
-
-        internal static MetadataImageReference CreateFromFile(
-            Stream peStream,
-            string path,
             PEStreamOptions options,
-            MetadataReferenceProperties properties = default,
+            MetadataReferenceProperties properties,
             DocumentationProvider? documentation = null)
         {
             // prefetch image, close stream to avoid locking it:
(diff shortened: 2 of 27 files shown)


jaredpar:
@dotnet/roslyn-compiler PTAL

src/Compilers/Core/Portable/MetadataReference/MetadataReference.cs
```diff
@@ -318,14 +332,10 @@ public static MetadataReference CreateFromAssembly(
...
         }
 
-        internal static PortableExecutableReference CreateFromAssemblyInternal(
+        internal static string GetAssemblyFilePath(
             Assembly assembly,
-            MetadataReferenceProperties properties,
-            DocumentationProvider? documentation = null)
+            MetadataReferenceProperties properties)
         {
-            // Note: returns MetadataReference and not PortableExecutableReference so that we can in future support assemblies that
```
333fred, line 326: Should this comment be moved?

333fred (commented):
Can we use the `BannedAPIAnalyzer` to ban the non-options-taking versions of these methods in our test assemblies so we don't end up here again in 5 years?

jaredpar:
> Can we use the BannedAPIAnalyzer to ban the non-options-taking versions of these methods in our test assemblies so we don't end up here again in 5 years?

I don't think we can. I definitely considered this but the problem is that every overload has an options parameter, it's just that in some cases it's a default value. Don't think the banned API analyzer can deal with that.

333fred:
Gotcha. This doesn't feel worth an analyzer to me, but we should document this somewhere; maybe a README in the scripting test project? Or a note in the base class?

src/Compilers/Core/Portable/MetadataReference/MetadataReference.cs
jaredpar: No. This comment was originally on `CreateFromAssemblyInternal` and I changed that return type to be `MetadataImageReference`. The comment was incorrect before this change anyways because the return was `MetadataReference` not `PortableExecutableReferenceb

jaredpar:
@dotnet/roslyn-compiler PTAL
@333fred doc added.

333fred (approved):

src/Scripting/CoreTestUtilities/README.md
```diff
@@ -0,0 +1,15 @@
+# Scripting Unit Tests
+
+## Native Memory Pressure
+
+The scripting engine in Roslyn makes heavy use of creating `MetadatImageReference` over a `Stream`. That is the way in which runtime assembly references are added. This form of `MetadataReference` results in a lot of native allocations as opposed to using an `ImmutableArray<byte>` which has none. By default these native allocations are collected by the finalizer as `MetadadataImageReference` is not disposable.
```
jjonescz, line 5:
```suggestion
The scripting engine in Roslyn makes heavy use of creating `MetadataImageReference` over a `Stream`. That is the way in which runtime assembly references are added. This form of `MetadataReference` results in a lot of native allocations as opposed to using an `ImmutableArray<byte>` which has none. By default these native allocations are collected by the finalizer as `MetadadataImageReference` is not disposable.
```

src/Scripting/CoreTestUtilities/README.md
```diff
@@ -0,0 +1,15 @@
...
+
```
jjonescz, line 6: Have we considered fixing this by using `ImmutableArray<byte>` instead of relying on `MetadataImageReference` over `Stream`?

src/Scripting/Core/Hosting/CommandLine/CommandLineRunner.cs
```diff
@@ -167,21 +174,29 @@ private static ScriptOptions GetScriptOptions(CommandLineArguments arguments, st
                 allowUnsafe: true,
                 checkOverflow: false,
                 warningLevel: 4,
-                parseOptions: arguments.ParseOptions);
+                parseOptions: arguments.ParseOptions,
+                createFromFileFunc: _createFromFileFunc);
         }
 
-        internal static MetadataReferenceResolver GetMetadataReferenceResolver(CommandLineArguments arguments, TouchedFileLogger loggerOpt)
+#nullable enable
```
jjonescz, line 181: This looks unnecessary since we removed `#nullable disable` at the beginning of the file.

jjonescz (approved):

src/Compilers/Core/Portable/CommandLine/CommonCompiler.cs
```diff
@@ -210,7 +211,7 @@ internal virtual Func<string, MetadataReferenceProperties, PortableExecutableRef
             return (path, properties) =>
             {
                 var peStream = FileSystem.OpenFileWithNormalizedException(path, FileMode.Open, FileAccess.Read, FileShare.Read);
-                return MetadataReference.CreateFromFile(peStream, path, properties);
+                return MetadataReference.CreateFromFile(peStream, path, PEStreamOptions.PrefetchEntireImage, properties, documentation: null);
```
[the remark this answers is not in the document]
jaredpar: On this particular overload there is no default parameter value so I had to pass one here.

src/Scripting/CoreTestUtilities/README.md
jaredpar: I _believe_ the reason scripting uses `Stream` is that in the normal use case the belief is that it should result in less overall allocations. The compiler lazily loads data from the PE file so in theory you could load a lot of assemblies with only targeted allocations. As opposed to the `ImmutableArray<byte>` case where you have to eagerly load them all into memory. 

The unit tests are just taking the downsides of this approach to the extreme.

jaredpar:
@dotnet/roslyn-ide can I get a review on the scripting side? This will resolve the OOM issues we've been seeing in CI

src/Scripting/Core/Script.cs
```diff
@@ -308,6 +309,32 @@ internal bool HasReturnValue()
...
+            return createFromFileFunc(filePath, PEStreamOptions.Default, properties);
+        }
+
+        /// <summary>
+        /// <see cref="MetadataReference.CreateFromFile(string, PEStreamOptions, MetadataReferenceProperties, DocumentationProvider)"/>
+        /// </summary>
+        /// <remarks>
+        /// This API exists as the default for reading <see cref="MetadataReference"/> from files. It is handy 
+        /// to have this separate method when trying to track down unit tests that aren't going through the 
+        /// proper heplers in ScriptTestBase to hook loading from file. Can set a breakpoint here, debug the
```
dibarbet, line 330:
```suggestion
        /// proper helpers in ScriptTestBase to hook loading from file. Can set a breakpoint here, debug the 
```

dibarbet (approved):
seems reasonable

PR #75404, opened 2024-10-05
Reduce allocations from creating MissingTokenWithTrivia objects

ToddGrun:
This method accounts for 0.7% of allocations in the roslyn csharp editing speedometer test profile I'm looking at. In addition to caching tokens < LastTokenWithWellKnownText similar to what is already done, the SyntaxKind.IdentifierToken case is handled as it's very commonly hit too. Local testing of opening LanguageParser and typing in a new method reduced the number of MissingTokenWithTrivia from over 1 million to zero.

*** allocations in speedometer profile ***
image


jjonescz (approved):

AlekseyTs (approved):
LGTM (commit 2)

PR #75414, opened 2024-10-07
Add more first-class Span tests

jjonescz:

### DIFF
--- a/src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
+++ b/src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
@@ -5,6 +5,7 @@
 using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis.CSharp.Symbols;
+using Microsoft.CodeAnalysis.CSharp.Symbols.Retargeting;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.CSharp.Test.Utilities;
 using Microsoft.CodeAnalysis.Operations;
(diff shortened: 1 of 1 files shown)


jjonescz:
@cston @333fred for test-only review, thanks

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -7759,6 +7861,178 @@ static class C
...
+    {
+        var source = """
+            using System;
+
+            C.M(new string[0]);
+            C.M(default(Span<string>));
+
+            class C
+            {
+                public static implicit operator C(string s) => new C();
```
cston, line 8017: >public static implicit operator C(string s) => new C(); [](http://example.com/codeflow?start=16&length=55)

Does this overload have any effect? (Or should the argument be `string[]` instead?)

cston (approved):

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -2418,6 +2480,46 @@ .maxstack  1
...
+            .ToMetadataReference();
+
+        var source2 = """
+            using System;
+            class C
+            {
+                ReadOnlySpan<int> M(Span<int> s) => s;
+            }
+            """;
+        var comp2 = CreateCompilation(source2, [comp1Ref]);
```
333fred, line 2505: When we're doing a retargeting test, it's usually good practice to explicitly specify the underlying TFM for the second compilation, so we can be sure that it's a different TFM from the assembly that will be retargeted.

333fred (approved):

PR #75423, opened 2024-10-07
EE: Handle notification when metadata has been invalidated

cston:
Implement notification interface in the EE to ensure assembly metadata is discarded before being invalidated.


src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmExceptionCode.cs
```diff
@@ -20,1179 +20,8 @@ public enum DkmExceptionCode
         E_FAIL = -2147467259,
         //
         // Summary:
-        //     A debugger is already attached.
-        E_ATTACH_DEBUGGER_ALREADY_ATTACHED = -2147221503,
```
cston, line 24: Only a few of these values were used, and many were missing. Reduced this to just those values that are used.

src/ExpressionEvaluator/Core/Source/ExpressionCompiler/ExpressionCompiler.cs
```diff
@@ -23,6 +22,7 @@ namespace Microsoft.CodeAnalysis.ExpressionEvaluator
     public abstract class ExpressionCompiler :
         IDkmClrExpressionCompiler,
         IDkmClrExpressionCompilerCallback,
+        IDkmMetaDataPointerInvalidatedNotification,
         IDkmModuleModifiedNotification,
```
gregg-miskelly, line 26: >IDkmModuleModifiedNotification [](http://example.com/codeflow?start=8&length=30)

You shouldn't need to implement this anymore (unless this is just a temporary thing to keep tests happy until you are running against a new debugger)

eng/Directory.Packages.props
```diff
@@ -122,8 +122,8 @@
       VS Debugger
     -->
     <PackageVersion Include="Microsoft.VisualStudio.Debugger.Contracts" Version="17.12.0-beta.24403.1" />
-    <PackageVersion Include="Microsoft.VisualStudio.Debugger.Engine-implementation" Version="17.8.1072001-preview" />
-    <PackageVersion Include="Microsoft.VisualStudio.Debugger.Metadata-implementation" Version="17.8.1072001-preview" />
+    <PackageVersion Include="Microsoft.VisualStudio.Debugger.Engine-implementation" Version="17.13.1100701-preview" />
+    <PackageVersion Include="Microsoft.VisualStudio.Debugger.Metadata-implementation" Version="17.13.1100701-preview" />
```
cston, line 126: Do we need to reference a newer version of Microsoft.VSSDK.Debugger.VSDConfigTool (at line 67) as well?

src/ExpressionEvaluator/Core/Source/ExpressionCompiler/ExpressionCompiler.cs
cston: I think I'd prefer keeping the implementation of `IDkmModuleModifiedNotification`, for robustness, assuming it doesn't conflict with `IDkmMetaDataPointerInvalidatedNotification`.

gregg-miskelly: They will not conflict - just a bit of a perf overhead since now Rosyln will be notified twice anytime, we modify an assembly.

cston:
@tmat, @dotnet/roslyn-compiler, please review.

cston:
The failures in the integration tests (`"Internal error in the C# compiler"`) are likely due to the EE depending on `Microsoft.VisualStudio.Debugger.*.dll` from the latest version of Visual Studio.

gregg-miskelly:
> The failures in the integration tests (`"Internal error in the C# compiler"`) are likely due to the EE depending on `Microsoft.VisualStudio.Debugger.*.dll` from the latest version of Visual Studio which is not available in CI.

FYI Debugger changes are in VS Main 35408.49

cston:
Locally, I was able to run the following integration tests successfully with Visual Studio 35408.184, with a `DEBUG` build of the Roslyn EE:
- `BasicExpressionEvaluator.*`
- `BasicEditAndContinue+CommonProjectSystem.*`
- `{Basic,CSharp}Immediate.DumpLocalVariableValue`

And with `BasicEditAndContinue+CommonProjectSystem.EditLambdaExpression`, I verified the EE implementation of `IDkmMetaDataPointerInvalidatedNotification.OnMetaDataPointerInvalidated()` was called and handled.

jjonescz (approved):

cston:
@dotnet/roslyn-compiler, @tmat, for a second review please.

jaredpar (approved):

Issue #74013, opened 2024-06-14
Long-running IAsyncEnumerable<T> leaks the last seen enumerated value.


hexawyz:
**Version Used**: .NET 8.0 / C# 12

**Steps to Reproduce**:

When using async iterators, the "live" last enumerated value will be kept rooted within the `IAsyncEnumerator`, preventing it from being collected, even though the user code does not explicitly reference it anymore.
While at the surface, it may seem like an innocuous problem, this can have severe implications when the object graph behind the enumerated values is quite huge (e.g. collectible assemblies) and the run time of the iterator is very long (think process lifetime).

Also, I believe it can be adressed with a reasonable change. (See the details later)

Run the following C# program:

````csharp
using System.Runtime.CompilerServices;

using (var cts = new CancellationTokenSource())
{
	ConsoleCancelEventHandler consoleCancelKeyPressEventHandler = (s, e) =>
	{
		if (e.SpecialKey == ConsoleSpecialKey.ControlC)
		{
			cts.Cancel();
		}
	};
	Console.CancelKeyPress += consoleCancelKeyPressEventHandler;
	try
	{
		var collectTask = CollectAfterTenSecondsAsync(cts.Token);
		try
		{
			await foreach (var it in WatchThemAsync(cts.Token))
			{
				Console.WriteLine($"Index #{it.Index}");
				GC.Collect(2, GCCollectionMode.Forced, true);
			}
		}
		finally
		{
			await collectTask;
		}
	}
	finally
	{
		Console.CancelKeyPress -= consoleCancelKeyPressEventHandler;
	}
}

async Task CollectAfterTenSecondsAsync(CancellationToken cancellationToken)
{
	await Task.Delay(10_000, cancellationToken).ConfigureAwait(false);
	GC.Collect(2, GCCollectionMode.Forced, true);
}

async IAsyncEnumerable<WatchMe> WatchThemAsync([EnumeratorCancellation] CancellationToken cancellationToken)
{
	await Task.Yield();
	yield return new(0);
	await Task.Delay(100, cancellationToken).ConfigureAwait(false);
	yield return new(1);
	await Task.Delay(100, cancellationToken).ConfigureAwait(false);
	yield return new(2);
	await Task.Delay(100, cancellationToken).ConfigureAwait(false);
	yield return new(3);
	await Task.Delay(Timeout.Infinite, cancellationToken).ConfigureAwait(false);
	yield return new(4);
}

internal sealed class WatchMe(int index)
{
[...]

PR #75426, opened 2024-10-07
Clear current value when awaiting in async-iterator

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/74013

The problem is that in a `yield return someReference; await LongWaitAsync();` the iterator holds on unnecessarily to the reference for a long time.
Some options discussed in the issue:
1. do nothing
2. clear `Current` when resuming from a yield
3. clear `Current` when about to suspend for an `await` (implemented by this PR)

Note: we don't need to clear the `Current` value when the type is unmanaged.
Note: there is a potential optimization where skip clearing the value for an `await` that definitely follows after another `await`. This is not included in this PR.


jcouv:
@333fred @dotnet/roslyn-compiler for review. Thanks

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
```diff
@@ -8666,5 +8666,331 @@ static async System.Threading.Tasks.Task Main()
...
+    {
+        yield return 42;
+        await t;
+        yield return 43;
+    }
+}
+""";
+            var comp = CreateCompilation(src, targetFramework: TargetFramework.Net80);
+            var verifier = CompileAndVerify(comp,
+                expectedOutput: ExecutionConditionUtil.IsMonoOrCoreClr ? "424243" : null,
```
AlekseyTs, line 8947: >424243 [](http://example.com/codeflow?start=74&length=6)

It feels to me that it would be better to invalidate the value of Current, or, perhaps, even throw InvalidOperationException from that property during the period of time starting with MoveNextAsync invocation and until the return task completes. And do that regardless of whether the element type is managed or not.

src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncIteratorMethodToStateMachineRewriter.cs
```diff
@@ -317,6 +317,17 @@ public override BoundNode VisitYieldBreakStatement(BoundYieldBreakStatement node
                 F.Goto(_currentDisposalLabel));
         }
 
+        protected override BoundStatement MakeAwaitPreamble()
+        {
+            if (_asyncIteratorInfo.CurrentField.Type.IsManagedTypeNoUseSiteDiagnostics)
```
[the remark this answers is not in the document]
AlekseyTs: Consider adding a test for a scenario like that.

AlekseyTs:

src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncMethodToStateMachineRewriter.cs
```diff
@@ -333,8 +333,14 @@ public sealed override BoundNode VisitBadExpression(BoundBadExpression node)
             return node;
         }
 
+#nullable enable
+        protected virtual BoundStatement? MakeAwaitPreamble() { return null; }
+#nullable disable
+
         private BoundBlock VisitAwaitExpression(BoundAwaitExpression node, BoundExpression resultPlace)
         {
+            BoundStatement preamble = MakeAwaitPreamble();
```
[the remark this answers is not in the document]
jcouv: That would indeed be possible and reasonable. But as I mentioned in OP I'm not planning such optimization at this point.

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
jcouv: That feels like addressing a different problem. The problem we're trying to solve here is not callers accessing the value of `Current` after the state machine has restarted, but simply that we're preventing GC of referenced values.

@stephentoub What do you think of clearing the `Current` value regardless of whether the element type is managed or not?

AlekseyTs: > but simply that we're preventing GC of referenced values

Thank you for making the primary motivation clear. However with the current approach we end up with an observable inconsistent behavior. Users are likely to build up a "muscle memory" with unmanaged types thinking that values are available, and then run into a problem with managed types while doing the same thing.

stephentoub: > What do you think of clearing the Current value regardless of whether the element type is managed or not?

I'm not particularly concerned with "consistency" here. The difference here is only observable when accessing Current at a time when you're not supposed to access Current, such that the behavior is undefined. We have a multitude of places where we have similar discrepancies, where we check RuntimeHelpers.IsReferenceOrContainsReference and clear based on it, e.g. only clearing arrays returned to the array pool if they contain a reference, or only zero'ing the data during `List<T>.Clear` if it contains a reference. Those are also similarly observable but only in cases where you're relying on undefined behavior / unsafe APIs, e.g. subsequently using CollectionsMarshal.SetCount to increase back the list's count, at which point it'll still contain the data from before the Clear but only if the data didn't contain references. And even if this does just always clear Current, there are still inconsistencies. For example, if the clearing happens as part of an await, and in particular as part of a suspension, then given the code:
```C#
yield return 42;
await t;
```
whether e.Current is 42 after the enumerator completes will be based on a race condition of whether `t` was completed by the time its IsCompleted was checked as part of the await.

Thus, I'd have no qualms if the zero'ing was only done for managed Ts. That said, in all but the most extreme cases, I also doubt there will be a measurable impact to the zero'ing, given that in the most common cases you're dealing with either a single reference or a small struct, such that zero'ing is just one or a few instructions, and it's only at a time when we're about to pay non-trivial costs as part of suspension.

jcouv:
@dotnet/roslyn-compiler for a second review and more thought on open question of whether to conditionally clear only for managed types. Thanks

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
jcouv: Jared pointed out that we have some precedent for clearing variables to avoid unnecessary reference retention: locals in nested blocks in async/iterator state machines.
See `MethodToStateMachineRewriter.AddVariableCleanup`.
So I think it's okay to keep the logic conditional for clearing `Current` as well.

Note: the logic `MightContainReferences` could be improved, but I'm planning to address that in a follow-up PR.

[sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQAYACdbAFgG4CD0BmMzEgYRIIG8CSezHz6AHlgAXAHwkAsgApgAezkAbEsACU7fLxJdNW3gHcAFlEUBTErNXc9PHTZvlSAZxIBeEgCIAbhEUBXUw8aXXtecmwyAHYSCkxg0K0WORgnJVMAOgB1ACcoEVNpJ1V4hJ4AejKSfXMAYzMIbJIAAycmkigwAAdFKBq8xQBPaxsAX2G9JJS0rNz86Q9sD2LxsfwRoA)
```
using System;

public class C 
{
    public IEnumerable<int> M(bool b) 
    {
        while (b)
        {
            string s = "value";
            yield return 42;
            Console.Write(s);
            // we clear `s` implicitly
        }
        Console.Write("1");
    }
}
```
[sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQAYACdbAVgG4CD0BmMzEgYRIIG8CSezH0AHGQBsAHlgAXAHwkAsgApgAeyUAbEsACU7fLxJdde3gHcAFlFUBTEos3cjPAw4flSAZxIBeEgCIAbhCqAK6WPjSGzrzoAJwiAHQAmlCWqgjymuGReixKMG5qlnEA6gBOUBKW8m4Z9lkkAPT1JMbWAMZWECUkAAZu3SRQYAAOqlCt5aoAnrVGAL4z2bn5VsVlFfI+2D41EUboAOwkACyYmbzz+LNAA==)
```
using System;
using System.Threading.Tasks;

public class C 
{
    public async Task<int> M(bool b) 
    {
        while (b)
        {
            string s = "value";
            await Task.Yield();
            Console.Write(s);
            // we clear `s` implicitly
        }
        Console.Write("1");
        return 42;
    }
}
```

AlekseyTs (approved):
LGTM (commit 2)

jcouv:

333fred (approved):

PR #75427, opened 2024-10-08
Do not use overload resolution for expression tree conversion

AlekseyTs:
Related to #74275.


src/Compilers/Core/Portable/WellKnownTypes.cs
```diff
@@ -337,6 +337,23 @@ internal enum WellKnownType
 
         System_Runtime_CompilerServices_ParamCollectionAttribute,
 
+        System_Linq_Expressions_BinaryExpression,
```
jjonescz, line 340: Because we use well-known types now, there might be some behavior changes, right? For example, having ambiguous `BinaryExpression`s (so we won't resolve the well-known type for it), we now won't be able to bind `Expression.Binary` whereas previously we could. I'm not sure it matters though.

src/Compilers/CSharp/Portable/Lowering/ClosureConversion/ExpressionLambdaRewriter.cs
```diff
@@ -1054,58 +1140,71 @@ private BoundExpression VisitUnaryOperator(BoundUnaryOperator node)
...
         // ======================================================
 
-        private BoundExpression ExprFactory(string name, params BoundExpression[] arguments)
-        {
-            return _bound.StaticCall(ExpressionType, name, disallowExpandedNonArrayParams: true, arguments);
-        }
-
-        private BoundExpression ExprFactory(string name, ImmutableArray<TypeSymbol> typeArgs, params BoundExpression[] arguments)
-        {
-            return _bound.StaticCall(_ignoreAccessibility ? BinderFlags.IgnoreAccessibility : BinderFlags.None, ExpressionType, name, disallowExpandedNonArrayParams: true, typeArgs, arguments);
```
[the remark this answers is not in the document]
AlekseyTs: > It looks like we no longer use `_ignoreAccessibility`. Is that an unintended behavior change?

Definitions of well known types and members must be accessible. I am not sure why this field was used. I assume that, if there were specific scenarios in mind, there would be tests that would fail. 

> Also consider removing the unused field.

Will do.

src/Compilers/Core/Portable/WellKnownTypes.cs
AlekseyTs: > Because we use well-known types now, there might be some behavior changes, right? For example, having ambiguous `BinaryExpression`s (so we won't resolve the well-known type for it), we now won't be able to bind `Expression.Binary` whereas previously we could.

That is correct.

jcouv (approved):
LGTM Thanks (iteration 2)

src/Compilers/Core/Portable/WellKnownTypes.cs
jaredpar: This only happens when customers define a their own types in the `System` namespace though. That is indeed a break but it's generally one that we are okay with. Defining types in the `System` namespace, particularly when they clash with the ones in corelib is not recommended for this very reason

jjonescz (approved):

Issue #75430, opened 2024-10-08
Nullability mismatch in non-`ref` foreach cannot be suppressed


jjonescz:
Is it expected that now you can suppress nullability mismatch for `foreach` with `ref` variables but not for `foreach` with regular variables?

It looks like the following test is supposed to cover that scenario, but it's missing the actual nullability mismatch because the parameter of the test method is `object[]` instead of `object?[]`

![image](https://github.com/user-attachments/assets/298a927d-8919-4794-9e4a-2f95fc1f22b3)

https://github.com/dotnet/roslyn/commit/9941c1fb15e4c7d9dbdd2b61faac9d975f96627a#diff-14fe026e6093e9522c01e6c377b26554dd6b663214884198dcc244f370548fa1R90883

The following code right now does have a warning and it cannot be suppressed:
```
void M2(object?[] a) { foreach (object item in a!) { } }
```

_Originally posted by @TessenR in https://github.com/dotnet/roslyn/issues/73928#issuecomment-2397368397_

alrz:
This looks very much like https://github.com/dotnet/roslyn/issues/38586

PR #75432, opened 2024-10-08
Allow suppressing foreach iteration variable nullability mismatch

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75430.


src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -8703,7 +8704,7 @@ private TypeWithState VisitConversion(
             bool canConvertNestedNullability = true;
             bool isSuppressed = false;
 
-            if (conversionOperand.IsSuppressed == true)
+            if (suppressed || conversionOperand.IsSuppressed)
```
jcouv, line 8707: Could you share some details on the set up of the situation? I gather that we have two stacked conversions that are not part of the same conversion group and the operand is suppressed, but I don't understand why.
If this setup is expected, should we just dig through conversions here when looking for the `IsSuppressed` flag?

jcouv (approved):
LGTM Thanks (iteration 1)

jcouv:

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
RikkiGibson, line 8707: Would it be possible to unify the `suppressed` parameter with the `isSuppressed` local?

jcouv:
@RikkiGibson Please take another look. Thanks

jcouv (approved):

jcouv:
@RikkiGibson Please take another look. Thanks

RikkiGibson (approved):

Issue #56256, opened 2021-09-08
[MemberNotNull] isn't respected on local functions


stephentoub:
**Version Used**: 
9525fa5

**Steps to Reproduce**:
https://sharplab.io/#v2:EYLgtghglgdgPgYhgVwDaosVBTABNmTHAWACgABABl3IEYA6AESggHMYB7AZwBcoBjLvQDCHACbYAgoVQBPLlC4BuMmXIBmGgCZcwsgG8yuY7gAOAJygA3CDzwdgAK2z8eAflwB9B45WkTuEYmGriwPLgAsgAUAJRBxob+ASYAkjBQPLF+ycEA7F4+9ADi2DwAEhBcABaiElnxyQ0BANoR2GDA2OYAchw83WioUYRg2BwAZlHeTjExALpNwQAsuGkZsbgAvAB8BU5buDDYAO64Pi6ZMdkmAL5kN0A===
```C#
#nullable enable
using System.Diagnostics.CodeAnalysis;

public class C
{
    private object? _obj;
    
    public int M()
    {
        Init();
        return _obj.GetHashCode();
        
        [MemberNotNull(nameof(_obj))]
        void Init() => _obj = new object();
    }
}
````

**Expected Behavior**:
No warning.

**Actual Behavior**:
`warning CS8602: Dereference of a possibly null reference.`

stephentoub:
Ah, dup of https://github.com/dotnet/roslyn/issues/47896

stephentoub:
Apparently I opened the same issue last year ;-)
https://github.com/dotnet/roslyn/issues/49234

Youssef1313:
@stephentoub #47896 was closed as fixed, but this issue still repros. This issue wasn't probably an exact dupe of #47896 and should be reopened. Can you reopen? Thanks!

RikkiGibson:
LDM did decide that we should do this and that `MemberNotNull` applied to the function applies to members of the captured `this`.

jcouv:
Implemented in 17.13 :-)

PR #75448, opened 2024-10-09
Respect [MemberNotNull] on local functions

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/56256

The `MemberNotNull` attribute takes effect if either:
- the context is non-static (ie. the containing method and the nested local functions are not static). The target member can be static or not
- the context is static (ie. either the containing method or the local function is static) and the target member is static

When the `MemberNotNull` takes effect, three things happen:
1. after the local function is called, the state of the target member is changed to not-null
2. when entering the local function, the state of the target member is changed to maybe-null
3. when existing the local function, we warn if the target member isn't not-null

Note: I didn't gate this change behind LangVer as it seems small and only affects nullability warnings, but I'm open to it.

### DIFF
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -3258,9 +3258,8 @@ namespace Microsoft.CodeAnalysis.CSharp
                 var oldPending = SavePending();
 
                 EnterParameters();
-                bool isLocalFunction = lambdaOrFunctionSymbol is LocalFunctionSymbol;
-                SyntaxNode? localFunctionSyntax = isLocalFunction ? ((LocalFunctionSymbol)lambdaOrFunctionSymbol).Syntax : null;
 
+                bool isLocalFunction = lambdaOrFunctionSymbol is LocalFunctionSymbol;
                 if (isLocalFunction)
                 {
                     MakeMembersMaybeNull(lambdaOrFunctionSymbol, lambdaOrFunctionSymbol.NotNullMembers);
@@ -3281,7 +3280,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 EnforceDoesNotReturn(syntaxOpt: null);
                 if (isLocalFunction)
                 {
-                    enforceMemberNotNull(localFunctionSyntax, this.State);
+                    enforceMemberNotNull(((LocalFunctionSymbol)lambdaOrFunctionSymbol).Syntax, this.State);
                 }
                 EnforceParameterNotNullOnExit(null, this.State);
 
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -2092,7 +2092,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
         }
 
-        static MethodSymbol? GetTopLevelMethod(MethodSymbol? method)
+        private static MethodSymbol? GetTopLevelMethod(MethodSymbol? method)
         {
             while (method is object)
             {
@@ -7015,6 +7015,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             if (method.MethodKind == MethodKind.LocalFunction
                 && GetTopLevelMethod(method) is { ThisParameter: { } thisParameter })
             {
+                Debug.Assert(receiverSlot == -1);
                 receiverSlot = GetOrCreateSlot(thisParameter);
             }
 
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -2012,7 +2012,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                     case BoundKind.ThisReference:
                     case BoundKind.BaseReference:
                         {
-                            var method = getTopLevelMethod(_symbol as MethodSymbol);
+                            var method = GetTopLevelMethod(_symbol as MethodSymbol);
                             var thisParameter = method?.ThisParameter;
                             return thisParameter is object ? GetOrCreateSlot(thisParameter) : -1;
                         }
@@ -2090,20 +2090,20 @@ namespace Microsoft.CodeAnalysis.CSharp
                 }
                 return -1;
             }
+        }
 
-            static MethodSymbol? getTopLevelMethod(MethodSymbol? method)
+        static MethodSymbol? GetTopLevelMethod(MethodSymbol? method)
+        {
+            while (method is object)
             {
-                while (method is object)
+                var container = method.ContainingSymbol;
+                if (container.Kind == SymbolKind.NamedType)
                 {
-                    var container = method.ContainingSymbol;
-                    if (container.Kind == SymbolKind.NamedType)
-                    {
-                        return method;
-                    }
-                    method = container as MethodSymbol;
+                    return method;
                 }
-                return null;
+                method = container as MethodSymbol;
             }
+            return null;
         }
 
         protected override int GetOrCreateSlot(Symbol symbol, int containingSlot = 0, bool forceSlotEvenIfEmpty = false, bool createIfMissing = true)
@@ -7011,6 +7011,12 @@ namespace Microsoft.CodeAnalysis.CSharp
                 method.IsStatic ? 0 :
                 receiverOpt is null ? -1 :
                 MakeSlot(receiverOpt);
+
+            if (method.MethodKind == MethodKind.LocalFunction
+                && GetTopLevelMethod(method) is { ThisParameter: { } thisParameter })
+            {
+                receiverSlot = GetOrCreateSlot(thisParameter);
+            }
 
             if (receiverSlot < 0)
             {
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -2012,7 +2012,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                     case BoundKind.ThisReference:
                     case BoundKind.BaseReference:
                         {
-                            var method = GetTopLevelMethod(_symbol as MethodSymbol);
+                            var method = getTopLevelMethod(_symbol as MethodSymbol);
                             var thisParameter = method?.ThisParameter;
                             return thisParameter is object ? GetOrCreateSlot(thisParameter) : -1;
                         }
@@ -2090,20 +2090,20 @@ namespace Microsoft.CodeAnalysis.CSharp
                 }
                 return -1;
             }
-        }
 
-        private static MethodSymbol? GetTopLevelMethod(MethodSymbol? method)
-        {
-            while (method is object)
+            static MethodSymbol? getTopLevelMethod(MethodSymbol? method)
             {
-                var container = method.ContainingSymbol;
-                if (container.Kind == SymbolKind.NamedType)
+                while (method is object)
                 {
-                    return method;
+                    var container = method.ContainingSymbol;
+                    if (container.Kind == SymbolKind.NamedType)
+                    {
+                        return method;
+                    }
+                    method = container as MethodSymbol;
                 }
-                method = container as MethodSymbol;
+                return null;
             }
-            return null;
         }
 
         protected override int GetOrCreateSlot(Symbol symbol, int containingSlot = 0, bool forceSlotEvenIfEmpty = false, bool createIfMissing = true)
@@ -7008,15 +7008,10 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
 
             int receiverSlot =
+                method.MethodKind == MethodKind.LocalFunction ? getReceiverSlotForLocalFunction(method) :
                 method.IsStatic ? 0 :
-                receiverOpt is null ? -1 :
-                MakeSlot(receiverOpt);
-
-            if (method.MethodKind == MethodKind.LocalFunction
-                && GetTopLevelMethod(method) is { ThisParameter: { } thisParameter })
-            {
-                receiverSlot = GetOrCreateSlot(thisParameter);
-            }
+                receiverOpt is not null ? MakeSlot(receiverOpt) :
+                -1;
 
             if (receiverSlot < 0)
             {
@@ -7024,6 +7019,45 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
 
             ApplyMemberPostConditions(receiverSlot, method);
+
+            int getReceiverSlotForLocalFunction(MethodSymbol method)
+            {
+                bool localFunctionIsStatic = method.IsStatic;
+
+                MethodSymbol? current = method;
+                while (current is object)
+                {
+                    if (current.IsStatic != localFunctionIsStatic)
+                    {
+                        // If an instance local function is a static method or vice versa, 
+                        // we cannot apply the member post-conditions
+                        return -1;
+                    }
+
+                    var container = current.ContainingSymbol;
+                    if (container.Kind == SymbolKind.NamedType)
+                    {
+                        if (localFunctionIsStatic)
+                        {
+                            return 0;
+                        }
+
+                        if (current.ThisParameter is { } thisParameter)
+                        {
+                            return GetOrCreateSlot(thisParameter);
+                        }
+
+                        // The ContainingSymbol on a substituted local function is incorrect (returns the containing type instead of the containing method)
+                        // so we can't find the proper `this` receiver to apply the member post-conditions to.
+                        // Tracked by https://github.com/dotnet/roslyn/issues/75543
+                        return -1;
+                    }
+
+                    current = container as MethodSymbol;
+                }
+
+                return -1;
+            }
         }
 
         private void ApplyMemberPostConditions(int receiverSlot, MethodSymbol method)
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -583,7 +583,6 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
             this.Diagnostics.Clear();
             this.regionPlace = RegionPlace.Before;
-            int containingSlot = GetReceiverSlotForMemberPostConditions(_symbol as MethodSymbol);
             if (!_isSpeculative)
             {
                 ParameterSymbol methodThisParameter = MethodThisParameter;
@@ -593,7 +592,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                     EnterParameter(methodThisParameter, methodThisParameter.TypeWithAnnotations);
                 }
 
-                makeNotNullMembersMaybeNull(containingSlot);
+                makeNotNullMembersMaybeNull();
                 // We need to create a snapshot even of the first node, because we want to have the state of the initial parameters.
                 _snapshotBuilderOpt?.TakeIncrementalSnapshot(methodMainNode, State);
             }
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7012,6 +7012,12 @@ private void ApplyMemberPostConditions(BoundExpression? receiverOpt, MethodSymbo
                 receiverOpt is null ? -1 :
                 MakeSlot(receiverOpt);
 
+            if (method.MethodKind == MethodKind.LocalFunction
+                && GetTopLevelMethod(method) is { ThisParameter: { } thisParameter })
+            {
+                receiverSlot = GetOrCreateSlot(thisParameter);
```
RikkiGibson, line 7018: Should we assert that receiverSlot was `-1` before this assignment?

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -2090,20 +2090,20 @@ int getPlaceholderSlot(BoundExpression expr)
                 }
                 return -1;
             }
+        }
 
-            static MethodSymbol? getTopLevelMethod(MethodSymbol? method)
+        static MethodSymbol? GetTopLevelMethod(MethodSymbol? method)
```
RikkiGibson, line 2095:
```suggestion
        private static MethodSymbol? GetTopLevelMethod(MethodSymbol? method)
```

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,406 @@ public class C
...
+                Diagnostic(ErrorCode.WRN_NullReferenceReceiver, "field3").WithLocation(21, 9),
+                // (22,9): warning CS8602: Dereference of a possibly null reference.
                 //         field4.ToString(); // 4
-                Diagnostic(ErrorCode.WRN_NullReferenceReceiver, "field4").WithLocation(16, 9)
-                );
+                Diagnostic(ErrorCode.WRN_NullReferenceReceiver, "field4").WithLocation(22, 9));
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/56256")]
+        public void MemberNotNull_LocalFunction_WithinExtensionMethod()
```
RikkiGibson, line 21182: It looks like an extension method is not used in this test.

jcouv: I couldn't come up with any interesting test involving extension methods after all. Removed

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7015,6 +7015,7 @@ private void ApplyMemberPostConditions(BoundExpression? receiverOpt, MethodSymbo
             if (method.MethodKind == MethodKind.LocalFunction
                 && GetTopLevelMethod(method) is { ThisParameter: { } thisParameter })
             {
+                Debug.Assert(receiverSlot == -1);
```
RikkiGibson, line 7018: Apparently a bunch of tests are failing this assert. Huh. https://dev.azure.com/dnceng-public/public/_build/results?buildId=842224&view=ms.vss-test-web.build-test-results-tab&runId=21841210&resultId=123811&paneView=debug

jcouv:

RikkiGibson (dismissed):

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,409 @@ public class C
...
+    public void M()
+    {
+        init();
+        field1.ToString();
+        field2.ToString();
+        field3.ToString();
+        field4.ToString(); // 1
+
+        [MemberNotNull(nameof(field1), nameof(field2))]
+        bool init() => throw null!;
```
cston, line 21074: >bool init() => throw null!; [](http://example.com/codeflow?start=8&length=27)

Consider testing within nested local function.
```csharp
void init()
{
    init2();
    [MemberNotNull(nameof(field1), nameof(field2))]
    void init2() { }
}
```

cston, line 21074: >bool init() => throw null!; [](http://example.com/codeflow?start=8&length=27)

Consider testing the equivalent case with a `static` local function, with the same instance members.
```csharp
[MemberNotNull(nameof(field1), nameof(field2))]
static bool init() => throw null!;
```

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,409 @@ public class C
...
+using System.Diagnostics.CodeAnalysis;
+
+init();
+
+[MemberNotNull(nameof(field1))]
+void init() => throw null!;
+""", MemberNotNullAttributeDefinition]);
+
+            c.VerifyDiagnostics(
+                // (6,23): error CS0103: The name 'field1' does not exist in the current context
```
cston, line 21224: >'field1' [](http://example.com/codeflow?start=50&length=8)

Consider testing with a defined field.
```csharp
...

partial class Program
{
    object? field1;
}
```

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -791,115 +792,13 @@ void checkMemberStateOnConstructorExit(MethodSymbol constructor, Symbol member,
...
-                            enforceMemberNotNullWhenIfAffected(returnStatement.Syntax, sense: true, method.ContainingType.GetMembers(memberName), pendingReturn.StateWhenTrue, pendingReturn.StateWhenFalse);
-                        }
-
-                        foreach (var memberName in method.NotNullWhenFalseMembers)
-                        {
-                            enforceMemberNotNullWhenIfAffected(returnStatement.Syntax, sense: false, method.ContainingType.GetMembers(memberName), pendingReturn.StateWhenFalse, pendingReturn.StateWhenTrue);
-                        }
-                    }
-                }
-                else if (returnStatement.ExpressionOpt is { ConstantValueOpt: { IsBoolean: true, BooleanValue: bool value } })
```
jcouv, line 834: :memo: this `if` block was removed (it was not reachable)

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -791,113 +791,6 @@ void checkMemberStateOnConstructorExit(MethodSymbol constructor, Symbol member,
                 }
             }
 
-            void enforceMemberNotNullOnMember(SyntaxNode? syntaxOpt, LocalState state, MethodSymbol method, string memberName)
```
jcouv, line 794: :memo: A number of local functions were switched to methods below

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7020,6 +7056,53 @@ private void ApplyMemberPostConditions(BoundExpression? receiverOpt, MethodSymbo
...
+            }
+
+            bool isStatic = method.IsStatic;
+
+            MethodSymbol? current = method;
+            while (current is object)
+            {
+                if (current.IsStatic != isStatic)
+                {
+                    // If an instance local function is a static method or vice versa,
```
cston, line 7076: >an instance local function is a static method [](http://example.com/codeflow?start=26&length=45)

Does "instance local function" mean "non-static local function"? If so, let's use "non-static" instead to avoid implying there's an instance.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7020,6 +7056,53 @@ private void ApplyMemberPostConditions(BoundExpression? receiverOpt, MethodSymbo
             ApplyMemberPostConditions(receiverSlot, method);
         }
 
+        // For an instance method, or a non-static local function in an instance method, returns the slot for the `this` parameter
+        // For a static method, or a static local function in a static method, returns 0
```
cston, line 7060: >For a static method, or a static local function in a static method [](http://example.com/codeflow?start=11&length=66)

Should this also include non-static local function in a static method? If so, perhaps: "For a static method, or a local function in a static method, or a static local function ...".

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,1366 @@ public class C
...
+                // (14,9): warning CS8602: Dereference of a possibly null reference.
+                //         field3.ToString(); // 1
+                Diagnostic(ErrorCode.WRN_NullReferenceReceiver, "field3").WithLocation(14, 9),
+                // (15,9): warning CS8602: Dereference of a possibly null reference.
+                //         field4.ToString(); // 2
+                Diagnostic(ErrorCode.WRN_NullReferenceReceiver, "field4").WithLocation(15, 9));
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/56256")]
+        public void MemberNotNull_InstanceLocalFunction_MissingArgument()
```
cston, line 20976: >InstanceLocalFunction [](http://example.com/codeflow?start=34&length=21)

Consider calling these "NonStaticLocalFunction" since I don't think we have a term "instance local function".

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
cston: "... **in** a static method"?

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7020,6 +7056,53 @@ private void ApplyMemberPostConditions(BoundExpression? receiverOpt, MethodSymbo
...
+        {
+            if (method is null)
+            {
+                return -1;
+            }
+
+            bool isStatic = method.IsStatic;
+
+            MethodSymbol? current = method;
+            while (current is object)
```
cston, line 7072: Consider using:
```
MethodSymbol current = method;
while (true)
```

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7020,6 +7056,53 @@ private void ApplyMemberPostConditions(BoundExpression? receiverOpt, MethodSymbo
...
+            {
+                if (current.IsStatic != isStatic)
+                {
+                    // If an instance local function is a static method or vice versa, 
+                    // we cannot apply the member post-conditions
+                    return -1;
+                }
+
+                var container = current.ContainingSymbol;
+                if (container.Kind == SymbolKind.NamedType)
```
cston, line 7082: >if (container.Kind == SymbolKind.NamedType) [](http://example.com/codeflow?start=16&length=43)

Perhaps `if (container.Kind != SymbolKind.Method) { ... } current = (MethodSymbol)container;`

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,1366 @@ public class C
...
+public class C
+{
+    public static string field1 = "";
+    public static string? field2;
+
+    public void M()
+    {
+        init();
+        field1.ToString();
+        field2.ToString(); // 1
```
cston, line 21319: >field2.ToString(); // 1 [](http://example.com/codeflow?start=8&length=23)

Is this warning expected? (It looks like we don't report a warning for the corresponding static method case.)

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,1366 @@ public class C
...
+public class C
+{
+    public static string field1 = "";
+    public static string? field2;
+
+    public static void M()
+    {
+        init();
+        field1.ToString();
+        field2.ToString(); // 1
```
cston, line 21375: >field2.ToString(); // 1 [](http://example.com/codeflow?start=8&length=23)

Reporting a warning seems incorrect here.

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,1366 @@ public class C
...
+        public void MemberNotNull_StaticLocalFunction_TopLevel()
+        {
+            var c = CreateCompilation(["""
+#nullable enable
+using System.Diagnostics.CodeAnalysis;
+
+init();
+
+[MemberNotNull(nameof(field1))]
+static void init() => throw null!;
```
cston, line 21529: Consider testing with a field.
```csharp
#nullable enable

init();
field1.ToString();

[MemberNotNull(nameof(field1))]
static void init() => throw null!;

partial class Program
{
    static object? field1;
}
```

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -3256,6 +3259,14 @@ private void AnalyzeLocalFunctionOrLambda(
 
                 EnterParameters();
 
+                bool isLocalFunction = lambdaOrFunctionSymbol is LocalFunctionSymbol;
+                if (isLocalFunction)
+                {
+                    MakeMembersMaybeNull(lambdaOrFunctionSymbol, lambdaOrFunctionSymbol.NotNullMembers);
+                    MakeMembersMaybeNull(lambdaOrFunctionSymbol, lambdaOrFunctionSymbol.NotNullWhenTrueMembers);
+                    MakeMembersMaybeNull(lambdaOrFunctionSymbol, lambdaOrFunctionSymbol.NotNullWhenFalseMembers);
```
cston, line 3267: Why set these members to maybe null rather than relying on the state from the calling method?

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,1366 @@ public class C
...
+    public string? field4;
+
+    public void M()
+    {
+        init();
+
+        [MemberNotNull(nameof(field1), nameof(field2))]
+        void init()
+        {
+            field1.ToString(); // 1
```
cston, line 21557: >field1.ToString(); // 1 [](http://example.com/codeflow?start=12&length=23)

Why report a warning for this case?

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20952,32 +20952,1366 @@ public class C
...
+        {
+        } // 1
+    }
+}
+""", MemberNotNullAttributeDefinition]);
+
+            // Note: we add a synthesized BoundReturnStatement in BindLocalFunctionStatement
+            //       using the block body as syntax, but we report the diagnostic on the closing brace in that case
+            c.VerifyEmitDiagnostics(
+                // (16,9): warning CS8774: Member 'field1' must have a non-null value when exiting.
```
cston, line 21679: >'field1' [](http://example.com/codeflow?start=50&length=8)

Why report a warning for `field1`?

jcouv: That's the same behavior as with methods. If you put the `MemberNotNull` attribute, you must set the field to a not-null value by the time you exit the method. [sharplab](https://sharplab.io/#v2:CYLg1APgAgDABFAjAOgCIEsCGBzAdgewGcAXdAY0OQGF9gBTAQV0wBsBPQ9QgbgFgAoKAGYEAJjhUBAbwFw5CEUngAzdHRbBEcALxwARHr795ChIhgB+OKvXBRRk8LMq1GkboMP5TpVZsaAFiNZEwBtAFk6AFsAIzoAJwA5fGJEgFcWFgAKZii6fGUs/00ASgAaOFz8wuLREpKAXRDvALh0XHRiLJLmuRljEwBfOAB6EbhEXtHxgHdMeI7cbAkAZQAOAHYNgJA4SNiEuAByYsQjuCi0kjgAC0wANzo4TEr8XABaXAyWOHvWNKeMxudFwcDoAA9Ou1sMgpmM4HMFtDVpttrt9nF4sdaudLtc7o9nq8Pl9Mr9/oDgaCIVClrD+IMgA)

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
jcouv: Yes, that's why.
For the enforcement of `MemberNotNull` attribute within a method body, we set the initial state of the field to maybe-null when entering the body and check that it's been changed to not-null by the time you exit the method. That way we know that the method did explicitly set a not-null state. 
I used the same design for local functions. That analysis/enforcement can be performed by looking at the local function in isolation, without regard to its invocations.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
jcouv: Addressed in another comment

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -1082,54 +975,164 @@ static Symbol getFieldSymbolToBeInitialized(Symbol requiredMember)
...
+
+            int containingSlot = GetReceiverSlotForMemberPostConditions(_symbol as MethodSymbol);
+
+            if (containingSlot < 0)
+            {
+                return -1;
+            }
 
-                return GetOrCreateSlot(member, containingSlot);
+            if (member.IsStatic != (containingSlot == 0))
```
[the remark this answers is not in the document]
jcouv: Yes
This is illustrated by `MemberNotNull_StaticMemberOnInstanceMethod`
The opposite scenario is illustrated by `MemberNotNull_InstanceMemberOnStaticMethod`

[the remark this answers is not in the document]
jcouv: Making a fix to align the logic in `ApplyMemberPostConditions` and `GetSlotForMemberPostCondition` regarding this scenario

jcouv:
@cston @RikkiGibson This is ready for another look. Thanks

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -20937,10 +20937,10 @@ public C()
                 );
         }
 
-        [Fact]
-        public void MemberNotNull_LocalFunction()
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/56256")]
+        public void MemberNotNull_InstanceLocalFunction()
```
cston, line 20941: >InstanceLocalFunction [](http://example.com/codeflow?start=34&length=21)

Consider naming this `_NonStaticLocalFunction`, for consistency with other tests, and because there's no requirement that the local function takes an instance (and doesn't in this case). There are a couple of other uses of `_InstanceLocalFunction` below as well.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7020,6 +7062,40 @@ private void ApplyMemberPostConditions(BoundExpression? receiverOpt, MethodSymbo
...
+            {
+                return -1;
+            }
+
+            MethodSymbol? current = method;
+            bool anyStatic = current.IsStatic;
+            while (current.ContainingSymbol is MethodSymbol container)
+            {
+                current = container;
+                anyStatic |= container.IsStatic;
```
cston, line 7080: >anyStatic |= container.IsStatic [](http://example.com/codeflow?start=16&length=31)

Consider removing `anyStatic` and change this to `if (container.IsStatic) return 0;`

cston (approved):

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
jcouv: Yup. Sorry, I missed those

cston (approved):

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -21083,11 +21083,199 @@ public void M()
...
+        static void local()
+        {
+            init();
+            field1.ToString();
+            field2.ToString();
+            field3.ToString();
+            field4.ToString();
+
+            [MemberNotNull(nameof(field1), nameof(field2))]
+            bool init() => throw null!;
```
RikkiGibson, line 21185: It would be interesting to have versions of a few of these tests (e.g. referencing instance members in 'static context'), where the end of the local function is reachable.

src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs
```diff
@@ -21002,7 +21002,7 @@ public void M()
         }
 
         [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/56256")]
-        public void MemberNotNull_LocalFunction_BaseField()
+        public void MemberNotNull_NonStaticLocalFunction_BaseField()
```
RikkiGibson, line 21005: Is there any test which is for 'MemberNotNull' on a local function within a constructor?

RikkiGibson (approved):
LGTM just minor questions/nits.

RikkiGibson (approved):

PR #75449, opened 2024-10-09
VB: Do not use overload resolution while lowering interpolated strings

AlekseyTs:
Related to #74275.

### DIFF
--- a/src/Compilers/VisualBasic/Portable/Lowering/LocalRewriter/LocalRewriter_InterpolatedString.vb
+++ b/src/Compilers/VisualBasic/Portable/Lowering/LocalRewriter/LocalRewriter_InterpolatedString.vb
@@ -46,7 +46,8 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
                 Return factory.StringLiteral(ConstantValue.Create(valueWithEscapes.Replace("{{", "{").Replace("}}", "}")))
 
             Else
-                Return InvokeInterpolatedStringFactory(node, node.Type, "Format", node.Type, factory)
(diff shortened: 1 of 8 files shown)


AlekseyTs:

AlekseyTs:

src/Compilers/VisualBasic/Portable/Lowering/LocalRewriter/LocalRewriter_InterpolatedString.vb
```diff
@@ -78,117 +79,7 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
             '
             ' (2) For the built-in types, we can use .ToString(string format) for some format strings.
             '     Detect those cases that can be handled that way and take advantage of them.
```
333fred, line 81: Should we remove this comment now, since it's done in initial binding? Or mention that this transformation was done during initial binding and the result is in `ConstructionOpt`?

333fred (approved):

src/Compilers/VisualBasic/Portable/Lowering/LocalRewriter/LocalRewriter_InterpolatedString.vb
AlekseyTs: > Should we remove this comment now, since it's done in initial binding? Or mention that this transformation was done during initial binding and the result is in `ConstructionOpt`?

I do not think there is a need to do either.

AlekseyTs:

jcouv:
Looking

jcouv (approved):
LGTM Thanks (iteration 2)

Issue #75024, opened 2024-09-09
Handling of infinite collection initialization chains could be unified


jjonescz:
ERR_CollectionInitializerInfiniteChainOfAddCalls and ERR_ParamsCollectionInfiniteChainOfConstructorCalls

See https://github.com/dotnet/roslyn/pull/74899#discussion_r1747550530:

> We are handling cycles through constructors (`ERR_ParamsCollectionInfiniteChainOfConstructorCalls`) elsewhere and with a different approach. I assume that at the time that handling was introduced, there was no way to get into a cycle through an `Add` method. It feels like we should unify and combine both checks, i.e. do them in the same place and use similar approaches for both.

PR #75464, opened 2024-10-10
Unify collection expression cycle handling

jjonescz:
Resolves https://github.com/dotnet/roslyn/issues/75024.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
@@ -660,10 +660,23 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             if (collectionTypeKind is CollectionExpressionTypeKind.ImplementsIEnumerable)
             {
+                if (targetType is NamedTypeSymbol namedType &&
+                    HasParamsCollectionTypeInProgress(namedType, out NamedTypeSymbol? inProgress, out MethodSymbol? inProgressConstructor))
+                {
+                    Debug.Assert(inProgressConstructor is not null);
+                    diagnostics.Add(ErrorCode.ERR_ParamsCollectionInfiniteChainOfConstructorCalls, syntax, inProgress, inProgressConstructor.OriginalDefinition);
+                    return BindCollectionExpressionForErrorRecovery(node, namedType, inConversion: true, diagnostics);
+                }
+
                 implicitReceiver = new BoundObjectOrCollectionValuePlaceholder(syntax, isNewInstance: true, targetType) { WasCompilerGenerated = true };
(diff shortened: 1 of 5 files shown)


src/Compilers/CSharp/Test/Emit3/Semantics/ParamsCollectionTests.cs
```diff
@@ -13878,16 +13878,7 @@ static void Test(params MyCollection a)
                 Diagnostic(ErrorCode.ERR_ParamsCollectionInfiniteChainOfConstructorCalls, "[1]").WithArguments("MyCollection", "MyCollection.MyCollection(params MyCollection)").WithLocation(21, 14),
                 // (21,15): error CS9223: Creation of params collection 'MyCollection' results in an infinite chain of invocation of constructor 'MyCollection.MyCollection(params MyCollection)'.
                 //         Test([1]);
```
cston, line 13880: Is this comment referring to the correct line?

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -660,11 +660,24 @@ private BoundCollectionExpression ConvertCollectionExpression(
...
+                    return BindCollectionExpressionForErrorRecovery(node, namedType, inConversion: true, diagnostics);
+                }
+
                 implicitReceiver = new BoundObjectOrCollectionValuePlaceholder(syntax, isNewInstance: true, targetType) { WasCompilerGenerated = true };
-                collectionCreation = BindCollectionExpressionConstructor(syntax, targetType, diagnostics);
+                collectionCreation = BindCollectionExpressionConstructor(syntax, targetType, constructor, diagnostics);
                 Debug.Assert((collectionCreation is BoundNewT && !isExpanded && constructor is null) ||
                              (collectionCreation is BoundObjectCreationExpression creation && creation.Expanded == isExpanded && creation.Constructor == constructor));
 
+                if (collectionCreation.HasErrors)
```
cston, line 676: >collectionCreation.HasErrors [](http://example.com/codeflow?start=20&length=28)

Just curious: Why is this check needed rather than continuing and binding the elements?

cston (approved):

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
jjonescz: It's not needed, but it prevents cascading diagnostics. For example, in call like `Test(2, 3);`, the cycle error would be reported on `Test(2, 3)`, `2`, and `3`.

AlekseyTs (approved):
LGTM (commit 2)

Issue #75453, opened 2024-10-09
Migrate the target framework in our builds


jaredpar:
Now that `main` is 17.13 and .NET 9 is hitting RTM it's time to update the target frameworks used in our build. 

Specifically in [TargetFramework.props](https://github.com/dotnet/roslyn/blob/main/eng/targets/TargetFrameworks.props). 

- `$(NetRoslyn)` should be net9.0
- `$(NetRoslynAll)` should be net8.0 and net9.0

For the moment `$(NetRoslynNext)` should remain `net9.0` if possible as we won't have a real `net10.` target for a few more months. 

Other items: 

- Make sure to check for `#if NET7` style directives that can be moved forward.
- Likely requires us to change YML files in a few places.
- The `NetVSShared` can likely move to simply `net8.0` now but we should check with IDE that they don't need `net7.0` in VS Code anymore

jjonescz:
> * Make sure to check for `#if NET7` style directives that can be moved forward.

Why would these need to move forward? I think they will still work, and it seems more resilient to let them be.

jaredpar:
When people see `#if NET7` in code but no `net7.0` in the TFM list then there is a tendency to ask "wait, is that stale? can we remove that?" Hence I like to roll them forward to make sure that we still need them and to make it clearer to readers that they are intended.

Issue #75205, opened 2024-09-22
remove `net7.0` target


WeihanLi:
![image](https://github.com/user-attachments/assets/4a8483d9-727c-4b28-a898-6cdb96b08435)

Since `net7.0` is out of date, maybe we should remove the `net7.0`

![image](https://github.com/user-attachments/assets/e997c85f-0eaf-4e3c-ae4c-c8fca665b6f5)

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

huoyaoyuan:
The previous major version seems used for source build: https://github.com/dotnet/roslyn/blob/main/eng/targets/TargetFrameworks.props

We should be able to remove this when moving source build to next version.

PR #75465, opened 2024-10-10
Update to net9.0

jjonescz:
Resolves https://github.com/dotnet/roslyn/issues/75453.

### DIFF
--- a/docs/contributing/Target Framework Strategy.md
+++ b/docs/contributing/Target Framework Strategy.md
@@ -110,3 +110,21 @@ When the .NET SDK RTMs and Roslyn adopts it all occurrences of `$(NetRoslynNext)
 
 **DO NOT** include both `$(NetRoslyn)` and `$(NetRoslynNext)` in the same project unless there is a very specific reason that both tests are adding value. The most common case is that the runtime has changed behavior and we simply need to update our baselines to match it. Adding extra TFMs for this just increases test time for very little gain.
 
+## Checklist for updating TFMs (once a year)
+
+- Update `TargetFrameworks.props`.
+- Change `$(NetRoslynNext)` references to `$(NetRoslyn)` in project files.
+- Update TFMs in:
+  - `.vscode\launch.json`
+  - `.vscode\tasks.json`
+  - `eng\build.ps1`
+  - `eng\build.sh`
+  - `eng\generate-compiler-code.ps1`
+  - `eng\prepare-tests.ps1`
+  - `eng\prepare-tests.sh`
+  - `eng\test-rebuild.ps1`
+  - `eng\pipelines\test-integration-helix.yml`
+  - `eng\pipelines\test-unix-job.yml`
+  - `eng\pipelines\test-windows-job.yml`
+  - `src\Tools\BuildBoss\CompilerNuGetCheckerUtil.cs`
+  - `src\Tools\Replay\README.md`
--- a/docs/contributing/Target Framework Strategy.md
+++ b/docs/contributing/Target Framework Strategy.md
@@ -113,6 +113,7 @@ When the .NET SDK RTMs and Roslyn adopts it all occurrences of `$(NetRoslynNext)
 ## Checklist for updating TFMs (once a year)
 
 - Update `TargetFrameworks.props`.
+- Might need updating `MicrosoftNetCompilersToolsetVersion` in `eng\Versions.props` to consume latest compiler features.
 - Change `$(NetRoslynNext)` references to `$(NetRoslyn)` in project files.
 - Update TFMs in:
   - `.vscode\launch.json`
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -52,7 +52,7 @@
     <MicrosoftExtensionsOptionsConfigurationExtensionVersion>8.0.0</MicrosoftExtensionsOptionsConfigurationExtensionVersion>
     <MicrosoftExtensionsOptionsVersion>8.0.0</MicrosoftExtensionsOptionsVersion>
     <MicrosoftExtensionsPrimitivesVersion>8.0.0</MicrosoftExtensionsPrimitivesVersion>
-    <MicrosoftNetCompilersToolsetVersion>4.11.0-2.24270.4</MicrosoftNetCompilersToolsetVersion>
+    <MicrosoftNetCompilersToolsetVersion>4.12.0-3.24473.3</MicrosoftNetCompilersToolsetVersion>
     <MicrosoftVSSDKBuildToolsVersion>17.9.3137-preview3</MicrosoftVSSDKBuildToolsVersion>
     <vswhereVersion>2.4.1</vswhereVersion>
     <!--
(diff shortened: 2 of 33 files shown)


DoctorKrolic:
Does this also fix https://github.com/dotnet/roslyn/issues/75205 (are there any mentions of `net7.0` in the codebase after this change)?

docs/contributing/Target Framework Strategy.md
```diff
@@ -110,3 +110,22 @@ When the .NET SDK RTMs and Roslyn adopts it all occurrences of `$(NetRoslynNext)
 
 **DO NOT** include both `$(NetRoslyn)` and `$(NetRoslynNext)` in the same project unless there is a very specific reason that both tests are adding value. The most common case is that the runtime has changed behavior and we simply need to update our baselines to match it. Adding extra TFMs for this just increases test time for very little gain.
 
+## Checklist for updating TFMs (once a year)
+
+- Update `TargetFrameworks.props`.
+- Might need updating `MicrosoftNetCompilersToolsetVersion` in `eng\Versions.props` to consume latest compiler features.
+- Change `$(NetRoslynNext)` references to `$(NetRoslyn)` in project files.
+- Update TFMs in:
```
jaredpar, line 118: Usually the way I approach this is I search for say `net7.0` using VSCode and exclude .cs,.vb. That typically gives the right set of files to update.

eng/Versions.props
```diff
@@ -52,7 +52,7 @@
     <MicrosoftExtensionsOptionsConfigurationExtensionVersion>8.0.0</MicrosoftExtensionsOptionsConfigurationExtensionVersion>
     <MicrosoftExtensionsOptionsVersion>8.0.0</MicrosoftExtensionsOptionsVersion>
     <MicrosoftExtensionsPrimitivesVersion>8.0.0</MicrosoftExtensionsPrimitivesVersion>
-    <MicrosoftNetCompilersToolsetVersion>4.11.0-2.24270.4</MicrosoftNetCompilersToolsetVersion>
+    <MicrosoftNetCompilersToolsetVersion>4.12.0-3.24473.3</MicrosoftNetCompilersToolsetVersion>
```
jaredpar, line 55: Curious: what if we just delete this and move to the compiler that comes with the .NET SDK.

docs/contributing/Target Framework Strategy.md
```diff
@@ -110,3 +110,22 @@ When the .NET SDK RTMs and Roslyn adopts it all occurrences of `$(NetRoslynNext)
 
 **DO NOT** include both `$(NetRoslyn)` and `$(NetRoslynNext)` in the same project unless there is a very specific reason that both tests are adding value. The most common case is that the runtime has changed behavior and we simply need to update our baselines to match it. Adding extra TFMs for this just increases test time for very little gain.
 
+## Checklist for updating TFMs (once a year)
```
jaredpar, line 113: One other item to check is that you don't regress the number of tests that are running in CI. Our logic for finding tests can include looking at TFM. In the past I messed this up and stopped running certain unit test assemblies.

eng/Versions.props
jjonescz: I can try that. But generally we sometimes want to consume new compiler features before we update our SDK, right? So then we would need to re-add this again.

docs/contributing/Target Framework Strategy.md
jjonescz: That's what I did as well. I will update the docs to just say that, thanks.

eng/Versions.props
jjonescz: Although I updated this to bring in the "better conversion from collection expression" feature that's needed because of some string.Join ambiguities - and that might not be even in RC1, right? Well, definitely it's not in preview 7 which we are currently at.

src/Compilers/Test/Core/Compilation/RuntimeUtilities.cs
```diff
@@ -40,10 +40,6 @@ internal static bool IsCoreClr8OrHigherRuntime
         internal static bool IsCoreClr9OrHigherRuntime
             => CoreClrRuntimeVersion is { } v && v >= 9;
 
-#if NET9_0_OR_GREATER
-#error Make the above check be an #if NET9_OR_GREATER when we add net8 support to build
-#endif
```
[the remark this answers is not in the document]
jaredpar: I'm guessing it was important at some point when I was converting locally but not by the time I merged the change.

jaredpar (approved):

333fred (approved):

src/Workspaces/MSBuildTest/Microsoft.CodeAnalysis.Workspaces.MSBuild.UnitTests.csproj
```diff
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <OutputType>Library</OutputType>
     <RootNamespace>Microsoft.CodeAnalysis.MSBuild.UnitTests</RootNamespace>
-    <TargetFrameworks>$(NetRoslyn);net472</TargetFrameworks>
+    <TargetFrameworks>$(NetVS);net472</TargetFrameworks>
```
JoeRobich, line 7: I wouldn't expect it to be necessary to hold this back. Was there a particular issue when moving this to 9.0?

src/Features/DiagnosticsTestUtilities/CodeActions/CSharpCodeFixVerifier`2+Test.cs
```diff
@@ -38,12 +38,14 @@ static Test()
                 // reasonable TLS protocol version for outgoing connections.
 #pragma warning disable CA5364 // Do Not Use Deprecated Security Protocols
 #pragma warning disable CS0618 // Type or member is obsolete
+#pragma warning disable SYSLIB0014 // 'ServicePointManager' is obsolete
```
jasonmalinowski, line 41: Is this obsoleted in a way that means this isn't working anymore? Is this just stale now?

jasonmalinowski: (I have no idea what this is, but might be worth filing a bug as a follow up to track what this is.)

src/Workspaces/MSBuildTest/Microsoft.CodeAnalysis.Workspaces.MSBuild.UnitTests.csproj
jasonmalinowski: We run the unit tests on a machine with VS installed, so that _might_ matter somehow if maybe the regular SDK isn't there. If that somehow broke, do file a bug @jjonescz and we can investigate further.

jasonmalinowski (approved):

src/Features/DiagnosticsTestUtilities/CodeActions/CSharpCodeFixVerifier`2+Test.cs
jjonescz: WebRequest, WebClient, and ServicePoint are obsolete since .NET 6. See https://learn.microsoft.com/en-us/dotnet/core/compatibility/networking/6.0/webrequest-deprecated. I'm not sure why the SYSLIB0014 warning wasn't reported previously, will investigate that.

Issue #57745, opened 2021-11-13
Obsolete diagnostic not reported on System.Index or System.Range


jcouv:
```

        [Fact]
        public void ObsoleteRangeType()
        {
            var source = @"
_ = new C()[..];

class C
{
    public int Length => 0;
    public int this[int i] => 0;
    public int Slice(int i, int j) => 0;
}

namespace System
{
    [Obsolete]
    public readonly struct Range
    {
        public Index Start { get; }

        public Index End { get; }

        public Range(Index start, Index end) => throw null;

        public static Range StartAt(Index start) => throw null;

        public static Range EndAt(Index end) => throw null;

        public static Range All => throw null;
    }
}
";
            // Note: we currently don't report Obsolete diagnostic on either Index or Range
            var comp = CreateCompilation(new[] { source, TestSources.Index });
            comp.VerifyDiagnostics();
        }
```

Youssef1313:
This should have fallen off of https://github.com/dotnet/roslyn/pull/53463

jcouv:
We discussed this topic earlier today with Jared/Aleksey/Fred and I wanted to record the team's stance: we're okay not to check Obsolete, UnmanagedCallersOnly or unexpected constraints on well-known types/members. I'm in process of [documenting](https://github.com/dotnet/roslyn/pull/75522) that.
Thanks @DoctorKrolic for looking into the issue. I'll go ahead and close it

PR #75473, opened 2024-10-10
Report obsolete diagnostics on `System.Index` and `System.Range`

DoctorKrolic:
Fixes: https://github.com/dotnet/roslyn/issues/57745

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
@@ -4,12 +4,6 @@
 
 #nullable disable
 
-using Microsoft.CodeAnalysis.Collections;
-using Microsoft.CodeAnalysis.CSharp.Symbols;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
-using Microsoft.CodeAnalysis.PooledObjects;
-using Microsoft.CodeAnalysis.Text;
-using Roslyn.Utilities;
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
@@ -18,6 +12,12 @@ using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using Microsoft.CodeAnalysis.Collections;
+using Microsoft.CodeAnalysis.CSharp.Symbols;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.PooledObjects;
+using Microsoft.CodeAnalysis.Text;
+using Roslyn.Utilities;
 
 namespace Microsoft.CodeAnalysis.CSharp
 {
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -2597,6 +2602,7 @@ private BoundExpression BindRangeExpression(RangeExpressionSyntax node, BindingD
             CheckFeatureAvailability(node, MessageID.IDS_FeatureRangeOperator, diagnostics);
 
             TypeSymbol rangeType = GetWellKnownType(WellKnownType.System_Range, diagnostics, node);
+            ReportDiagnosticsIfObsolete(diagnostics, rangeType, node, hasBaseReceiver: false);
```
jcouv, line 2605: I looked around for other references to `WellKnownType.System_Range/System_Index` to see if more scenarios have a similar issue.
The ones related to implicit index/range indexers seem fine.
But a few more places need closer look (likely are failing to report obsolete diagnostic):
- slice patterns (see `BindSlicePattern`),
- inline array index (see `tryImplicitConversionToInlineArrayIndex` and `bindInlineArrayElementAccess`). It looks like the scenario may require an implicit user-defined conversion from the operand type to `System.Index` or `System.Range`
- `BindRangeExpressionOperand` uses `WellKnownType.System_Index`
Could you add a test for these scenarios and either fix them or file a tracking issue?

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -2588,6 +2589,10 @@ private BoundExpression BindFromEndIndexExpression(PrefixUnaryExpressionSyntax n
 
             BoundExpression boundConversion = CreateConversion(boundOperand, conversion, intType, diagnostics);
             MethodSymbol symbolOpt = GetWellKnownTypeMember(WellKnownMember.System_Index__ctor, diagnostics, syntax: node) as MethodSymbol;
+            if (symbolOpt is not null)
+            {
+                ReportDiagnosticsIfObsolete(diagnostics, symbolOpt, node, hasBaseReceiver: false);
```
333fred, line 2594: This should probably also check `UnmanagedCallersOnly`

jcouv (commented):
Done with review pass (iteration 1). Looks good but may be some more interesting scenarios to test

DoctorKrolic:
@jcouv @333fred A thought came to my mind: what if we move "report if obsolete" logic down to `GetWellKnownType`/`GetWellKnownTypeMember`? They already receive a syntax node to report a use site diagnostics on, well, they can use this node to also eportobsolete diagnostics as well. This way such logic will implicitly apply to new well-known types/members as they are added to the compiler and we want such behaviour, right?

jcouv:
> what if we move "report if obsolete" logic down to GetWellKnownType/GetWellKnownTypeMember?

Discussed further with @333fred and @AlekseyTs following this question. 
It would indeed be beneficial to simplify the code and be more systematic, but the churn, review and possible complications don't seem worthwhile.
Going in the other direction, we think it would be better to close this category of issues and make it clear that we don't generally not bother reporting Obsolete/UnmanagedCallersOnly diagnostics for well-known types/members, unless there is a strong reason to.
I'll do a search to close such issues. I also took a stab at some [documentation](https://github.com/dotnet/roslyn/pull/75522) to clarify this for future work.

As far as the present PR, I would recommend closing it. Sorry for the trouble

PR #75487, opened 2024-10-11
Test case for analyzer dependency loading

jaredpar:
Teams were asking about whether or not the compiler will handle multiple analyzers taking a dependency on the same utility DLL at different versions (when it's strong named). This adds a test that goes through that scenario to validate it will work.

### DIFF
--- a/src/Compilers/Test/Core/AssemblyLoadTestFixture.cs
+++ b/src/Compilers/Test/Core/AssemblyLoadTestFixture.cs
@@ -180,6 +180,11 @@ namespace Delta
         {
             sb.AppendLine(""Delta: "" + s);
         }
+
+        public void M1()
+        {
+
+        }
     }
 }
 ";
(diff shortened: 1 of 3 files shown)


jaredpar:
@RikkiGibson, @dotnet/roslyn-compiler PTAL

src/Compilers/Test/Core/AssemblyLoadTestFixture.cs
```diff
@@ -189,6 +194,36 @@ public void Write(StringBuilder sb, string s)
...
+namespace Delta
+{
+    public class D
+    {
+        public void Write(StringBuilder sb, string s)
+        {
+            sb.AppendLine(""Delta.2: "" + s);
+        }
+
+        public void M2()
```
RikkiGibson, line 212: It looks like the only real change in here is that 'M1' and 'M2' were added, right?

RikkiGibson (approved):

PR #75489, opened 2024-10-11
Build and publish .ni.pdb for R2R images

tmat:


src/Workspaces/Remote/ServiceHub.CoreComponents/CoreComponents.Shared.targets
```diff
@@ -104,7 +105,21 @@
...
 
+  <!--
+    Compiles all R2R images and packages the symbol files (*.ni.pdb)
+  -->
   <Target Name="CompileReadyToRun"
+          DependsOnTargets="CompileReadyToRunImage"
+          Condition="'$(GenerateReadyToRun)' == 'true'">
+    
+    <MakeDir Directories="$(ArtifactsNonShippingPackagesDir)" />
+    <ZipDirectory SourceDirectory="$(PdbWorkDir)" DestinationFile="$(ArtifactsNonShippingPackagesDir)$(ProjectName).symbols.nupkg" />
```
genlu, line 116: I thought the fix would be much complicated than that, thanks! By the way how did you figured this out? Is there any documents related to this?

genlu (approved):

arkalyanms (approved):

src/Workspaces/Remote/ServiceHub.CoreComponents/CoreComponents.Shared.targets
tmat: Arcade is publishing symbol packages here: https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/Publish.proj#L89

Typically a symbol package is created by nuget, but I think just zipping up the files might also work. I don't think arcade publishing specifically needs a nuget pacakge, just a zip container.

PR #75490, opened 2024-10-11
VB: Do not use overload resolution for expression tree conversion

AlekseyTs:
Related to #74275.

### DIFF
--- a/src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter.vb
+++ b/src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter.vb
@@ -194,6 +194,11 @@ lSelect:
             _factory.Syntax = node.Syntax
             Dim result = VisitInternal(node)
             _factory.Syntax = old
+
+            If result.Type Is _expressionType Then
+                Return result
+            End If
+
             Return _factory.Convert(_expressionType, result)
         End Function
 
@@ -333,7 +338,7 @@ lSelect:
                 parameters.Add(parameterReference)
                 _parameterMap(p) = parameterReference
 
-                Dim parameter As BoundExpression = ConvertRuntimeHelperToExpressionTree("Parameter",
+                Dim parameter As BoundExpression = ConvertRuntimeHelperToExpressionTree(WellKnownMember.System_Linq_Expressions_Expression__Parameter,
                                                                                         _factory.[Typeof](
                                                                                             p.Type.InternalSubstituteTypeParameters(_typeMap).Type,
                                                                                             _factory.WellKnownType(WellKnownType.System_Type)),
@@ -352,7 +357,7 @@ lSelect:
             Dim result = _factory.Sequence(locals.ToImmutableAndFree(),
                                            initializers.ToImmutableAndFree(),
                                            ConvertRuntimeHelperToExpressionTree(
-                                               "Lambda",
+                                               WellKnownMember.System_Linq_Expressions_Expression__Lambda_OfTDelegate,
                                                ImmutableArray.Create(Of TypeSymbol)(delegateType),
                                                translatedBody,
                                                _factory.Array(parameterExpressionType, parameters.ToImmutableAndFree())))
@@ -384,11 +389,11 @@ lSelect:
             End If
 
             If method.MethodKind = MethodKind.DelegateInvoke Then
-                Return ConvertRuntimeHelperToExpressionTree("Invoke",
+                Return ConvertRuntimeHelperToExpressionTree(WellKnownMember.System_Linq_Expressions_Expression__Invoke,
                                                             receiverOpt,
                                                             ConvertArgumentsIntoArray(node.Arguments))
             Else
-                Return ConvertRuntimeHelperToExpressionTree("Call",
+                Return ConvertRuntimeHelperToExpressionTree(WellKnownMember.System_Linq_Expressions_Expression__Call,
                                                             If(method.IsShared, _factory.Null(_expressionType), receiverOpt),
                                                             _factory.MethodInfo(method, _factory.WellKnownType(WellKnownType.System_Reflection_MethodInfo)), ConvertArgumentsIntoArray(node.Arguments))
             End If
@@ -402,7 +407,7 @@ lSelect:
 
             Dim rewrittenReceiver As BoundExpression = Nothing
             If fieldIsShared Then
-                rewrittenReceiver = _factory.Null()
+                rewrittenReceiver = _factory.Null(_expressionType)
             Else
                 Debug.Assert(origReceiverOpt IsNot Nothing)
                 If origReceiverOpt.Kind = BoundKind.MyBaseReference Then
@@ -415,7 +420,10 @@ lSelect:
                 End If
             End If
 
-            Return ConvertRuntimeHelperToExpressionTree("Field", rewrittenReceiver, _factory.FieldInfo(field))
+            Return ConvertRuntimeHelperToExpressionTree(
+                WellKnownMember.System_Linq_Expressions_Expression__Field,
+                rewrittenReceiver,
+                _factory.FieldInfo(field))
         End Function
 
         Private Function VisitPropertyAccess(node As BoundPropertyAccess) As BoundExpression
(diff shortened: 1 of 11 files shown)


src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter_BinaryOperator.vb
```diff
@@ -32,7 +22,6 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
                      BinaryOperatorKind.Divide,
                      BinaryOperatorKind.Modulo,
                      BinaryOperatorKind.IntegerDivide,
-                     BinaryOperatorKind.Concatenate,
```
jcouv, line 35: nit: If it's the same situation as the `Like`, consider moving this case below and adding to the comment (line 44)

AlekseyTs:
@dotnet/roslyn-compiler Please review.

src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter.vb
```diff
@@ -735,32 +761,40 @@ lSelect:
...
+                        WellKnownMember.System_Linq_Expressions_Expression__Call,
+                        receiver,
+                        _factory.MethodInfo(method, _factory.WellKnownType(WellKnownType.System_Reflection_MethodInfo)),
+                        _factory.Array(_expressionType, params)))
         End Function
 
         ' Emit a Default node for a specific type
         Private Function [Default](type As TypeSymbol) As BoundExpression
-            Return ConvertRuntimeHelperToExpressionTree("Default", _factory.[Typeof](type, _factory.WellKnownType(WellKnownType.System_Type)))
+            Return _factory.Convert(
```
jcouv, line 775: Could you provide some context on why some calls to `ConvertRuntimeHelperToExpressionTree` are now wrapped in a `Convert` but not all?

src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter.vb
```diff
@@ -735,32 +761,40 @@ lSelect:
...
         ' Emit a call node
         Private Function [Call](receiver As BoundExpression, method As MethodSymbol, ParamArray params As BoundExpression()) As BoundExpression
-            Dim factoryArgs(0 To params.Length + 1) As BoundExpression
-            factoryArgs(0) = receiver
-            factoryArgs(1) = _factory.MethodInfo(method, _factory.WellKnownType(WellKnownType.System_Reflection_MethodInfo))
-            Array.Copy(params, 0, factoryArgs, 2, params.Length)
-            Return ConvertRuntimeHelperToExpressionTree("Call", factoryArgs)
+            Return _factory.Convert(
+                _expressionType,
+                    ConvertRuntimeHelperToExpressionTree(
```
jcouv, line 766: nit: The indentation is off (two arguments should be on same level like below)

src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter.vb
AlekseyTs: > Could you provide some context on why some calls to `ConvertRuntimeHelperToExpressionTree` are now wrapped in a `Convert` but not all?

Normally, ```Visit``` does this conversion. However, we are not getting here due to a ```Visit``` invocation and this conversion was applied during overload resolution, which we no longer perform. Therefore, the conversion is applied exp[licitly because types should match exactly.

jcouv: From offline discussion, `ConvertRuntimeHelperToExpressionTree` that are returned from a `Visit...` method are already converted, we only need to apply the conversion explicitly for helper/factory methods.
FWIW, I didn't observe any break in `CodeGenExprLambda.vb` (on either `net472` or `net9.0`) or in `Build.cmd -testCompilerOnly -test` when removing `Convert` in `Default` or `New`.

src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter.vb
```diff
@@ -735,32 +761,40 @@ lSelect:
...
         Private Function Negate(expr As BoundExpression) As BoundExpression
-            Return ConvertRuntimeHelperToExpressionTree("Negate", expr)
+            Return _factory.Convert(_expressionType, ConvertRuntimeHelperToExpressionTree(WellKnownMember.System_Linq_Expressions_Expression__Negate_Expression, expr))
         End Function
 
         Private Function InitWithParameterlessValueTypeConstructor(type As TypeSymbol) As BoundExpression
             ' The "New" overload without a methodInfo automatically generates the parameterless constructor for us.
             Debug.Assert(type.IsValueType)
-            Return ConvertRuntimeHelperToExpressionTree("New", _factory.[Typeof](type, _factory.WellKnownType(WellKnownType.System_Type)))
+            Return ConvertRuntimeHelperToExpressionTree(WellKnownMember.System_Linq_Expressions_Expression__New_Type, _factory.[Typeof](type, _factory.WellKnownType(WellKnownType.System_Type)))
```
jcouv, line 797: It is surprising that we don't need `Convert` here since `InitWithParameterlessValueTypeConstructor` and `[Default]` are used in the same place in the same way.

jcouv (approved):
Done with review pass (iteration 1)

src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter.vb
AlekseyTs: I removed this conversion, apparently changes elsewhere made it unnecessary

src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter.vb
AlekseyTs: > It is surprising that we don't need `Convert` here since `InitWithParameterlessValueTypeConstructor` and `[Default]` are used in the same place in the same way.

I was adding conversions on as needed basis, when tests were failing. I also wasn't changing all ```ConvertRuntimeHelperToExpressionTree``` calls at the same time, so the conversions were necessary to pass tests at some point. It looks like the need to add conversions for some call sites got eliminated when all call sites were adjusted. I removed those. Thanks.

jcouv (approved):
LGTM Thanks (iteration 3)

AlekseyTs:

AlekseyTs:

AlekseyTs:

jjonescz (approved):

src/Compilers/VisualBasic/Portable/Lowering/ExpressionLambdaRewriter/ExpressionLambdaRewriter_BinaryOperator.vb
```diff
@@ -317,34 +391,40 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
...
+                If helper Is Nothing Then ' Don't know how to do 'BinaryOperatorKind.Power' without the method
+                    Return _factory.BadExpression(Visit(node.Left), Visit(node.Right))
+                End If
             ElseIf resultUnderlyingSpecialType = SpecialType.System_Decimal Then
                 helper = GetHelperForDecimalBinaryOperation(opKind)
-
-            ElseIf opKind = BinaryOperatorKind.Concatenate Then
-                helper = DirectCast(_factory.SpecialMember(SpecialMember.System_String__ConcatStringString), MethodSymbol)
-
+                If helper Is Nothing Then ' Don't know how to do 'BinaryOperatorKind.Power' without the method
```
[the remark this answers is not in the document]
AlekseyTs: > This comment and the one above look inaccurate. We might not be doing 'Power' operation.

I think the comment accurately reflects the motivation for adding the check. It is an error to fail to find the helper for any operator. The question is what the rewriter should do to recover. The error has been reported, but the rewriter needs to produce a bound node to return. It used to be that the rewriter would continue as though the helper is not needed and would produce a node of that shape, there are expression tree factories to do that (despite the fact that that wouldn't accurately reflect semantics of the operator). In order to maintain that recovery strategy, we would have to make use of a "Power" factory method that is never used in a success scenario, which, in my opinion, would lead to a confusion of anyone who would want to figure out what is a minimal set of factories required by VB. The new error recovery approach is going to work well for all operators, but the primary motivation is to exclude usage of a "Power" factory method that would never be needed in a success scenario.

PR #75502, opened 2024-10-14
Run BinSkim once on artifacts

jjonescz:


azure-pipelines-official.yml
```diff
@@ -135,27 +140,49 @@ extends:
         displayName: Official Build
         timeoutInMinutes: 360
         templateContext:
+          # This ensures that the SDL analysis runs only once and not for each output separately.
+          # https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/features/outputs#multiple-outputs
+          outputParentDirectory: '$(Build.SourcesDirectory)/artifacts'
+
           outputs:
 
+          - output: pipelineArtifact
```
jjonescz, line 149: These two outputs are just converted from below where they were defined as steps. They need to be defined as outputs so `outputParentDirectory` set above has an effect for them.

azure-pipelines-official.yml
```diff
@@ -135,27 +140,49 @@ extends:
         displayName: Official Build
         timeoutInMinutes: 360
         templateContext:
+          # This ensures that the SDL analysis runs only once and not for each output separately.
+          # https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/features/outputs#multiple-outputs
+          outputParentDirectory: '$(Build.SourcesDirectory)/artifacts'
```
jjonescz, line 145: Technically this is orthogonal to fixing binskim (that could be done just by setting the `analyzeTargetGlob` parameter above), but without `outputParentDirectory`, the SDL analysis (including 10min of BinSkim analysis) would run for each output we upload separately.

azure-pipelines-official.yml
```diff
@@ -95,6 +95,7 @@ extends:
   parameters:
     featureFlags:
       autoBaseline: true
+      useArtifactDropTaskV1: true
```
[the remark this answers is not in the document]
jjonescz: I added this simply because I converted `- task: 1ES.PublishArtifactsDrop@1` into `- output: artifactsDrop` and the 1ES docs say that without `useArtifactDropTaskV1: true`, the output would use `@0` version of the task.

azure-pipelines-official.yml
```diff
@@ -109,6 +110,10 @@ extends:
       tsa:
         enabled: true
         configFile: '$(Build.SourcesDirectory)/eng/TSAConfig.gdntsa'
+      binskim:
+        # BinSkim from 1ES analyzes only uploaded outputs by default,
+        # but our outputs are nupkg/zip files and BinSkim cannot process those.
+        analyzeTargetGlob: '$(Build.SourcesDirectory)/artifacts/**.dll'
```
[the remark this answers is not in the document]
jjonescz: > Also does this mean it's not scanning in say `src/` anymore?

It never did, it only scanned the uploaded outptus.

> Presumably when it added suppor for NuPkg we'd need to change this to be everything?

Right. But I think I can actually remove this setting, because the `outputParentDirectory` setting already ensures we scan the whole `artifacts/` dir, so the default binskim pattern in 1ES should work again (this `analyzeTargetGlob` is a leftover from a config that I had before I set the `outputParentDirectory`, but that made binskim scan all artifacts repeatedly on each output upload).

azure-pipelines-official.yml
```diff
@@ -135,27 +136,51 @@ extends:
...
             condition: succeeded()
-            targetPath: '$(Build.SourcesDirectory)\artifacts\OptProf\$(BuildConfiguration)\Data'
+            targetPath: '$(Build.SourcesDirectory)/artifacts\OptProf\$(BuildConfiguration)\Data'
             artifactName: 'OptProf Data Files'
 
           - output: pipelineArtifact
             displayName: 'Publish Logs'
             condition: succeededOrFailed()
-            targetPath: '$(Build.SourcesDirectory)\artifacts\log\$(BuildConfiguration)'
+            targetPath: '$(Build.SourcesDirectory)/artifacts\log\$(BuildConfiguration)'
```
jaredpar, line 176: Why did you switch the directory separator char in only one place?

jjonescz: The `outputParentDirectory: '$(Build.SourcesDirectory)/artifacts'` prefix is compared as string with each target path, so the directory separator needs to match there. I can update them all to match though.

jaredpar: Where does that comparison happen?

jjonescz: https://dev.azure.com/dnceng/1ESPipelineTemplates/_git/1ESPipelineTemplates?path=/v1/Steps/SDLBinaryAnalysis.yml&version=GBmain&line=65&lineEnd=66&lineStartColumn=1&lineEndColumn=132&lineStyle=plain&_a=contents

PR #75503, opened 2024-10-14
[WIP] Pass `runningProject` to `GetUpdateAsync` and get projectToRebuild and projectToRestart back

LittleLittleCloud:
cc @tmat

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -3,6 +3,7 @@
 <configuration>
   <packageSources>
     <clear />
+    <add key="local" value="C:\Users\xiaoyuz\source\repos\VSDebugCore\artifacts\packages\Debug\Shipping" />
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet6" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" />
--- a/eng/Directory.Packages.props
+++ b/eng/Directory.Packages.props
@@ -92,7 +92,7 @@
     <!--
       Subset of Microsoft.VisualStudio.Sdk meta-package (run `csi generate-vssdk-versions.csx` to update based on VSSDK meta-package).
     -->
-    <PackageVersion Include="Microsoft.ServiceHub.Framework" Version="4.5.31" />
+    <PackageVersion Include="Microsoft.ServiceHub.Framework" Version="4.5.35" />
     <PackageVersion Include="Microsoft.VisualStudio.Composition" Version="17.10.37" />
     <PackageVersion Include="Microsoft.VisualStudio.Composition.Analyzers" Version="17.10.37" />
     <PackageVersion Include="Microsoft.VisualStudio.CoreUtility" Version="17.10.191" />
@@ -104,7 +104,7 @@
     <PackageVersion Include="Microsoft.VisualStudio.Language.StandardClassification" Version="17.10.191" />
     <PackageVersion Include="Microsoft.VisualStudio.LanguageServer.Client" Version="17.10.124" />
     <PackageVersion Include="Microsoft.VisualStudio.RemoteControl" Version="16.3.52" />
-    <PackageVersion Include="Microsoft.VisualStudio.RpcContracts" Version="17.10.21" />
+    <PackageVersion Include="Microsoft.VisualStudio.RpcContracts" Version="17.11.8" />
     <PackageVersion Include="Microsoft.VisualStudio.Shell.15.0" Version="17.10.40170" />
     <PackageVersion Include="Microsoft.VisualStudio.Telemetry" Version="17.11.8" />
     <PackageVersion Include="Microsoft.VisualStudio.Text.Data" Version="17.10.191" />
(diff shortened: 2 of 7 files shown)


NuGet.config
```diff
@@ -3,6 +3,7 @@
 <configuration>
   <packageSources>
     <clear />
+    <add key="local" value="C:\Users\xiaoyuz\source\repos\VSDebugCore\artifacts\packages\Debug\Shipping" />
```
tmat, line 6: Remove

eng/Directory.Packages.props
```diff
@@ -92,7 +92,7 @@
     <!--
       Subset of Microsoft.VisualStudio.Sdk meta-package (run `csi generate-vssdk-versions.csx` to update based on VSSDK meta-package).
     -->
-    <PackageVersion Include="Microsoft.ServiceHub.Framework" Version="4.5.31" />
+    <PackageVersion Include="Microsoft.ServiceHub.Framework" Version="4.5.35" />
```
tmat, line 95: Shouldn't be updated without updating Microsoft.VisualStudio.SDK. Probably better to bump Microsoft.VisualStudio.SDK in a separate PR.

src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
```diff
@@ -309,7 +311,7 @@ public async ValueTask<bool> HasChangesAsync(string? sourceFilePath, Cancellatio
         }
     }
 
-    public async ValueTask<ManagedHotReloadUpdates> GetUpdatesAsync(CancellationToken cancellationToken)
+    public async ValueTask<ManagedHotReloadUpdates> GetUpdatesAsync(string[] runningProjects, CancellationToken cancellationToken)
```
tmat, line 314: Should be `ImmutableArray<string>`

src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
```diff
@@ -321,6 +323,10 @@ public async ValueTask<ManagedHotReloadUpdates> GetUpdatesAsync(CancellationToke
         var activeStatementSpanProvider = GetActiveStatementSpanProvider(solution);
         var result = await GetDebuggingSession().EmitSolutionUpdateAsync(solution, activeStatementSpanProvider, cancellationToken).ConfigureAwait(false);
 
+        var projectsToRestart = new HashSet<Project>();
```
tmat, line 326: Use pooled hashsets: 
```
using var _1 = PooledHashSet.GetInstance(out var projectsToRestart);
using var _2 = PooledHashSet.GetInstance(out var projectsToRebuild);
```

src/Features/Core/Portable/EditAndContinue/EmitSolutionUpdateResults.cs
```diff
@@ -74,6 +74,127 @@ internal ImmutableArray<ManagedHotReloadDiagnostic> GetAllDiagnostics()
 
             return builder.ToImmutableAndClear();
         }
+
```
tmat, line 77: Not sure why this is shown as added. Did you rebase against the latest main?

src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
```diff
@@ -329,6 +335,9 @@ public async ValueTask<ManagedHotReloadUpdates> GetUpdatesAsync(CancellationToke
 
         UpdateApplyChangesDiagnostics(result.Diagnostics);
 
-        return new ManagedHotReloadUpdates(result.ModuleUpdates.Updates.FromContract(), result.GetAllDiagnostics().FromContract());
+        var projectsToRestartString = projectsToRestart.Select(p => p.AssemblyName).AsImmutable();
```
tmat, line 338: Wouldn't the project system need project path rather than assembly name?

src/Features/Core/Portable/EditAndContinue/EmitSolutionUpdateResults.cs
LittleLittleCloud: Not quite follow, what you mean here

eng/Directory.Packages.props
LittleLittleCloud: How to determine the right VS SDK version to use. I saw the current version is `17.10.234-preview.1` but that one is not available on nuget?

src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
```diff
@@ -20,7 +22,7 @@
 namespace Microsoft.CodeAnalysis.EditAndContinue;
 
 [Shared]
-[Export(typeof(IManagedHotReloadLanguageService))]
+[Export(typeof(IManagedHotReloadLanguageService2))]
```
tmat, line 25: The type also needs to implement the interface.

src/Features/Core/Portable/EditAndContinue/EmitSolutionUpdateResults.cs
tmat: Looks like the code was copied from https://github.com/dotnet/roslyn/pull/75503/files#diff-95af090e3bc672c6352c9245c5b43e6d1561613fc81afee422e462dc9f17a49fR266.

LittleLittleCloud: Yeah that's true, would be better to find out a way to clean it up

tmat: Let me look into that

PR #75505, opened 2024-10-14
Exclude two manually zipped servicehub symbol nupkgs from being signed

genlu:
Fixing signed build after theses two packages are being added in https://github.com/dotnet/roslyn/pull/75489

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -15,7 +15,11 @@
   <!--
     Non-default certificates.
   -->
-  <ItemGroup>
+  <ItemGroup>    
+    <!-- Exclude two manually zipped servicehub symbol nupkgs from being signed -->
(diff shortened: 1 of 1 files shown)


eng/Signing.props
```diff
@@ -15,7 +15,11 @@
   <!--
     Non-default certificates.
   -->
-  <ItemGroup>
+  <ItemGroup>    
+    <!-- Exclude two manually zipped servicehub symbol nupkgs from being signed -->
+    <ItemsToSign Remove="@(ItemsToSign)" Condition="'%(FileName)%(Extension)' == 'Microsoft.CodeAnalysis.Remote.ServiceHub.CoreComponents.arm64.symbols.nupkg'" />
```
tmat, line 20: You should be able to use an absolute path:

`<ItemsToSign Remove="$(ArtifactsNonShippingPackagesDir)Microsoft.CodeAnalysis.Remote.ServiceHub.CoreComponents.arm64.symbols.nupkg"

tmat (approved):

Issue #75484, opened 2024-10-11
The compiler should report an error about references to out-of-scope stack memory generated by [UnscopedRef], defensively copied struct and pattern matching


TessenR:
**Version Used**: 
```
Microsoft Visual Studio Professional 2022
Version 17.12.0 Preview 2.1
VisualStudio.17.Preview/17.12.0-pre.2.1+35323.107
Microsoft .NET Framework
Version 4.8.09037
```

**Steps to Reproduce**:

Compile and run the following code:
```cs
using System;
using System.Runtime.CompilerServices;
using System.Diagnostics.CodeAnalysis;
using System.Runtime.InteropServices;

namespace Net7Test;

internal class Program
{
  private static readonly Vec4 ReadOnlyVec = new Vec4(1, 2, 3, 4);

  static void Main(string[] args)
  {
    Console.WriteLine(RuntimeInformation.FrameworkDescription);

    // This refers to stack memory that has already been left out.
    ref Vec4 foo = ref Foo().x;

    foo = new Vec4(0, 0, 0, 0);
    Console.WriteLine(foo.X);
    MakeStackDirty();
    Console.WriteLine(foo.X);
    Console.WriteLine(foo);
  }

  [MethodImpl(MethodImplOptions.NoInlining)]
  private static Wrapper Foo()
  {
    // Defensive copy occurs and it is placed in stack memory implicitly.
    // The method returns a reference to the copy, which happens invalid memory access.
    if (ReadOnlyVec is var (x, _)) return x;

    throw null!;
  }

  [MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
  private static void MakeStackDirty()
  {
    Span<byte> mem = stackalloc byte[1028];
    mem.Clear();
  }
}

public struct Vec4
{
  public float X, Y, Z, W;
  public Vec4(float x, float y, float z, float w) => (X, Y, Z, W) = (x, y, z, w);

  [UnscopedRef]
  public ref Vec4 Self => ref this;

  [UnscopedRef]
  public void Deconstruct(out Wrapper x, out Vec4 y)
  {
    x = new Wrapper(ref this);
    y = Self;
  }

  public override string ToString() => $"<{X}, {Y}, {Z}, {W}>";
}

public ref struct Wrapper
{
  public ref Vec4 x;

  public Wrapper(ref Vec4 X)
  {
    x = ref X;
  }

  [UnscopedRef]
  public ref Vec4 Self => ref x;

[...]

PR #75511, opened 2024-10-15
Handle deconstruct pattern in ref safety analysis

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75484.


jjonescz:

jaredpar:
@jcouv, @cston PTAL

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -5982,6 +5982,223 @@ public static class Extensions
...
+                }
+                """;
+            CreateCompilation([source, UnscopedRefAttributeDefinition]).VerifyDiagnostics(
+                // (7,20): error CS8352: Cannot use variable 'x' in this context because it may expose referenced variables outside of their declaration scope
+                //             return x;
+                Diagnostic(ErrorCode.ERR_EscapeVariable, "x").WithArguments("x").WithLocation(7, 20),
+                // (13,20): error CS8352: Cannot use variable 'x2' in this context because it may expose referenced variables outside of their declaration scope
+                //             return x2;
+                Diagnostic(ErrorCode.ERR_EscapeVariable, "x2").WithArguments("x2").WithLocation(13, 20));
+        }
```
jcouv, line 6200: nit: Consider also testing a nested deconstruction scenario: `((var x, var y), var z)`. I expect it should just fall out by composition

jcouv (approved):
LGTM Thanks (iteration 1)

jcouv:
@cston for second review. Thanks

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
```diff
@@ -594,6 +594,63 @@ static uint getDeclarationValEscape(BoundTypeExpression typeExpression, uint val
...
+                {
+                    if (x.Pattern is BoundObjectPattern { VariableAccess: { } variableAccess })
+                    {
+                        return variableAccess;
+                    }
+
+                    var placeholder = new BoundDeconstructValuePlaceholder(
+                        x.Syntax,
+                        variableSymbol: null,
+                        isDiscardExpression: true,
```
RikkiGibson, line 624: I didn't understand why we should pass 'isDiscardExpression: true' here.

jjonescz: Because this placeholder corresponds to an `out _` discard parameter in the lowered `Deconstruct` call.

jjonescz: Originally this code path was written to handle discards in the pattern which correspond to `out _` discard arguments of the equivalent `Deconstruct` call.

But now I realize there can be also nested subpatterns and we need to be smarter about extracting all variable accesses from that.

jjonescz:
@jcouv @RikkiGibson for another look, thanks

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -5982,6 +5982,514 @@ public static class Extensions
...
+        [Fact]
+        public void Deconstruction_UnscopedRef_Assignment()
+        {
+            var source = """
+                using System.Diagnostics.CodeAnalysis;
+                class C
+                {
+                    R M()
+                    {
+                        (var x, _) = new S();
```
RikkiGibson, line 6109: It feels like we should also have tests where the Deconstruct receiver is something we can return a reference to. For example, if it is a by-ref parameter.

jjonescz: Thanks, that actually doesn't work as expected.

jjonescz:

jjonescz:
@jcouv @cston for reviews, thanks

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
```diff
@@ -594,6 +594,21 @@ static uint getDeclarationValEscape(BoundTypeExpression typeExpression, uint val
         public override BoundNode? VisitRecursivePattern(BoundRecursivePattern node)
         {
             SetPatternLocalScopes(node);
+
+            // If the equivalent `Deconstruct` call has unscoped receiver,
+            // we narrow the escape scope of the pattern input
+            // which flows as the escape scope of all ref-struct declaration subpatterns
+            // and so the ref safety of the pattern is equivalent to a `Deconstruct(out var ...)` invocation
+            // where "safe-context inference of declaration expressions" would have the same effect.
+            if (node.DeconstructMethod is { HasUnscopedRefAttribute: true } or
```
jcouv, line 603: >{ HasUnscopedRefAttribute: true } [](http://example.com/codeflow?start=42&length=33)

nit: I'm wondering whether this could also be written in terms of `EffectiveScope`, for regularity with second case. Would something like `{ IsExtensionMethod: false, ThisParameter: { EffectiveScope: ScopedKind.None } }` work?

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -5982,6 +5982,589 @@ public static class Extensions
                 Diagnostic(ErrorCode.ERR_EscapeVariable, "local2").WithArguments("local2").WithLocation(11, 18));
         }
 
+        [Fact]
+        public void Deconstruction_UnscopedRef()
```
jcouv, line 5986: nit: it's a bit difficult to check that every scenario with `Deconstruct` invocation has an equivalent deconstruction-assignment scenario. Consider merging those pairs. 
For example, in `Deconstruction_UnscopedRef` this would mean adding:
```
                    R M2()
                    {
                        (var x, _) = new S();
                        return x;
                    }
```

jcouv (approved):
LGTM Thanks (iteration 4)

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
```diff
@@ -594,6 +594,21 @@ static uint getDeclarationValEscape(BoundTypeExpression typeExpression, uint val
...
+                { IsExtensionMethod: true, Parameters: [{ EffectiveScope: ScopedKind.None }, ..] })
+            {
+                using (new PatternInput(this, _localScopeDepth))
+                {
+                    return base.VisitRecursivePattern(node);
+                }
+            }
```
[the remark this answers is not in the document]
jcouv: Looking through the spec some more, makes me wonder: what if the `Deconstruct` method was compiled with an older compiler? Then the receiver, which is treated as a `ref` parameter, would be considered unscoped by default.

Here's the relevant [section](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#refsafetyrulesattribute):
> To reduce the chance of breaking changes when recompiling with C#11, we will update the C#11 compiler to use the ref safe context rules for method invocation that match the rules that were used to analyze the method declaration. Essentially, when analyzing a call to a method compiled with an older compiler, the C#11 compiler will use C#7.2 ref safe context rules.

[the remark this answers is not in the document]
jjonescz: > Let me try to summarize my understanding. Does that match your understanding?

Yes, your summary matches my understanding.

> what if the `Deconstruct` method was compiled with an older compiler? Then the receiver, which is treated as a `ref` parameter, would be considered unscoped by default.

Good point. I think if I rewrite the condition in terms of EffectiveScope as you suggested in the other comment, and we take https://github.com/dotnet/roslyn/pull/75930, this scenario will work (because that PR makes EffectiveScope consider old/updated rules).

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
jjonescz: `ThisParameter` property looks like it could throw, so added a local function instead of the pattern.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
jcouv: Thanks for the change. It's much easier to follow now :-)

jcouv (commented):
Done with review pass (iteration 6) Let's wait for the [PR](https://github.com/dotnet/roslyn/pull/75930) related to EffectiveScope to go in and add a test for the intersection scenario that'll fix.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -5982,6 +5982,454 @@ public static class Extensions
...
+        public void Deconstruction_UnscopedRef_RefReturnableReceiver()
+        {
+            var source = """
+                using System.Diagnostics.CodeAnalysis;
+                class C
+                {
+                    R M1(ref S s)
+                    {
+                        s.Deconstruct(out var x, out _);
+                        return x; // 1
```
RikkiGibson, line 6179:
```suggestion
                        return x;
```

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -5982,6 +5982,454 @@ public static class Extensions
...
+                {
+                    R M1()
+                    {
+                        if (new S() is ({ Prop: (var x2, _) } x, _))
+                            return x;
+                        return default;
+                    }
+                    R M2()
+                    {
+                        if (new S() is ({ Prop: (var x2, _) } x, _))
```
RikkiGibson, line 6410: It wasn't clear to me why we are also testing this here, it seems the same as the `M1` case.

RikkiGibson (commented):
LGTM, just want to wait on answers to questions and upstream effects from other PR

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -5982,6 +5982,454 @@ public static class Extensions
...
+                        return default;
+                    }
+                }
+                struct S
+                {
+                    [UnscopedRef] public void Deconstruct(out R x, out int y) => throw null;
+                }
+                ref struct R;
+                """;
+            // Pattern matching emits a copy of the receiver `s`, so the error in M3 is expected.
```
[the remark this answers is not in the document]
jjonescz: Yes, `M3` is lowered to the following which would be a ref safety error as well at the Deconstruct call:

```cs
R M3(ref S s)
{
    S s2 = s;
    R x;
    int y;
    s2.Deconstruct(out x, out y);
    if (true)
    {
        return x;
    }
    return default(R);
}
```

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
jjonescz: It's not the same - `M1` returns `x` whereas `M2` returns `x2`

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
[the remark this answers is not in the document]
jcouv: Using `EffectiveScope` seems a good solution to abstract away from those differences.
Were you able to add a test for this old rules vs. new rules scenario?

jcouv (commented):
Done with review pass (iteration 7)

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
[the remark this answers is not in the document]
jjonescz: I'm waiting for the other PR to be merged before adding a test.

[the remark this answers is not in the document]
jjonescz: > what if the `Deconstruct` method was compiled with an older compiler? Then the receiver, which is treated as a `ref` parameter, would be considered unscoped by default.

Actually this does not seem to be true. It's hard to tell from the spec how it should be because the old rules have no concept of scoped/unscoped. Anyway, I added tests that verify the behavior is consistent between explicit calls and patterns, so it behaves as expected wrt this PR, I think.

jjonescz:
@jcouv for another look, thanks

jcouv (approved):
LGTM Thanks (iteration 9)

PR #75513, opened 2024-10-15
[release/dev17.12] Field-backed properties: [field:] should not be considered a backing field reference (#75461)

### LINKED ISSUE #75459: Synthesized backing field generated for non-auto property with `[field:]` attribute

cston:
*Port https://github.com/dotnet/roslyn/pull/75461 to `release/dev17.12`.*

An attribute with a `field:` target used on a *property accessor* was incorrectly considered a reference to the backing field, resulting in a synthesized field for the property.

Fixes #75459.

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbol.cs
@@ -47,6 +47,9 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                 out var getSyntax,
                 out var setSyntax);
 
+            Debug.Assert(!usesFieldKeyword ||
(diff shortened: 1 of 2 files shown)


RikkiGibson (approved):
This change is already approved in main, is there any sign-off needed besides from @jaredpar?

PR #75514, opened 2024-10-15
Create nuspec for servicehub r2r symbol packages

genlu:


src/Workspaces/Remote/ServiceHub.CoreComponents/CoreComponents.Shared.targets
```diff
@@ -111,9 +111,34 @@
...
+    <id>$(_SymbolPackageId)</id>
+    <version>$(PackageVersion)</version>
+    <authors>Microsoft</authors>
+    <license type="expression">MIT</license>
+    <licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
+    <projectUrl>https://github.com/dotnet/roslyn</projectUrl>
+    <description>Package Description
+      More details at https://aka.ms/roslyn-packages</description>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <tags>Roslyn CodeAnalysis Compiler CSharp VB VisualBasic Parser Scanner Lexer Emit CodeGeneration Metadata IL Compilation Scripting Syntax Semantics</tags>
```
tmat, line 131: Can remove the tags element

eng/Signing.props
```diff
@@ -17,8 +17,8 @@
   -->
   <ItemGroup>    
     <!-- Exclude two manually zipped servicehub symbol nupkgs from being signed -->
-    <ItemsToSign Remove="$(ArtifactsNonShippingPackagesDir)Microsoft.CodeAnalysis.Remote.ServiceHub.CoreComponents.x64.Symbols.nupkg"/>
-    <ItemsToSign Remove="$(ArtifactsNonShippingPackagesDir)Microsoft.CodeAnalysis.Remote.ServiceHub.CoreComponents.arm64.Symbols.nupkg"/>
+    <ItemsToSign Remove="$(ArtifactsNonShippingPackagesDir)Microsoft.CodeAnalysis.Remote.ServiceHub.CoreComponents.x64.Symbols.*.nupkg"/>
```
tmat, line 20: Shouldn't it be `{project-name}.{version}.Symbols.nupkg`?

genlu: I followed the current pattern, e.g.
`Microsoft.CodeAnalysis.CodeStyle.Fixes.Symbols.4.13.0-1.24511.9.nupkg`

RikkiGibson:
Do we want to move forward by just merging and seeing if official build is fixed?

RikkiGibson (approved):

eng/Signing.props
tmat: Arcade is looking for *.symbols.nupkg here:

https://github.com/dotnet/arcade/blob/ad64a76fd00a641116e79f77d48b42c6e7ed9b0a/src/Microsoft.DotNet.Arcade.Sdk/tools/Publish.proj#L98-L99

wonder if the other symbol package we produce is broken

tmat: Re `Microsoft.CodeAnalysis.CodeStyle.Fixes.Symbols.nupkg` -- are we building that? Why would code style fixes need a symbol package?

tmat: NuGet symbol package naming is documented here: https://learn.microsoft.com/en-us/nuget/create-packages/symbol-packages

tmat (approved):

eng/Signing.props
tmat: Let's fix the naming in a follow up PR.

PR #75522, opened 2024-10-15
Clarify the compiler's policy towards well-known types/members

jcouv:


DoctorKrolic:
I think you should also alter [this comment](https://github.com/dotnet/roslyn/blob/5771a500694f5eb192dc0e6cf8639321c443961c/src/Compilers/CSharp/Test/Emit/CodeGen/IndexAndRangeTests.cs#L3743-L3744) or remove that test entirely

AlekseyTs (approved):

PR #75525, opened 2024-10-16
Fix servicehub symbol package name

genlu:


eng/publish-assets.ps1
```diff
@@ -69,6 +69,11 @@ function Publish-Nuget($publishData, [string]$packageDir) {
         throw "$nupkg does not exist"
       }
 
+      if ($nupkg.EndsWith(".symbols.nupkg")) {
```
jaredpar, line 72: Can we add a comment on how they are published if they aren't published here?

genlu: tbh I'm still feeling perplexed on how our symbol publishing works. From the log, it seems we generate a bunch of {packageId}.Symbols.{version}.nupkg packages, then use `*.symbols.nupkg` to try to find them and failed to find any,  therefore we copy those {packageId}.Symbols.{version}.nupkg to {packageId}*.Symbols.{version}.symbols.nupkg and use them instead... @tmat and I decided to at least have the name of these two servicehub packages right and figure out what to do for others later.

PR #73393, opened 2024-05-08
Remove compile-time dependency from Workspace.MSBuild on Workspace.MSBuild.BuildHost

tmat:
Also removes usage of Microsoft.Extensions.Logging from the BuildHost. We just need to write text to Console.Error.

Removes `Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.exe` from `Roslyn.VisualStudio.Setup.vsix`.

Microsoft.CodeAnalysis.Workspaces.MSBuild.nupkg content diff:

core:
image

netfx:
image

Reduces C# extension size by 49 files (3MB uncompressed).

PR #75528, opened 2024-10-16
Revert "Update BuildHost STJ to 8.0.4"

tmat:
This reverts commit 6f65bfca9ce7afc4b3a081b8f7406d046301d27e.

System.Text.Json.dll was eliminated from BuildHost by https://github.com/dotnet/roslyn/pull/73393 but then re-added back by flow from internal branch.

### DIFF
--- a/src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
+++ b/src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
@@ -30,7 +30,6 @@
     <PackageReference Include="Microsoft.Build.Tasks.Core" VersionOverride="17.3.4" ExcludeAssets="Runtime" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" VersionOverride="17.3.4" ExcludeAssets="Runtime" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Build.Locator" PrivateAssets="All" />
-    <PackageReference Include="System.Text.Json" VersionOverride="8.0.4" PrivateAssets="All" Condition="'$(DotNetBuildSourceOnly)' != 'true'"/>
     <PackageReference Include="System.CommandLine" />
     <PackageReference Include="System.Collections.Immutable" />
     <PackageReference Include="Newtonsoft.Json" />


arkalyanms (approved):

arkalyanms:
@jaredpar and @jasonmalinowski fyi

ericstj:
It looks like `Microsoft.Build` brings it in, then CPM will update it.
```json
      "Microsoft.Build/17.3.4": {
        "type": "package",
        "dependencies": {
          "Microsoft.Build.Framework": "17.3.4",
          "Microsoft.IO.Redist": "6.0.0",
          "Microsoft.NET.StringTools": "17.3.4",
          "System.Collections.Immutable": "6.0.0",
          "System.Configuration.ConfigurationManager": "6.0.0",
          "System.Memory": "4.5.5",
          "System.Reflection.MetadataLoadContext": "6.0.0",
          "System.Text.Json": "6.0.0",
          "System.Threading.Tasks.Dataflow": "6.0.0"
        },
```

ericstj:
The MSBuild reference is `ExcludeAssets="Runtime"` which means this fix did remove the dll from the net472 output which is good, but the package is still mentioned in the deps file, which we've been trying to avoid.
```json

      "System.Text.Json/8.0.4": {
        "dependencies": {
          "System.Runtime.CompilerServices.Unsafe": "6.0.0",
          "System.Text.Encodings.Web": "8.0.0"
        }
      },
```

marcpopMSFT:
@ericstj do we need roslyn to update the STJ in the directory.packages.props now to resolve the deps.json entry coming from MSBuild? Do we need a 17.12 msbuild first to be able to fix this (which we can't do until we have VS updated?)?

ericstj:
You could either:
1. Update STJ to 8.0.5 only for `net6.0` and exclude it -- and sign up to keep updating it every time there's an update in 8.0.  👎 (means ongoing servicing coordination between 8 and 9 where we waterfall 8.0 changes through roslyn then back into 9.0)
2. Make the SDK responsible for rewriting the deps file like it does for other tools.  That would erase the mention of the stale dependencies.
3. Use a different mechanism for excluding MSBuild's package dependency closure.  I have a slightly hacky way for doing this -- https://github.com/dotnet/runtime/blob/43813ac73242fa78c463d456bf755e3a6622b5d7/eng/PackageDownloadAndReference.targets#L5

tmat:
@ericstj 
Re [2] - can you point me to the targets in SDK that do so for other tools?

ericstj:
I pulled down a binlog and found it's actually this project that's doing it: https://github.com/dotnet/sdk/blob/ea38ebafb21205f84fe0536a3255c030afaf2f79/src/Layout/redist/redist.csproj#L26

It looks like it actually produces a deps file for the entire SDK "folder" (which will reference the latest packages) then it removes instances of the tools from that deps file, then it copies that deps file to each tool.deps.json instead of using the one provided by the tool:
https://github.com/dotnet/sdk/blob/ce0abad026fcae9d670952aed61c915a1028355a/src/Layout/redist/targets/GenerateLayout.targets#L231-L252

You can see [here](https://dev.azure.com/dnceng-public/_apis/resources/Containers/43307004/TestBuild_windows_x64?itemPath=TestBuild_windows_x64%2Flog%2FRelease%2FBuild.binlog) -- search for $target GenerateCliRuntimeConfigurationFiles

tmat:
Oh, I think the redist stuff is just for setting up dogfooding env and testing. I don't think these are for production.

ericstj:
Also -- not sure if you caught the distinction -- the SDK doesn't remove libraries - it just removes mention of some files.  The way it's actually making "live version" deps files for tools is by generating an entirely new deps file that's produced by referencing a superset of packages and then replacing the tool's deps file with that one renamed.

PR #75548, opened 2024-10-17
ClrOnly condition is not working on mono

giritrivedi:
This test case tests the Clr behaviour Only. ConditionalFact(typeof(ClrOnly) should skip the execution of the test case if it is executed on Mono. However, the condition doesn't hold good. 
The fix takes care of skipping the execution tests on Mono which are meant for ClrOnly.


giritrivedi:
r/azp

giritrivedi:
@jjonescz  Can you please restart azure pipeline. This was failing due to dependency with previous commit.

giritrivedi:
\cc @uweigand

src/Compilers/CSharp/Test/Emit/Emit/DeterministicTests.cs
```diff
@@ -373,7 +373,7 @@ static ImmutableArray<INamedTypeSymbol> getForwardedTypes(AssemblySymbol assembl
             }
         }
 
-        [ConditionalFact(typeof(ClrOnly), Reason = "Static execution is runtime defined and this tests Clr behavior only")]
+        [ConditionalFact(typeof(NotOnMonoCore), Reason = "Static execution is runtime defined and this tests Clr behavior only")]
```
jjonescz, line 376: With `NotOnMonoCore`, the test would get executed on Mono Desktop, right? But it presumably wouldn't work there since it used to be marked `ClrOnly`. Should we change it to `CoreClrOnly` instead?

jjonescz: Oh, right, is that because `ExecutionConditionUtil.IsCoreClr` is `true` when running on mono core?

Well, still, we want to skip this test on both MonoCore and MonoDesktop, don't we? Perhaps we should have a `NotOnAnyMono` attribute?

giritrivedi: Yes you are right . IsCoreClr is just a negation of IsDesktop which just check "RuntimeUtilities.IsDesktopRuntime" which is true when running on mono core. 
yes, We should have NotOnAnyMono attribute. Will make these changes .

src/Compilers/Test/Core/Assert/ConditionalFactAttribute.cs
```diff
@@ -341,6 +342,12 @@ public class NotOnMonoCore : ExecutionCondition
         public override string SkipReason => "Test not supported on Mono Core";
     }
 
+    public class NotOnAnyMono: ExecutionCondition
```
jjonescz, line 345:
```suggestion
    public class NotOnAnyMono : ExecutionCondition
```

jjonescz (approved):

giritrivedi:
Hi @jjonescz , Can this be merged ?

jjonescz:
Thanks @giritrivedi for the contribution

PR #75550, opened 2024-10-17
Add ETW events when ALCs are created or disposed

RikkiGibson:
Trying to get more data about increase in allocs from #75233.
@jaredpar @CyrusNajmabadi 

Appears to be working when compiling Microsoft.CodeAnalysis thru PerfView. I am assuming that merely compiling does not actually go thru the teardown process, so the lack of Stop events is not a surprise.
image

### DIFF
--- a/src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.Core.cs
+++ b/src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.Core.cs
@@ -74,6 +74,7 @@ namespace Microsoft.CodeAnalysis
             {
                 if (!_loadContextByDirectory.TryGetValue(fullDirectoryPath, out loadContext))
                 {
+                    CodeAnalysisEventSource.Log.CreateAssemblyLoadContext(fullDirectoryPath);
                     loadContext = new DirectoryLoadContext(fullDirectoryPath, this);
                     _loadContextByDirectory[fullDirectoryPath] = loadContext;
                 }
@@ -110,6 +111,7 @@ namespace Microsoft.CodeAnalysis
             {
                 try
                 {
+                    CodeAnalysisEventSource.Log.DisposeAssemblyLoadContext(context.Directory);
                     context.Unload();
                 }
                 catch (Exception ex) when (FatalError.ReportAndCatch(ex, ErrorSeverity.Critical))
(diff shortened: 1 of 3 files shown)


src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerAssemblyLoader.Core.cs
```diff
@@ -110,6 +111,7 @@ private partial void DisposeWorker()
             {
                 try
                 {
+                    CodeAnalysisEventSource.Log.DisposeAssemblyLoadContext(context.Directory);
```
chsienki, line 114: Consider putting this in a finally with a status, or after the actual call to `.Unload` so we don't end up with potentially misleading events.

chsienki (approved):

CyrusNajmabadi (approved):

RikkiGibson:
I believe I am not causing the integration test failures, as different tests are failing in each attempt on different issues: once a lightbulb being unexpectedly dismissed, another an NRE in TestTraceListener.
https://dev.azure.com/dnceng-public/public/_build/results?buildId=846617&view=ms.vss-test-web.build-test-results-tab

src/Workspaces/Core/Portable/Microsoft.CodeAnalysis.Workspaces.csproj
```diff
@@ -40,6 +40,7 @@
     <Compile Include="..\..\..\Compilers\Core\Portable\DiagnosticAnalyzer\DefaultAnalyzerAssemblyLoader.cs" Link="Diagnostics\CompilerShared\DefaultAnalyzerAssemblyLoader.cs" />
     <Compile Include="..\..\..\Compilers\Core\Portable\DiagnosticAnalyzer\ShadowCopyAnalyzerAssemblyLoader.cs" Link="Diagnostics\CompilerShared\ShadowCopyAnalyzerAssemblyLoader.cs" />
     <Compile Include="..\..\..\Compilers\Core\Portable\DiagnosticAnalyzer\IAnalyzerAssemblyResolver.cs" Link="Diagnostics\CompilerShared\IAnalyzerAssemblyResolver.cs" />
+    <Compile Include="..\..\..\Compilers\Core\Portable\CodeAnalysisEventSource.cs" Link="Diagnostics\CompilerShared\CodeAnalysisEventSource.cs" />
```
RikkiGibson, line 43: I suspect that this source-include of the CodeAnalysisEventSource is causing regression https://dev.azure.com/devdiv/DevDiv/_workitems/edit/2289777

PR #75561, opened 2024-10-18
Use PackageDownloadAndReference to avoid including msbuild dependencies

tmat:


tmat:
@ericstj ptal

eng/targets/PackageDownloadAndReference.targets
```diff
@@ -0,0 +1,35 @@
...
+
+  <!--
+  Based on https://github.com/dotnet/runtime/blob/9893d145ebb61ae592afa1a64baa2f6f094b1f6b/eng/PackageDownloadAndReference.targets
+  
+  These file supports using PackageDownloadAndReference items.
+
+  The PackageDownloadAndReference item is used to download a package and reference it in the project, without restoring the package's dependency closure.
+
+  When using PackageDownloadAndReference you are responsible for selecting the correct assets from the package and ensuring that the package and it's
+  dependencies are available at runtime.
```
jasonmalinowski, line 11: Any chance a feature for this is on NuGet's backlog?

jasonmalinowski (approved):
Feels like we're missing a feature in NuGet or the SDK somewhere for this...it'll work though!

eng/targets/PackageDownloadAndReference.targets
ericstj: There's https://github.com/NuGet/Home/issues/13821 as a very specific case (which also applies here), but we've also been discussing a more general story for plugins / excludeassets=runtime case.  In that case what you really want is reference assemblies for API provided by the host, then package-knockout for any package that overlaps.  It's on the list of things we want to solve for 10.0 - we think it's essential for making package updates work.

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
```diff
@@ -23,16 +23,20 @@
...
+    <PackageDownloadAndReference Include="Microsoft.Build" Version="17.3.4" Folder="ref/$(TargetFramework)" />
+    <PackageDownloadAndReference Include="Microsoft.Build.Framework" Version="17.3.4" Folder="ref/$(TargetFramework)" />
+    <PackageDownloadAndReference Include="Microsoft.Build.Tasks.Core" Version="17.3.4" Folder="ref/$(TargetFramework)" />
+    <PackageDownloadAndReference Include="Microsoft.Build.Utilities.Core" Version="17.3.4" Folder="ref/$(TargetFramework)" />
+    
     <PackageReference Include="Microsoft.Build.Locator" PrivateAssets="All" />
-    <PackageReference Include="System.Text.Json" VersionOverride="8.0.4" PrivateAssets="All" Condition="'$(DotNetBuildSourceOnly)' != 'true'"/>
     <PackageReference Include="System.CommandLine" />
     <PackageReference Include="System.Collections.Immutable" />
+    <PackageReference Include="System.Reflection.Metadata" />
```
ericstj, line 39: Seems this previously came in transitively from MSBuild which was marked with `ExcludeAssets="Runtime"`.  Will this add a copy to your package where it wasn't before?

ericstj (approved):
Please double check the package content before and after this change.

One trick I use is 
```
build -pack /p:NuSpecOutputPath=artifacts\specs\before
git checkout myBranch
build -pack /p:NuSpecOutputPath=artifacts\specs\after
```
Then I diff the folders.

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
tmat: Yes, it does add it. It seems odd it wasn't there before since we have System.Collections.Immutable in the package as well.

tmat:
Source build isn't happy :(

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
tmat: I see. We don't actually end up using SRM types directly in the build host. SRM is just needed to compile some util code that is actually unused.

tmat:
> Please double check the package content before and after this change.

Yup, I did verify it's the same content.

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
```diff
@@ -21,16 +21,28 @@
...
     <PlatformTarget>AnyCPU</PlatformTarget>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+    <!--
+      We specifically reference Microsoft.Build 17.3.4, since it's the oldest version still compatible $(NetRoslynBuildHostNetCoreVersion).
+      The version targeted is only used for build time, since we use MSBuildLocator to discover the proper version at runtime.
+    -->
+    <_MsbuildVersion>17.3.4</_MsbuildVersion>
   </PropertyGroup>
+  <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
```
ericstj, line 31: There's no reason why you need `PackageReference` for sourcebuild.  It should work just the same there.

ericstj: I see what happened, in SBRP the `net472` asset is missing: https://github.com/dotnet/source-build-reference-packages/tree/main/src/referencePackages/src/microsoft.build.framework/17.3.4/ref

tmat: Would you mind filing an issue? We can remove the condition here once fixed.

eng/targets/PackageDownloadAndReference.targets
```diff
@@ -0,0 +1,35 @@
...
+    - Folder: The folder in the package where the assembly is located.
+    - AssemblyName: The name of the assembly to reference.
+    - Private: Whether the reference should be private (copied to the output directory) or not. Default is false.
+
+  A common use case for PackageDownloadAndReference is to reference assemblies provided by MSBuild or the .NET SDK.
+  -->
+
+  <ItemDefinitionGroup>
+    <PackageDownloadAndReference>
+      <Folder>lib/$(TargetFramework)</Folder>
```
jaredpar, line 23: I assume this is guaranteed to be lower case so no need for a defensive `ToLowerInvariant()` call.

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
```diff
@@ -39,6 +51,11 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\..\..\Compilers\Core\Portable\InternalUtilities\*.cs" Link="InternalUtilities\%(FileName).cs" />
+
+    <!-- Exclude utilities that use System.Reflection.Metadata -->
+    <Compile Remove="..\..\..\Compilers\Core\Portable\InternalUtilities\BlobBuildingStream.cs" />
+    <Compile Remove="..\..\..\Compilers\Core\Portable\InternalUtilities\IncrementalHashExtensions.cs" />
```
jaredpar, line 57: Why did we need to do these removes?

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
```diff
@@ -94,4 +111,5 @@
   </Target>
   <Import Project="..\..\..\Dependencies\PooledObjects\Microsoft.CodeAnalysis.PooledObjects.projitems" Label="Shared" />
   <Import Project="..\..\..\Dependencies\Collections\Microsoft.CodeAnalysis.Collections.projitems" Label="Shared" />
+  <Import Project="$(RepositoryEngineeringDir)targets\PackageDownloadAndReference.targets"/>
```
jaredpar, line 114: Is there any reason we couldn't put this is the core `Imports.targets`? Imagine we will need to use this in other placse at some point. If it's just to limit impact of change that's fine.

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
tmat: We would need to add reference to SRM, which is otherwise not needed. These helpers would bring it in but they are not used by the build host.

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
tmat: I'd do that once we need it elsewhere. Ideally, we'll switch to the new msbuild/nuget feature that implements the real solution and remove this workaround.

eng/targets/PackageDownloadAndReference.targets
tmat: We set the TFM, so I'd hope we use lower case.

src/Workspaces/Core/MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj
ericstj: It would be OK to `PackageDownloadAndReference`  the SRM provided by MSBuild, but no need to do so if you don't actually need the dependency.

ericstj (approved):

jaredpar:
Remaining test failure was flaky, merged over it so we can get .NET 9 SDK moving.

PR #75566, opened 2024-10-20
Field-backed properties: recognize field in interpolated string

cston:
Parse `field` as a keyword when used in an interpolated string in an accessor.
 
See https://github.com/dotnet/csharplang/issues/140#issuecomment-2425018091

### DIFF
--- a/src/Compilers/CSharp/Portable/Parser/LanguageParser_InterpolatedString.cs
+++ b/src/Compilers/CSharp/Portable/Parser/LanguageParser_InterpolatedString.cs
@@ -109,7 +109,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Syntax.InternalSyntax
                         originalTextSpan[currentContentStart..interpolation.OpenBraceRange.Start]));
 
                     // Now parse the interpolation itself.
-                    var interpolationNode = ParseInterpolation(this.Options, originalText, interpolation, kind);
+                    var interpolationNode = ParseInterpolation(this.Options, originalText, interpolation, kind, IsInFieldKeywordContext);
 
                     // Make sure the interpolation starts at the right location.
                     var indentationError = getInterpolationIndentationError(indentationWhitespace, interpolation);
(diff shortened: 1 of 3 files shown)


333fred (commented):
Needs tests for field used in the alignment or format specifier sections of an interpolation. In particular, what will happen when there's a constant named field and field is used in the alignment section.

333fred (approved):

src/Compilers/CSharp/Portable/Parser/LanguageParser_InterpolatedString.cs
```diff
@@ -377,6 +378,7 @@ private static InterpolationSyntax ParseInterpolation(
 
             // Now create a parser to actually handle the expression portion of the interpolation
             using var tempParser = new LanguageParser(tempLexer, oldTree: null, changes: null);
+            using var __ = new FieldKeywordContext(tempParser, isInFieldKeywordContext);
```
jaredpar, line 381: Why `__` instead of simply `_`?

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -10171,6 +10171,297 @@ partial class B
...
+                    // (5,61): error CS0103: The name 'field' does not exist in the current context
+                    //     public object P2 { get; set { field = value; field = $"{field}"; } }
+                    Diagnostic(ErrorCode.ERR_NameNotInContext, "field").WithArguments("field").WithLocation(5, 61));
+            }
+            else
+            {
+                var verifier = CompileAndVerify(comp, verify: Verification.Skipped, expectedOutput: IncludeExpectedOutput("""
+                    P1: True
+                    2
+                    """));
```
jjonescz, line 10227: Consider calling `verifier.VerifyDiagnostics`. Applies to the tests below as well.

jjonescz (approved):

Issue #75583, opened 2024-10-22
Assert in SyntaxNode.ContainsDirective


ToddGrun:
**Version Used**: 17.13 Preview 1

**Steps to Reproduce**:

1. Open VS with debug Roslyn installed
2. Open LanguageParser.cs in Roslyn.sln
3. Navigate to inside the IsVarType method
4. type "if (#if)" on an empty line -- Assert hit indicating either invalid parse tree or assumption about parse tree structure

![Image](https://github.com/user-attachments/assets/c3d0a6e6-84a1-4017-aaa7-dc4fdb92d279)

PR #75595, opened 2024-10-23
Handle directives in trailing trivia

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75583.


AlekseyTs:

src/Compilers/Core/Portable/Syntax/SyntaxNode.cs
```diff
@@ -479,7 +479,16 @@ public bool ContainsDirective(int rawKind)
                             Debug.Assert(!triviaContainsMatch(current.GetLeadingTriviaCore(), rawKind), "Should not have a match if the token doesn't even have leading trivia");
                         }
 
-                        Debug.Assert(!triviaContainsMatch(current.GetTrailingTriviaCore(), rawKind), "Should never have a match in trailing trivia");
+                        // no need to look within if this token doesn't even have trailing trivia.
+                        if (current.HasTrailingTrivia)
+                        {
+                            if (triviaContainsMatch(current.GetTrailingTriviaCore(), rawKind))
```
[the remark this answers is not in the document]
jjonescz: > In the problematic scenario, `#if` is in the middle of the line. Perhaps the parser is not supposed to treat it as a directive?

Sound good, I haven't realized that's an invariant. Thanks.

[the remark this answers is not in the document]
AlekseyTs: > I haven't realized that's an invariant. 

I am not stating this is a fact, but the nature of the assert makes me think that is probably what is going on.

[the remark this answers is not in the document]
CyrusNajmabadi: Yes.

> If that is correct, it might be reasonable to assume that a directive can never be in a trailing trivia. In the problematic scenario, #if is in the middle of the line. Perhaps the parser is not supposed to treat it as a directive?

I strongly align with this approach. 

I believe we intended to have the invariant that directives were only leading trivia.  As such, if they do appear at the end, while it is fine to be *resilient* to seeing it (And give good messages) we shoudl still represent the problematic code in some other way (like simply making it into 'skipped token trivia').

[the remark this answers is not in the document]
CyrusNajmabadi: It looks like we have this:

```c#
                    case '#':
                        if (_allowPreprocessorDirectives)
                        {
                            this.LexDirectiveAndExcludedTrivia(afterFirstToken, isTrailing || !onlyWhitespaceOnLine, ref triviaList);
                            break;
                        }
                        else
                        {
                            return;
                        }
```

We can either say that we don't go into LexDirectiveAndExcludedTrivia if isTrailing is false.  Or we go in, get the result, but then convert somehow into a skipped 'bad token' that acts as the trailing trivia of hte current token.  I'm fine with either approach.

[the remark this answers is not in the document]
jjonescz: I've tried this approach but it seems some parts of Roslyn assume there are directives in trailing trivia, for example this test stops working:

https://github.com/dotnet/roslyn/blob/09a7b388974b48af9ef28a539de10cb463014e15/src/Workspaces/CSharpTest/Formatting/FormattingTests.cs#L5757-L5773

I'm not sure if we should continue with this approach and try to make it work, or we could say that directives can be in trailing trivia in error scenarios - and make `SyntaxNode.ContainsDirective` API ignore those (so we would just remove the assert).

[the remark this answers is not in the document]
CyrusNajmabadi: It's fine with me for that test to stop working.

src/Compilers/CSharp/Portable/Parser/DirectiveParser.cs
```diff
@@ -32,7 +32,8 @@ public CSharpSyntaxNode ParseDirective(
             bool isActive,
             bool endIsActive,
             bool isAfterFirstTokenInFile,
-            bool isAfterNonWhitespaceOnLine)
+            bool isAfterNonWhitespaceOnLine,
+            bool trailing)
```
AlekseyTs, line 36: >trailing [](http://example.com/codeflow?start=17&length=8)

Do we really need this additional information that looks redundant? Isn't ```isAfterNonWhitespaceOnLine``` sufficient?

src/EditorFeatures/CSharpTest/QuickInfo/DiagnosticAnalyzerQuickInfoSourceTests.cs
```diff
@@ -32,9 +32,11 @@ public async Task ErrorTitleIsShownOnDisablePragma()
     {
         await TestInMethodAsync(
             """
+
             #pragma warning disable CS0219$$
```
AlekseyTs, line 36: It doesn't look like there are meaningful changes in this file. Consider excluding it from the PR

AlekseyTs:

CyrusNajmabadi:
It looks like we're still ending up with a directive in trailing trivia. Is that correct?

src/EditorFeatures/CSharpTest/QuickInfo/DiagnosticAnalyzerQuickInfoSourceTests.cs
jjonescz: These are needed. `TestInMethod` injects the code into a method but without newlines between the opening/closing braces of the method, so the `#pragma` was in trailing position and the test stopped working.

jjonescz:
> It looks like we're still ending up with a directive in trailing trivia. Is that correct?

Yes, I chose to parse these trailing directives as BadDirectiveTrivia. Is that a problem?

I first tried to attach a skipped trivia onto a hash token and put that into the trailing trivia. But that failed some checks because hash token is not a trivia. Which makes sense. BadDirectiveTrivia is a trivia, so I used that instead. But if you have other syntax tree shape in mind for these scenarios, I'm open to changing this.

AlekseyTs:
> But that failed some checks because hash token is not a trivia.

With this approach the hash token should be a skipped trivia as well, i.e. not a token

CyrusNajmabadi:
> Yes, I chose to parse these trailing directives as BadDirectiveTrivia. Is that a problem?

My *preference* (but I could be convinced otherwise) is that we never have any sort of directive trivia as trailing trivia. 

That is then very easy to remember and have invariants around. 

If you guys feel otherwise, I'm happy to discuss and compromise on that. However, given that it's never legal for the language, and adds complexity in all the upper layers to consider this case, my preference instead is to make these to skipped tokens. 

If you do think allowing BadDirective as trailing is preferable to skipped tokens, I'm ok with that. 

Lmk what you all think! :-)

src/Compilers/CSharp/Test/Syntax/Syntax/SyntaxNodeTests.cs
```diff
@@ -469,20 +469,55 @@ class D
...
+                // (1,9): error CS1026: ) expected
+                // if (#if)
+                TestBase.Diagnostic(ErrorCode.ERR_CloseParenExpected, "").WithLocation(1, 9),
+                // (1,9): error CS1733: Expected expression
+                // if (#if)
+                TestBase.Diagnostic(ErrorCode.ERR_ExpressionExpected, "").WithLocation(1, 9),
+                // (1,9): error CS1002: ; expected
+                // if (#if)
+                TestBase.Diagnostic(ErrorCode.ERR_SemicolonExpected, "").WithLocation(1, 9),
+                // (1,9): error CS1027: #endif directive expected
```
AlekseyTs, line 514: > // (1,9): error CS1027: #endif directive expected [](http://example.com/codeflow?start=15&length=50)

This error feels unexpected. There are similar unexpected  changes to other tests in commit 8.

AlekseyTs:

src/Compilers/CSharp/Portable/Parser/Lexer.cs
```diff
@@ -1980,7 +1980,31 @@ private void LexSyntaxTrivia(bool afterFirstToken, bool isTrailing, ref SyntaxLi
                     case '#':
                         if (_allowPreprocessorDirectives)
                         {
-                            this.LexDirectiveAndExcludedTrivia(afterFirstToken, isTrailing || !onlyWhitespaceOnLine, ref triviaList);
+                            if (isTrailing || !onlyWhitespaceOnLine)
+                            {
+                                // Directives cannot be in trailing trivia - parse as skipped tokens trivia instead.
+
+                                var directiveBuilder = new SyntaxListBuilder(10);
+                                this.LexDirectiveAndExcludedTrivia(afterFirstToken, afterNonWhitespaceOnLine: true, ref directiveBuilder);
```
AlekseyTs, line 1988: >this.LexDirectiveAndExcludedTrivia(afterFirstToken, afterNonWhitespaceOnLine: true, ref directiveBuilder); [](http://example.com/codeflow?start=32&length=106)

Is it possible that this process is going to accumulate some side information which we might not want to accumulate?

jjonescz: It accumulates diagnostics attached to the tokens, but those seem useful, I think we *want* to accumulate those.

src/Compilers/CSharp/Test/Syntax/Syntax/SyntaxNodeTests.cs
jjonescz: This is the same set of diagnostics reported today.

Before commit 8, we missed some diagnostics because we parsed the directive as BadDirective. In commit 8, we parse it again as normal `#if` directive, just put it into skipped tokens afterwards, but the diagnostics are the same.

AlekseyTs: > In commit 8, we parse it again as normal `#if `directive, just put it into skipped tokens afterwards

Calling the parse helper is an implementation detail. The tree reflects how we parsed the directive and we don't have the directive in the tree. Effectively, that means that the tokens were simply skipped (the mechanism used to skip them is not important), not treated as a directive. Therefore, it is unexpected and undesired (in my opinion) to produce any diagnostics that might imply otherwise.

> but the diagnostics are the same.

We change parsing result, diagnostics should change accordingly, I think.

src/Compilers/CSharp/Portable/Parser/Lexer.cs
AlekseyTs: > It accumulates diagnostics attached to the tokens

It looks like we somehow get diagnostics about missing end constructs for directives that we ignore here. Is that diagnostic reported here? Otherwise, this method has side effects beyond just tokens that it returns. I do no think we want those side effects.

AlekseyTs: I think it would be fine to report a regular unexpected token error, skip just the `#` token, and continue parsing as if it wasn't there. What we are trying to do right now feels unnecessary complicated to me.

jjonescz: > I think it would be fine to report a regular unexpected token error, skip just the `#` token, and continue parsing as if it wasn't there. What we are trying to do right now feels unnecessary complicated to me.

Sounds good, I will do that, thanks.

jjonescz: > I think it would be fine to report a regular unexpected token error, skip just the `#` token, and continue parsing as if it wasn't there. What we are trying to do right now feels unnecessary complicated to me.

@AlekseyTs I have implemented this. The disadvantage of this approach is that it makes error recovery considerably worse.

To summarize the options:
- Skip `#` token, continue parsing as if it wasn't there. Worse error recovery than before.
- Parse the trailing directives into skipped tokens. Should we parse the directive as normal, including disabled text, error reporting, then just convert the tokens to skipped tokens? If not, how should we parse?
- Parse as BadDirective. A bit worse error recovery because we parse as a generic directive. Technically, it's still a directive in trailing trivia (so doesn't hold the invariant).
- Allow directives in trailing trivia in error scenarios. Very simple implementation, just remove the assert and everything continues working (error recovery, IDE refactoring features). We still have the invariant that there are no directives in trailing trivia in success scenarios. (Currently, this seems as the best option to me.)

CyrusNajmabadi: You can still parse it like before, but represent in the tree however you want.  This is very common in the parser. 

My main preference is simply that we do not have this structured trivia as trailing trivia.

CyrusNajmabadi: > I don't know how I would use it here directly because I don't have any SyntaxToken to attach the SkippedTokensTrivia onto.

Aren't you producing the token?  So you would attach to the token that you're just about to make :)

> If it's so common

By common, i mean the concept of: have the parser take steps to determine what is going on, but then determine how best to fit that into the syntactic model.

In other words, just because you parsed out a directive, doesn't mean you need to place it in the tree as a directive.

> Also Aleksey https://github.com/dotnet/roslyn/pull/75595#discussion_r1822535828 the fact we are forwarding the diagnostics from parsing the directive onto the skipped tokens

Yeah.  I don't have strong feelings/thoughts about that.  My general point is simply that i would prefer teh *final syntactic model* not have directives within trailing trivia.

--

Note: it might be easier for us to chat.  But i'll try to share my thoughts clearly here.  If you need more clarity on my views, please feel free to ask :)

--

The lexer/parser is effectively a complex program that can execute arbitrary logic to try to decide what the user code means and how it should represent it.  So, to me, it's fine for it to do things like actually parse out directives wherever it finds them, to have good error recovery.  However, that execution logic is orthogonal to the decisions it then makes in terms of final syntactic tree representation (including diagnostics).

So, *personally*, i think it would be fine to:

1. parse out as a directive, for the best recovery experience.
2. take the results of that parsing, and re-represent them as a skipped token when placing in the tree.  Note: for this i would make this extremely simple.  Literally take the entire contents of the parsed directive, and just make a single '.BadToken' whose contents are teh entirety of the text of the entire directive.

The end representation then never has a directive in trailing trivia.  Nor does it expose that we produced special directive diagnostics.  All teh user sees is a message saying `#` is not expected there.  And all the tree has is a skipped token representing that entire chunk that we don't feel is valid there.

Issue #75560, opened 2024-10-18
`!` does not suppress nullable warning from spread element type


cston:
**Compile:**
```csharp
#nullable enable
using System.Collections.Generic;

class Program
{
    static void Main()
    {
        IEnumerable<object>? x = null;
        IEnumerable<object?> y = [null];
        object[] z;
        z = [..x!];
        z = [..y!]; // warning CS8601: Possible null reference assignment.
    }
}
```

**Expected Behavior**:
No warnings.

**Actual Behavior**:
Warning for `..y!`.
```
(12,16): warning CS8601: Possible null reference assignment.
```

See https://github.com/dotnet/roslyn/pull/74686#discussion_r1806800774

PR #75596, opened 2024-10-23
Propagate spread suppression

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75560.

### DIFF
--- a/src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
+++ b/src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
@@ -14184,14 +14184,19 @@ namespace System
             var source = """
                 #nullable enable
                 using System.Collections.Generic;
-                IEnumerable<object?> a = [null];
-                object[] b = [..a];
-                object[] c = [..a!];
+                {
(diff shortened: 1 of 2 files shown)


cston (approved):

src/Compilers/CSharp/Test/Emit3/Semantics/CollectionExpressionTests.cs
```diff
@@ -14178,6 +14178,56 @@ public void Add(T t) { }
...
 
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75560")]
+        public void Nullable_Spread_01()
+        {
+            var source = """
+                #nullable enable
+                using System.Collections.Generic;
+                IEnumerable<object?> a = [null];
+                object[] b = [..a];
+                object[] c = [..a!];
```
AlekseyTs, line 14189: >a [](http://example.com/codeflow?start=32&length=1)

I would strengthen the tests and used a different local here in order to avoid a chance that the error was not reported here because it was reported for the same local on the previous line.

AlekseyTs (approved):
LGTM (commit 1)

PR #75613, opened 2024-10-23
Use distinct EventSource names for compiler and workspace layer events

RikkiGibson:
Related VS regression bug: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2289777

We think this issue may be causing CodeAnalysis events to fail to appear in IDE traces.

VS PR to enable new event source in RPS/Speedometer: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/587560

Alternatively, we could expose the CodeAnalysisEventSource as a public experimental API.

### DIFF
--- a/src/Compilers/Core/Portable/CodeAnalysisEventSource.Compiler.cs
+++ b/src/Compilers/Core/Portable/CodeAnalysisEventSource.Compiler.cs
@@ -1,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.CodeAnalysis
+{
+    [EventSource(Name = "Microsoft-CodeAnalysis-General")]
+    internal sealed partial class CodeAnalysisEventSource
+    {
+        public static readonly CodeAnalysisEventSource Log = new CodeAnalysisEventSource();
+    }
+}
--- a/src/Workspaces/Core/Portable/Diagnostics/CodeAnalysisEventSource.Workspaces.cs
+++ b/src/Workspaces/Core/Portable/Diagnostics/CodeAnalysisEventSource.Workspaces.cs
@@ -1,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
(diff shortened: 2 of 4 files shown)


src/Compilers/Core/Portable/CodeAnalysisEventSource.Compiler.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.CodeAnalysis
+{
+    [EventSource(Name = "Microsoft-CodeAnalysis-General")]
+    internal sealed partial class CodeAnalysisEventSource
+    {
+        public static readonly CodeAnalysisEventSource Log = new CodeAnalysisEventSource();
```
RikkiGibson, line 13: The goal of putting this member here is to make it so you can't log events at all if you forget to declare a partial which specifies the event source name.

src/Workspaces/Core/Portable/Diagnostics/CodeAnalysisEventSource.Workspaces.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.CodeAnalysis
+{
+    [EventSource(Name = "Microsoft-CodeAnalysis-Workspaces")]
```
333fred, line 10: Are we going to need to ask the RPS folks to turn this collector on?

RikkiGibson: I am expecting the following PR to enable this collector: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/587560

333fred (approved):

jjonescz (approved):

jaredpar (approved):

RikkiGibson:
> Pretty sure that we have documentation on how to collect traces for the compiler / IDE. Should we add both of the event source names into that document to make it easier for customers to give us traces?

Besides in the original attribute, I only found the provider name mentioned in `Tools/Replay/README.md`: https://github.com/search?q=repo%3Adotnet%2Froslyn%20microsoft-codeanalysis-general&type=code

Since Replay is only handling a compilation scenario, it feels reasonable to leave that doc as is.

CyrusNajmabadi (approved):

PR #75615, opened 2024-10-24
Allow the client to update the server logging level dynamically

dibarbet:
Client side change - https://github.com/dotnet/vscode-csharp/pull/7688/

Allows us to update the log level dynamically as the server is running.

### DIFF
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Program.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Program.cs
@@ -44,16 +44,18 @@ static async Task RunAsync(ServerConfiguration serverConfiguration, Cancellation
     // Create a console logger as a fallback to use before the LSP server starts.
     using var loggerFactory = LoggerFactory.Create(builder =>
     {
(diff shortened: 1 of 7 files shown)


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Program.cs
```diff
@@ -44,16 +44,18 @@ static async Task RunAsync(ServerConfiguration serverConfiguration, Cancellation
     // Create a console logger as a fallback to use before the LSP server starts.
     using var loggerFactory = LoggerFactory.Create(builder =>
     {
-        builder.SetMinimumLevel(serverConfiguration.MinimumLogLevel);
+        // The actual logger is responsible for deciding whether to log based on the current log level.
+        // The factory should be configured to log everything.
+        builder.SetMinimumLevel(LogLevel.Trace);
```
dibarbet, line 49: As best as I can tell it is not possible to change this level after creation.  So we have to check the log level in the logger itself.

JoeRobich (approved):

Issue #75592, opened 2024-10-23
Cyclic assignment should generate a ref safety error


jaredpar:
Consider the following code:

```csharp
S G()
{
    S local = new();
    M(ref local);
    return local;
}

void M([UnscopedRef] ref S s)
{
    // This should error as `[UnscopedRef]` at most promotes `ref` to `ref struct` to *return only* 
    s.f = ref s.i;
}

ref struct S {
    public int i;
    public ref int f;
}
```

Relevant portion of [the spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#silly-cyclic-assignment):

> It is also why `[UnscopedRef]` only promotes the *ref-safe-context* of any `ref` to `ref struct` values to *return-only* and not *caller-context*. Consider that using *caller-context* allows for cyclic assignment and would force a viral use of `[UnscopedRef]` for a `ref struct`:

Likely we should conisder making the initial spec section on `UnscopedRef` a bit clearer about this as this detail is a bit further on

jjonescz:
> This should error as `[UnscopedRef]` at most promotes `ref` to `ref struct` to *return only*

Is that correct? `ref` parameters have *ref-safe-context* of *return-only* per [this part](https://github.com/dotnet/csharplang/blob/bf7756ade305e3094d1492c6f664c606160a5677/proposals/csharp-11.0/low-level-struct-improvements.md#return-only-safe-context):

> A `ref` or `in` parameter will have a *ref-safe-context* of *return-only*. This is done in part for `ref struct` to prevent silly cyclic assignment issues. It is done uniformly though to simplify the model as well as minimize compat changes.

So `[UnscopedRef]` promotes *ref-safe-context* of `ref S s` to *caller-context*.

It seems this part of the spec might be wrong:

> It is also why `[UnscopedRef]` only promotes the *ref-safe-context* of any `ref` to `ref struct` values to *return-only* and not *caller-context*. Consider that using *caller-context* allows for cyclic assignment and would force a viral use of `[UnscopedRef]` for a `ref struct`:

jjonescz:
The example further below

> Note: this cyclic assignment problem does continue to exist for `[UnscopedRef] out` to `ref struct` because that causes the *safe-context* and *ref-safe-context* to be equivalent. 

also doesn't make sense to me; it's not possible to have a cyclic assignment problem since the parameter is `out`, so we cannot even access it to do the self-assignment.

[...]

PR #75618, opened 2024-10-24
Fix ref analysis of self-assignment

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75592.

### DIFF
--- a/src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
+++ b/src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
@@ -26477,6 +26477,12 @@ class Program<S> where S : allows ref struct
 
             var comp = CreateCompilation(source, targetFramework: s_targetFrameworkSupportingByRefLikeGenerics);
             comp.VerifyEmitDiagnostics(
+                // (12,26): error CS8350: This combination of arguments to 'Program<S>.F1(ref S)' is disallowed because it may expose variables referenced by parameter 'x1' outside of their declaration scope
+                //         ref var y2 = ref F1(ref x2);
+                Diagnostic(ErrorCode.ERR_CallArgMixing, "F1(ref x2)").WithArguments("Program<S>.F1(ref S)", "x1").WithLocation(12, 26),
+                // (12,33): error CS8166: Cannot return a parameter by reference 'x2' because it is not a ref parameter
+                //         ref var y2 = ref F1(ref x2);
+                Diagnostic(ErrorCode.ERR_RefReturnParameter, "x2").WithArguments("x2").WithLocation(12, 33),
                 // (13,20): error CS8157: Cannot return 'y2' by reference because it was initialized to a value that cannot be returned by reference
                 //         return ref y2; // 1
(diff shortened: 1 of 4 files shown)


src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -10170,5 +10170,78 @@ public void Utf8Addition()
...
+                {
+                    int field;
+                    ref int refField;
+
+                    public static void M([UnscopedRef] ref S s)
+                    {
+                        s.refField = ref s.field;
+                    }
+                }
+                """;
```
jjonescz, line 10214: Without the fix, this compiled without an error.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -10170,5 +10170,78 @@ public void Utf8Addition()
...
+                S.M(ref s);
+
+                ref struct S
+                {
+                    int field;
+                    ref int refField;
+
+                    public static void M([UnscopedRef] ref S s)
+                    {
+                        s.refField = ref s.field;
```
jaredpar, line 10211: This is the code that should error as the _ref-safe-context_ is _return-only_.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -10170,5 +10170,78 @@ public void Utf8Addition()
...
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75592")]
+        public void SelfAssignment_CallerContext()
+        {
+            var source = """
+                using System.Diagnostics.CodeAnalysis;
+
+                S s = default;
+                S.M(ref s);
```
jaredpar, line 10202: Consider that these two forms are equivalent 

```csharp
ref struct S {
  [UnscopedRef]
  void M1() { }

  static void M2([UnscopedRef] ref S p) { }
}

S local;
local.M1();
S.M2(ref local);
```

So if the `S.M(ref s)` is an error then virtually any instance method call on a `ref struct` with `[UnscopedRef]` is too..

jaredpar (changes requested):
Think the error is in the wrong place here.

RikkiGibson (approved):

src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs
```diff
@@ -26477,6 +26477,12 @@ static ref S F2(S x2)
 
             var comp = CreateCompilation(source, targetFramework: s_targetFrameworkSupportingByRefLikeGenerics);
             comp.VerifyEmitDiagnostics(
+                // (12,26): error CS8350: This combination of arguments to 'Program<S>.F1(ref S)' is disallowed because it may expose variables referenced by parameter 'x1' outside of their declaration scope
```
jaredpar, line 26480: Can we add a variation of this test where the signature of `F1` is 

```csharp
static ref S F1(ref S x1)
```

jaredpar (approved):

Issue #65353, opened 2022-11-11
Unexpected ref safety error assigning ref of captured local


cston:
An error is reported for the ref assignment in the more nested lambda expression:
```csharp
delegate void D();

class Program
{
    static void F(D d) { }

    static void Main()
    {
        {
            int x = 1;
            F(() =>
                {
                    int y = 2;
                    ref int r = ref y;
                    r = ref x;
                });
        }
        {{{
            int x = 1;
            F(() =>
                {
                    int y = 2;
                    ref int r = ref y;
                    r = ref x; // error CS8374: ... 'x' has a narrower escape scope than 'r'
                });
        }}}
    }
}
```
Result:
```
(24,21): error CS8374: Cannot ref-assign 'x' to 'r' because 'x' has a narrower escape scope than 'r'.
```

This is not a recent regression. The same error is reported in csc 3.11.

cston:
The same issue exists with local functions:
```csharp
#pragma warning disable 8321

class Program
{
    static void Main()
    {
        {{{
        int x = 1;
        void F()
        {
            int y = 2;
            ref int r = ref y;
            r = ref x; // error CS8374: ... 'x' has a narrower escape scope than 'r'
        }
        }}}
    }
}
```

PR #75627, opened 2024-10-25
Fix ref safety analysis of captured locals

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/65353.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
+++ b/src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
@@ -179,9 +179,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             return _localEscapeScopes?.TryGetValue(local, out var scopes) == true
                 ? scopes
-                : _symbol != local.ContainingSymbol // captured local
-                ? (CurrentMethodScope, CallingMethodScope)
-                : (CallingMethodScope, CallingMethodScope);
+                : (CurrentMethodScope, CallingMethodScope);
         }
 
         private void SetLocalScopes(LocalSymbol local, uint refEscapeScope, uint valEscapeScope)
--- a/src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
+++ b/src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
@@ -73,8 +73,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             MethodSymbol symbol,
             bool inUnsafeRegion,
             bool useUpdatedEscapeRules,
-            BindingDiagnosticBag diagnostics,
-            Dictionary<LocalSymbol, (uint RefEscapeScope, uint ValEscapeScope)>? localEscapeScopes = null)
+            BindingDiagnosticBag diagnostics)
         {
             _compilation = compilation;
             _symbol = symbol;
@@ -85,7 +84,6 @@ namespace Microsoft.CodeAnalysis.CSharp
             // outermost. To ensure that locals in the outermost block are considered at
             // the same depth as parameters, _localScopeDepth is initialized to one less.
             _localScopeDepth = CurrentMethodScope - 1;
-            _localEscapeScopes = localEscapeScopes;
         }
 
         private ref struct LocalScope
@@ -177,10 +175,12 @@ namespace Microsoft.CodeAnalysis.CSharp
 
         private (uint RefEscapeScope, uint ValEscapeScope) GetLocalScopes(LocalSymbol local)
         {
-            Debug.Assert(_localEscapeScopes?.ContainsKey(local) == true);
+            Debug.Assert(_localEscapeScopes?.ContainsKey(local) == true || _symbol != local.ContainingSymbol);
 
             return _localEscapeScopes?.TryGetValue(local, out var scopes) == true
                 ? scopes
+                : _symbol != local.ContainingSymbol // captured local
+                ? (CurrentMethodScope, CallingMethodScope)
                 : (CallingMethodScope, CallingMethodScope);
         }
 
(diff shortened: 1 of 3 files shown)


src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
```diff
@@ -177,10 +175,12 @@ public void Dispose()
...
         private (uint RefEscapeScope, uint ValEscapeScope) GetLocalScopes(LocalSymbol local)
         {
-            Debug.Assert(_localEscapeScopes?.ContainsKey(local) == true);
+            Debug.Assert(_localEscapeScopes?.ContainsKey(local) == true || _symbol != local.ContainingSymbol);
 
             return _localEscapeScopes?.TryGetValue(local, out var scopes) == true
                 ? scopes
+                : _symbol != local.ContainingSymbol // captured local
+                ? (CurrentMethodScope, CallingMethodScope)
                 : (CallingMethodScope, CallingMethodScope);
```
cston, line 184: >(CallingMethodScope, CallingMethodScope) [](http://example.com/codeflow?start=18&length=40)

Is this last value just a fallback in the unexpected case where the local is not in the dictionary and not captured? If so, then `CurrentMethodScope` seems like a safer value than `CallingMethodScope` for ref escape scope. In short, should we simply return `(CurrentMethodScope, CallingMethodScope)` whenever the local is not in the dictionary?

cston (approved):

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
```diff
@@ -177,11 +175,11 @@ public void Dispose()
 
         private (uint RefEscapeScope, uint ValEscapeScope) GetLocalScopes(LocalSymbol local)
         {
-            Debug.Assert(_localEscapeScopes?.ContainsKey(local) == true);
+            Debug.Assert(_localEscapeScopes?.ContainsKey(local) == true || _symbol != local.ContainingSymbol);
 
             return _localEscapeScopes?.TryGetValue(local, out var scopes) == true
                 ? scopes
-                : (CallingMethodScope, CallingMethodScope);
+                : (CurrentMethodScope, CallingMethodScope);
```
RikkiGibson, line 182: BTW, does entering this branch usually mean we are in some kind of "use before definition" case?

RikkiGibson, line 182: It looks like this has the effect of always saying that a local in a method/function has these ref/value lifetimes? Is that safe?

I think this is OK for the value lifetime because you can't capture a ref struct in a lambda. But I'm not so sure for the ref lifetime. Trying to think of a test.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -8630,6 +8630,62 @@ static R F()
...
+                    int x = 1;
+                    F(ref x);
+                    void F(ref int z)
+                    {
+                        z = ref x;
+                    }
+                {{new string('}', nesting)}}
+                """;
+            CreateCompilation(source).VerifyDiagnostics(
+                // (6,9): error CS8374: Cannot ref-assign 'x' to 'z' because 'x' has a narrower escape scope than 'z'.
```
RikkiGibson, line 8684: Something about this is not clicking for me. I think it's the fact that passing `ref x` as an argument is very similar to assigning `z = ref x`. Yet one is permitted and the other is not.

However, this seems like a pre-existing situation, which is not being changed by this PR.

It might be good to show what happens when using `scoped ref int z` and various levels of nesting.

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
RikkiGibson: I think I am comfortable with the captured local ref having `CurrentMethod` lifetime.

It feels like in reality, the captured local has some "external" lifetime relative to the current function. It is always still live after the local function call returns. It does compile down to a ref parameter, after all, when the function is not converted to delegate, etc.

But, it is safe for us to, in the context of a capture, implicitly convert that captured local's lifetime to CurrentMethod of the current function, since converting from a wider to narrower top-level lifetime is generally safe.

It means the captured local might not be convertible to some things that it maybe ought to be, like, a ref parameter or return, but this doesn't seem like a coding pattern people really need. Imagine something like the following, which I think is safe, but that can't be done as written, either before or after this PR.

```cs
bool flag = false;
int x = 0;
int y = 0;
ref int z = fn();
ref int fn()
    => ref flag ? ref x : ref y;
```

I think also if we wanted the ability to capture ref struct locals, we would have to reckon with this problem a little more, but I am also not seeing that as something people are urgently asking to do, so am not worried about it.

RikkiGibson (approved):

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
RikkiGibson: https://github.com/dotnet/roslyn/pull/75627#discussion_r1819874912

src/Compilers/CSharp/Portable/Binder/RefSafetyAnalysis.cs
jjonescz: Right. From implementation perspective, the captured locals could have the CallingMethod ref lifetime, and everything should work since the equivalent lowered code would be ref safe.

But from the language specification perspective, I'm not sure we can make that assumption. Perhaps there are alternative lowering strategies where the ref safety would be violated.

If we wanted to change this, I think we would need to change the ref safety speclet to specify lifetimes of captured locals (or propose a new feature for that).

PR #75636, opened 2024-10-25
Adjust `ContainingSymbol` for substituted local function

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/75543

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/SubstitutedMethodSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/SubstitutedMethodSymbol.cs
@@ -20,7 +20,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
     // C<X>.M<Y> is a ConstructedMethodSymbol.
     internal class SubstitutedMethodSymbol : WrappedMethodSymbol
     {
-        private readonly NamedTypeSymbol _containingType;
+        private readonly Symbol _containingSymbol;
         private readonly MethodSymbol _underlyingMethod;
         private readonly TypeMap _inputMap;
         private readonly MethodSymbol _constructedFrom;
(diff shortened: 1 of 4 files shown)


src/Compilers/CSharp/Portable/Symbols/SubstitutedMethodSymbol.cs
```diff
@@ -190,7 +190,9 @@ public sealed override Symbol ContainingSymbol
         {
             get
             {
-                return _containingType;
+                return MethodKind is MethodKind.LocalFunction
+                    ? OriginalDefinition.ContainingSymbol
```
AlekseyTs, line 194: >OriginalDefinition.ContainingSymbol [](http://example.com/codeflow?start=22&length=35)

This will be valid only when we are dealing with local function symbol with container that is not substituted itself. At the moment this assumption is probably true because we never create such a symbol for a local function, not that it is technically impossible. At the same time we have an invariant that a constructed generic symbol must maintain ```ContainingSymbol === ConstructedFrom.ContainingSymbol```. It is not obvious that the conditional logic around local function is going to maintain the invariant under all possible scenarios now and in the fututre. Consider if it would be more intuitive and more maintainable in the long run to do the following instead:
1. Override ```ContainingSymbol``` in ```ConstructedMethodSymbol``` as follows:
```
        public override Symbol ContainingSymbol
        {
            get
            {
                return ConstructedFrom.ContainingSymbol;
            }
        }
```

2. Add the following assert to ```protected SubstitutedMethodSymbol(NamedTypeSymbol containingSymbol, TypeMap map, MethodSymbol originalDefinition, MethodSymbol constructedFrom)```:

```
Debug.Assert(this is ConstructedMethodSymbol || originalDefinition.MethodKind is not (MethodKind.LocalFunction or MethodKind.Lambda));
```

AlekseyTs: Here is another even more general alternative approach:
1. Replace ```NamedTypeSymbol _containingType``` with ```Symbol _containingSymbol```
2. Change signature of the second constructor to ```protected SubstitutedMethodSymbol(Symbol containingSymbol, TypeMap map, MethodSymbol originalDefinition, MethodSymbol constructedFrom)```
3. Adjust implementation as follows:
```
public sealed override Symbol ContainingSymbol
{
    get
    {
        return _containingSymbol;
    }
}

public override NamedTypeSymbol ContainingType
{
    get
    {
        return _containingSymbol is NamedTypeSymbol nt ? nt : containingSymbol.ContainingType;
    }
}
```
4. Adjust constructor of ```ConstructedMethodSymbol``` to do the following:
```base(containingSymbol: constructedFrom.ContainingSymbol,```

jcouv: Thanks much for the alternative proposals

cston (approved):

AlekseyTs (approved):
LGTM (commit 2)

PR #75639, opened 2024-10-25
Reduce allocations in AbstractAsynchronousTaggerProvider.ComputeDifference

ToddGrun:
This method is showing up in a speedometer typing profile I'm looking at. This method computes the difference between two TagSpanIntervalTrees given an updated snapshot.

To do so, it currently creates a new heap allocated TagSpan for each item in both given interval trees. However, it need not do that, as that generated TagSpan isn't passed out of this method. Instead, TagSpanIntervalTree now has an AddAllSpans method that populates to a collection of value tuples, which ComputeDifference can now use.

This doesn't save a ton of allocations, but it's fairly simple and contained. The amount of allocations reduced by this change varies from 0.1% to 0.5% in a typing scenario I'm looking at depending on the profile.

*** allocations that are removed from the worse of the profiles of a typing scenarios in a speedometer test ***
image


ToddGrun:
@dotnet/roslyn-ide ptal

ToddGrun:
@dotnet/roslyn-ide ptal

PR #75646, opened 2024-10-26
Reduce allocations in ProjectSystemProjectFactory.ConvertMetadataReferencesToProjectReferences_NoLock

ToddGrun:
This method is pretty high traffic and it's use of the OfType Linq method is showing up as 0.2% of allocations in the speedometer profile I'm looking at of a solution open. Instead of using linq here, it's just as easy to do the ""is" check inline avoid the yield return state machinery.

*** relevant allocations from the speedometer profile ***
image


ToddGrun:
@dotnet/roslyn-ide ptal

CyrusNajmabadi (approved):

PR #75647, opened 2024-10-26
Ref safety: replace uint scope with struct SafeContext

RikkiGibson:
Update: renamed the new type to *SafeContext* to match the language standard. https://github.com/dotnet/roslyn/pull/75647#issuecomment-2463118930

---

The 'struct Lifetime' represents a shift in philosophy toward a type-like solution for ref safety. Rather than directly using uints, equality/relational operators, Math.Min/Max, etc. we express things in terms of "convertibility" of these semi-opaque Lifetime values as well as 'Intersect' and 'Union' operators. Hopefully you will find that this improves the clarity of the code quite a bit.

I am adopting the term "lifetime" rather than "escape scope", "safe to escape", "ref-safe-scope", etc. very deliberately here. I am coming around to the opinion that using the term "lifetime" is much clearer than any of the above.

Separately, I am hoping to publish a spec for this area which will hopefully demonstrate that improvement in clarity. That spec will probably be an expansion upon [low-level-struct-improvements.md#annotations](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#annotations), but as a standalone document. The goal of that document would be to succinctly and fully define all the ref safety rules in the language, and provide a basis for deciding if any given program is "ref-safe".

Because the dynamics of how ref lifetimes can be nested and related is much, much simpler than for types in C#, you can think of this as more of a metaphor for how things work than anything else. We are not "reifying" the concept of a lifetime as a type, we are not appending this concept to the symbol model in any way, etc. In the compiler it's just a "temporary" concept which is introduced for the purpose of these safety checks and diagnostics and then dropped.


RikkiGibson:
@cston @jjonescz @jaredpar for review.

(I think there are other PRs modifying this code at the same time. This one can wait till after they go in and handle the conflicts.)

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,130 @@
...
+        public bool Equals(Lifetime other)
+            => this._value == other._value;
+
+        public override bool Equals(object? obj)
+            => obj is Lifetime other && this.Equals(other);
+
+        public override int GetHashCode()
+            => unchecked((int)_value);
+
+        public static bool operator ==(Lifetime lhs, Lifetime rhs)
```
RikkiGibson, line 115: I found that BoundNodes generation expects this operator to exist for value types used as fields. So I ended up filling in all the equality methods rather than just having an 'Equals(Lifetime)'.

I did avoid using this operator in manually written code but don't feel extremely strongly about that.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
jjonescz, line 1: > I am adopting the term "lifetime" rather than "escape scope", "safe to escape", "ref-safe-scope", etc. very deliberately here. I am coming around to the opinion that using the term "lifetime" is much clearer than any of the above.

Shouldn't we strive to use terminology from the language spec, though? The terminology we landed upon is "safe-context" / "ref-safe-context" and we also modified some speclets to use this new teminology: https://github.com/dotnet/csharplang/pull/7606

That being said, the code wasn't updated anyway (e.g., it uses calling-method and current-method instead of caller-context and function-member), so that's out of scope for this PR. I'm also fine with the name `Lifetime` for the struct. I guess it's a bit different concept than "safe-context" / "ref-safe-context" anyway - "lifetime" is the value of "safe-context" / "ref-safe-context" of a given variable. But perhaps the doc comments could mention the spec terms.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        /// <summary>
+        /// The "current method" scope that is just inside the containing method/lambda.
+        /// If something can escape to this scope, it can escape to any scope in a given method, but cannot be returned.
+        /// </summary>
+        public static Lifetime CurrentMethod => new Lifetime(CurrentMethodRaw);
+
+        /// <summary>
+        /// Gets a lifetime which is "empty". i.e. which refers to a variable whose storage is never allocated.
+        ///</summary>
```
jjonescz, line 52:
```suggestion
        /// </summary>
```

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,130 @@
...
+        public override int GetHashCode()
+            => unchecked((int)_value);
+
+        public static bool operator ==(Lifetime lhs, Lifetime rhs)
+            => lhs._value == rhs._value;
+
+        public static bool operator !=(Lifetime lhs, Lifetime rhs)
+            => lhs._value != rhs._value;
+
+        private string GetDebuggerDisplay()
```
jjonescz, line 121: Could we override ToString instead? Might be useful if a Lifetime instance becomes part of an Exception message (e.g., via `ExceptionUtilities.UnexpectedValue(lifetime)`).

RikkiGibson: Sounds good.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
RikkiGibson: There is indeed a mix of terms in the implementation in this PR. My goal for this PR was to use what are in my opinion the best terms for new members, and mostly leave existing names alone for reviewability.

*safe-context* is definitely an improvement over the terms I mentioned originally. But I think we can improve things even more.

I want the spec language to emphasize that it is not the lifetimes themselves that are *safe* or not, rather it is the conversions between lifetimes that are safe or not. Also, I want to introduce the concept of a variable being *live*, and the correspondence between *liveness* of a variable and *lifetime* of a reference is useful.

jaredpar: The term needs to work well with both value and ref contexts. Basically think about how it's *safe-context* and *ref-safe-context*. I think that *lifetime* and *ref-lifetime* work well. I agree that the terminology is all over the map here.

Bill Wagner moved the spec term to the current state cause he felt it was the best way to describe it to users and in the spec. I do want us to get to a place where the spec and the code agree on terms. So before committing to this lets check with @BillWagner if this term is acceptable or not.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,129 @@
...
+            Debug.Assert(!result.IsReturnable);
+            return result;
+        }
+
+        /// <summary>
+        /// Gets a lifetime which is wider than the given lifetime.
+        /// Used to "exit" a nested local scope.
+        /// </summary>
+        public Lifetime Wider()
+        {
```
jaredpar, line 71:
```suggestion
        {
            Debug.Assert(this._value > 0);
```

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// </summary>
+        public static Lifetime Empty => new Lifetime(uint.MaxValue);
+
+        /// <summary>
+        /// Gets a lifetime which is narrower than the given lifetime.
+        /// Used to "enter" a nested local scope.
+        /// </summary>
+        public Lifetime Narrower()
+        {
+            var result = new Lifetime(this._value + 1);
```
jaredpar, line 60: Why `this._value` vs. `_value` in a few places?

RikkiGibson: Some of these differences are unintentional. Mostly I felt that in methods where both `this._value` and `other._value` are used, for example, keeping the `this` prefix reads a little clearly.

In here, and in `Wider()`, it would be reasonable to drop the `this` prefix.

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -2768,7 +2745,7 @@ private bool ShouldInferDeclarationExpressionValEscape(BoundExpression argument,
                 _ => null
             };
             if (symbol is SourceLocalSymbol local &&
-                GetLocalScopes(local).ValEscapeScope == CallingMethodScope)
+                GetLocalScopes(local).ValEscapeScope.IsCallingMethod)
```
jjonescz, line 2748: At some places we use `scope.IsXYZ` and in others we use `scope.Equals(Lifetime.XYZ)`. Do we even need the `IsXYZ` properties; wouldn't it be better to use `.Equals` everywhere? (When I see the property I have to wonder if it returns `true` for multiple possible scopes like `IsReturnable` whereas `.Equals` is clear that we are checking for exactly one scope.)

jjonescz (approved):

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,129 @@
...
+namespace Microsoft.CodeAnalysis.CSharp
+{
+    /// <summary>
+    /// A representation of the program region in which a 'ref' can be used,
+    /// which is limited to what is expressible in C#.
+    /// </summary>
+    /// <remarks>
+    /// For example, in this design, all lifetimes have a known relationship to all other lifetimes.
+    /// </remarks>
+    internal readonly struct Lifetime
```
jcouv, line 16: >Lifetime [](http://example.com/codeflow?start=29&length=8)

nit: consider `RefLifetime` to narrow down the concept a bit

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
RikkiGibson: The main challenge without `bool IsReturnOnly` prop, for example, is: how to check in a pattern, if a lifetime equals some particular well-known lifetime. These could be moved to 'when' clauses or some such, though.

```cs
Lifetime li = ...;
li switch { { IsReturnOnly: true } => ... };
```

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
RikkiGibson: I am hesitant to use Ref in the type name at this time, because there is a temptation to continue using the naming scheme in the spec of *lifetime* versus *ref-lifetime* of an expression (which currently are called *safe-context* and *ref-safe-context*).

I am adjusting the doc comment to try to improve clarity.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
RikkiGibson: `!IsReturnable` already implies this, but I can add this assertion also if you feel it improves clarity.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// escaping to a wider scope, we can use simple depth numbering without ambiguity.
+        /// </summary>
+        private readonly uint _value;
+        private Lifetime(uint value) => _value = value;
+
+        /// <summary>
+        /// The "calling method" scope that is outside of the containing method/lambda.
+        /// If something can escape to this scope, it can escape to any scope in a given method through a ref parameter or return.
+        /// </summary>
+        public static Lifetime CallingMethod => new Lifetime(CallingMethodRaw);
```
jcouv, line 37: nit: consider switching to a static field, saves the constructor call.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,129 @@
...
+        public static Lifetime Empty => new Lifetime(uint.MaxValue);
+
+        /// <summary>
+        /// Gets a lifetime which is narrower than the given lifetime.
+        /// Used to "enter" a nested local scope.
+        /// </summary>
+        public Lifetime Narrower()
+        {
+            var result = new Lifetime(this._value + 1);
+            // Narrower() operator should always result in a local lifetime
```
jcouv, line 63: What if we apply `Narrower` to `CallingMethod`? Then we'd get `ReturnOnly` which is returnable

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
```diff
@@ -0,0 +1,129 @@
...
+            return new Lifetime(this._value - 1);
+        }
+
+        public bool IsCallingMethod => _value == CallingMethodRaw;
+        public bool IsReturnOnly => _value == ReturnOnlyRaw;
+        public bool IsReturnable => _value is CallingMethodRaw or ReturnOnlyRaw;
+
+        /// <summary>Returns true if a 'ref' with this lifetime can be converted to the 'other' lifetime. Otherwise, returns false.</summary>
+        /// <remarks>Generally, a wider lifetime is convertible to a narrower lifetime.</remarks>
+        public bool IsConvertibleTo(Lifetime other)
```
jcouv, line 85: nit: I had a bit of trouble with the concept of "convertibility" and thus the name here. Would `IsWiderThan` work?

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -216,29 +216,6 @@ public void Deconstruct(out ParameterSymbol? parameter, out BoundExpression argu
                 ? p.ToString()
                 : Argument.ToString();
         }
-
-        /// <summary>
```
jcouv, line 220: It feels like this comment could be preserved on `Lifetime`

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -1294,13 +1271,13 @@ private bool CheckParameterValEscape(SyntaxNode node, ParameterSymbol parameter,
...
-        private bool CheckParameterRefEscape(SyntaxNode node, BoundExpression parameter, ParameterSymbol parameterSymbol, uint escapeTo, bool checkingReceiver, BindingDiagnosticBag diagnostics)
+        private bool CheckParameterRefEscape(SyntaxNode node, BoundExpression parameter, ParameterSymbol parameterSymbol, Lifetime escapeTo, bool checkingReceiver, BindingDiagnosticBag diagnostics)
         {
             var refSafeToEscape = GetParameterRefEscape(parameterSymbol);
-            if (refSafeToEscape > escapeTo)
+            if (!refSafeToEscape.IsConvertibleTo(escapeTo))
             {
                 var isRefScoped = parameterSymbol.EffectiveScope == ScopedKind.ScopedRef;
-                Debug.Assert(parameterSymbol.RefKind == RefKind.None || isRefScoped || refSafeToEscape == ReturnOnlyScope);
+                Debug.Assert(parameterSymbol.RefKind == RefKind.None || isRefScoped || refSafeToEscape.Equals(Lifetime.ReturnOnly));
```
jcouv, line 1280: >Equals(Lifetime.ReturnOnly) [](http://example.com/codeflow?start=103&length=27)

nit: `refSafeToEscape.IsReturnOnly`?

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -1241,46 +1218,46 @@ static void reportReadOnlyParameterError(ParameterSymbol parameterSymbol, Syntax
 
     internal partial class RefSafetyAnalysis
     {
-        private static EscapeLevel? EscapeLevelFromScope(uint scope) => scope switch
+        private static EscapeLevel? EscapeLevelFromScope(Lifetime lifetime) => lifetime switch
```
jcouv, line 1221: >EscapeLevelFromScope [](http://example.com/codeflow?start=36&length=20)

nit: doesn't have to be addressed in this PR, but we now have two names for seemingly the same thing: "Scope" and "Lifetime"

jcouv (commented):
Done with review pass (iteration 8). The change looks good and helps readability. Some minor nits to consider. Thanks

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
RikkiGibson: I tried to migrate pieces of this comment to the appropriate members of Lifetime. For example, the lines which are describing the special scopes, are now moved to the properties for accessing those special scopes.

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
RikkiGibson: Yes I am hoping that once we have a standalone spec for this area (think something with a similar level of precision and "self-containedness" as we have for definite assignment analysis), we will adopt the terms from it uniformly throughout.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
RikkiGibson: I ended up microbenching it and the static field seems very slightly better.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
RikkiGibson: Have rewritten these asserts.

src/Compilers/CSharp/Portable/Binder/Lifetime.cs
RikkiGibson: Have adjusted the asserts to be in terms of the raw value.

I made the asserts intolerant of calling Narrower() on `Lifetime.CallingMethod`, for example, because we don't have a reason to do that.

jcouv (approved):
LGTM Thanks (iteration 10)

RikkiGibson:
Based on offline discussion and investigation, I am thinking this is not the right time or order to introduce a new term. Terms which we want to match specification, should be introduced first in specification. ECMA has written quite a bit already about these features in terms of *safe-context*, so, backing that train up also requires a proportional level of justification and work.

The goal of this PR is to simply improve understandability of this area. So let's do that on a standalone basis for now.

jaredpar:
> Update: renamed the new type to SafeContext to match the language standard

Semes like the _safe_ choice in this context

src/Compilers/CSharp/Portable/Binder/Binder_Operators.cs
```diff
@@ -4377,13 +4377,13 @@ private void ValidateRefConditionalOperator(SyntaxNode node, BoundExpression tru
             var currentScope = _localScopeDepth;
 
             // val-escape must agree on both branches.
-            uint whenTrueEscape = GetValEscape(trueExpr, currentScope);
-            uint whenFalseEscape = GetValEscape(falseExpr, currentScope);
+            SafeContext whenTrueEscape = GetValEscape(trueExpr, currentScope);
+            SafeContext whenFalseEscape = GetValEscape(falseExpr, currentScope);
 
-            if (whenTrueEscape != whenFalseEscape)
+            if (!whenTrueEscape.Equals(whenFalseEscape))
```
jaredpar, line 4383:
```suggestion
            if (whenTrueEscape != whenFalseEscape)
```

Given that we have operators here seems like we should use them.

src/Compilers/CSharp/Portable/Binder/SafeContext.cs
```diff
@@ -0,0 +1,132 @@
...
+            return new SafeContext(_value + 1);
+        }
+
+        /// <summary>
+        /// Gets a SafeContext which is wider than the given SafeContext.
+        /// Used to "exit" a nested local scope.
+        /// </summary>
+        public SafeContext Wider()
+        {
+            Debug.Assert(_value >= CurrentMethodRaw);
```
jaredpar, line 76: Why `CurrentMethodRaw vs. ReturnOnly`? It should be valid to get a wider scope from a `ReturnOnly` context.

RikkiGibson: The purpose of the method is to "exit" a local scope, alongside `Narrower()` to "enter" a local scope. It's not intended to, for example, transition between the well-known "returnable" scopes. (We do allow this method to return `ReturnOnly`, though, to allow for uniformity when exiting the "top-level" block of a method in the visitor.)

jjonescz (approved):

jaredpar (approved):

RikkiGibson:
It looks like following tests are failing in Release_32, which appear unrelated to my changes:
- `Microsoft.CodeAnalysis.Editor.UnitTests.IntelliSense.VisualBasicCompletionCommandHandlerTests.TestMatchWithTurkishIWorkaround9`
- `Microsoft.CodeAnalysis.Editor.UnitTests.IntelliSense.VisualBasicCompletionCommandHandlerTests.TestMatchWithTurkishIWorkaround10`

PR #75650, opened 2024-10-28
update configs for 17.13p2 snap

akhera99:
Draft until snap time

### DIFF
--- a/eng/config/PublishData.json
+++ b/eng/config/PublishData.json
@@ -215,6 +215,15 @@
       "vsMajorVersion": 17,
       "insertionTitlePrefix": "[d17.12]"
     },
+    "release/dev17.13": {
+      "nugetKind": [
+        "Shipping",
+        "NonShipping"
+      ],
+      "vsBranch": "main",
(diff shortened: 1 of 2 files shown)


eng/config/PublishData.json
```diff
@@ -215,16 +215,25 @@
...
     },
-    "main": {
+    "release/dev17.13": {
       "nugetKind": [
         "Shipping",
         "NonShipping"
       ],
       "vsBranch": "main",
       "vsMajorVersion": 17,
-      "insertionCreateDraftPR": false,
```
RikkiGibson, line 225: nit: would prefer to leave this in so I don't have to think about what the default is.

RikkiGibson (approved):

eng/config/PublishData.json
akhera99: the diff is misleading, it's still in for main

RikkiGibson: ah, think I see what you mean, we generally don't put this flag for release branches.

PR #75655, opened 2024-10-28
Ensure discards are initially soft selected in VSCode

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/7623

Unfortunately LSP doesn't give servers much control over filtering.  Servers do not generally get callbacks as filter text is typed - meaning we generally are not able to modify hard vs soft selection after the initial list is presented.

However, we can force VSCode to call us back in some scenarios using the `isIncomplete` flag on the completion list.  For this case, when we see only punctuation typed (`_`), we set the `isIncomplete` flag and unset the commit characters to mimic soft selection.

As the user continues typing (e.g. `_o`) we get called again and we can update to hard selection if they are completing a variable.  

However - this does not cover every case of switching between hard/soft selection.  It could be covered by always setting `isIncomplete` but that would be expensive as we're re-computing and reserializing the entire list.

### DIFF
--- a/src/Features/Core/Portable/Completion/CompletionService.cs
+++ b/src/Features/Core/Portable/Completion/CompletionService.cs
@@ -363,69 +363,6 @@ public abstract partial class CompletionService : ILanguageService
         }
     }
 
-    internal static bool IsHardSelection(CompletionItem item, bool matchedFilterText, bool hasSuggestionModeItem, string filterText)
-    {
-        if (hasSuggestionModeItem)
-        {
-            return false;
-        }
-
-        // We don't have a builder and we have a best match.  Normally this will be hard
-        // selected, except for a few cases.  Specifically, if no filter text has been
-        // provided, and this is not a preselect match then we will soft select it.  This
-        // happens when the completion list comes up implicitly and there is something in
-        // the MRU list.  In this case we do want to select it, but not with a hard
-        // selection.  Otherwise you can end up with the following problem:
-        //
-        //  dim i as integer =<space>
-        //
-        // Completion will comes up after = with 'integer' selected (Because of MRU).  We do
-        // not want 'space' to commit this.
-
-        // If all that has been typed is punctuation, then don't hard select anything.
-        // It's possible the user is just typing language punctuation and selecting
-        // anything in the list will interfere.  We only allow this if the filter text
-        // exactly matches something in the list already. 
-        if (filterText.Length > 0 && IsAllPunctuation(filterText) && filterText != item.DisplayText)
-        {
-            return false;
-        }
-
-        // If the user hasn't actually typed anything, then don't hard select any item.
-        // The only exception to this is if the completion provider has requested the
-        // item be preselected.
-        if (filterText.Length == 0)
-        {
-            // Item didn't want to be hard selected with no filter text.
-            // So definitely soft select it.
-            if (item.Rules.SelectionBehavior != CompletionItemSelectionBehavior.HardSelection)
-            {
-                return false;
-            }
-
-            // Item did not ask to be preselected.  So definitely soft select it.
-            if (item.Rules.MatchPriority == MatchPriority.Default)
-            {
-                return false;
-            }
-        }
-
-        // The user typed something, or the item asked to be preselected.  In 
-        // either case, don't soft select this.
-        Debug.Assert(filterText.Length > 0 || item.Rules.MatchPriority != MatchPriority.Default);
-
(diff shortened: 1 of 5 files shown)


src/Features/Core/Portable/Completion/CompletionService.cs
```diff
@@ -363,6 +363,82 @@ internal static void FilterItems(
         }
     }
 
+    internal static bool IsHardSelection(CompletionItem item, bool matchedFilterText, bool hasSuggestionModeItem, string filterText)
```
dibarbet, line 366: everything in this file is moved unchanged

src/LanguageServer/Protocol/Handler/Completion/CompletionHandler.cs
```diff
@@ -253,5 +278,67 @@ static CompletionFilterReason GetFilterReason(CompletionTrigger trigger)
                 };
             }
         }
+
+        private static MatchResult GetBestCompletionItemSelectionFromFilteredResults(IReadOnlyList<MatchResult> filteredMatchResults)
```
genlu, line 282: is this copied from ItemManager? We uses this to actually determine the selected item in VS, but here this is at best a heuristic which might not match VSCode selection.

src/LanguageServer/Protocol/Handler/Completion/CompletionHandler.cs
```diff
@@ -219,6 +219,29 @@ private static (CompletionList CompletionList, bool IsIncomplete) FilterCompleti
             // Next, we sort the list based on the pattern matching result.
             matchResultsBuilder.Sort(MatchResult.SortingComparer);
 
+            // Determine if the list should be hard selected or soft selected.
+            bool isHardSelection;
+            if (matchResultsBuilder.Count > 0)
+            {
+                var bestResult = GetBestCompletionItemSelectionFromFilteredResults(matchResultsBuilder);
+                isHardSelection = CompletionService.IsHardSelection(bestResult.CompletionItem, bestResult.ShouldBeConsideredMatchingFilterText, completionList.SuggestionModeItem != null, filterText);
```
genlu, line 227: is this part needed to handle soft-selecting `_`? I think we could just use the result of `IsAllPunctuation(filterText)` to decide both selection mode and isIncomplete

src/LanguageServer/Protocol/Handler/Completion/CompletionHandler.cs
dibarbet: Yes this is copied.  Understood that it is a heuristic which may not match, but this definitely works better than before

src/LanguageServer/Protocol/Handler/Completion/CompletionHandler.cs
dibarbet: Ah interesting, I can make that change

src/LanguageServer/Protocol/Handler/Completion/CompletionHandler.cs
dibarbet: reverted

src/LanguageServer/Protocol/Handler/Completion/CompletionHandler.cs
```diff
@@ -219,6 +219,21 @@ private static (CompletionList CompletionList, bool IsIncomplete) FilterCompleti
...
+            // If they type something that is not punctuation, we may need to update the hard vs soft selection.
+            // For example, typing '_' should initially be soft selection, but if the user types 'o' we should hard select '_otherVar' (if it exists).
+            // This isn't perfect - ideally we would make this determination every time a filter character is typed, but we do not get called back
+            // for typing filter characters in LSP (unless we always set isIncomplete, which is expensive).
+            var isHardSelection = completionList.SuggestionModeItem is null && !isFilterTextAllPunctuation;
+            var isIncomplete = isFilterTextAllPunctuation;
+
+            // If our completion list hasn't hit the max size, we don't need to do anything filtering
+            if (completionListMaxSize < 0 || completionListMaxSize >= completionList.ItemsList.Count)
+                return (completionList, isIncomplete, isHardSelection);
```
genlu, line 235: I think this part can be moved to the beginning of the method, before calculating MatchResults

dibarbet: ah yup now that it no longer needs it.  moved

genlu (approved):

Issue #68027, opened 2023-04-29
IntelliSense cannot detect IAsyncEnumerable's yield return type when that parameter is used


datvm:
**Version Used**: 4.5.2-3.23171.7 (d17f7415)

**Steps to Reproduce**:

Put in the following code (in this example it's in the top-level `Program.cs`):

```cs
Console.WriteLine("Hello");
_ = BarAsync();

static async IAsyncEnumerable<Foo> BarAsync()
{
    var tmp = new string[10];
    foreach (var item in tmp)
    {
        yield return new(async s =>
        {
            // Happen here
            await Task.CompletedTask;
            throw new NotImplementedException();
        });
    }
}

record Foo(Func<string, Task<string>> func); // Happen to both Func<> and delegate
```

The code above compiles but at the `// Happen here` line, any attempt to access `s` results in `s` being unknown `?` even though without it, the compiler and IntelliSense correctly detect `s` as `string`.

**Diagnostic Id**: N/A

**Expected Behavior**:

`s` should maintain its type as `string` while typing.

**Actual Behavior**:

Whenever there is an attempt to access `s`, it becomes `?`. This happens to both Visual Studio and VS Code (latest version). If you put in valid code without IntelliSense, it still compiles, for example:

```
yield return new(async s =>
{
    return s.ToUpper(); // Have to manually type it
    await Task.CompletedTask;
    throw new NotImplementedException();            
});
```

Correct behavior when `s` is not being typed:

![image](https://user-images.githubusercontent.com/6388546/235317576-22c6b799-0e03-4b55-893a-f489607b807d.png)

In both VS and VS Code, `s` becomes unknown when any attempt to access it is made:

![image](https://user-images.githubusercontent.com/6388546/235317592-355b9948-fc42-4aa1-9607-a26157380d33.png)

datvm:
Attempt to "navigate" around the issue by declaring a variable does not work as well (note that `s2` now has type of `??` instead of `?`):

![image](https://user-images.githubusercontent.com/6388546/235317764-980f013b-bc5c-4205-adca-f1210b6f2784.png)

However this issue does not happen if `s2` is explicitly typed, maybe because at this time IntelliSense trusts that the error would be thrown earlier and `s2` doesn't need to be guessed:

![image](https://user-images.githubusercontent.com/6388546/235317800-5d2a0996-2c71-4574-867c-2e8fcf97bb00.png)

datvm:
[...]

PR #75660, opened 2024-10-29
Bind lambda in `yield return` despite errors

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/68027

The problem was that `BindYieldReturnStatement` gave up too easily on trying to convert the lambda. Without conversion, the lambda state didn't account for the target delegate type, which provides the parameter types for the lambda. So the lookup result for identifier `s` (which finds a candidate for type in local declaration `s await`) lacked a type.

In fixing this, I noticed that the binding diagnostics for the `yield return` and the `return` scenarios were different. 
The `ERR_BadSKknown` and `ERR_IllegalStatement` binding errors were now reported in the `yield return` scenario, but not in the `return` scenario. 
In `BindYieldReturn`, we use `GenerateConversionForAssignment` which collects the conversion diagnostics from `GenerateImplicitConversionError` unless `expression.HasAnyErrors && expression.Kind != BoundKind.UnboundLambda`.
In `BindReturn`, we use `CreateReturnConversion`, which skipped reporting from `GenerateImplicitConversionError` if `argument.HasAnyErrors`.
So I made an adjustment to the `return` binding logic to also report the diagnostics from converting/binding the lambda (`BoundKind.UnboundLambda`).

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
@@ -3167,7 +3167,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             diagnostics.Add(syntax, useSiteInfo);
 
-            if (!argument.HasAnyErrors)
+            if (!argument.HasAnyErrors || argument.Kind == BoundKind.UnboundLambda)
             {
                 if (returnRefKind != RefKind.None)
                 {
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
@@ -239,14 +239,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 ? BadExpression(node).MakeCompilerGenerated()
                 : BindValue(node.Expression, diagnostics, BindValueKind.RValue);
 
-            if (!argument.HasAnyErrors)
-            {
-                argument = GenerateConversionForAssignment(elementType, argument, diagnostics);
-            }
-            else
-            {
-                argument = BindToTypeForErrorRecovery(argument);
-            }
+            bool hasErrors = false;
 
             // NOTE: it's possible that more than one of these conditions is satisfied and that
             // we won't report the syntactically innermost.  However, dev11 appears to check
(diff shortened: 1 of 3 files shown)


src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -239,40 +239,53 @@ private BoundStatement BindYieldReturnStatement(YieldStatementSyntax node, Bindi
...
-            }
+            bool hasErrors = false;
 
             // NOTE: it's possible that more than one of these conditions is satisfied and that
             // we won't report the syntactically innermost.  However, dev11 appears to check
             // them in this order, regardless of syntactic nesting (StatementBinder::bindYield).
             if (this.Flags.Includes(BinderFlags.InFinallyBlock))
             {
                 Error(diagnostics, ErrorCode.ERR_BadYieldInFinally, node.YieldKeyword);
+                hasErrors = true;
```
AlekseyTs, line 250: >hasErrors = true; [](http://example.com/codeflow?start=16&length=17)

Why would we want to disable conversion binding for this and the other modified code paths. It feels to me that the conversion checks and the "unsupported" context checks are orthogonal, therefore cannot be considered as cascading for one another

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -239,40 +239,53 @@ private BoundStatement BindYieldReturnStatement(YieldStatementSyntax node, Bindi
...
             }
             else if (InUnsafeRegion && Compilation.IsFeatureEnabled(MessageID.IDS_FeatureRefUnsafeInIteratorAsync))
             {
                 Error(diagnostics, ErrorCode.ERR_BadYieldInUnsafe, node.YieldKeyword);
+                hasErrors = true;
             }
 
             CheckRequiredLangVersionForIteratorMethods(node, diagnostics);
+
+            if (argument != null)
```
AlekseyTs, line 275: >argument != null [](http://example.com/codeflow?start=16&length=16)

Is this condition ever `false`?

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -3154,7 +3167,7 @@ internal BoundExpression CreateReturnConversion(
 
             diagnostics.Add(syntax, useSiteInfo);
 
-            if (!argument.HasAnyErrors)
+            if (!argument.HasAnyErrors || argument.Kind == BoundKind.UnboundLambda)
```
AlekseyTs, line 3170: >!argument.HasAnyErrors || argument.Kind == BoundKind.UnboundLambda [](http://example.com/codeflow?start=16&length=66)

It looks like according to PR description this change is motivated by consistency with `BindYieldReturnStatement`. However, the conditions do not look the same to me. My preference would be to not touch this code, or use the same condition. Specifically ```!(argument.HasErrors || ((object)argument.Type != null && argument.Type.IsErrorType()))```

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
```diff
@@ -8672,5 +8673,154 @@ static async System.Threading.Tasks.Task Main()
...
+        {
+            await Task.CompletedTask;
+            throw new NotImplementedException();
+        };
+    }
+}
+""";
+            var comp = CreateCompilation(src, targetFramework: TargetFramework.Net80);
+            comp.VerifyDiagnostics(
+                // (7,63): warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.
```
AlekseyTs, line 8699: >warning CS1998 [](http://example.com/codeflow?start=27&length=14)

Consider suppressing this warning if it is not a goal of the test to verify its presence

AlekseyTs:

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
jcouv: I woke up this morning with the same thought, but you'd already reviewed. Thanks

AlekseyTs (approved):
LGTM (commit 6)

jcouv:

RikkiGibson (approved):

PR #75661, opened 2024-10-29
Reduce memory and CPU costs due to SegmentedList usage

ToddGrun:
Currently, the SegmentedList class suffers from two potential inefficiencies:

1) Upon growth, it doubles the SegmentedArray size. This is necessary for normal List like collections to get constant time amortized growth, but isn't necessary for segmented data structures. 
2) Upon growth, it reallocates and copies over the existing pages.

This PR addresses 2).
Instead, if we reuse existing segments when growing the list, we can save significant CPU and allocation costs.

*** Benchmark.NET data ***
OLD:
image

NEW:
image


src/Dependencies/Collections/SegmentedArray`1.cs
```diff
@@ -94,6 +94,39 @@ public SegmentedArray(int length)
             }
         }
 
+        public SegmentedArray<T> Resize(int newLength)
+        {
+            // For now, only allow growing resizes
+            if (newLength <= _length)
+                throw new ArgumentOutOfRangeException(nameof(newLength));
+
+            var newItems = new T[(newLength + SegmentSize - 1) >> SegmentShift][];
```
ToddGrun, line 103: >var newItems = new T[(newLength + SegmentSize - 1) >> SegmentShift][]; [](http://example.com/codeflow?start=12&length=70)

I considered using Array.Resize for this too, but in the case where the size doesn't change, it just returns the input array. Which we would then mutate. If we're fine modifying this SegmentedArray as part of this operation (which I think I might be) then this would simplify this a bit and improve the perf slightly.

CyrusNajmabadi:
Oh. I didn't scroll down far enough.

CyrusNajmabadi (approved):
I'm approving the concept. Haven't looked deeply at the code change yet.

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -503,6 +498,14 @@ internal void Grow(int capacity)
             // Capacities exceeding Array.MaxLength will be surfaced as OutOfMemoryException by Array.Resize.
             if (newCapacity < capacity)
                 newCapacity = capacity;
+            else
```
CyrusNajmabadi, line 501: Note: we don't mix bracelets and braced statements.

CyrusNajmabadi (approved):
Code looks ok.  I think dedicated tests would be good.

src/Dependencies/Collections/SegmentedArray`1.cs
```diff
@@ -94,6 +94,39 @@ public SegmentedArray(int length)
             }
         }
 
+        public SegmentedArray<T> Resize(int newLength)
```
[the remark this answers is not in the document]
ToddGrun: Add some doc comments

src/Dependencies/Collections/SegmentedArray`1.cs
```diff
@@ -94,6 +94,39 @@ public SegmentedArray(int length)
...
+                throw new ArgumentOutOfRangeException(nameof(newLength));
+
+            var newItems = new T[(newLength + SegmentSize - 1) >> SegmentShift][];
+            var lastPageSize = newLength - ((newItems.Length - 1) << SegmentShift);
+
+            // Copy over all old pages
+            for (var i = 0; i < _items.Length; i++)
+                newItems[i] = _items[i];
+
+            // If the previous last page is still the last page, resize it to lastPageSize.
```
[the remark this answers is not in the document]
ToddGrun: added some tests to cover each of the methods changed

src/Dependencies/Collections/SegmentedArray`1.cs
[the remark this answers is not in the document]
sharwell: This heavily deviates from expected array semantics and should not be added. There are other internal members for converting between `T[][]` and `SegmentedArray<T>` which should instead be used for crafting this specialized resize under the tight constraints of `SegmentedList<T>`.

src/Dependencies/Collections/SegmentedArray`1.cs
```diff
@@ -419,6 +465,8 @@ public void Reset()
         internal static class TestAccessor
         {
             public static int SegmentSize => SegmentedArray<T>.SegmentSize;
+            public static T[][] GetSegments(SegmentedArray<T> array)
+                => array._items;
```
sharwell, line 469: This already exists as a helper method (not sure where, would need to find it)

src/Tools/IdeCoreBenchmarks/SegmentedListBenchmarks_InsertRange.cs
```diff
@@ -40,6 +41,17 @@ public void GlobalSetup()
             _segmentedInsertValuesObject = new SegmentedArray<object?>(100);
         }
 
+        [Benchmark(Description = "AddToSegmentedList<object>", Baseline = true)]
+        public void AddList()
```
sharwell, line 45: This should move to `SegmentedListBenchmarks_Add`

ToddGrun: I was going to undo the benchmarks and not merge them, unless you think I should go ahead and move this test and include it if this merges.

src/Dependencies/Collections/SegmentedArray`1.cs
ToddGrun: Found it, PrivateMarshal.AsSegments

src/Dependencies/Collections/SegmentedArray`1.cs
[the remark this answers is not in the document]
ToddGrun: Would love to come up with something that fits your request here, while still sharing the duplicated segments between the old and new. Can you maybe give a bit more detail about what you are proposing?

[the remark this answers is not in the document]
ToddGrun: Lots of possibilities, first one to come to mind:

1) Move this code to PrivateMarshal
2) Change it's name to something like Grow to indicate it's differs from the standard Array.Resize mechanism
3) Modify this SegmentedArray's _items directly and not return a new SegmentedArray. This would allow the simplify this method a bit.

Or are there additional concerns that I'm not covering?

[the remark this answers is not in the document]
sharwell: Rather, half of what you need is there. In addition to `AsSegments`, you'll need a corresponding helper to turn segments back into the full array.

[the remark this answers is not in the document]
sharwell: The helpers would be mirrored after `ImmutableCollectionsMarshal`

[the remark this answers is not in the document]
ToddGrun: @sharwell -- Moved some things around to try and match what you've mentioned. What did I miss?

src/Dependencies/Collections/SegmentedArray`1+PrivateMarshal.cs
```diff
@@ -14,5 +16,17 @@ internal static class PrivateMarshal
...
             => array._items;
+
+        public static SegmentedArray<T> AsSegmentedArray(T[][] segments)
+        {
+            if (segments is null)
+                throw new ArgumentNullException(nameof(segments));
+
+            var length = 0;
+            foreach (var segment in segments)
+                length += segment.Length;
```
sharwell, line 27: All segments except for the last segment have a known size. The length of the array should be calculated by adding the final segment length to the a multiple of the known segment size.

I'm undecided on whether argument validation should go here or in the caller. I lean towards placing it exclusively in the caller, and potentially omitting it (or making it debug assertions) since this is an unsafe class.

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -135,12 +135,39 @@ public int Capacity
...
-                            SegmentedArray.Copy(_items, newItems, _size);
+                            Array.Resize(
+                                ref segments[oldSegmentCount - 1],
+                                oldSegmentCount == newSegmentCount ? lastPageSize : segmentSize);
                         }
-                        _items = newItems;
+
+                        // Create all new pages (except the last one which is done separately)
+                        for (var i = oldSegmentCount; i < newSegmentCount - 1; i++)
+                            segments[i] = new T[segmentSize];
```
sharwell, line 164: `Array.Resize` works even when the argument is null. This should handle all cases:

```
for (var i = oldSegmentCount - 1; i < newSegmentCount - 1; i++)
  Array.Resize(ref segments[i], segmentSize);

Array.Resize(ref segments[newSegmentCount - 1], lastPageSize);
```

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -135,12 +135,39 @@ public int Capacity
...
+
+                        // Create a new last page if necessary
+                        if (oldSegmentCount < newSegmentCount)
+                            segments[newSegmentCount - 1] = new T[lastPageSize];
```
sharwell, line 168: `Array.Resize` works even when the argument is null. This should handle all cases:

```
for (var i = oldSegmentCount - 1; i < newSegmentCount - 1; i++)
  Array.Resize(ref segments[i], segmentSize);

Array.Resize(ref segments[newSegmentCount - 1], lastPageSize);
```

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -135,12 +135,39 @@ public int Capacity
...
                     {
-                        var newItems = new SegmentedArray<T>(value);
-                        if (_size > 0)
+                        // Rather than creating a copy of _items, instead reuse as much of it's data as possible.
+                        // This saves as much as 50% of allocations and 70% of CPU cost of Add in large collections.
+                        // See SegmentedListBenchmarks_Add for repro details.
+                        var segments = SegmentedCollectionsMarshal.AsSegments(_items);
+
+                        var segmentSize = SegmentedArrayHelper.GetSegmentSize<T>();
+                        var segmentShift = SegmentedArrayHelper.GetSegmentShift<T>();
```
sharwell, line 144: Inlining these values will likely improve performance. They are meticulously JIT-optimized.

ToddGrun: Dustin just explained this, very cool!

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -502,7 +529,17 @@ internal void Grow(int capacity)
...
+            {
                 newCapacity = capacity;
+            }
+            else
+            {
+                var segmentSize = SegmentedArrayHelper.GetSegmentSize<T>();
+
+                // If caller didn't request a large capacity increase, limit the increase to a single page
+                if (newCapacity > segmentSize)
+                    newCapacity = (((capacity - 1) / segmentSize) + 1) * segmentSize;
```
sharwell, line 541: Let's move this entire section, plus any changes to the logic, to instead be part of the initial calculation of `newCapacity`.

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -502,7 +529,17 @@ internal void Grow(int capacity)
             // If the computed capacity is still less than specified, set to the original argument.
```
sharwell, line 529: Requested changes:

1. Document the algorithm for initialization of `newCapacity`
2. For initial length greater than or equal to half the segment size but less than one full segment size, set `newCapacity` to one segment size.
3. For any initial length where the final segment is not a full segment, set `newCapacity` to the length it would be with a full-size final segment. This guarantees that the outer array will not be reallocated, and also guarantees that the single inner array allocation performed during the resize will not need to be performed a second time during the next resize. (Can be modified and treated as a generalization of the preceding point)
4. If the calculated `newCapacity` ends up being less than `capacity` _and_ `capacity` is greater than the segment size, apply a final ceiling operation so the final segment is full size.

These rules are all relevant regardless of whether we increase by doubling or by a page at a time. I am still reviewing the choice of expansion size.

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -135,12 +135,25 @@ public int Capacity
...
+
+                        var oldSegmentCount = segments.Length;
+                        var newSegmentCount = (value + SegmentedArrayHelper.GetSegmentSize<T>() - 1) >> SegmentedArrayHelper.GetSegmentShift<T>();
+
+                        // Grow the array of segments, if necessary
+                        Array.Resize(ref segments, newSegmentCount);
+
+                        // Resize all segments to full segment size from the last old segment to the next to last
+                        // new segment.
+                        for (var i = oldSegmentCount > 0 ? oldSegmentCount - 1 : 0; i < newSegmentCount - 1; i++)
```
sharwell, line 149: 💭 What about calling out the case of empty at the beginning of this block and just using `new SegmentedArray<T>` for it? It took me a bit to figure out how `oldSegmentCount - 1` could be a problem.

ToddGrun: Seems reasonable. It took me a while to figure out that oldSegmentCount -1 was a problem too.

ToddGrun: I also went ahead and reduced the nesting in here, that was muddling things up for me, at least.

src/Dependencies/Collections/SegmentedArray`1+PrivateMarshal.cs
ToddGrun: Missed this one earlier. Removed the arg check, as you mentioned this is unsafe code, you should know what you're doing if you use this.

src/Dependencies/Collections/SegmentedArray`1+PrivateMarshal.cs
```diff
@@ -14,5 +16,14 @@ internal static class PrivateMarshal
         /// <inheritdoc cref="SegmentedCollectionsMarshal.AsSegments{T}(SegmentedArray{T})"/>
         public static T[][] AsSegments(SegmentedArray<T> array)
             => array._items;
+
+        public static SegmentedArray<T> AsSegmentedArray(T[][] segments)
+        {
+            var length = segments.Length == 0 ? 0 : segments[segments.Length - 1].Length;
```
sharwell, line 22: 💭 Given the advanced use case, would it be better to make length an argument?

ToddGrun: I could go either way. I assume you would prefer it so I'll go ahead and switch to that.

jaredpar (approved):
Reviewed compiler test code

src/Dependencies/Collections/SegmentedArray`1.cs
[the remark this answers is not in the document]
sharwell: ➡️ Now resolved.

src/Tools/IdeCoreBenchmarks/SegmentedListBenchmarks_InsertRange.cs
sharwell: ➡️ Now resolved

src/Dependencies/Collections/SegmentedArray`1+PrivateMarshal.cs
sharwell: ➡️ Now resolved

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -135,12 +135,39 @@ public int Capacity
                 {
                     if (value > 0)
                     {
-                        var newItems = new SegmentedArray<T>(value);
-                        if (_size > 0)
+                        // Rather than creating a copy of _items, instead reuse as much of it's data as possible.
+                        // This saves as much as 50% of allocations and 70% of CPU cost of Add in large collections.
+                        // See SegmentedListBenchmarks_Add for repro details.
```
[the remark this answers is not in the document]
sharwell: ➡️ Now resolved

src/Dependencies/Collections/SegmentedList`1.cs
sharwell: ➡️ Now resolved

src/Dependencies/Collections/SegmentedList`1.cs
sharwell: ➡️ Now resolved

src/Dependencies/Collections/SegmentedList`1.cs
sharwell: ➡️ Changing the capacity selection algorithm for resize has been deferred to a later pull request.

src/Dependencies/Collections/SegmentedList`1.cs
sharwell: ➡️ See https://github.com/dotnet/roslyn/pull/75661#discussion_r1824714647

src/Dependencies/Collections/SegmentedList`1.cs
sharwell: ➡️ Now simplified

src/Dependencies/Collections/SegmentedArray`1+PrivateMarshal.cs
sharwell: ➡️ Now simplified

sharwell (approved):

PR #75665, opened 2024-10-29
VB: Recognize and check an `unmanaged` constraint

AlekseyTs:


AlekseyTs:

AlekseyTs:

jaredpar:
@333fred, @cston PTAL

src/Compilers/Core/Portable/Binding/UseSiteInfo.cs
```diff
@@ -320,6 +320,21 @@ public void AddDiagnostics(ImmutableArray<DiagnosticInfo> diagnostics)
...
+            {
+                return;
+            }
+
+            _diagnostics ??= new HashSet<DiagnosticInfo>();
+
+            if (_diagnostics.Add(diagnosticInfo) && diagnosticInfo?.Severity == DiagnosticSeverity.Error)
+            {
+                RecordPresenceOfAnError();
+            }
```
cston, line 335: This is repeated several times. Consider extracting a helper method.

src/Compilers/Core/Portable/Symbols/INamedTypeSymbolInternal.cs
```diff
@@ -16,5 +17,69 @@ internal interface INamedTypeSymbolInternal : ITypeSymbolInternal
 
         ImmutableArray<ISymbolInternal> GetMembers();
         ImmutableArray<ISymbolInternal> GetMembers(string name);
+
+        /// <summary>
+        /// True if this type or some containing type has type parameters.
+        /// </summary>
+        public bool IsGenericType { get; }
```
cston, line 24: >public [](http://example.com/codeflow?start=8&length=6)

Is the `public` qualifier needed?

src/Compilers/Core/Portable/Symbols/INamedTypeSymbolInternal.cs
```diff
@@ -16,5 +17,69 @@ internal interface INamedTypeSymbolInternal : ITypeSymbolInternal
...
+
+        internal static class Helpers
+        {
+            /// <summary>
+            /// Returns True or False if we can determine whether the type is managed
+            /// without looking at its fields and Unknown otherwise.
+            /// Also returns whether or not the given type is generic.
+            /// </summary>
+            public static (ThreeState isManaged, bool hasGenerics) IsManagedTypeHelper(INamedTypeSymbolInternal type)
```
cston, line 33: >IsManagedTypeHelper [](http://example.com/codeflow?start=67&length=19)

Consider use a top-level class and make this an extension method.

src/Compilers/VisualBasic/Portable/Symbols/NamedTypeSymbol.vb
```diff
@@ -1195,6 +1195,121 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
             Return True
         End Function
 
+        Friend Overrides Function GetManagedKind(ByRef useSiteInfo As CompoundUseSiteInfo(Of AssemblySymbol)) As ManagedKind
+            Return GetManagedKind(Me, useSiteInfo)
+        End Function
+
+        ' <summary>
+        ' IsManagedType is simple for most named types:
```
cston, line 1203: >IsManagedType [](http://example.com/codeflow?start=10&length=13)

GetManagedKind()?

src/Compilers/VisualBasic/Portable/Symbols/NamedTypeSymbol.vb
```diff
@@ -1195,6 +1195,121 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
...
+
+                    Dim field = TryCast(member, FieldSymbol)
+
+                    ' Only instance fields (including field-like events) affect the outcome.
+                    If field Is Nothing OrElse field.IsShared Then
+                        Continue For
+                    End If
+
+                    useSiteInfo.Add(field.GetUseSiteInfo())
+                    Dim fieldType As TypeSymbol = field.Type.GetTupleUnderlyingTypeOrSelf()
```
cston, line 1265: >GetTupleUnderlyingTypeOrSelf() [](http://example.com/codeflow?start=61&length=30)

Are we testing this case?

src/Compilers/VisualBasic/Portable/Symbols/TypeSymbolExtensions.vb
```diff
@@ -1308,6 +1308,17 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
                 namedTypeSymbol.IsContainedInNamespace(NameOf(System), NameOf(System.Threading))
         End Function
 
+        ' Keep in sync with C# equivalent.
+        <Extension>
+        Friend Function IsWellKnownTypeUnmanagedType(typeSymbol As TypeSymbol) As Boolean
+            Dim namedTypeSymbol = TryCast(typeSymbol, NamedTypeSymbol)
+            Return namedTypeSymbol IsNot Nothing AndAlso
+                namedTypeSymbol.Name = "UnmanagedType" AndAlso
+                namedTypeSymbol.Arity = 0 AndAlso
```
cston, line 1317: >namedTypeSymbol.Arity = 0 AndAlso [](http://example.com/codeflow?start=16&length=33)

It looks like the C# version and related C# extension methods are not checking `Arity == 0`. Should we add that check to those methods?

src/Compilers/Core/Portable/Symbols/INamedTypeSymbolInternal.cs
AlekseyTs: > Consider use a top-level class and make this an extension method.

I prefer to keep the code as is. I do not foresee us using this helper much in the future. Therefore, in my opinion there is no good reason to "pollute" name lookup binding space with it. If we could use Default Interface Implementations feature in our code, I would make this method a member of ```INamedTypeSymbolInternal```.

src/Compilers/VisualBasic/Portable/Symbols/NamedTypeSymbol.vb
AlekseyTs: > Are we testing a case where this call makes a difference?

I cannot think of a situation when it would make an observable difference in behavior. That is not to say that that it is impossible to run into one in some edge error scenario. Given the difference in the way C# and VB compilers represent tuple symbols, I prefer to be conservative and explicitly unwrap tuple types.

cston:
>         Private Function CreateBinderForTypeDeclaration() As Binder

What does "ForTypeDeclaration" mean here? Should this method be named "ForMemberDeclaration" instead?

And would it make sense to pass the `BindingLocation` in as an optional argument to allow this method to be used in `ComputeType` as well?

---
Refers to: src/Compilers/VisualBasic/Portable/Symbols/Source/SourceEventSymbol.vb:741 in a55f298. [](commit_id = a55f298b9f6dd91b7381188e9da73f5fbb2435f6, deletion_comment = False)

cston:
>         Private Function CreateBinderForTypeDeclaration() As Binder

Same questions as in `SourceEventSymbol`:

Should this method be named "ForMemberDeclaration" instead?

And would it make sense to pass the `BindingLocation` in as an optional argument to allow this method to be used in `ComputeType` as well?

---
Refers to: src/Compilers/VisualBasic/Portable/Symbols/Source/SourcePropertySymbol.vb:1020 in a55f298. [](commit_id = a55f298b9f6dd91b7381188e9da73f5fbb2435f6, deletion_comment = False)

AlekseyTs:
>         Private Function CreateBinderForTypeDeclaration() As Binder

> What does "ForTypeDeclaration" mean here? (It looks like the resulting binder is more specific than for the containing type.) Should this method be named "ForMemberDeclaration" instead?

>And would it make sense to pass the BindingLocation in as an optional argument to allow this method to be used in ```ComputeType``` as well?

This is an existing method that new code is not using. I do not intend to rename it or change its behavior. 


---
In reply to: [2450165726](https://github.com/dotnet/roslyn/pull/75665#issuecomment-2450165726) [](http://example.com/codeflow?ancestors=2450165726)

---
Refers to: src/Compilers/VisualBasic/Portable/Symbols/Source/SourceEventSymbol.vb:741 in a55f298. [](commit_id = a55f298b9f6dd91b7381188e9da73f5fbb2435f6, deletion_comment = False)

AlekseyTs:
>         Private Function CreateBinderForTypeDeclaration() As Binder

Same response

---
In reply to: [2450168231](https://github.com/dotnet/roslyn/pull/75665#issuecomment-2450168231) [](http://example.com/codeflow?ancestors=2450168231)

---
Refers to: src/Compilers/VisualBasic/Portable/Symbols/Source/SourcePropertySymbol.vb:1020 in a55f298. [](commit_id = a55f298b9f6dd91b7381188e9da73f5fbb2435f6, deletion_comment = False)

src/Compilers/VisualBasic/Portable/Symbols/TypeSymbolExtensions.vb
AlekseyTs: Indeed, added a test and opened an issue to follow up - #75681

src/Compilers/VisualBasic/Test/Symbol/SymbolsTests/UnmanagedTypeConstraintTests.vb
```diff
@@ -0,0 +1,1799 @@
...
+            Dim compilation = CreateCompilation(source, references:={reference})
+            AssertTheseDiagnostics(compilation, <expected><![CDATA[
+BC32105: Type argument 'Program' does not satisfy the 'Structure' constraint for type parameter 'T'.
+        o.M(New Program())
+          ~
+BC37332: Type argument 'Program' must be a non-nullable value type, along with all fields at any level of nesting, in order to use it as type parameter 'T'.
+        o.M(New Program())
+          ~
+                                                ]]></expected>)
+        End Sub
```
cston, line 983: >End Sub [](http://example.com/codeflow?start=8&length=7)

Are we testing type parameters?
```
Sub M1(Of T1, T2 As Class, T3 As Structure)(t1 as T1, t2 As T2, t3 As T3)
    Dim o As New Test
    o.M(t1)
    o.M(t2)
    o.M(t3)
End Sub
```

src/Compilers/VisualBasic/Test/Symbol/SymbolsTests/UnmanagedTypeConstraintTests.vb
```diff
@@ -0,0 +1,1799 @@
...
+        o.M(New S4(Of Program)())
+          ~
+                                                ]]></expected>)
+
+            Dim s1 = compilation.GetTypeByMetadataName("S1")
+            Dim s2 = compilation.GetTypeByMetadataName("S2")
+            Dim s5 = compilation.GetTypeByMetadataName("S5`1")
+            Assert.Equal(ManagedKind.Managed, s1.GetManagedKind(Nothing))
+            Assert.Equal(ManagedKind.Unmanaged, s2.GetManagedKind(Nothing))
+            Assert.Equal(ManagedKind.UnmanagedWithGenerics, s5.GetManagedKind(Nothing))
```
cston, line 1043: >Assert.Equal [](http://example.com/codeflow?start=12&length=12)

Check `S3` and `S4` as well?

cston (approved):

src/Compilers/VisualBasic/Test/Symbol/SymbolsTests/UnmanagedTypeConstraintTests.vb
AlekseyTs: > Are we testing type parameters?

Type parameters are covered in `UnmanagedCheck_TypeParameter` and indirectly on several other tests.

AlekseyTs:
@333fred, @dotnet/roslyn-compiler For the second review.

AlekseyTs:
@333fred, @dotnet/roslyn-compiler For the second review.

src/Compilers/VisualBasic/Portable/Symbols/NamedTypeSymbol.vb
```diff
@@ -1195,6 +1195,121 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
...
+        ''' If the closure is infinite, we disregard all but a representative of any expanding cycle.
+        ''' 
+        ''' Intuitively, this will only return true if there's a specific type we can point to that is would
+        ''' be managed even if it had no fields.  e.g. struct S { S s; } is not managed, but struct S { S s; object o; }
+        ''' is because we can point to object.
+        ''' </summary>
+        Private Overloads Shared Function GetManagedKind(type As NamedTypeSymbol, ByRef useSiteInfo As CompoundUseSiteInfo(Of AssemblySymbol)) As ManagedKind
+            type = DirectCast(type.GetTupleUnderlyingTypeOrSelf(), NamedTypeSymbol)
+
+            ' The code below is a clone of BaseTypeAnalysis.GetManagedKind from C# it should be kept in sync going forward
```
333fred, line 1221:
```suggestion
            ' The code below is a clone of BaseTypeAnalysis.GetManagedKind from C#. It should be kept in sync going forward
```

333fred (commented):
Generally lgtm. Couple of small comments.

src/Compilers/VisualBasic/Portable/Symbols/Metadata/PE/PETypeParameterSymbol.vb
```diff
@@ -149,9 +155,18 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols.Metadata.PE
             End Get
         End Property
 
-        Private Function GetDeclaredConstraints() As ImmutableArray(Of TypeParameterConstraint)
+        Friend Overrides ReadOnly Property HasUnmanagedTypeConstraint As Boolean
+            Get
+                EnsureAllConstraintsAreResolved()
+                Return CType(Volatile.Read(_lazyHasIsUnmanagedConstraint), ThreeState).Value()
```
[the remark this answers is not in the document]
AlekseyTs: I  prefer to keep the ```Volatile.Read```. The "distance" between this operation and the interlocked exchange is several method calls and some of them are virtual calls. In one of the methods we check value of ```_lazyConstraintTypes``` without any memory barrier operation in the same body. Assuming that that read and this ```_lazyHasIsUnmanagedConstraint``` can be reordered, we might see the old value of ```_lazyHasIsUnmanagedConstraint``` here while the ```_lazyConstraintTypes``` check in ```EnsureAllConstraintsAreResolved``` sees the new value of ```_lazyConstraintTypes```. If that is too paranoid, that is fine with me,  ```Volatile.Read``` won't make things worse.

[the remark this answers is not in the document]
AlekseyTs: I remain unconvinced. The documentation says: "Reads the value of the specified field." That is exactly what I am going after. The call reads the specified address. I am not concerned about what happens before or after the call. I am concerned about that specific read happening exactly when the call is executed. Not before the call, not after the call.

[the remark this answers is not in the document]
333fred: Alright, after some extensive offline conversation here, I've opened https://github.com/dotnet/roslyn/issues/75759 to follow up on the race condition here; given that it's present in C# version as well, and has been present for a while, I'm going to approve for now, but we'll be following up on this next week to discuss a better long-term approach for the compiler.

333fred (approved):

cston (approved):

src/Compilers/VisualBasic/Portable/LanguageVersion.vb
```diff
@@ -21,6 +21,7 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
         VisualBasic15_5 = 1505
         VisualBasic16 = 1600
         VisualBasic16_9 = 1609
+        VisualBasic17_13 = 1713
```
jcouv, line 24: Please add the new language version and feature in:
- [https://github.com/dotnet/vblang/blob/main/Language-Version-History.md](https://github.com/dotnet/vblang/blob/main/Language-Version-History.md)
- [https://github.com/dotnet/roslyn/blob/main/docs/Language%20Feature%20Status.md](https://github.com/dotnet/roslyn/blob/main/docs/Language%20Feature%20Status.md)

paul1956: Is there an explanation as to what this enables VB to do?

PR #75671, opened 2024-10-29
Update global.json with updated msbuild version

akhera99:


dibarbet (approved):
if CI passes who am i to judge

PR #75695, opened 2024-10-31
Bind lambda in object initializer despite errors

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/72571

The issue is that we get to `BindAssignment` but the LHS (produced in `BindObjectInitializerMemberCommon`) has the `hasError` flag set because the implicit receiver (a placeholder with the whole object creation as its syntax) has an error.
See `hasErrors = boundMember.HasAnyErrors || implicitReceiver.HasAnyErrors;` in `BindObjectInitializerMemberCommon`.

Note: I also tried the approach of attaching a narrower syntax node to the placeholder (commit 2), but I thought it's better to limit the impact of attached syntax on placeholders more generally.

### DIFF
--- a/src/Compilers/CSharp/Portable/BoundTree/BoundNode.cs
+++ b/src/Compilers/CSharp/Portable/BoundTree/BoundNode.cs
@@ -102,10 +102,19 @@ namespace Microsoft.CodeAnalysis.CSharp
         {
             get
             {
-                // NOTE: check Syntax rather than WasCompilerGenerated because sequence points can have null syntax.
-                if (this.HasErrors || this.Syntax != null && this.Syntax.HasErrors)
+                if (this.HasErrors)
                 {
                     return true;
+                }
+
+                // The syntax attached to placeholders should not limit usage of those placeholders
+                if (this is not BoundValuePlaceholderBase)
+                {
+                    // NOTE: check Syntax rather than WasCompilerGenerated because sequence points can have null syntax.
+                    if (this.Syntax != null && this.Syntax.HasErrors)
+                    {
+                        return true;
+                    }
                 }
                 var expression = this as BoundExpression;
                 return expression?.Type?.IsErrorType() == true;
--- a/src/Compilers/CSharp/Portable/BoundTree/BoundNode.cs
+++ b/src/Compilers/CSharp/Portable/BoundTree/BoundNode.cs
@@ -102,19 +102,10 @@ namespace Microsoft.CodeAnalysis.CSharp
         {
             get
             {
-                if (this.HasErrors)
+                // NOTE: check Syntax rather than WasCompilerGenerated because sequence points can have null syntax.
+                if (this.HasErrors || this.Syntax != null && this.Syntax.HasErrors)
                 {
                     return true;
-                }
-
-                // The syntax attached to placeholders should not limit usage of those placeholders
-                if (this is not BoundValuePlaceholderBase)
-                {
-                    // NOTE: check Syntax rather than WasCompilerGenerated because sequence points can have null syntax.
-                    if (this.Syntax != null && this.Syntax.HasErrors)
-                    {
-                        return true;
-                    }
                 }
                 var expression = this as BoundExpression;
                 return expression?.Type?.IsErrorType() == true;
--- a/src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenRefLocalTests.cs
+++ b/src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenRefLocalTests.cs
@@ -4168,24 +4168,24 @@ public class C
             string expectedOperationTree = @"
 IInvalidOperation (OperationKind.Invalid, Type: ?, IsInvalid) (Syntax: 'new ref[] { 1 }')
   Children(1):
-      IObjectOrCollectionInitializerOperation (OperationKind.ObjectOrCollectionInitializer, Type: ?[]) (Syntax: '{ 1 }')
+      IObjectOrCollectionInitializerOperation (OperationKind.ObjectOrCollectionInitializer, Type: ?[], IsInvalid) (Syntax: '{ 1 }')
         Initializers(1):
-            IInvalidOperation (OperationKind.Invalid, Type: ?, IsImplicit) (Syntax: '1')
-              Children(2):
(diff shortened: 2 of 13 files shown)


jcouv:
@dotnet/roslyn-compiler for review. Thanks

src/Compilers/CSharp/Portable/BoundTree/BoundNode.cs
```diff
@@ -102,11 +102,20 @@ public bool HasAnyErrors
...
             {
-                // NOTE: check Syntax rather than WasCompilerGenerated because sequence points can have null syntax.
-                if (this.HasErrors || this.Syntax != null && this.Syntax.HasErrors)
+                if (this.HasErrors)
                 {
                     return true;
                 }
+
+                // The syntax attached to placeholders should not limit usage of those placeholders
+                if (this is not BoundValuePlaceholderBase)
```
AlekseyTs, line 111: >if (this is not BoundValuePlaceholderBase [](http://example.com/codeflow?start=16&length=41)

While I do not object to making this change, it feels like this is not a fix for the problem the PR is claiming to address. In order for SemanticModel to work properly, every syntax node that is normally bindable should be bound, regardless of errors, That means that the value returned by this helper shouldn't matter. The lambda should be bound even if this helper returns true.

AlekseyTs: We might suppress additional diagnostics based on what this helper returns, but we should not be suppressing the binding.

AlekseyTs:

src/Compilers/CSharp/Portable/BoundTree/BoundNode.cs
jcouv: I can adjust the PR title if that's what you're suggesting. Like in another lambda-related PR last week, the question is whether we do `BindToTypeForErrorRecovery` (which results in poor lambda binding) or `GenerateConversionForAssignment` (which binds a conversion so results in better lambda information).
Maybe "Adjust lambda binding in object initializer despite errors"?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenRefLocalTests.cs
```diff
@@ -4180,10 +4180,10 @@ public static void Main()
 
             var expectedDiagnostics = new[]
             {
-                // file.cs(6,28): error CS8386: Invalid object creation
+                // (6,28): error CS8386: Invalid object creation
```
AlekseyTs, line 4183: >// (6,28): error CS8386: Invalid object creation [](http://example.com/codeflow?start=16&length=48)

It doesn't look like there are meaningful changes in this file. Consider reverting.

src/Compilers/CSharp/Test/IOperation/IOperation/IOperationTests_IEventAssignmentExpression.cs
```diff
@@ -967,21 +967,24 @@ void M(C c, EventHandler handler)
...
+                            Instance Receiver:
                               IParameterReferenceOperation: c (OperationKind.ParameterReference, Type: C, IsInvalid) (Syntax: 'c')
-                        Handler: 
+                        Handler:
                           IParameterReferenceOperation: handler (OperationKind.ParameterReference, Type: System.EventHandler, IsInvalid) (Syntax: 'handler')
-              Right: 
-                ILiteralOperation (OperationKind.Literal, Type: System.Int32, Constant: 0) (Syntax: '0')
-
+              Right:
+                IConversionOperation (TryCast: False, Unchecked) (OperationKind.Conversion, Type: System.Void, IsImplicit) (Syntax: '0')
```
AlekseyTs, line 983: >System.Void [](http://example.com/codeflow?start=98&length=11)

Minor. Consider excluding `void` type as a meaningful target type for conversion.

src/Compilers/CSharp/Test/IOperation/IOperation/IOperationTests_IVariableDeclaration.cs
```diff
@@ -4082,25 +4082,22 @@ void M()
     Statements (0)
     Next (Regular) Block[B1]
         Entering: {R1}
-
 .locals {R1}
```
AlekseyTs, line 4085: It doesn't look like there are meaningful changes in this file. Consider reverting

src/Compilers/CSharp/Test/Semantic/Semantics/SemanticErrorTests.cs
```diff
@@ -16515,10 +16515,7 @@ static void Main()
                 Diagnostic(ErrorCode.ERR_CloseParenExpected, "").WithLocation(11, 34),
                 // (11,34): error CS1002: ; expected
                 //             Console.WriteLine(");
-                Diagnostic(ErrorCode.ERR_SemicolonExpected, "").WithLocation(11, 34),
-                // (9,13): error CS1688: Cannot convert anonymous method block without a parameter list to delegate type 'OutParam' because it has one or more out parameters
-                //         o = delegate  // CS1688
-                Diagnostic(ErrorCode.ERR_CantConvAnonMethNoParams, "delegate").WithArguments("OutParam").WithLocation(9, 13));
```
AlekseyTs, line 16521: >ERR_CantConvAnonMethNoParams [](http://example.com/codeflow?start=37&length=28)

Let's make sure we have other tests covering this error. The name of the test seems to indicate that this error was the purpose for creating it.

AlekseyTs:

src/Compilers/CSharp/Test/IOperation/IOperation/IOperationTests_IEventAssignmentExpression.cs
jcouv: It looks like we call `GenerateConversionForAssignment` with `void` in other scenarios too (`VoidCall_ImplicitVoidConversion_DiscardAssignment` which is a conditional operator, and `CS0670ERR_FieldCantHaveVoidType01` which is a field initializer).
I'm leaning to leave as-is. But I'd also be okay to change and file an issue to align/tighten things up.

jcouv:

cston (approved):

PR #75708, opened 2024-11-02
Modify Capacity growth rate to be linear instead of doubling on resize.

ToddGrun:
Modify algorithm in SegmentedList.Grow to grow by a single segment at a time instead of doubling in size.  This is an extension from some work done in https://github.com/dotnet/roslyn/pull/75661.

Note that when allocating the outer array in the Capacity setter, we still utilize the double size allocation mechanism to avoid an n^2 allocation pattern for it. However, we *don't* allocate all the inner arrays, only what is needed. If a subsequent request comes in to Grow, we'll utilize those empty array entries.

*** Benchmark.NET data ***
Note: The AddExtraItem flag is used to demonstrate that the old algorithm had a large allocation occurrence at that count. The new algorithm doesn't have such a steep allocation cliff at those discrete points. Generally,

Allocation changes from PR1 => PR2 => this PR:
Purple went from n => n/2 => n/2
Blue went from n => n/2 => n/2 => n/4

CPU changes are a little harder to summarize, but generally went down anywhere from 20%-70%

Before https://github.com/dotnet/roslyn/pull/75661
image

Ater https://github.com/dotnet/roslyn/pull/75661, before this PR:
image

With both PRs:
image

### DIFF
--- a/src/Dependencies/Collections/SegmentedList`1.cs
+++ b/src/Dependencies/Collections/SegmentedList`1.cs
@@ -124,54 +124,57 @@ namespace Microsoft.CodeAnalysis.Collections
         public int Capacity
         {
             get => _items.Length;
-            set
+            set => SetCapacityInternal(value, isExactCapacity: true);
+        }
+
+        private void SetCapacityInternal(int value, bool isExactCapacity)
+        {
+            if (value < _size)
             {
-                if (value < _size)
+                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.value, ExceptionResource.ArgumentOutOfRange_SmallCapacity);
+            }
+
+            if (value == _items.Length)
+                return;
+
+            if (value <= 0)
+            {
+                _items = s_emptyArray;
+                return;
+            }
+
+            if (_items.Length == 0)
+            {
+                // No data from existing array to reuse, just create a new one.
+                _items = new SegmentedArray<T>(value);
+            }
+            else
+            {
+                // Rather than creating a copy of _items, instead reuse as much of it's data as possible.
+                var segments = SegmentedCollectionsMarshal.AsSegments(_items);
+
+                var oldSegmentCount = segments.Length;
+                var newSegmentCount = (value + SegmentedArrayHelper.GetSegmentSize<T>() - 1) >> SegmentedArrayHelper.GetSegmentShift<T>();
+
+                if (newSegmentCount > segments.Length)
                 {
-                    ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.value, ExceptionResource.ArgumentOutOfRange_SmallCapacity);
+                    // Grow the array of segments, if necessary. If isExactCapacity is false, then this array may end up
+                    // having null entries at it's end when this method completes. Minimally doubling the outer array size
+                    // allows amortized linear cost growth properties.
+                    var newSegmentAllocCount = isExactCapacity ? newSegmentCount : Math.Max(newSegmentCount, segments.Length * 2);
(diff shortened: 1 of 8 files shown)


src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -153,8 +153,13 @@ public int Capacity
...
                     var newSegmentCount = (value + SegmentedArrayHelper.GetSegmentSize<T>() - 1) >> SegmentedArrayHelper.GetSegmentShift<T>();
 
-                    // Grow the array of segments, if necessary
-                    Array.Resize(ref segments, newSegmentCount);
+                    if (newSegmentCount > segments.Length)
+                    {
+                        // Grow the array of segments, if necessary. Note that this array may end up having null entries
+                        // at it's end when this method completes. Minimally doubling the outer array size allows amortized
+                        // linear cost growth properties.
+                        Array.Resize(ref segments, Math.Max(newSegmentCount, segments.Length * 2));
```
sharwell, line 161: 📝 I don't like this being on the general `Capacity.set` path, since it could penalize users who are setting a capacity specifically because they know the exact size of the resulting collection.

ToddGrun: That makes sense, let me move this to Grow

ToddGrun: Tried to move this to Grow and not modify the Capacity setter, but it ended up needing to duplicate pretty much all the logic in the Capacity setter. So, I created a helper method that both places can call into that takes in a flag to specify whether the doubling is desired for the segment array.

src/Dependencies/Collections/SegmentedList`1.cs
```diff
@@ -124,49 +124,57 @@ public SegmentedList(IEnumerable<T> collection)
...
-                    var newSegmentCount = (value + SegmentedArrayHelper.GetSegmentSize<T>() - 1) >> SegmentedArrayHelper.GetSegmentShift<T>();
+                var oldSegmentCount = segments.Length;
+                var newSegmentCount = (value + SegmentedArrayHelper.GetSegmentSize<T>() - 1) >> SegmentedArrayHelper.GetSegmentShift<T>();
 
-                    // Grow the array of segments, if necessary
-                    Array.Resize(ref segments, newSegmentCount);
+                if (newSegmentCount > segments.Length)
+                {
+                    // Grow the array of segments, if necessary. If isExactCapacity is false, then this array may end up
+                    // having null entries at it's end when this method completes. Minimally doubling the outer array size
```
ToddGrun, line 162: >having null entries at it's end when this method completes [](http://example.com/codeflow?start=23&length=58)

@sharwell 

As you mentioned offline, this could be problematic as various code assumes that entries in the array aren't null. Still looking into this, but to me it seems like it may require changing SegmentedArray to indicate that it's inner arrays can be null.

ToddGrun: Made changes to SegmentedArray _items array to explicitly acknowledge that it may now contain nulls. Commented the _items declaration to indicate which parts of it may be null. By making this nullable, it catches all the locations that I missed that could have null ref'ed with this change.

ToddGrun: Updated the benchmark numbers with the changes from commit 4. Not a big difference from the numbers taken without the nullability changes, still significantly better in the measured scenarios than without this PR.

ToddGrun:
Closed out in favor of https://github.com/dotnet/roslyn/pull/75756, which was probably *slightly* less efficient, but didn't require adding support to SegmentedArray to support null arrays in the outer array.

PR #75713, opened 2024-11-04
Use absolute path for CompilerGeneratedFilesOutputPath

jjonescz:
Workaround for https://github.com/dotnet/roslyn/issues/75680 until the fix flows into VS IntPreview.

### DIFF
--- a/src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj
+++ b/src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj
@@ -17,7 +17,7 @@
 
     <!-- Enable output of generated files back into our Generated directory -->
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
-    <CompilerGeneratedFilesOutputPath>Generated</CompilerGeneratedFilesOutputPath>
+    <CompilerGeneratedFilesOutputPath>$(MSBuildThisFileDirectory)\Generated</CompilerGeneratedFilesOutputPath>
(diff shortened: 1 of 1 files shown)


src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj
```diff
@@ -17,7 +17,7 @@
 
     <!-- Enable output of generated files back into our Generated directory -->
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
-    <CompilerGeneratedFilesOutputPath>Generated</CompilerGeneratedFilesOutputPath>
+    <CompilerGeneratedFilesOutputPath>$(MSBuildThisFileDirectory)\Generated</CompilerGeneratedFilesOutputPath>
```
jaredpar, line 20: Can we add a note to the issue to undo this change and put a link to the issue here?

jaredpar (approved):

src/Compilers/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.csproj
jjonescz: We don't necessarily need to undo it, absolute path should work always

chsienki: Yeah, I'm fine with leaving it as absolute as in the project file it means the same thing ultimately.

chsienki (approved):

Issue #36377, opened 2019-06-12
Go To Definition does not work correctly with delegate and operator


vsfeedback:
VSF_TYPE_MARKDOWNVS2019 goes to the operator definition when I perform &quot;Go to Defenition&quot; on &quot;Foo.Test&quot;, but I expect it to show me the method &quot;Test&quot;

```
public sealed class Foo
    {
        public delegate int IntDelegate();

public static int Test()
        {
            return 1;
        }

public static void Bar1()
        {
            Console.WriteLine((Foo)Foo.Test);
        }

public static explicit operator Foo(IntDelegate intDelegate)
        {
            return new Foo();
        }
    }
```

_This issue has been moved from https://developercommunity.visualstudio.com/content/problem/587895/go-to-definition-does-not-work-correctly-with-dele.html
VSTS ticketId: 901152_
_These are the original issue comments:_

Visual Studio Feedback System on 5/29/2019, 11:39 PM (13 days ago): <p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

_These are the original issue solutions:_
(no solutions)

CyrusNajmabadi:
@jcouv The above crashes for me in Debug VS.  I'm using:

![image](https://user-images.githubusercontent.com/4564579/198149900-ecad3034-69bc-4b97-8b82-fe8dca44843b.png)

Crash is in this stack:

```
>	Microsoft.CodeAnalysis.CSharp.dll!Microsoft.CodeAnalysis.CSharp.CSharpSemanticModel.GetMethodGroupSemanticSymbols(Microsoft.CodeAnalysis.CSharp.BoundMethodGroup boundNode, Microsoft.CodeAnalysis.CSharp.BoundNode boundNodeForSyntacticParent, Microsoft.CodeAnalysis.CSharp.Binder binderOpt, out Microsoft.CodeAnalysis.CSharp.LookupResultKind resultKind, out bool isDynamic, out System.Collections.Immutable.ImmutableArray<Microsoft.CodeAnalysis.CSharp.Symbol> methodGroup) Line 4290	C#
[...]

PR #75719, opened 2024-11-04
Adjust semantic model for method group conversion

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/36377

In the scenario, we have a user-defined conversion for cast, stacked with a method group conversion.

### DIFF
--- a/src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
+++ b/src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
@@ -4294,17 +4294,18 @@ namespace Microsoft.CodeAnalysis.CSharp
                         // we want to get the symbol that overload resolution chose for M, not the whole method group M.
                         var conversion = (BoundConversion)boundNodeForSyntacticParent;
 
-                        if (conversion.ConversionKind is not ConversionKind.MethodGroup && conversion.Operand is BoundConversion nestedConversion)
+                        MethodSymbol method = null;
+                        if (conversion.ConversionKind == ConversionKind.MethodGroup)
                         {
-                            Debug.Assert(conversion.ConversionKind is ConversionKind.NoConversion || conversion.ExplicitCastInCode);
-                            conversion = nestedConversion;
+                            method = conversion.SymbolOpt;
+                        }
+                        else if (conversion.Operand is BoundConversion { ConversionKind: ConversionKind.MethodGroup } nestedMethodGroupConversion)
+                        {
+                            method = nestedMethodGroupConversion.SymbolOpt;
                         }
 
-                        var method = conversion.SymbolOpt;
                         if ((object)method != null)
                         {
-                            Debug.Assert(conversion.ConversionKind == ConversionKind.MethodGroup);
-
                             if (conversion.IsExtensionMethod)
                             {
                                 method = ReducedExtensionMethodSymbol.Create(method);
--- a/src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
+++ b/src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
@@ -4294,6 +4294,12 @@ namespace Microsoft.CodeAnalysis.CSharp
                         // we want to get the symbol that overload resolution chose for M, not the whole method group M.
                         var conversion = (BoundConversion)boundNodeForSyntacticParent;
 
+                        if (conversion.ConversionKind is not ConversionKind.MethodGroup && conversion.Operand is BoundConversion nestedConversion)
+                        {
+                            Debug.Assert(conversion.ConversionKind is ConversionKind.NoConversion || conversion.ExplicitCastInCode);
+                            conversion = nestedConversion;
+                        }
+
                         var method = conversion.SymbolOpt;
                         if ((object)method != null)
                         {
--- a/src/Compilers/CSharp/Test/Symbol/Symbols/ConversionTests.cs
+++ b/src/Compilers/CSharp/Test/Symbol/Symbols/ConversionTests.cs
@@ -430,7 +430,7 @@ class Program
         public void GetSymbolInfo_ExplicitCastOnMethodGroup()
         {
             var src = """
-public sealed class C
+public class C
 {
     public static void M()
     {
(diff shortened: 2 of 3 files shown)


src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
```diff
@@ -4294,6 +4294,12 @@ private OneOrMany<Symbol> GetMethodGroupSemanticSymbols(
                         // we want to get the symbol that overload resolution chose for M, not the whole method group M.
                         var conversion = (BoundConversion)boundNodeForSyntacticParent;
 
+                        if (conversion.ConversionKind is not ConversionKind.MethodGroup && conversion.Operand is BoundConversion nestedConversion)
```
333fred, line 4297: Nit:
```suggestion
                        if (conversion is { ConversionKind: not ConversionKind.MethodGroup, Operand: BoundConversion nestedConversion })
```

333fred (approved):

AlekseyTs:
@jcouv Was this a regression?

src/Compilers/CSharp/Test/Symbol/Symbols/ConversionTests.cs
```diff
@@ -425,6 +426,112 @@ static void Main()
             Assert.True(conversion.IsNumeric);
         }
 
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/36377")]
+        public void GetSymbolInfo_ExplicitCastOnMethodGroup()
```
AlekseyTs, line 430: Consider adding a similar test for VB

src/Compilers/CSharp/Test/Symbol/Symbols/ConversionTests.cs
```diff
@@ -425,6 +426,112 @@ static void Main()
...
+public sealed class C
+{
+    public static void M()
+    {
+        C x = (C)C.Test;
+    }
+
+    public static int Test() => 1;
+
+    public static explicit operator C(System.Func<int> intDelegate)
```
AlekseyTs, line 442: >explicit [](http://example.com/codeflow?start=18&length=8)

Consider testing with implicit operator, including a scenario when the cast operator, `(C)`,  is not present in code.

src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
AlekseyTs, line 4297: >nestedConversion [](http://example.com/codeflow?start=129&length=16)

It feels like we should also check how this node related to the conversion syntactically

src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
```diff
@@ -4294,6 +4294,12 @@ private OneOrMany<Symbol> GetMethodGroupSemanticSymbols(
...
+                        {
+                            Debug.Assert(conversion.ConversionKind is ConversionKind.NoConversion || conversion.ExplicitCastInCode);
```
AlekseyTs, line 4299: >Debug.Assert(conversion.ConversionKind is ConversionKind.NoConversion || conversion.ExplicitCastInCode); [](http://example.com/codeflow?start=28&length=104)

It is not obvious why is it reasonable to make this assumption. And more importantly why do we need to make it. If we are going after a specific bound tree shape, I think we should be checking for that instead. For example, we might want to confirm that the nested conversion is the immediate conversion for the ```BoundMethodGroup boundNode```.

src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
```diff
@@ -4294,6 +4294,12 @@ private OneOrMany<Symbol> GetMethodGroupSemanticSymbols(
...
+                            conversion = nestedConversion;
```
AlekseyTs, line 4300: >conversion = nestedConversion; [](http://example.com/codeflow?start=28&length=30)

Consider adding a comment about why/when we need to do this transformation.

AlekseyTs:

src/Compilers/CSharp/Test/Symbol/Symbols/ConversionTests.cs
```diff
@@ -425,6 +426,207 @@ static void Main()
...
+            var comp = CreateCompilation(src);
+            comp.VerifyEmitDiagnostics(
+                // (5,17): error CS0428: Cannot convert method group 'Test' to non-delegate type 'C'. Did you intend to invoke the method?
+                //         C x = C.Test;
+                Diagnostic(ErrorCode.ERR_MethGrpToNonDel, "Test").WithArguments("Test", "C").WithLocation(5, 17));
+
+            var tree = comp.SyntaxTrees.Single();
+            var model = comp.GetSemanticModel(tree);
+            var memberAccess = GetSyntax<MemberAccessExpressionSyntax>(tree, "C.Test");
+            Assert.Equal("C C.op_Implicit(System.Func<System.Int32> intDelegate)", model.GetSymbolInfo(memberAccess).Symbol.ToTestDisplayString());  // Unexpected: Should be "void C.Test()"
```
AlekseyTs, line 485: >// Unexpected: Should be "void C.Test()" [](http://example.com/codeflow?start=149&length=40)

Are we planning to address this?

src/Compilers/CSharp/Test/Symbol/Symbols/ConversionTests.cs
jcouv: Done. There are some existing unexpected behaviors there. Filing issues for those

src/Compilers/VisualBasic/Test/Semantic/Semantics/Conversions.vb
```diff
@@ -5154,5 +5154,33 @@ True
...
+End Class
+    ]]></file>
+    </compilation>)
+
+            compilation.AssertTheseEmitDiagnostics(<expected>
+BC30581: 'AddressOf' expression cannot be converted to 'C' because 'C' is not a delegate type.
+        Dim x As C = DirectCast(AddressOf C.Test, C)
+                                ~~~~~~~~~~~~~~~~
+</expected>)
+        End Sub
```
AlekseyTs, line 5183: >End Sub [](http://example.com/codeflow?start=8&length=7)

It doesn't look like we are testing SemanticModel behavior

AlekseyTs:

AlekseyTs (approved):
LGTM (commit 4)

333fred (approved):

PR #75724, opened 2024-11-05
Parse directives in trailing trivia as skipped tokens trivia

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75583.
Alternative to https://github.com/dotnet/roslyn/pull/75595.
- Keeps parsing the directive for best recovery experience. But I simplified the code by creating a single BadToken with the parsed text and putting that into SkippedTokensTrivia.
- Doesn't keep diagnostics and other state from parsing the directive. Hence also doesn't parse disabled text afterwards.


src/Compilers/CSharp/Portable/Parser/Lexer.cs
```diff
@@ -1980,7 +1980,29 @@ private void LexSyntaxTrivia(bool afterFirstToken, bool isTrailing, ref SyntaxLi
...
+
+                                var savePosition = TextWindow.Position;
+
+                                this.LexDirectiveAndExcludedTrivia(afterFirstToken, afterNonWhitespaceOnLine: true, ref triviaList);
+
+                                var text = TextWindow.Text.GetSubText(TextSpan.FromBounds(savePosition, TextWindow.Position));
+
+                                var error = new SyntaxDiagnosticInfo(
+                                    offset: 0,
+                                    width: Math.Min(1, text.Length),
```
CyrusNajmabadi, line 1995: long widths like this are undesirable.  what would be better would just be a diagnostic of length 1 under the `#`.

src/Compilers/CSharp/Portable/Parser/Lexer.cs
```diff
@@ -2381,8 +2403,12 @@ private CSharpSyntaxNode LexSingleDirective(
 
             directive = _directiveParser.ParseDirective(isActive, endIsActive, afterFirstToken, afterNonWhitespaceOnLine);
 
-            this.AddTrivia(directive, ref triviaList);
-            _directives = directive.ApplyDirectives(_directives);
+            if (!afterNonWhitespaceOnLine)
```
CyrusNajmabadi, line 2406: what happens to the directive if we are after non-whitespace?  we don't call .AddTrivia below.  is that ok?  can you comment why this is correct in the code?

src/Compilers/CSharp/Test/Syntax/Parsing/ParsingErrorRecoveryTests.cs
```diff
@@ -8110,5 +8110,109 @@ public void RazorCommentRecovery_NoStart()
             }
             EOF();
         }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75583")]
+        public void PreprocessorDirective_Trailing_01()
+        {
+            UsingTree("""
+                if (#if)
```
CyrusNajmabadi, line 8118: can you have variants with things like whitespace after the `(` and before/after the `#`.  i want to make sure all the positions/text are correct for all the tokens.

src/EditorFeatures/CSharpTest/Classification/SyntacticClassifierTests_Preprocessor.cs
```diff
@@ -1282,6 +1282,33 @@ await TestAsync(code,
...
+            #endif
+            ) { }
+            """;
+
+        await TestAsync(code,
+            testHost,
+            ControlKeyword("if"),
+            Punctuation.OpenParen,
+            Keyword("true"),
+            PPKeyword("#"),
```
CyrusNajmabadi, line 1302: confused by this.  shouldn't we just see this as one big skipped token now?

src/Compilers/CSharp/Portable/Parser/Lexer.cs
jjonescz: That's what it is; the min just ensures it goes to 0 if the text is empty (not sure if that can actually happen)

src/EditorFeatures/CSharpTest/Classification/SyntacticClassifierTests_Preprocessor.cs
jjonescz: We see the `#if false` as one skipped token. But we don't continue parsing disabled text, we act as if the directive wasn't there (because it's invalid), so the rest is parsed as if the `#if false` wasn't there.

src/Compilers/CSharp/Portable/Parser/Lexer.cs
jjonescz: Yes, we don't want the directive in the trivia, we just want to parse it and then discard it and just take the text we parsed and put it into a BadToken which we put into SkippedTokensTrivia - and on *that* we call AddTrivia.

src/Compilers/CSharp/Portable/Parser/Lexer.cs
CyrusNajmabadi: 1.  Oh duh, i misread min/max
2. You clearly can't have 0 length text, since you literally saw the # character :).  You can just use '1' here.

Do make sure that the error shows up on the right char though if there are things like spaces before the `#`.

src/Compilers/CSharp/Portable/Parser/Lexer.cs
CyrusNajmabadi: please doc.

src/EditorFeatures/CSharpTest/Classification/SyntacticClassifierTests_Preprocessor.cs
CyrusNajmabadi: i see.

src/Compilers/CSharp/Portable/Parser/Lexer.cs
AlekseyTs, line 1995: >width: Math.Min(1, text.Length) [](http://example.com/codeflow?start=36&length=31)

Why do we need this logic? Shouldn't we at least consume the `#` character?

src/Compilers/CSharp/Portable/Parser/Lexer.cs
```diff
@@ -2323,9 +2345,9 @@ private void LexDirectiveAndExcludedTrivia(
         {
             var directive = this.LexSingleDirective(true, true, afterFirstToken, afterNonWhitespaceOnLine, ref triviaList);
 
-            // also lex excluded stuff            
-            var branching = directive as BranchingDirectiveTriviaSyntax;
-            if (branching != null && !branching.BranchTaken)
+            // also lex excluded stuff
+            if (!afterNonWhitespaceOnLine &&
```
AlekseyTs, line 2349: >if (!afterNonWhitespaceOnLine && [](http://example.com/codeflow?start=12&length=32)

It feels strange to call `LexDirective**AndExcludedTrivia**` and then modify its code to not lex the excluded trivia. Consider reverting this change and calling ```LexSingleDirective``` instead.

src/Compilers/CSharp/Portable/Parser/Lexer.cs
AlekseyTs, line 2406: >if (!afterNonWhitespaceOnLine) [](http://example.com/codeflow?start=12&length=30)

I would like to suggest getting rid of the conditional logic around `afterNonWhitespaceOnLine` and stopping passing it around by refactoring the code a little bit:
1. Extract the code that produces ```directive``` above, plus ```_mode = saveMode;``` into a reusable helper with signature ```CSharpSyntaxNode ParseDirective(bool isActive, bool endIsActive, bool afterFirstToken)```. 
2. Call the helper here and from ```LexSyntaxTrivia``` when we are parsing the directive for error recovery.   
3. In ```LexSyntaxTrivia```, pass `false` for afterFirstToken to the helper. We need to decide what are the right values for ```isActive```, ```endIsActive```. Right now they both are implicitly `true` for that code path, but we need to make an explicit decision about each of them. We might end up using the same values, just need to understand why these are the right values for the case.
4. Get rid of the ```afterNonWhitespaceOnLine``` parameter from all the helpers and adjust their code as if its value was always false. This includes the target of ```_directiveParser.ParseDirective``` call.

src/Compilers/CSharp/Test/Syntax/Parsing/ParsingErrorRecoveryTests.cs
```diff
@@ -8110,5 +8110,109 @@ public void RazorCommentRecovery_NoStart()
...
+                        }
+                    }
+                }
+                N(SyntaxKind.EndOfFileToken);
+            }
+            EOF();
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75583")]
+        public void PreprocessorDirective_Trailing_02()
```
AlekseyTs, line 8165: Please test a scenarios that verify that skipped directives do not stop exclusion of code started by the previous directive. For example:
```
a();
#if false 
b();
/* comment */ #else
c();
#endif

Comment (AlekseyTs) [inline src/Compilers/CSharp/Test/Syntax/Parsing/ParsingErrorRecoveryTests.cs:8165]: >PreprocessorDirective_Trailing_02 [](http://example.com/codeflow?start=20&length=33)

Consider adding a test for each distinct directive kind and observe that their presence at the wrong position in line has no observable effects with exception of reported diagnostics. For example, ```#define``` doesn't define a symbol, etc.

**Comment by AlekseyTs** [comment]

File: src/Compilers/CSharp/Portable/Parser/Lexer.cs [hunk 7]
```diff
@@ -1980,7 +1980,31 @@ private void LexSyntaxTrivia(bool afterFirstToken, bool isTrailing, ref SyntaxLi
...
+                            if (isTrailing || !onlyWhitespaceOnLine)
+                            {
+                                // Directives cannot be in trailing trivia.
+                                // We parse the directive (ignoring its effects like disabled text or defined symbols).
+                                // We extract the text corresponding to the parsed directive
+                                // and add it as a BadToken in SkippedTokensTrivia.
+
+                                var savePosition = TextWindow.Position;
+
+                                this.ParseDirective(isActive: false, endIsActive: false, isFollowingToken);
```
AlekseyTs, line 1992: >isActive: false, endIsActive: fals [](http://example.com/codeflow?start=52&length=34)

It would be good to explain why we use these values. What better effect do they have by comparison to other values.

AlekseyTs, line 1992: >isFollowingToken [](http://example.com/codeflow?start=89&length=16)

I think this value is not consistent with the way we were parsing directive before. And it doesn't match what was suggested in the earlier comment.

AlekseyTs:
Done with review pass (commit 14)

src/Compilers/CSharp/Portable/Parser/Lexer.cs
jjonescz: They don't have any effect - they are only written into the resulting directive syntax node but we are discarding it.

jjonescz: However, since the directive is at invalid location, it seems better to parse it as "not active".

AlekseyTs: Please add a comment reflecting the rational behind the choice

AlekseyTs (approved):
LGTM (commit 15)

src/Compilers/CSharp/Portable/Parser/Lexer.cs
```diff
@@ -1980,7 +1980,33 @@ private void LexSyntaxTrivia(bool afterFirstToken, bool isTrailing, ref SyntaxLi
                     case '#':
                         if (_allowPreprocessorDirectives)
                         {
-                            this.LexDirectiveAndExcludedTrivia(afterFirstToken, isTrailing || !onlyWhitespaceOnLine, ref triviaList);
+                            if (isTrailing || !onlyWhitespaceOnLine)
```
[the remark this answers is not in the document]
333fred: Oh, you definitely are, but it's _fine_.

jjonescz:

cston (approved):

src/Compilers/CSharp/Portable/Parser/Lexer.cs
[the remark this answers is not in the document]
333fred: What are the rules for trailing newlines with this change? Does a newline get a separate token after, or is the newline consumed as part of the directive trivia?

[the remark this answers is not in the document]
jjonescz: It's part of the directive trivia:

image

You can try it out in the [lab](https://syntax-trivia.dotnetinternals.pages.dev/#4xLl4gooyk8vSszVSy4WYjdUUM5M4zISYjU3NTcyAQA) (the link is setup to use the compiler build from this PR)

PR #75764, opened 2024-11-06
EE: Support compact name in `IDkmLanguageInstructionDecoder.GetMethodName()` implementation

cston:
The compact name is simply the method metadata name without arguments or type arguments, or for an accessor, the property or event name.

The table below includes differences between the strings returned from `GetMethodName()` with and without `DkmVariableInfoFlags.CompactName`.

|Member|with CompactName|without|
|:---:|:---:|:---:|
|method|`M`|`C.M`|
|property accessor|`P`|`C.P.get`|
|indexer accessor|`this[]`|`C.this[int].get`|
|constructor|`C`|`C.C`|
|destructor|`~C`|`C.~C`|
|unary operator|`operator +`|`C.operator +`|
|conversion operator|`implicit operator int`|`C.implicit operator int`|
|...|...|...|


src/ExpressionEvaluator/Core/Source/ExpressionCompiler/LanguageInstructionDecoder.cs
```diff
@@ -38,16 +38,24 @@ string IDkmLanguageInstructionDecoder.GetMethodName(DkmLanguageInstructionAddres
                 // but it was ignored.  Furthermore, it's not clear what FullNames would mean with respect
                 // to argument names in C# or Visual Basic.  For consistency with the old behavior, we'll
                 // just ignore the flag as well.
-                Debug.Assert((argumentFlags & (DkmVariableInfoFlags.FullNames | DkmVariableInfoFlags.Names | DkmVariableInfoFlags.Types)) == argumentFlags,
+                Debug.Assert((argumentFlags & (DkmVariableInfoFlags.FullNames | DkmVariableInfoFlags.Names | DkmVariableInfoFlags.Types | DkmVariableInfoFlags.CompactName)) == argumentFlags,
```
jjonescz, line 41: Curious where will the CompactName be used exactly in VS/EE UI?

src/ExpressionEvaluator/CSharp/Source/ExpressionCompiler/CSharpInstructionDecoder.cs
```diff
@@ -36,7 +36,8 @@ private CSharpInstructionDecoder()
 
         internal override string GetCompactName(MethodSymbol method)
         {
-            return (method.AssociatedSymbol ?? method).Name;
+            var symbol = method.AssociatedSymbol ?? method;
+            return symbol.ToDisplayString(CompactNameFormat);
```
jjonescz, line 40: Why did you change this from `Name` to DisplayString?

cston: We're using `ToDisplayString()` so the member name in the string returned from `IDkmLanguageInstructionDecoder.GetMethodName()` is consistent regardless of whether `DkmVariableInfoFlags.CompactName` is used. See table in PR description.

src/ExpressionEvaluator/Core/Source/ExpressionCompiler/LanguageInstructionDecoder.cs
cston: TBD.

jjonescz (approved):

cston:

src/ExpressionEvaluator/Core/Source/ExpressionCompiler/LanguageInstructionDecoder.cs
jcouv, line 41: Could we be stricter than this? 
1. It looks like `DkmVariableInfoFlags.CompactName` should only be used by itself, is that right?
2. If so, we could throw (instead of merely asserting) if it's used in conjunction with other flags (since we'd ignore them anyways).

src/ExpressionEvaluator/Core/Source/ExpressionCompiler/InstructionDecoder.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.CodeAnalysis.PooledObjects;
 using Microsoft.CodeAnalysis.Symbols;
 using Microsoft.VisualStudio.Debugger.Clr;
+using Microsoft.VisualStudio.Debugger.Evaluation;
```
jcouv, line 10: Seems we don't need this

src/ExpressionEvaluator/Core/Source/ExpressionCompiler/LanguageInstructionDecoder.cs
jcouv: @jjonescz FWIW, here's the doc for the new flag:
```
  // Summary:
  //     If specified, the expression evaluator will create a very compact name for the
  //     frame. This is intended for UI presentation in cases where screen space is very
  //     limited. The suggested implementation is to return only the method name, with
  //     no namespace, class name, or generic arguments.
```

src/ExpressionEvaluator/CSharp/Test/ExpressionCompiler/InstructionDecoderTests.cs
```diff
@@ -402,6 +402,144 @@ class C
...
+                    static void M(this string x) { }
+                }
+                """;
+            var compilation = CreateCompilation(source);
+            var containingType = compilation.GlobalNamespace.GetTypeMember("C");
+            VerifyMethodName(compilation, containingType.GetMethod("F"), "C.F()", "F");
+            VerifyMethodName(compilation, containingType.GetMethod("get_P1"), "C.P1.get()", "P1");
+            VerifyMethodName(compilation, containingType.GetMethod("set_P2"), "C.P2.set(value)", "P2");
+            VerifyMethodName(compilation, containingType.GetMethod("set_P3"), "C.P3.init(value)", "P3");
+            VerifyMethodName(compilation, containingType.GetMethod("get_Item"), "C.this[int].get(i)", "this");
```
jcouv, line 440: >this [](http://example.com/codeflow?start=103&length=4)

nit: I'm a bit concerned about the indexer case. `this` doesn't really shout "indexer"...

src/ExpressionEvaluator/CSharp/Test/ExpressionCompiler/InstructionDecoderTests.cs
```diff
@@ -402,6 +402,144 @@ class C
...
+            VerifyMethodName(compilation, containingType.GetMethod("set_Item"), "C.this[int].set(i, value)", "this");
+            VerifyMethodName(compilation, containingType.GetMethod("add_E"), "C.E.add(value)", "E");
+            VerifyMethodName(compilation, containingType.GetMethod("remove_E"), "C.E.remove(value)", "E");
+            VerifyMethodName(compilation, compilation.GlobalNamespace.GetTypeMember("E").GetMethod("M"), "E.M(x)", "M");
+            VerifyMethodName(compilation, containingType.GetMethod(".cctor"), "C.C()", "C");
+            VerifyMethodName(compilation, containingType.GetMethod(".ctor"), "C.C(x)", "C");
+            VerifyMethodName(compilation, containingType.GetMethod("Finalize"), "C.~C()", "~C");
+            VerifyMethodName(compilation, containingType.GetMethod("op_UnaryPlus"), "C.operator +(c)", "operator +");
+            VerifyMethodName(compilation, containingType.GetMethod("op_Implicit"), "C.implicit operator int(c)", "implicit operator int");
```
jcouv, line 449: >implicit operator int [](http://example.com/codeflow?start=114&length=21)

Since you can't define both an implicit and an explicit overload of the same operator in C#, could we spare the `implicit`/`explicit` modifier?
Also, consider testing with `checked` user-defined operator (that one does come in pairs, so we may want to include `checked` modifier)

jcouv (commented):
Done with review pass (iteration 2)

src/ExpressionEvaluator/CSharp/Test/ExpressionCompiler/InstructionDecoderTests.cs
cston: > Also, consider testing with `checked` user-defined operator (that one does come in pairs, so we may want to include `checked` modifier)

I don't think we need to include modifiers to disambiguate one method from another. We're not including parameters for example, so overloaded methods will be ambiguous.

src/ExpressionEvaluator/Core/Source/ExpressionCompiler/LanguageInstructionDecoder.cs
jcouv: Also, it looks like we don't have any direct tests on this method. Is that something we could cover?

cston: We could be stricter and throw an exception if certain values are used in combination with `DkmVariableInfoFlags.CompactName`. But there might be reasons in the future to combine `CompactName` with other values, including existing values, that we'd disallow as a result.

src/ExpressionEvaluator/CSharp/Test/ExpressionCompiler/InstructionDecoderTests.cs
cston: > Since you can't define both an implicit and an explicit overload of the same operator in C#, could we spare the `implicit`/`explicit` modifier?

Removing `implicit` and `explicit` would require modifying the set of tokens generated by `SymbolDisplay`. For now, let's keep the modifier.

src/ExpressionEvaluator/CSharp/Test/ExpressionCompiler/InstructionDecoderTests.cs
cston: Replaced with `this[]`.

src/ExpressionEvaluator/CSharp/Source/ExpressionCompiler/CSharpInstructionDecoder.cs
```diff
@@ -34,6 +34,18 @@ private CSharpInstructionDecoder()
             AddMemberOptions(SymbolDisplayMemberOptions.IncludeParameters).
             WithParameterOptions(SymbolDisplayParameterOptions.IncludeType);
 
+        internal static readonly SymbolDisplayFormat s_indexerCompactNameFormat = CompactNameFormat.
+            WithMemberOptions(SymbolDisplayMemberOptions.IncludeParameters);
```
jcouv, line 38: nit: may be worth a comment or making `SymbolDisplayParameterOptions.None` explicit for clarity

jcouv (approved):
LGTM Thanks (iteration 4)

PR #75789, opened 2024-11-06
Update breaking change docs

BillWagner:
Offline, we had a discussion that the title and H1 were verbose and could be misunderstood.

I've updated the most recent 4 documents to match that discussion.

Notes for reviewers:

All edits in each file should be the same (modulo typos and version numbers). Let's agree on the language on one of them, then make sure they all match.

### DIFF
--- a/docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
+++ b/docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
@@ -1,6 +1,6 @@
 # Roslyn breaking changes from .NET 9.0.100 through .NET 10.0.100
 
-This document lists known breaking changes in Roslyn after .NET 9 general release (.NET SDK version 9.0.100) through .NET 10 general release (.NET SDK version 10.0).100).
+This document lists known breaking changes in Roslyn after .NET 9 general release (.NET SDK version 9.0.100) through .NET 10 general release (.NET SDK version 10.0.100).
 
 ## `Span<T>` and `ReadOnlySpan<T>` overloads are applicable in more scenarios in C# 14 and newer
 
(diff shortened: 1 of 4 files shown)


BillWagner:
PTAL @jcouv @333fred

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -1,4 +1,6 @@
-# This document lists known breaking changes in Roslyn after .NET 9 all the way to .NET 10.
+# Roslyn breaking changes from .NET 9.0.100 through .NET 10.0.100
```
jcouv, line 1: I think we should say "after .NET 9.0.100" (9.0.100 is included in previous document, but excluded from this document)

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -1,4 +1,6 @@
...
+
+This document lists known breaking changes in Roslyn after .NET 9 general release (.NET SDK version 9.0.100) through .NET 10 general release (.NET SDK version 10.0.100).
```
jcouv, line 3: As a side note, Bill, I don't remember whether we'd decided on chronological or reverse chronological order for entries in each document. We've been appending at the end (ie. chronological). Does that match your expectation?

BillWagner: That works. We've had no concerns on the order.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 7.md
```diff
@@ -1,5 +1,7 @@
 # This document lists known breaking changes in Roslyn after .NET 6 all the way to .NET 7.
```
333fred, line 1: Is this title going to be updated too?

jcouv (approved):
LGTM Thanks (iteration 4)

333fred (approved):

PR #75794, opened 2024-11-06
Bump Razor to 9.0.0-preview.24555.12

davidwengier:
The previous bump was because of a serialization change from v6 to v7. Sadly RPS meant that change got reverted, so this takes us back to v6.

I suspect we'll be back at v7 soon, but preparing this so Razor insertions are not otherwise blocked.

### DIFF
--- a/eng/Directory.Packages.props
+++ b/eng/Directory.Packages.props
@@ -144,7 +144,7 @@
     <PackageVersion Include="NuGet.ProjectModel" Version="6.8.0-rc.112" />
     <PackageVersion Include="Microsoft.TestPlatform.TranslationLayer" Version="$(_MicrosoftTestPlatformVersion)" />
     <PackageVersion Include="Microsoft.TestPlatform.ObjectModel" Version="$(_MicrosoftTestPlatformVersion)" />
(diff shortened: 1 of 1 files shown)


jjonescz (approved):

dibarbet:
@davidwengier does this need to go in before we can insert roslyn into vscode?  theres a high pri change we need to get in before snap on friday

davidwengier:
No, current inserted razor works with current builds of Roslyn. This will be needed if we have to take a newer razor though

dibarbet:
> No, current inserted razor works with current builds of Roslyn. This will be needed if we have to take a newer razor though

We will be taking a newer version of Razor for the telemetry fix as well - this is the change - https://github.com/dotnet/razor/pull/11181

dibarbet (approved):

davidwengier:
I think @ryzngard did the right thing with that one, I think it should be fine to not take this. In fact, taking this would break that!

ryzngard:
> I think @ryzngard did the right thing with that one, I think it should be fine to not take this. In fact, taking this would break that!

They in fact did not....

PR #75795, opened 2024-11-07
Make MaxSupportedLangVersion calculation dynamic

kasperk81:
Uses historical trend to calculate language version based on the .NET version, removing hardcoded mappings. Ensures future .NET versions automatically align with expected C# version.

### DIFF
--- a/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
+++ b/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
@@ -7,31 +7,20 @@
     <!-- .NETCoreApp < 3.0, .NETStandard < 2.1, or any other target framework -->
     <_MaxSupportedLangVersion Condition="('$(TargetFrameworkIdentifier)' != '.NETCoreApp' OR '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0') AND
                                          ('$(TargetFrameworkIdentifier)' != '.NETStandard' OR '$(_TargetFrameworkVersionWithoutV)' &lt; '2.1')">7.3</_MaxSupportedLangVersion>
-    
+
     <!-- .NETCoreApp < 5.0, .NETStandard == 2.1 -->
     <_MaxSupportedLangVersion Condition="(('$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(_TargetFrameworkVersionWithoutV)' &lt; '5.0') OR
                                           ('$(TargetFrameworkIdentifier)' == '.NETStandard' AND '$(_TargetFrameworkVersionWithoutV)' == '2.1')) AND
                                           '$(_MaxSupportedLangVersion)' == ''">8.0</_MaxSupportedLangVersion>
 
-    <!-- .NETCoreApp == 5.0 -->
-    <_MaxSupportedLangVersion Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(_TargetFrameworkVersionWithoutV)' == '5.0' AND
-                                         '$(_MaxSupportedLangVersion)' == ''">9.0</_MaxSupportedLangVersion>
-
-    <!-- .NETCoreApp == 6.0 -->
-    <_MaxSupportedLangVersion Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(_TargetFrameworkVersionWithoutV)' == '6.0' AND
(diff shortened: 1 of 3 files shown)


src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
```diff
@@ -7,31 +7,35 @@
...
+      Based on the empirical pattern, each major .NET version release supports 
+      a new language version. For example:
+      - .NET 5.0 -> C# 9.0
+      - .NET 6.0 -> C# 10.0
+      - .NET 7.0 -> C# 11.0
+      - .NET 8.0 -> C# 12.0
+      This pattern suggests a consistent, incremental increase in supported language versions.
+      Starting at 9 for .NET 5.0, we can compute the language version automatically for future versions.
+    -->
+    <_BaseLangVersion>9</_BaseLangVersion>
```
jjonescz, line 27: Why do we have a variable for the base C# version, but not the base .NET version? Consider simplifying by inlining the "9" into the formula.

src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
```diff
@@ -7,31 +7,20 @@
...
-    <_MaxSupportedLangVersion Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(_TargetFrameworkVersionWithoutV)' == '9.0' AND
-                                         '$(_MaxSupportedLangVersion)' == ''">13.0</_MaxSupportedLangVersion>
+    <!-- 
+      Automatically calculate the maximum supported C# language version based on the .NET Target Framework.
+      - Pattern: .NET 5.0 uses C# 9.0, .NET 6.0 uses C# 10.0, and so on.
+      - Starting from C# 9.0 for .NET 5.0, we add the difference between the major .NET version and 5 
+        to determine the correct language version.
+    -->
+    <_MaxSupportedLangVersion Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND
+                                         '$(_MaxSupportedLangVersion)' == ''">$([MSBuild]::Add(9, $([MSBuild]::Subtract($(_TargetFrameworkVersionWithoutV), 5)))).0</_MaxSupportedLangVersion>
```
jaredpar, line 23: Won't this break once users start targeting `net10.0`? That would lead to the language version being 14.0 and that will not exist, by explicit design, until June 2025.

jaredpar (changes requested):
Pretty sure this will break the moment people start targeting `net10.0`

jaredpar (approved):

jaredpar:
@jcouv want to make sure you've seen this. MSBuild logic looks good but you often do the lang version updates so want to make sure you see the new setup.

src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
```diff
@@ -7,31 +7,25 @@
...
+
+    <!-- Cap _MaxSupportedLangVersion if it exceeds _MaxAvailableLangVersion -->
+    <_MaxAvailableLangVersion>13.0</_MaxAvailableLangVersion>
```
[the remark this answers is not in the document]
333fred: If we still need to update this file every year when we increase the max language version, then what do we actually save by making this change over the current system?

jaredpar:
@jcouv for second sign off

jcouv (dismissed):
LGTM Thanks (iteration 4)

src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
jcouv, line 26: The comment on the C# test named `LanguageVersionAdded_Canary` should be updated to say `_MaxAvailableLangVersion` instead of `MaxSupportedLangVersion`. Those are the instructions we follow when adding a new language version.

jcouv (commented):
Done with review pass (iteration 4)

src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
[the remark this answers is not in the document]
jcouv: > one [update] when a new framework version is released (around October/November, managed by dotnet/sdk)

@kasperk81 Would you mind sharing a link to one of those previous dotnet/sdk changes? That'll help provide more context

jcouv (approved):
LGTM Thanks (iteration 5)

PR #75816, opened 2024-11-08
Add runtime async design doc

333fred:
Adding an initial design document for runtime async. I'm not making a branch for this yet so I don't have to worry about keeping it up to date until we're actually ready to start implementation work; that should begin soon, but I think getting this document into `main` and starting collaboration with the runtime and debugger folks will help that proceed.

### DIFF
--- a/docs/compilers/CSharp/Runtime Async Design.md
+++ b/docs/compilers/CSharp/Runtime Async Design.md
@@ -1,0 +1,268 @@
+# Runtime Async Design
+
+See also the ECMA-335 specification change for this feature: https://github.com/dotnet/runtime/pull/104063, https://github.com/dotnet/runtime/blob/main/docs/design/specs/runtime-async.md (when the PR completes).
+
+This document goes over the general design of how Roslyn works with the Runtime Async feature to produce IL. In general, we try to avoid exposing this feature at the user level; initial binding is almost entirely
+unaffected by runtime async. Exposed symbols do not give direct information about whether they were compiled with runtime async, and indeed the compiler has no idea whether a method from a referenced assembly is
+compiled with runtime async or not.
+
+## Supporting runtime apis
+
+We use the following runtime flag to drive whether feature can be used. This flag must be defined in the same assembly that defines `object`, and the assembly cannot reference any other assemblies. In terms of
+CoreFX, this means it must be defined in the `System.Runtime` reference assembly.
+
+TODO: Determine whether just the presence of this flag will cause the compiler to generate in runtime async mode.
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+public static class RuntimeFeature
+{
+    public const string RuntimeAsync = nameof(RuntimeAsync);
+}
+```
+
+We use the following helper APIs to indicate suspension points to the runtime, in addition to the runtime async call syntax:
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+// TODO: We need to determine what type contains these methods
+// These methods are used to await things that cannot use runtime async signature form, whether that's a local/parameter/field/etc of type `Task`/`ValueTask`, or a method that returns an awaitable that is
+// not `Task` or `ValueTask`.
+public static Task AwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter) where TAwaiter : INotifyCompletion
+public static Task UnsafeAwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter) where TAwaiter : ICriticalNotifyCompletion
+```
+
+Additionally, we use the following helper attributes to indicate information to the runtime. If these attributes are not present in the reference assemblies, we will generate them; the runtime matches by full
+name, not by type identity, so we do not need to care about using the "canonical" versions.
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+// Used to tell the runtime to generate the async state machinery for this method
+[AttributeUsage(AttributeTargets.Method)]
+public class RuntimeAsyncMethodAttribute() : Attribute();
+
+// Used to mark locals that should be hoisted to the generated async closure. Note that the runtime does not guarantee that all locals marked with this attribute will be hoisted; if it can prove that it
+// doesn't need to hoist a variable, it may avoid doing so.
+public class HoistedLocal();
+```
+
+For experimentation purposes, we recognize an attribute that can be used to force the compiler to generate the runtime async code, or to force the compiler to generate a full state machine. This attribute is not
+defined in the BCL, and exists as an escape hatch for experimentation. It may be removed when the feature ships in stable.
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+[AttributeUsage(AttributeTargets.Method)]
+public class RuntimeAsyncMethodGenerationAttribute(bool runtimeAsync) : Attribute();
+```
+
+## Transformation strategy
+
+As mentioned previously, we try to expose as little of this to initial binding as possible. The one major exception to this is our handling of the `RuntimeAsyncMethodAttribute`; we do not let this be applied to
+user code, and will issue an error if a user tries to do this by hand.
+
+Compiler generated async state machines and runtime generated async share some of the same building blocks. Both need to have `await`s with in `catch` and `finally` blocks rewritten to pend the exceptions,
+perform the `await` outside of the `catch`/`finally` region, and then have the exceptions restored as necessary.
+
+TODO: Go over `IAsyncEnumerable` and confirm that the initial rewrite to a `Task`-based method produces code that can then be implemented with runtime async, rather than a full compiler state machine.
+
+TODO: Clarify with the debugger team where NOPs need to be inserted for debugging/ENC scenarios.
+
+### Example transformations
+
+Below are some examples of what IL is generated for specific examples.
+
+TODO: Include debug versions
+
+#### Await `Task`-returning method
+
+```cs
+class C
+{
+    static Task M();
+}
+
+await C.M();
+```
+
+```il
+call modreq(class [System.Runtime]System.Threading.Tasks.Task) void C::M()
+```
+
+---------------------------
+
+```cs
+class C
+{
+    Task M();
+}
+
+var c = new();
+await c.M();
+```
+
+```il
+newobj instance void C::.ctor()
+callvirt instance modreq(class [System.Runtime]System.Threading.Tasks.Task) void C::M()
+```
+
+#### Await a `Task<T>`-returning method
+
+```cs
+class C
+{
+    static Task<int> M();
+}
+
+await C.M();
+```
+
+```il
+call modreq(class [System.Runtime]System.Threading.Tasks.Task`1<int32>) int32 C::M()
+```
+
+---------------------------
+
+```cs
+class C
+{
+    Task<int> M();
+}
+
+var c = new();
+await c.M();
+```
+
+```il
+newobj instance void C::.ctor()
+callvirt instance modreq(class [System.Runtime]System.Threading.Tasks.Task`1<int32>) int32 C::M()
+```
+
+#### Await local of type `Task`
+
+```cs
+class C
+{
+    static Task M();
+}
+
+var local = M();
+await local;
+```
+
+```il
+{
+    .locals init (
+        [0] valuetype [System.Runtime]System.Runtime.CompilerServices.TaskAwaiter
+    )
+
+    IL_0000: call class [System.Runtime]System.Threading.Tasks.Task C::M()
+    IL_0005: callvirt instance valuetype [System.Runtime]System.Runtime.CompilerServices.TaskAwaiter class [System.Runtime]System.Threading.Tasks.Task::GetAwaiter()
+    IL_000a: stloc.0
+    IL_000b: ldloca.s 0
+    IL_000d: call modreq([System.Runtime]System.Threading.Tasks.Task) void <TODO>::AwaitAwaiterFromRuntimeAsync<valuetype [System.Runtime]System.Runtime.CompilerServices.TaskAwaiter>(!0)
+    IL_0012: ret
+}
+```
+
+#### Awaiting in a `catch` block
+
+```cs
+class C
+{
+    static Task M();
+}
+
+try
+{
+    throw new Exception();
+}
+catch (Exception ex)
+{
+    await C.M();
+    throw;
+}
+```
+
+```il
+{
+    .locals init (
+        [0] int32 pendingCatch,
+        [1] class [System.Runtime]System.Exception pendingException
+    )
+
+    .try
+    {
+        IL_0000: newobj instance void [System.Runtime]System.Exception::.ctor()
+        IL_0005: throw
+    }
+    catch [System.Runtime]System.Exception
+    {
+        IL_0006: stloc.1
+        IL_0007: ldc.i4.1
+        IL_0008: stloc.0
+        IL_0009: leave.s IL_000b
+    }
+
+    IL_000b: ldloc.0
+    IL_000c: ldc.i4.1
+    IL_000d: bne.un.s IL_0017
+
+    IL_000f: ldloc.1
+    IL_0010: call modreq(class [System.Runtime]System.Threading.Tasks.Task) void C::M()
+    IL_0015: pop
+    IL_0016: throw
+
+    IL_0017: ret
+}
+```
+
+#### Awaiting in a `finally` block
+
+```cs
+class C
+{
+    static Task M();
+}
+
+try
+{
+    throw new Exception();
+}
+finally
+{
+    await C.M();
+}
+```
+
+```il
+{
+    .locals init (
+        [0] class [System.Runtime]System.Exception pendingException
+    )
+
+    .try
+    {
+        IL_0000: newobj instance void [System.Runtime]System.Exception::.ctor()
+        IL_0005: throw
+    }
+    catch [System.Runtime]System.Exception
+    {
+        IL_0006: stloc.0
+        IL_0007: leave.s IL_0009
+    }
+
+    IL_0009: call modreq(class [System.Runtime]System.Threading.Tasks.Task) void C::M()
+    IL_000e: pop
+    IL_000f: ldloc.0
+    IL_0010: brfalse.s IL_0014
+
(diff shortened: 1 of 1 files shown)


thaystg:
I'm not super sure you can correct me if I'm wrong, but as far as I understood we can keep generating debug information as they are already generated by the older async.
I tested that keeping the "Method Stepping Information" helps VS to support the stepping on an await instruction by checking the information generated in the PDB. I kept the generation of AwaitYieldPoint and AwaitResumePoint on roslyn to test that this would help the debugger experience.
I'm not sure about hot reload as we didn't study it yet, but I think that if we can keep the same information generated for the old async will also help the hot reload support.

333fred:
@thaystg we can certainly keep those NOOPs, but it would be good to define precisely where you expect them to be emitted; that's one of the purposes of this document. For example, the awaiting a `Task` variable example; where in that IL do you want to see NOOPs to inform the debugger?

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+    IL_000e: pop
+    IL_000f: ldloc.0
+    IL_0010: brfalse.s IL_0014
+
+    IL_0012: ldloc.0
+    IL_0013: throw
+
+    IL_0014: ret
+}
+```
```
jcouv, line 456: Just to take a note so we don't forget: the compiler implements some (imperfect) clearing of hoisted locals when exiting scopes. I'm currently looking at strengthening that. We should keep track of what we want to do about that in async2 (will the runtime handle or do we still want the compiler to handle)?
For reference, see `MethodToStateMachineRewriter.AddVariableCleanup`

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+
+Additionally, we use the following helper attributes to indicate information to the runtime. If these attributes are not present in the reference assemblies, we will generate them; the runtime matches by full
+name, not by type identity, so we do not need to care about using the "canonical" versions.
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+// Used to tell the runtime to generate the async state machinery for this method
+[AttributeUsage(AttributeTargets.Method)]
+public class RuntimeAsyncMethodAttribute() : Attribute();
```
jcouv, line 51: It would be good to have an example of C# code and corresponding emitted signature with this attribute.
I think that signature should also be emitted with the Experimental attribute in .NET 10.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+public class RuntimeAsyncMethodGenerationAttribute(bool runtimeAsync) : Attribute();
+```
+
+## Transformation strategy
+
+As mentioned previously, we try to expose as little of this to initial binding as possible. The one major exception to this is our handling of the `RuntimeAsyncMethodAttribute`; we do not let this be applied to
+user code, and will issue an error if a user tries to do this by hand.
+
+Compiler generated async state machines and runtime generated async share some of the same building blocks. Both need to have `await`s with in `catch` and `finally` blocks rewritten to pend the exceptions,
+perform the `await` outside of the `catch`/`finally` region, and then have the exceptions restored as necessary.
```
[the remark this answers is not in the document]
333fred: @jakobbotsch this is the equivalent straight-line code. I think you're saying that those `ldflda` calls will need to be redone across the suspension point?

```il
        IL_0000: ldarg.0
        IL_0001: ldflda valuetype S C::S
        IL_0006: ldflda int32 S::X
        IL_000b: dup
        IL_000c: ldind.i4
        IL_000d: call int32 Program::Bar()
        IL_0012: add
        IL_0013: stind.i4
        IL_0014: ret
```

docs/compilers/CSharp/Runtime Async Design.md
333fred: > It would be good to have an example of C# code and corresponding emitted signature with this attribute.

Not sure what you're asking for here. The method signature is unchanged, the attribute is just added.

> I think that signature should also be emitted with the Experimental attribute in .NET 10.

Presumably, we still need to decide exactly what the semantics will be here.

> Are we planning an update (or ask for an update) to ILVerify?

I believe that was discussed, though I don't remember who was taking it on. @agocke, do you know?

docs/compilers/CSharp/Runtime Async Design.md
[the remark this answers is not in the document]
jcouv: From what I recall, a good way to see the ref scenario is `localArray[B] += await C;`. In normal code, we save the ref from indexing into array, but in async code, we'll obtain that ref before the suspension for `await` and again after resuming.

[sharplab](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQAwAIpwKwG5lQDMmM6AwugN7Lq2bFQAcmAbOgLIAUAlOjXdSR1h6AJYA7AC4BtALroAhugC86cQFMA7mKlyq6AL74hI2gHoz6ACoALdTuDqAHhIDmOyQHtFAJx8KATzEAZ3R1ADcFABsAVwVJdWB0SU1RAGN7flMLdGCJDPRNew1E5Lt0H3UAM3QAI2rPSsVxJIUqhJ8y+2CY4IAHdXE8z3E6mMl0NIVxcU8Jm09RYInRSSyRBWl2GB55MFUoAE5WAB4JSQA+ADoAMR9PAFsAJXUeqMlOABYYbmNhA3WREwcDY5w4O14lEwAHZ0HBcIZkAYgA=)
```csharp
using System;
using System.Threading.Tasks;
public class C {
    public async Task M() 
    {
        int[] a = new int[] { };
        // The indexing into array is evaluated twice 
        // since we need the ref before and after the suspension but cannot hoist it
        a[M2()] += await Task<int>.FromResult(42);
    }
    public static int M2() { return 1; }
}
```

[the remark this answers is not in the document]
jakobbotsch: > @jakobbotsch this is the equivalent straight-line code. I think you're saying that those `ldflda` calls will need to be redone across the suspension point?
> 
> ```
>         IL_0000: ldarg.0
>         IL_0001: ldflda valuetype S C::S
>         IL_0006: ldflda int32 S::X
>         IL_000b: dup
>         IL_000c: ldind.i4
>         IL_000d: call int32 Program::Bar()
>         IL_0012: add
>         IL_0013: stind.i4
>         IL_0014: ret
> ```

Right, exactly. There are more complex cases like @jcouv pointed out, or even impossible cases like `SomethingThatReturnsRefInt() += await Foo()` (this raises a compiler error today).

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+
+```cs
+
+class C
+{
+    static Task M();
+}
+
+Func<Task> d = C.M;
+await d;
```
jcouv, line 309: I didn't follow. Why is this expected to work?
This would not work in compiler-async today: [sharplab](https://sharplab.io/#v2:D4AQDABCCMCsDcBYAUCATNFKSwDwgDYA+CAEwgF4IBhAOgFklUBOMp7NGlAbxQn6gBmKNAJQx9ABQBKSiQAuACwBOAewDuEAHYBXADZ6mAXyzIgA)

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+
+---------------------------
+
+```cs
+class C
+{
+    Task M();
+}
+
+var c = new();
```
jcouv, line 110: `new()` => `new C()`

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+
+#### Await a concrete `T` `Task<T>`-returning method
+
+```cs
+class C
+{
+    static Task<int> M();
+}
+
+await C.M();
```
jcouv, line 127: nit: Consider assigning the resulting `int` to a local.
Same comment applies to other `Task'1` scenarios

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+
+#### Await local of concrete type `Task<T>`
+
+```cs
+class C
+{
+    static Task<int> M();
+}
+
+var local = M();
```
jcouv, line 208: `M()` =>`C.M()`
Also, consider moving the statements above the type

docs/compilers/CSharp/Runtime Async Design.md
jcouv, line 456: How will we lower an `await` on a custom awaitable type with proper `GetAwaiter()`? I assume the same way described in section "Await local of concrete type `Task<T>`". Is that right?

We'll have to figure out what triggers the new special behavior for Task/Task<T>: are those special types or well-known types? I expect the runtime will need that we recognize only the special types (ie. Task/Task<T> from corlib)

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+defined in the BCL, and exists as an escape hatch for experimentation. It may be removed when the feature ships in stable.
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+[AttributeUsage(AttributeTargets.Method)]
+public class RuntimeAsyncMethodGenerationAttribute(bool runtimeAsync) : Attribute();
+```
+
+## Transformation strategy
```
jcouv, line 68: It would be good to describe the transformation strategy for `await`, which would explain the examples before.
From my understanding we have the following cases:
1. invocation of method returning the special types Task/Task<T>: emit as a call, but with special signature/convention
2. all other cases: use helper

Consider calling out that compiler-async methods require no changes, and can call and use the results of runtime-async methods.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
+# Runtime Async Design
+
+See also the ECMA-335 specification change for this feature: https://github.com/dotnet/runtime/pull/104063, https://github.com/dotnet/runtime/blob/main/docs/design/specs/runtime-async.md (when the PR completes). https://github.com/dotnet/runtime/issues/109632 tracks open issues for the feature in the runtime.
```
jcouv, line 3: >(when the PR completes) [](http://example.com/codeflow?start=187&length=23)

nit: The PR was merged now

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+name, not by type identity, so we do not need to care about using the "canonical" versions.
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+// Used to tell the runtime to generate the async state machinery for this method
+[AttributeUsage(AttributeTargets.Method)]
+public class RuntimeAsyncMethodAttribute() : Attribute();
+
+// Used to mark locals that should be hoisted to the generated async closure. Note that the runtime does not guarantee that all locals marked with this attribute will be hoisted; if it can prove that it
```
jcouv, line 53: >locals marked with this attribute [](http://example.com/codeflow?start=128&length=33)

"this attribute" was confusing. This will be `modreq(HoistedLocal)`.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+This document goes over the general design of how Roslyn works with the Runtime Async feature to produce IL. In general, we try to avoid exposing this feature at the user level; initial binding is almost entirely
+unaffected by runtime async. Exposed symbols do not give direct information about whether they were compiled with runtime async, and indeed the compiler has no idea whether a method from a referenced assembly is
+compiled with runtime async or not.
+
+## Supporting runtime apis
+
+We use the following runtime flag to drive whether feature can be used. This flag must be defined in the same assembly that defines `object`, and the assembly cannot reference any other assemblies. In terms of
+CoreFX, this means it must be defined in the `System.Runtime` reference assembly.
+
+TODO: Determine whether just the presence of this flag will cause the compiler to generate in runtime async mode.
```
jaredpar, line 14: My 2 cents: I think the driving scenario here is going to be multi-targeting. Consider for example if I have an application that targets `net9.0` and `net11.0`. The build needs to be functional when having this multi-target and an explicit langversion of 14.0. Ideally for the `net9.0` target async v1 code generation occurs and for `net11.0` we get async v2 generation. 

Given that I believe that we should gate async v2 code generation on effectively

```csharp
bool IsAsyncV2 => LangVersion >= 14 && IsRuntimeFeatureFlagPresent(Async);
```


Note: my comment here specifically uses `net11.0` because that is where the feature flag will be present in the standard core lib. I would expect the experience with `net10.0` + use the async2 preview build flag matches `net11.0` 

Note2: I suspect that in the short term we will want a feature flag to explicitly use async v1 code generation. Basically override the defaults. Having that will allow us to quickly toggle between async v1 and v2 generation. Suspect that will be useful for quickly ruling out / in whether a behavior is tied to the new model or not.

jaredpar: The async v2 model is going to be a _highly compatible_ design but not a _fully compatible_ one. There will be a few rough edges that will some segment of customers up. From that angle I can see this as being a language version issue because it allows a customer to separate out movement to a new runtime vs. a new language version of `async`. 

Irrespective of whether it's a language version feature, I do think we are going to want a feature flag that allows us to tweak the new codegen off. Particularly during `net10.0` and early `net11.0` it will be an effective way to help diagnose issues.

agocke: Regardless of our goals I also suspect there will be bugs that will push for a flag to turn things on/off. I think an attribute like `MethodImplOptions` at the method level might even be appropriate, although I haven't given it much thought.

333fred: > I think an attribute like MethodImplOptions at the method level might even be appropriate

Yes, I am assuming we will have this, and it's included in this document.

333fred: > I do think we are going to want a feature flag that allows us to tweak the new codegen off. Particularly during net10.0 and early net11.0 it will be an effective way to help diagnose issues.

Oh, definitely. I figure for "turn off for the entire assembly", it'll be a `Features` element, and for individual methods, we'll use the attribute I defined in this doc.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,456 @@
...
+
+## Transformation strategy
+
+As mentioned previously, we try to expose as little of this to initial binding as possible. The one major exception to this is our handling of the `RuntimeAsyncMethodAttribute`; we do not let this be applied to
+user code, and will issue an error if a user tries to do this by hand.
+
+Compiler generated async state machines and runtime generated async share some of the same building blocks. Both need to have `await`s with in `catch` and `finally` blocks rewritten to pend the exceptions,
+perform the `await` outside of the `catch`/`finally` region, and then have the exceptions restored as necessary.
+
+TODO: Go over `IAsyncEnumerable` and confirm that the initial rewrite to a `Task`-based method produces code that can then be implemented with runtime async, rather than a full compiler state machine.
```
jcouv, line 76: FWIW, I know this is not blocking early work, but for peace of mind I took a stab at this to show how it could work: [https://gist.github.com/jcouv/aa5359db1b15522013cf19ef52dcc18f](https://gist.github.com/jcouv/aa5359db1b15522013cf19ef52dcc18f)

docs/compilers/CSharp/Runtime Async Design.md
333fred: > How will we lower an await on a custom awaitable type with proper GetAwaiter()? I assume the same way described in section "Await local of concrete type Task<T>". Is that right?

Correct. I'll add that.

jcouv (approved):
LGTM Thanks (iteration 7)

docs/compilers/CSharp/Runtime Async Design.md
VSadov: I do not think clearing hoisted locals is going to be interesting for runtime async. 
* you would not know exactly what JIT has hoisted (could even be different between quick JIT and when re-jitted with more optimizations). 
* you would be clearing actual locals, not the storage in a display type. To the JIT it will appear as dead stores - storing null values that are not subsequently used. Most likely the JIT will just optimize them away.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,543 @@
...
+int i = await C.M();
+
+class C
+{
+    static Task<int> M();
+}
+```
+
+```il
+call modreq(class [System.Runtime]System.Threading.Tasks.Task`1<int32>) int32 C::M()
```
VSadov, line 161: This would be just 
```
modreq(class [System.Runtime]System.Threading.Tasks.Task`1)  int32 . . . 
```

I.E. in generic case the modreq uses an open type. 

Otherwise we would need to think about mismatched combinations like `modreq<int> long Foo()` .
Not specifying element type in modreq avoids the problem.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -0,0 +1,543 @@
...
+
+#### Await a generic `T` `Task<T>`-returning method
+
+```cs
+await C.M<int>();
+
+class C
+{
+    static Task<T> M<T>();
+}
```
VSadov, line 309: This is just a regular case. Substituting the element type is not controversial.   (unlike substituting the entire return type)

VSadov:
Looks good! With a couple of comments.

docs/compilers/CSharp/Runtime Async Design.md
333fred: If that's the case, then we need to update the runtime document. It currently says this:

> * Otherwise, the return type is the type argument of the return type (either Task`1 or ValueTask`1) with the first custom modifier on the return type being the original type.

I interpret "the original type" to be the full original return type, which is also why I haven't filled in the `Task<T>` generic scenario. If what we really want is the open type, that's perfectly fine, but the spec needs to be clear :slightly_smiling_face:.

PR #75819, opened 2024-11-08
Enable TypeDefinitionProvider LSP capability

JoeRobich:
Also changes GoToTypeDefinition handler's behavior to lookup the symbol's type and return its location

https://github.com/user-attachments/assets/538d0e32-38c0-45c5-897c-a9510664fa1a

### DIFF
--- a/src/LanguageServer/ProtocolUnitTests/Definitions/GoToTypeDefinitionTests.cs
+++ b/src/LanguageServer/ProtocolUnitTests/Definitions/GoToTypeDefinitionTests.cs
@@ -4,10 +4,12 @@
 
 #nullable disable
 
+using System.Collections.Immutable;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Roslyn.Test.Utilities;
+using Roslyn.Test.Utilities.TestGenerators;
 using Xunit;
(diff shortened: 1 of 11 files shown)


src/LanguageServer/ProtocolUnitTests/Definitions/GoToTypeDefinitionTests.cs
```diff
@@ -81,6 +156,96 @@ class B
...
+                Dim {|caret:a|} As A
+            End Class
+        </Document>
+    </Project>
+</Workspace>";
+            await using var testLspServer = await CreateXmlTestLspServerAsync(markup, mutatingLspWorkspace);
+
+            var results = await RunGotoTypeDefinitionAsync(testLspServer, testLspServer.GetLocations("caret").Single());
+            AssertLocationsEqual(testLspServer.GetLocations("definition"), results);
+        }
```
dibarbet, line 247: nice tests! can we potentially add one for metadata?  Some examples here in case you need it - https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/ProtocolUnitTests/Metadata/LspMetadataAsSourceWorkspaceTests.cs

dibarbet (approved):

Issue #67435, opened 2023-03-22
A possible ref safety "hole" around UnscopedRef method invoked on an expression classified as a value of a struct type 


AlekseyTs:
Compile and run the following code:

https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4GIaoDZ4TB4CmABACZQDORpAsAFBMuMACATGQMJMDeTMkLJsAjADYyAZTIBZUQAoAlGQC8APjIwSAd2nKyfMgDE1ZURbIBfANythIiWVgI5HBchIAzZzFcAPOGkyAE8lQWEBRgcHNgB2Mn87aOEre2ExSRc5AGZlCKEomIyE2XdPH3llADoAJW9lIIpvCHwEJWSHNOYUoUzfV1kAFnzewwKY+LcPbzlFJTqGpSDtPSkDI1NVcw4c6w6J7on+7NkAVlGHIuKyADcIZESzVf0VTbNRDiHrTpupspmlXmiy8jUSBzG3QcxycbG+smgMEukQmsVE1TEAE4FLI8koITdHBjRNjhsoCX90VicRd8b8hN0jj02HtqAg0ABjVxSfjHPbZYzJCYAbTE1QAIlAIABzGAAezZUA51Gq3DlzQAgjAIHgQtQaNUAKowagcuUABxIFHqXgAunyyBUBmQbcjCqiSo7ZoLDkxukA=
```
class C
{
    static S M1() => new S() { F = 111 };

    static int M2(ref int x, S y)
    {
        return x;
    }

    static int M3()
    {
        return M2(ref M1().Ref(), default);
    }

    static int M4()
    {
        return M2(ref M1().Ref(), new S() { F = 123 });
    }

    static int M5()
    {
        var x = new S() { F = 124 };
        return M2(ref M1().Ref(), x);
    }
    
    static void Main()
    {
        System.Console.WriteLine(M3());
        System.Console.WriteLine(M4());
        System.Console.WriteLine(M5());
    }
}


public struct S
{
    public int F;

    [System.Diagnostics.CodeAnalysis.UnscopedRef]
    public ref int Ref()
    {
        return ref F;
    }
}
```

Observed:
```
0
123
111
```

Expected either an error for `Ref` invocations, or the following output:
```
111
111
111
```

[...]

PR #75830, opened 2024-11-08
Disallow escaping refs to implicit rvalue temporaries

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/67435.

TODO:
- [ ] Dedicated error code.
- [ ] Warning only in unsafe regions.
- [ ] Breaking change doc.


jjonescz:
I would like to know if the approach looks good before I continue (there are several remaining details to be implemented, see the PR description).
cc @cston @RikkiGibson

RikkiGibson:
Let's try to reach consensus that we want to solve this by adding an error rather than by preventing the temp from being shared. Perhaps we could talk through it in an upcoming compiler team meeting.

jjonescz:
Talked with Jared today; I misunderstood the discussion under the issue, we already seem to have reached a consensus to NOT report errors and instead prevent the temps from being reused by the emit layer. (The former would be too big of a break.) I will look into that.

PR #67736, opened 2023-04-11
Improve detection of receiver cloning in ref safety analysis

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/67697.
Fixes https://github.com/dotnet/roslyn/issues/67626.

The regression was likely introduced in https://github.com/dotnet/roslyn/pull/64890.

PR #75851, opened 2024-11-11
Move HasHome out of Binder

jjonescz:
According to https://github.com/dotnet/roslyn/pull/67736#issuecomment-1505177626, `Binder.HasHome` should be used only on lowered nodes, so I'm moving it out of `Binder` to avoid its misuse. (Prior to the linked PR, the helper was used inside `Binder` but it's not anymore.)

The first commit does not have any changes except code move.

cc @AlekseyTs


cston (approved):

AlekseyTs (approved):
LGTM (commit 3)

Issue #75714, opened 2024-11-04
The VBCSCompiler pipe name needs to normalize the path name


jaredpar:
The pipe name for VBCSCompiler is created by putting a few components in a string (path to compiler, user name, etc ...) and taking a checksum of that string. The path component need to be normalized, likely by making the strings lower case, as part of this process. 

Not doing this can lead to multiple VBCSCompiler processes being created. This was observed in a few customer traces.

PR #75852, opened 2024-11-11
Make VBCSCompiler pipe name insensitive to casing of the path

jjonescz:
Resolves https://github.com/dotnet/roslyn/issues/75714.


chsienki (approved):

src/Compilers/Shared/BuildServerConnection.cs
```diff
@@ -564,6 +564,9 @@ internal static string GetPipeName(
             // of this method.
             clientDirectory = clientDirectory.TrimEnd(Path.DirectorySeparatorChar);
 
+            // Similarly, we don't want multiple servers if the path differs in casing.
```
Youssef1313, line 567: Is that applicable only on Windows?

jaredpar: Any OS can have case insensitive paths, or even paths for which parts have case insensitivity.

jaredpar, line 567:
```suggestion
            // Similarly, we don't want multiple servers if the provided launch path differs in casing.
```

cston (approved):

PR #75862, opened 2024-11-11
VB: Check overload resolution priority for basic scenarios involving methods

AlekseyTs:
Speclet - https://github.com/dotnet/vblang/pull/627


src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
```diff
@@ -1187,6 +1259,248 @@ ResolutionComplete:
...
+                        toDeprioritize.State = CandidateAnalysisResultState.LessApplicable
+                        candidates(info.Index) = toDeprioritize
+                    End If
+                Next
+            Next
+
+            candidateInfoByDeclaringType.Free()
+        End Sub
+
+        Private Shared Sub ApplyTieBreakingRulesSkippedByCombineCandidates(
```
AlekseyTs, line 1322: >ApplyTieBreakingRulesSkippedByCombineCandidates [](http://example.com/codeflow?start=27&length=47)

Effect of changes to `CombineCandidates` with this fixup was tested across all unit-tests that we have right now - [https://github.com/dotnet/roslyn/pull/75792](https://github.com/dotnet/roslyn/pull/75792).

AlekseyTs:
@333fred, @cston, @dotnet/roslyn-compiler Please review

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
333fred, line 1: I was expecting to see tests for some interesting metadata scenarios, like double-application (AttributeAppliedTwiceMethod_Metadata). Are you planning on adding more tests in the exceptional scenarios like this in a later commit?

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
```diff
@@ -227,6 +231,28 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
 
                 Return False
             End Function
+
+            Public Overrides ReadOnly Property OverloadResolutionPriority As Integer
+                Get
+                    Return m_Method.OriginalDefinition.OverloadResolutionPriority
```
333fred, line 237: This is probably wrong; we're using this method for filtering whether overload resolution priority is present on any methods, but `m_Method` isn't the method that has the priority we care about. We should only be examining the least overridden method for overload resolution priority. Please make sure we have tests covering overridden methods where the overridable method has priority, as well as methods where invalid IL resets the overridding method's priority to 0.

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
```diff
@@ -915,13 +963,18 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
 
             Dim applicableNarrowingCandidateCount As Integer = 0
             Dim applicableInstanceCandidateCount As Integer = 0
+            Dim someCandidatesHaveOverloadResolutionPriority As Boolean = False
 
             ' First collect instance methods.
             If instanceCandidates.Count > 0 Then
 
+                someCandidatesHaveOverloadResolutionPriority = instanceCandidates.Any(Function(candidate) candidate.OverloadResolutionPriority <> 0)
```
333fred, line 971: Specifically, I expect this filter to fail in the presence of an override.

333fred (commented):
Done review pass

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
AlekseyTs: This doesn't affect correctness and is suitable for a cheap check, overrides return 0, given how symbols implement the property

AlekseyTs: Given the way VB name lookup works, the least derived forms are also among the candidates. Therefore we get to them as well

333fred: Thanks for the clarification, that wasn't apparent to me at first read.

333fred (approved):

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
AlekseyTs: > Specifically, I expect this filter to fail in the presence of an override.

Given the way VB name lookup works, the least derived forms are also among the candidates. Therefore we get to them as well without explicitly traversing the overrides hierarchy.

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
AlekseyTs: > I was expecting to see tests for some interesting metadata scenarios, like double-application (AttributeAppliedTwiceMethod_Metadata). Are you planning on adding more tests in the exceptional scenarios like this in a later commit?

I am planning to clone all C# tests for the feature. I assume there are tests like that among them.

AlekseyTs:

AlekseyTs:

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
cston, line 237: >OriginalDefinition [](http://example.com/codeflow?start=36&length=18)

Are we testing cases that require `OriginalDefinition` in `OverloadResolutionPriority` or `GetOverloadResolutionPriorityInfo()`?

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
```diff
@@ -444,6 +470,28 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
 
                 Return False
             End Function
+
+            Public Overrides ReadOnly Property OverloadResolutionPriority As Integer
+                Get
+                    Return _property.OriginalDefinition.OverloadResolutionPriority
```
cston, line 476: >OriginalDefinition [](http://example.com/codeflow?start=37&length=18)

Are we testing cases that require `OriginalDefinition` in `OverloadResolutionPriority` or `GetOverloadResolutionPriorityInfo()`?

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
AlekseyTs: This is not required for correctness.

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
AlekseyTs: This is not required for correctness. The effect is not observable, other than us doing less work.

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
```diff
@@ -1187,6 +1275,254 @@ ResolutionComplete:
             Return New OverloadResolutionResult(candidates.ToImmutable(), resolutionIsLateBound, narrowingCandidatesRemainInTheSet, asyncLambdaSubToFunctionMismatch)
         End Function
 
+        Private Shared ReadOnly s_poolInstance As ObjectPool(Of PooledDictionary(Of NamedTypeSymbol, NamedTypeSymbol)) =
```
cston, line 1278: >s_poolInstance [](http://example.com/codeflow?start=32&length=14)

Is this used?

AlekseyTs: > Is this used?

Thanks for catching this. It is meant to be used for ```candidateInfoByDeclaringType``` in the method below.

AlekseyTs: This wouldn't affect the result though since we are comparing definitions

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
```diff
@@ -4080,15 +4400,17 @@ Bailout:
                     '       This rule also applies to the types that extension methods are defined on.
                     '7.2.	If M and N are extension methods and the target type of M is a class or
                     '       structure and the target type of N is an interface, eliminate N from the set.
-                    If ShadowBasedOnReceiverType(existingCandidate, newCandidate, existingWins, newWins, useSiteInfo) Then
+                    If (Not someCandidatesHaveOverloadResolutionPriority OrElse
+                        (signatureMatch AndAlso Not (existingCandidate.Candidate.IsExtensionMethod OrElse newCandidate.Candidate.IsExtensionMethod))) AndAlso
```
cston, line 4404: What effect does this `OrElse` clause have, and are we testing it?

src/Compilers/VisualBasic/Portable/Symbols/Metadata/PE/PEPropertySymbol.vb
```diff
@@ -45,6 +46,7 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols.Metadata.PE
         Private _lazyObsoleteAttributeData As ObsoleteAttributeData = ObsoleteAttributeData.Uninitialized
 
         Private _lazyIsRequired As ThreeState = ThreeState.Unknown
+        Private _lazyOverloadResolutionPriority As StrongBox(Of Integer)
```
cston, line 49: >StrongBox [](http://example.com/codeflow?start=51&length=9)

Should we add an `UncommonFields` class to hold most of the `_lazy*` fields?

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
AlekseyTs: > What effect does this ```OrElse``` clause have, and are we testing it?

EarlyFilteringOnReceiverType_01 fails if we remove it. The meaning of it is pretty much similar to what the comment before the previous `If` says. I.e. priority filtering for extension methods must come before filtering by receiver. Because for extension methods the filtering by receiver is not about shadowing by signature between methods declared in base and derived types.

src/Compilers/VisualBasic/Portable/Symbols/Metadata/PE/PEPropertySymbol.vb
AlekseyTs: I don't think we have to, and I decided against doing this  work under umbrella of this feature.

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
cston: It looks like `EarlyFilteringOnReceiverType_01` still passes of the entire `OrElse ...` clause is removed.

AlekseyTs: I am not sure what you did, but when I remove ``` AndAlso Not (existingCandidate.Candidate.IsExtensionMethod OrElse newCandidate.Candidate.IsExtensionMethod)``` part. The test fails for me. Perhaps that should be taken offline.

cston: As we discussed, I was referring to the entire `OrElse (signatureMatch AndAlso Not (...))`.

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
```diff
@@ -0,0 +1,987 @@
...
+        End Set
+    End Property
+End Module
+"
+            Dim compilation = CreateCompilation({compilationDef, OverloadResolutionPriorityAttributeDefinitionVB}, options:=TestOptions.ReleaseExe)
+            CompileAndVerify(compilation, expectedOutput:="2")
+        End Sub
+
+        <Fact>
+        Public Sub DefaultProperty_01()
```
cston, line 770: >DefaultProperty_01 [](http://example.com/codeflow?start=19&length=18)

Is this test affected by the compiler change? Should `Main()` contain a call to the default property?

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
```diff
@@ -0,0 +1,987 @@
...
+            compilation.AssertTheseDiagnostics(
+<expected>
+BC31048: Properties with no required parameters cannot be declared 'Default'.
+    Default WriteOnly Property M1 As Integer
+                               ~~
+</expected>)
+        End Sub
+
+        <Fact>
+        Public Sub DefaultProperty_02()
```
cston, line 801: >DefaultProperty_02 [](http://example.com/codeflow?start=19&length=18)

Is this test affected by the compiler change?

cston (approved):

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
AlekseyTs: > Is this test affected by the compiler change? 

No. But the test demonstrates that there are no interesting scenarios when parameter-less property is marked default

> Should ```Main()``` contain a call to the default property?

I do not find it interesting, given the declaration error

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
AlekseyTs: > Is this test affected by the compiler change?

No. But the test demonstrates that there are no interesting scenarios when a parameter-less property is overloaded with default property

src/Compilers/VisualBasic/Portable/Semantics/OverloadResolution.vb
AlekseyTs: Addressed in #75941

PR #75872, opened 2024-11-12
Check for BackingField instead of IsAutoProperty in NullableWalker

### LINKED ISSUE #75245: `NullableWalker.Scan().getFieldSymbolToBeInitialized()`: Is checking IsAutoProperty sufficient or should we check the appropriate accessors are missing or auto-implemented
I will try and resolve this comment in my upcoming PR. I think we should just check for a `BackingField` and return it here.

_Originally posted by @RikkiGibson in https://github.com/dotnet/roslyn/pull/75116#discussion_r1769280504_
            

Code being referenced:
https://github.com/dotnet/roslyn/blob/bdebb57e8817bdfa1b43f53e905db9b4dfa95736/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs#L1076-L1077

It feels like this is another piece of our attempts to correctly substitute the field for the property and vice-versa when initializing/reading/managing flow state, and this check/branch can probably be consolidated into another area. But it doesn't appear to be urgent or impacting correctness.

RikkiGibson:
Closes #75245

It looks like this function is still needed for some complex cases involving required properties, MemberNotNull applied to setters, and constructor chaining. I adjusted the function to behave correctly with properties using the field keyword as well.

~~It looks like this function is no longer needed because we recently added a mechanism to share a slot between property and backing field under appropriate conditions. Specifically, changes to `NullableWalker.GetOrCreateSlot` in #75246.~~

### DIFF
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -917,6 +917,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                             (includeAllMembers: true, includeCurrentTypeRequiredMembers: _, includeBaseRequiredMembers: false)
                                 => containingType.GetMembersUnordered(),
+
                             (includeAllMembers: true, includeCurrentTypeRequiredMembers: true, includeBaseRequiredMembers: true)
                                 => getAllTypeAndRequiredMembers(containingType),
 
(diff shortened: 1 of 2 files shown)


RikkiGibson:
Looks like there are some cases where preconditions are not properly set in absence of this call. Likely we need to add some similar tests with properties using field keyword and make sure behavior is consistent.

AlekseyTs (approved):
LGTM (commit 3)

cston (approved):

PR #75874, opened 2024-11-12
[EnC] Implement support for restarting subset of projects affected by rude edits

tmat:
Implements `IManagedHotReloadLanguageService2` interface that allows the debugger to restart a subset of processes when a rude edit is encountered instead of terminating the entire debugging session.

This includes a new overload of `GetUpdatesAsync` that takes a list of running projects and returns sets of projects to restart and to rebuild based on dependency analysis of updated projects. This functionality has already been exposed to dotnet-watch, now it will be available to VS debugger as well.

Adds a new API `UpdateBaselines` that discards the initial baselines of specified projects and commits the current solution snapshot. This is to be called when projects affected by rude edits are rebuilt. The next update will start with a new baseline read from newly built metadata.

### DIFF
--- a/src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
+++ b/src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Immutable;
 using System.Composition;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.BrokeredServices;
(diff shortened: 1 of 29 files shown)


src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
```diff
@@ -309,7 +371,11 @@ public async ValueTask<bool> HasChangesAsync(string? sourceFilePath, Cancellatio
         }
     }
 
-    public async ValueTask<ManagedHotReloadUpdates> GetUpdatesAsync(CancellationToken cancellationToken)
+    [Obsolete]
+    public ValueTask<ManagedHotReloadUpdates> GetUpdatesAsync(CancellationToken cancellationToken)
+        => GetUpdatesAsync(runningProjects: [], cancellationToken);
```
davidwengier, line 376: Does this mean things are broken until the debugger respond? Would it be better to pass all something like `currentCompileTimeSolution.Projects.Select(p=>p.FilePath)` as the first parameter?

tmat: No, shouldn't be broken. The debugger doesn't currently consume the results that are affected by this input.

Issue #75871, opened 2024-11-12
Extension method lookup results in unexpected behavior with OverloadResolutionPriority


tannergooding:
`Compiler version: '4.13.0-1.24551.2 (be036403)'. Language version: 13.0.`

**Steps to Reproduce**:

1. Create a new C# Console Application targeting .NET 9
2. Set Program.cs to the following code:
```csharp
using System.Runtime.CompilerServices;

internal class Program
{
    private static void Main(string[] args)
    {
        int x = 5;
        x.R(); // E1.R(int)
    }
}

public static class E1
{
    public static void R(this int x) => Console.WriteLine("E1.R(int)");
}

public static class E2
{
    [OverloadResolutionPriority(-1)]
    public static void R(this int x) => Console.WriteLine("E2.R(int)");

    public static void R(this string x) => Console.WriteLine("E2.R(string)");
}
```

**Expected Behavior**:

The compiler should surface a diagnostic similar to `The call is ambiguous between the following methods or properties: 'E1.R(int)' and 'E2.R(int)'`

**Actual Behavior**:

The code compiles and selects `E1.R(int)` as the correct extension.

** Additional Notes**:

This works as expected if `E2.R(string)` doesn't exist. However, seemingly adding any additional overload (whether applicable or not) to the type which defines one of the ambiguous methods allows it to start working and for `OverloadResolutionPriority` to effectively work across types.

This started out as a repro using `E1.R(this Span<string> x)`, `E2.R(this Span<string> x)`, and `E2.R(this ReadOnlySpan<string> x)`; calling `args.R()`. Apparently there may be some consideration (which I don't entirely understand) where this other case is "expected" due to how applicability filtering and overload resolution prioritization interplay. But where it is unexpected for the repro given above.

tannergooding:
I also tested with `LangVersion=preview`. The compiler is the one that shipped with `VS 16.13.0 Preview 1`, but it also repros with the version that shipped with the .NET 9 SDK: `Compiler version: '4.12.0-3.24523.4 (f3348c2a)'`

PR #75878, opened 2024-11-12
Don't let inapplicable members participate in ORPA filtering

333fred:
Fixes https://github.com/dotnet/roslyn/issues/75871.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Operators/BinaryOperatorAnalysisResult.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Operators/BinaryOperatorAnalysisResult.cs
@@ -36,6 +36,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             get { return this.Kind != OperatorAnalysisResultKind.Undefined; }
         }
 
+        bool IMemberResolutionResultWithPriority<MethodSymbol>.IsApplicable => IsValid;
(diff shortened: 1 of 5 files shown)


AlekseyTs (approved):
LGTM (commit 1)

src/Compilers/CSharp/Portable/Binder/Semantics/Operators/BinaryOperatorAnalysisResult.cs
```diff
@@ -36,6 +36,7 @@ public bool HasValue
             get { return this.Kind != OperatorAnalysisResultKind.Undefined; }
         }
 
+        bool IMemberResolutionResultWithPriority<MethodSymbol>.IsApplicable => IsValid;
```
jcouv, line 39: nit: consider renaming existing `IsValid` property to `IsApplicable` instead. There's not much benefit of having two names for the same thing.

jcouv (approved):
LGTM Thanks (iteration 1)

src/Compilers/CSharp/Portable/Binder/Semantics/Operators/BinaryOperatorAnalysisResult.cs
333fred: I'd rather keep this scoped. If you want to come back with a refactoring of this I would be happy to review, but I don't want to conflate it with this change.

PR #75891, opened 2024-11-13
Simplify shape of APIs used for Overload Resolution Priority feature

AlekseyTs:


333fred (approved):

AlekseyTs:

AlekseyTs:

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
```diff
@@ -1192,9 +1192,9 @@ internal virtual void AddSynthesizedReturnTypeAttributes(PEModuleBuilder moduleB
         /// <remarks>
         /// Do not call this method from early attribute binding, cycles will occur.
         /// </remarks>
-        internal int OverloadResolutionPriority => CanHaveOverloadResolutionPriority ? (TryGetOverloadResolutionPriority() ?? 0) : 0;
+        internal int OverloadResolutionPriority => CanHaveOverloadResolutionPriority ? TryGetOverloadResolutionPriority() : 0;
 
-        internal abstract int? TryGetOverloadResolutionPriority();
+        internal abstract int TryGetOverloadResolutionPriority();
```
cston, line 1197: >TryGetOverloadResolutionPriority [](http://example.com/codeflow?start=30&length=32)

Should this method be called `GetOverloadResolutionPriority` now rather than `TryGet...` since we're no longer distinguishing between unset and set to default?

AlekseyTs: I am fine with the name as it is.

cston: Or should this be a property? `internal abstract int OverloadResolutionPriority { get; }`

cston (approved):

src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs
AlekseyTs: I do not think there is a requirement for that.

Issue #75666, opened 2024-10-29
Consider clearing state of iterator state machine when disposal completes


jcouv:
In the example below from @stephentoub, the local `values2` in the iterator state machine holds a reference to a large allocation but the reference is kept alive past the `foreach`.
```csharp
using System.Reflection;
using System.Collections.Generic;
using System;
using System.Linq;

IEnumerable<int> values = Produce();

Console.Write("here");
foreach (int value in values)
{
}

while (true)
{
    GC.Collect();
    Console.WriteLine(((int[])values.GetType().GetField("<values2>5__1", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)).Length);
}

static IEnumerable<int> Produce()
{
    int[] values2 = Enumerable.Range(0, 100_000_000).ToArray();
    foreach (int value in values2) yield return value;
}
``` 

The `foreach` calls `GetEnumerator` which may return `this`.
```c#
        IEnumerator<int> IEnumerable<int>.GetEnumerator()
        {
            if (<>1__state == -2 && <>l__initialThreadId == Environment.CurrentManagedThreadId)
            {
                <>1__state = 0;
                return this;
            }
            return new <<<Main>$>g__Produce|0_0>d(0);
        }
```
Then the state machine does the allocation:
```c#
    [CompilerGenerated]
    private sealed class <<<Main>$>g__Produce|0_0>d : IEnumerable<int>, IEnumerable, IEnumerator<int>, IEnumerator, IDisposable
    {
        private int[] <values2>5__1;

        private bool MoveNext()
        {
                ...
                <values2>5__1 = Enumerable.ToArray(Enumerable.Range(0, 100000000));
                ...
       }
```

That field for hoisted local `values2` never gets reset, so we hold onto the allocation for as long as the enumerable `values` is kept around.
The proposal would be for the `Dispose` method to reset all the enumerator state in the state machine once disposal completes.

jcouv:
Taking some notes from investigation:
We already have logical to clear unmanaged hoisted locals when we exit scopes. But that logic has a number of limitations:
1. we don't clear top-level locals on normal exit
2. we don't clear any locals on yield break
3. we don't clear any locals on early exit (from yield return, ie. `foreach (var i in coll) { break; }`)
4. we don't clear any locals on throw (won't fix, see below)

So the current design only works for nested locals on normal exit from the scope...

[...]

PR #75908, opened 2024-11-14
Clear hoisted locals when disposing iterator and async-iterator state machines

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/75666

## Background

We already have logic to clear unmanaged hoisted locals when we exit scopes. But that logic has a number of limitations:
1. we don't clear top-level locals on normal exit
2. we don't clear any locals on yield break
3. we don't clear any locals on early exit (from yield return, ie. `foreach (var i in coll) { break; }`)
4. we don't clear any locals on throw 

```csharp
static IEnumerable<int> Produce()
{
    {
        string hoistedNestedLocal = "";
        // existing clearing logic doesn't work for:
        // `yield break`
        // `yield return` and the caller breaks `foreach (var i in coll) { break; }`
        // `throw`
    }

    string hoistedTopLevelLocal = "";
    // existing clearing local doesn't work for:
    // normal exit (because a `yield break` is injected, see `FlowAnalysisPass.AppendImplicitReturn`, so the cleanup is added after a `yield break`)
    // `yield break`
    // `yield return` and the caller breaks `foreach (var i in coll) { break; }`
    // `throw`
}
```

So the current design only works for nested locals on normal exit from their scope.

## Change

This PR clears the hoisted fields for all the locals (top-level and nested) in the part of the code that handles disposal: in `Dispose` method for iterators and at the end of `MoveNext` for async-iterators.
There is no need to add such clearing for async state machines, as it is non-trivial for a user to get a direct reference to those.

The proposed design fixes scenarios 1, 2, 3, and most of 4 (a throw during disposal still results in uncleared state).
This remains "best effort", as we don't want to add the overhead of an additional `try`/`finally` to handle that last scenario.

Look at the diff between commit 1 (baseline) and commit 2 (code change) to see the expectedOutput and IL changes.


src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncIteratorMethodToStateMachineRewriter.cs
```diff
@@ -88,11 +88,22 @@ internal AsyncIteratorMethodToStateMachineRewriter(
 
             return (asyncDispatch != null) ? F.Block(asyncDispatch, iteratorDispatch) : iteratorDispatch;
         }
+
+        protected override BoundStatement GenerateHoistedLocalsCleanup(ImmutableArray<StateMachineFieldSymbol> hoistedLocals)
+        {
+            // We need to clean nested hoisted local variables too (not just top-level ones)
```
AlekseyTs, line 94: >We need to clean nested hoisted local variables too (not just top-level ones) [](http://example.com/codeflow?start=15&length=77)

This isn't obvious. It looks like there are two call sites of this method and it is not clear why we want to have this behavior for all of them. Instead of completely ignoring the passed argument, consider adjusting the call sites to construct the right set of locals to clear. We can discuss offline in more details.

src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/MethodToStateMachineRewriter.cs
```diff
@@ -434,6 +429,50 @@ private void AddVariableCleanup(ArrayBuilder<BoundExpression> cleanup, FieldSymb
             }
         }
 
+#nullable enable
+        protected BoundBlock GenerateAllHoistedLocalsCleanup()
+        {
+            var variableCleanup = ArrayBuilder<BoundExpression>.GetInstance();
+            var alreadyCleaned = PooledHashSet<string>.GetInstance();
+            foreach (var symbol in HoistedVariables)
```
AlekseyTs, line 437: >var [](http://example.com/codeflow?start=21&length=3)

Please spell out the type

src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/MethodToStateMachineRewriter.cs
```diff
@@ -434,6 +429,50 @@ private void AddVariableCleanup(ArrayBuilder<BoundExpression> cleanup, FieldSymb
...
+                {
+                    Debug.Assert(replacement is CapturedToStateMachineFieldReplacement or CapturedToExpressionSymbolReplacement);
+
+                    if (replacement is CapturedToStateMachineFieldReplacement fieldReplacement)
+                    {
+                        addVariableCleanupOnce(variableCleanup, fieldReplacement.HoistedField, alreadyCleaned);
+                    }
+                    else if (replacement is CapturedToExpressionSymbolReplacement expressionReplacement)
+                    {
+                        foreach (var field in expressionReplacement.HoistedFields)
```
AlekseyTs, line 450: >var [](http://example.com/codeflow?start=33&length=3)

Please spell out the type

src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/MethodToStateMachineRewriter.cs
```diff
@@ -434,6 +429,50 @@ private void AddVariableCleanup(ArrayBuilder<BoundExpression> cleanup, FieldSymb
             }
         }
 
+#nullable enable
+        protected BoundBlock GenerateAllHoistedLocalsCleanup()
```
AlekseyTs, line 433: >GenerateAllHoistedLocalsCleanup [](http://example.com/codeflow?start=29&length=31)

This code looks very different from original implementation of ```GenerateHoistedLocalsCleanup```. Is it based on some other existing code? Also, isn't there a more direct and robust way to enumerate fields used for hoisting? I mean without going through locals, interpreting their replacements, etc. If not, perhaps we can build the list in advance, when the fields are added.

src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncIteratorMethodToStateMachineRewriter.cs
AlekseyTs: This behavior could be a source of bugs in the future, since the method doesn't do what a reasonable dev would expect it to do.

src/Compilers/CSharp/Portable/Lowering/IteratorRewriter/IteratorMethodToStateMachineRewriter.cs
```diff
@@ -171,6 +175,7 @@ internal void GenerateMoveNextAndDispose(BoundStatement body, SynthesizedImpleme
                                     ImmutableArray.Create<LocalSymbol>(stateLocal),
                                     F.Assignment(F.Local(stateLocal), F.Field(F.This(), stateField)),
                                     EmitFinallyFrame(rootFrame, state),
+                                    GenerateAllHoistedLocalsCleanup(),
```
AlekseyTs, line 178: >GenerateAllHoistedLocalsCleanup(), [](http://example.com/codeflow?start=36&length=34)

Are there tests observing effect of this on IL?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
```diff
@@ -9426,15 +9432,15 @@ public void AddVariableCleanup_IntLocal()
...
 await foreach (int value in values) { }
-System.Console.Write(((int)values.GetType().GetField("<values2>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)));
+System.Console.Write(((int)values.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)));
 
 class C
 {
     public static async System.Collections.Generic.IAsyncEnumerable<int> Produce()
     {
-        int values2 = 42;
+        int s = 42;
```
AlekseyTs, line 9441: >int s = 42; [](http://example.com/codeflow?start=8&length=11)

It doesn't look like there are meaningful changes in this unit-test. Consider reverting.

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
```diff
@@ -9448,22 +9454,251 @@ public void AddVariableCleanup_StringLocal()
...
+    System.Console.Write(((string)values.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)));
+}
+System.Console.Write(((string)values.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)) is null);
 
 class C
 {
     public static async System.Collections.Generic.IAsyncEnumerable<int> Produce()
     {
-        string values2 = "value ";
+        string s = "value ";
```
AlekseyTs, line 9467: >string s = "value "; [](http://example.com/codeflow?start=8&length=20)

Why is this change necessary?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
```diff
@@ -9494,9 +9731,9 @@ public static async System.Collections.Generic.IAsyncEnumerable<int> Produce(boo
     {
         while (b)
         {
-            string values2 = "value ";
+            string s = "value ";
```
AlekseyTs, line 9734: >string s = "value "; [](http://example.com/codeflow?start=12&length=20)

Why is this change necessary?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
```diff
@@ -9545,9 +9899,9 @@ public static async System.Collections.Generic.IAsyncEnumerable<int> Produce(boo
     {
         while (b)
         {
-            S values2 = new S { field = 42 };
+            S s = new S { field = 42 };
```
AlekseyTs, line 9902: >S s = new S { field = 42 }; [](http://example.com/codeflow?start=12&length=27)

Why is this change necessary?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncTests.cs
```diff
@@ -6071,20 +6071,20 @@ public void AddVariableCleanup_NestedStringLocal()
...
+System.Console.Write((string)stateMachine.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(stateMachine) is null);
 
 class C
 {
     public static async System.Threading.Tasks.Task<int> ProduceAsync(bool b, System.Threading.Tasks.Task task)
     {
         while (b)
         {
-            string values2 = "value ";
+            string s = "value ";
```
AlekseyTs, line 6082: >string s = "value "; [](http://example.com/codeflow?start=12&length=20)

Why is this change necessary?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncTests.cs
```diff
@@ -6114,20 +6114,20 @@ public struct S
...
+System.Console.Write((S)stateMachine.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(stateMachine));
 
 class C
 {
     public static async System.Threading.Tasks.Task<int> ProduceAsync(bool b, System.Threading.Tasks.Task task)
     {
         while (b)
         {
-            S values2 = new S { field = 42 };
+            S s = new S { field = 42 };
```
AlekseyTs, line 6125: >S s = new S { field = 42 }; [](http://example.com/codeflow?start=12&length=27)

Why is this change necessary?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenIterators.cs
```diff
@@ -3032,90 +3039,602 @@ public void AddVariableCleanup_StringLocal()
...
+    System.Console.Write(((string)values.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)));
+}
+System.Console.Write(((string)values.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)) is null);
 
 class C
 {
     public static System.Collections.Generic.IEnumerable<int> Produce()
     {
-        string values2 = "ran";
+        string s = "ran ";
```
AlekseyTs, line 3052: >string s = "ran "; [](http://example.com/codeflow?start=8&length=18)

Why is this change necessary?

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenIterators.cs
```diff
@@ -3032,90 +3039,602 @@ public void AddVariableCleanup_StringLocal()
...
+{
+    System.Console.Write(((int[])values.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)).Length);
+}
+System.Console.Write(((int[])values.GetType().GetField("<s>5__2", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(values)) is null);
+
+class C
+{
+    public static System.Collections.Generic.IEnumerable<int> Produce()
+    {
+        int[] s = Enumerable.Range(0, 100).ToArray();
```
AlekseyTs, line 3126: >int[] s = Enumerable.Range(0, 100).ToArray(); [](http://example.com/codeflow?start=8&length=45)

Why is this change necessary?

AlekseyTs:

jcouv:
>         protected BoundCatchBlock GenerateExceptionHandling(LocalSymbol exceptionLocal, ImmutableArray<StateMachineFieldSymbol> hoistedLocals)

:memo: consider renaming or making a local function, consider commenting. This is only for the "master catch"

---
Refers to: src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncMethodToStateMachineRewriter.cs:227 in 386a056. [](commit_id = 386a0561f3183b5dea065a59a325714f3a539261, deletion_comment = False)

src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncMethodToStateMachineRewriter.cs
```diff
@@ -258,7 +259,7 @@ protected BoundCatchBlock GenerateExceptionHandling(LocalSymbol exceptionLocal,
                 isSynthesizedAsyncCatchAll: true);
         }
 
-        protected BoundStatement GenerateHoistedLocalsCleanup(ImmutableArray<StateMachineFieldSymbol> hoistedLocals)
+        protected virtual BoundStatement GenerateHoistedLocalsCleanup(ImmutableArray<StateMachineFieldSymbol> hoistedLocals)
```
jcouv, line 262: :memo: `CleanupForExit` and `rootHoistedLocals`

src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/MethodToStateMachineRewriter.cs
```diff
@@ -127,10 +127,11 @@ public MethodToStateMachineRewriter(
             _hoistedVariables = hoistedVariables;
             _synthesizedLocalOrdinals = synthesizedLocalOrdinals;
             _nextFreeHoistedLocalSlot = nextFreeHoistedLocalSlot;
+            _hoistedFieldsForCleanup = new ArrayBuilder<FieldSymbol>();
 
             foreach (var proxy in nonReusableLocalProxies)
             {
-                this.proxies.Add(proxy.Key, proxy.Value);
+                this.AddProxyAndRegisterForCleanup(proxy.Key, proxy.Value);
```
AlekseyTs, line 134: >this.AddProxyAndRegisterForCleanup(proxy.Key, proxy.Value); [](http://example.com/codeflow?start=16&length=59)

Is this order deterministic?

src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncMethodToStateMachineRewriter.cs
```diff
@@ -223,48 +261,13 @@ protected virtual BoundStatement GenerateSetResultCall()
...
-                body: F.Block(
-                    assignFinishedState, // _state = finishedState;
-                    GenerateHoistedLocalsCleanup(hoistedLocals),
-                    callSetException, // builder.SetException(ex);  OR  _promiseOfValueOrEnd.SetException(ex);
-                    GenerateReturn(false)), // return;
-                isSynthesizedAsyncCatchAll: true);
-        }
-
-        protected BoundStatement GenerateHoistedLocalsCleanup(ImmutableArray<StateMachineFieldSymbol> hoistedLocals)
+        protected virtual BoundStatement GenerateHoistedLocalsCleanupForExit(ImmutableArray<StateMachineFieldSymbol> rootHoistedLocals)
```
AlekseyTs, line 264: >GenerateHoistedLocalsCleanupForExit [](http://example.com/codeflow?start=41&length=35)

Consider removing "HoistedLocals" from the name to completely avoid any confusion that the clean up is limited to the locals in the ```rootHoistedLocals``` parameter.
```
        protected virtual BoundStatement GenerateCleanupForExit(ImmutableArray<StateMachineFieldSymbol> rootHoistedLocals)
```

src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/MethodToStateMachineRewriter.cs
```diff
@@ -133,6 +135,12 @@ public MethodToStateMachineRewriter(
                 this.proxies.Add(proxy.Key, proxy.Value);
             }
 
+            _fieldsForCleanup = new ArrayBuilder<FieldSymbol>();
+            foreach (FieldSymbol fieldForCleanup in nonReusableFieldsForCleanup)
+            {
+                _fieldsForCleanup.Add(fieldForCleanup);
```
AlekseyTs, line 141: >Add [](http://example.com/codeflow?start=34&length=3)

I think we have an ```AddRange``` helper to add all items from ```nonReusableFieldsForCleanup``` in one call

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenIterators.cs
```diff
@@ -3022,57 +3022,126 @@ public static System.Collections.Generic.IEnumerable<int> Produce()
...
+  // Code size        1 (0x1)
+  .maxstack  0
+  IL_0000:  ret
+}
+""");
         }
 
         [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75666")]
+
         public void AddVariableCleanup_StringLocal()
```
AlekseyTs, line 3037: This shows as an unnecessary empty line between an attribute and the target unit-test.

AlekseyTs (approved):
LGTM (commit 19)

jcouv:

jaredpar:
@333fred ptal

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
```diff
@@ -9743,5 +10097,439 @@ .locals init (int V_0,
...
+        {
+            string s = "value ";
+            try
+            {
+                await System.Threading.Tasks.Task.CompletedTask;
+                yield return 42;
+            }
+            finally
+            {
+                System.Console.Write(s);
```
333fred, line 10200: Is there a version of this with a `throw` in the `try`?

333fred (approved):

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenAsyncIteratorTests.cs
jcouv: No, the only test with a `throw` in the `try` is `AddVariableCleanup_NestedStringLocal_InTryFinally_WithThrow_EarlyIterationExit`.

PR #75909, opened 2024-11-14
Don't map spans for aspx files

davidwengier:
Follow up to https://github.com/dotnet/roslyn/pull/75460
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2288477

When I looked at the code I thought "Who would want to map a span, and then not use the mapped span?!". Turns out the answer is "aspx files".


ToddGrun (approved):
:shipit:

PR #75918, opened 2024-11-14
Include projects to restart/rebuild in ManagedHotReloadUpdates

tmat:


src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
```diff
@@ -371,15 +371,15 @@ public async ValueTask<bool> HasChangesAsync(string? sourceFilePath, Cancellatio
         }
     }
 
-    [Obsolete]
+    [Obsolete("Use other overload", false)]
```
phil-allen-msft, line 374: I don't remember encountering ObsoleteAttribute strings like this in the past.  Is there a more common pattern to switch to?
(non-blocking)

phil-allen-msft (approved):

src/EditorFeatures/Core/EditAndContinue/EditAndContinueLanguageService.cs
tmat: I double checked and the default is "false", so this is not really needed. Odd, I'm getting build errors in SDK repo. I guess they have warnings as errors on for all warnings :-/

Issue #75828, opened 2024-11-08
`UnscopedRef` should not affect code under the old ref safety rules


jjonescz:
**Version Used**: .NET 9 RC2

**Steps to Reproduce**:

Compile with C# 10 and TFM net6.0:

```cs
struct S
{
    public int F;

    [UnscopedRef] public ref int Ref() => ref F;
}

namespace System.Diagnostics.CodeAnalysis
{
    [AttributeUsage(AttributeTargets.All, AllowMultiple = false, Inherited = false)]
    public sealed class UnscopedRefAttribute : Attribute
    {
    }
}
```

**Expected Behavior**: Error at `ref F` same as if the UnscopedRef attribute wasn't present, because the code should use [the old ref safety rules](https://github.com/dotnet/csharplang/blob/8e11f7b7fe1f79adcf02d015b92a8361de63856e/proposals/csharp-11.0/low-level-struct-improvements.md#refsafetyrulesattribute) when compiled with C# 10 and TFM net6.0.

```
error CS8170: Struct members cannot return 'this' or other instance members by reference
```

**Actual Behavior**: No errors.

PR #75930, opened 2024-11-15
Consider RefSafetyRules with UnscopedRef

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75828


AlekseyTs:
The issue that this PR is claiming to fix sounds as though the problem is related to using a specific language version, but the code changes and the tests look like the language version makes no difference on its own. If that is the case, consider clarifying the issue, clearly describing all the factors involved. And, perhaps, adjusting its title.

jjonescz:
> If that is the case, consider clarifying the issue, clearly describing all the factors involved. And, perhaps, adjusting its title.

I've updated the issue's title and description.

src/Compilers/CSharp/Test/Semantic/Semantics/RefFieldTests.cs
```diff
@@ -22323,6 +22323,194 @@ public void ScopedReserved_Alias_Escaped()
...
+                struct S
+                {
+                    public int F;
+
+                    [UnscopedRef] public ref int Ref() => ref F;
+                }
+                """;
+            CreateCompilation([source, UnscopedRefAttributeDefinition],
+                parseOptions: TestOptions.Regular10).VerifyDiagnostics(
+                // (6,47): error CS8170: Struct members cannot return 'this' or other instance members by reference
```
cston, line 22340: >error CS8170: Struct members cannot return 'this' or other instance members by reference [](http://example.com/codeflow?start=27&length=88)

This looks like a breaking change, although it also looks like users of this library would ignore the `[UnscopedRef]` attribute on the method since the assembly is not marked with `[RefSafetyRules]`, so perhaps the breaking change when compiling this library is appropriate.

Consider adding a note to the Breaking Change document.

Is this the only breaking change in the PR? (From the tests below, it looks like in other cases we're reporting other errors where `[UnscopedRef]` is used (and ignored) with C#10 rules.)

cston: Should the compiler report a diagnostic when `[UnscopedRef]` is applied to a source method that is compiled with earlier ref safety rules? Otherwise, the library author may not realize the annotation will be ignored by callers. For instance, would the following still compile with earlier ref safety rules without warnings or errors?
```csharp
struct S
{
    int F;
    [UnscopedRef] public ref int Ref() => ref Unsafe.AsRef(in F);
}
static class C
{
    static ref int M()
    {
        S s = default;
        return ref s.Ref();
    }
}
```

cston: cc @jaredpar, @RikkiGibson

RikkiGibson: It feels like an oversight that we did not report an error when UnscopedRef was used in presence of the old ref safety rules.

jjonescz: > For instance, would the following still compile with earlier ref safety rules without warnings or errors?

No, it wouldn't compile. There will be errors because you are still using `this` by reference inside `S.Ref` which is disallowed there.

jjonescz:
@cston @RikkiGibson @jcouv for reviews, thanks

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -89,3 +89,42 @@ class C
...
 ```
+
+## `UnscopedRefAttribute` cannot be used with old ref safety rules
+
+
+The `UnscopedRefAttribute` accidentally affected code compiled by new Roslyn versions
+even when the code was compiled in the context of the old ref safety rules
+(i.e., targeting C# 10 or older with net6.0 or older).
+However, the attribute should not have an effect in that context, and that is now fixed.
```
cston, line 100: Minor suggestions:
```suggestion
The `UnscopedRefAttribute` unintentionally affected code compiled by new Roslyn versions
even when the code was compiled in the context of the earlier ref safety rules
(i.e., targeting C# 10 or earlier with net6.0 or earlier).
However, the attribute should not have an effect in that context, and that is now fixed.
```

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -89,3 +89,42 @@ class C
...
+
+```cs
+using System.Diagnostics.CodeAnalysis;
+struct S
+{
+    public int F;
+
+    // previously accidentally allowed in C# 10 with net6.0
+    // now fails with the same error as if the [UnscopedRef] wasn't there
+    // (struct members cannot return 'this' or other instance members by reference)
```
cston, line 112:
```suggestion
    // previously allowed in C# 10 with net6.0
    // now fails with the same error as if the [UnscopedRef] wasn't there:
    // error CS8170: Struct members cannot return 'this' or other instance members by reference
```

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -89,3 +89,42 @@ class C
...
+    [UnscopedRef] public ref int Ref() => ref F;
+}
+```
+
+To prevent misunderstanding (thinking the attribute has an effect
+but it actually does not because your code is compiled with the old ref safety rules),
+an error is reported when the attribute is used in C# 10 or older with net6.0 or older:
```
cston, line 119:
```suggestion
To prevent misunderstanding (thinking the attribute has an effect
but it actually does not because your code is compiled with the earlier ref safety rules),
an error is reported when the attribute is used in C# 10 or earlier with net6.0 or earlier:
```

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -89,3 +89,42 @@ class C
...
+
+```cs
+using System.Diagnostics.CodeAnalysis;
+struct S
+{
+    // both are errors in C# 10 with net6.0:
+    // UnscopedRefAttribute cannot be applied to this parameter because it is unscoped by default.
```
cston, line 126:
```suggestion
    // error CS9063: UnscopedRefAttribute cannot be applied to this parameter because it is unscoped by default.
```

RikkiGibson:
Will try to take a look today

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -2348,7 +2348,7 @@ lambdaParameter is not SourceComplexParameterSymbolBase
                     }
 
                     // If synthesizing a delegate with an [UnscopedRef] parameter, check the attribute is available.
-                    if (delegateParameter.HasUnscopedRefAttribute)
+                    if (delegateParameter.HasUnscopedRefAttribute && delegateParameter.UseUpdatedEscapeRules)
```
cston, line 2351: >delegateParameter.UseUpdatedEscapeRules [](http://example.com/codeflow?start=69&length=39)

It looks like we're still emitting the `[UnscopedRef]` attribute in `SynthesizedParameterSymbol.AddSynthesizedAttributes()` regardlelss of `UseUpdatedEscapeRules`. Is that expected?

cston: Are any tests affected by this change? If not, consider adding a test similar to `SynthesizedDelegateTypes_UnscopedRefAttribute_MissingConstructor_01` which seems to test the call below when `UseUpdatedEscapeRules` is set.

src/Compilers/CSharp/Test/Semantic/Semantics/RefFieldTests.cs
```diff
@@ -22323,6 +22323,300 @@ public void ScopedReserved_Alias_Escaped()
...
+                struct S
+                {
+                    public int F;
+
+                    [UnscopedRef] public ref int Ref() => ref F;
+                }
+                """;
+            CreateCompilation([source, UnscopedRefAttributeDefinition],
+                parseOptions: TestOptions.Regular10).VerifyDiagnostics(
+                // (6,6): error CS9063: UnscopedRefAttribute cannot be applied to this parameter because it is unscoped by default.
```
cston, line 22340: >error CS9063: UnscopedRefAttribute cannot be applied to this parameter because it is unscoped by default. [](http://example.com/codeflow?start=26&length=105)

This doesn't feel like the right error message for earlier language versions because `this` is not unscoped by default, as indicated by the next error: "error CS8170: Struct members cannot return 'this' or other instance members by reference".

Similarly, for the following, when compiled with C#10 and .NET 6, we're reporting that the parameter (in this case, `ref int i`) is unscoped by default. But that also seems surprising since callers can assume the `ref` is not captured in the `Span<int>` when `UseUpdatedEscapeRules` is `false`.
```csharp
static Span<int> F([UnscopedRef] ref int i) { ... }
```

Instead, should we report a diagnostic that `[UnscopedRef]` is not supported with C#10 and .NET 6 and earlier? Should this be a warning rather than an error so it's not a breaking change?

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -89,3 +89,42 @@ class C
     }
 }
 ```
+
+## `UnscopedRefAttribute` cannot be used with old ref safety rules
+
+
+The `UnscopedRefAttribute` unintentionally affected code compiled by new Roslyn versions
```
jaredpar, line 97:
```suggestion
The `UnscopedRefAttribute` unintentionally affected code compiled by new Roslyn compiler versions
```

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
```diff
@@ -92,7 +92,7 @@ public override Conversion GetMethodGroupDelegateConversion(BoundMethodGroup sou
                 }
 
                 // If synthesizing a delegate with an [UnscopedRef] parameter, check the attribute is available.
-                if (invoke.Parameters.Any(p => p.HasUnscopedRefAttribute))
+                if (invoke.Parameters.Any(p => p.HasUnscopedRefAttribute && p.UseUpdatedEscapeRules))
```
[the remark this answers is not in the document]
jjonescz: Adding tests SynthesizedDelegateTypes_UnscopedRefAttribute_MissingConstructor_03. And reverting the change here - similarly to https://github.com/dotnet/roslyn/pull/75930#discussion_r1856737766, we don't want to check UseUpdatedEscapeRules here, we have already checked it when synthesizing the parameter.

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
jjonescz: > Are any tests affected by this change?

I think this one cannot be tested - `HasUnscopedRefAttribute` will only be `true` if `UseUpdatedEscapeRules` is `true` since the lambda and the delegate live in the same compilation. That being said, I'm reverting the change here for the same reason as in https://github.com/dotnet/roslyn/pull/75930#discussion_r1856799309.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -89,3 +89,42 @@ class C
     }
 }
 ```
+
+## `UnscopedRefAttribute` cannot be used with old ref safety rules
```
jaredpar, line 93: What happens if we use `[InlineArray]` in an older language version? I would expect that we have symetry between that behavior and this behavior. Basically compiler should have a uniform policy of how we approach seeing attributes in older language versions (unless there are specific reasons to deviate). Pretty our policy is to ignore but want to double check. 

Can we also update the general compiler documentation / feature check list to call out what behavior we want for attributes in this context?

RikkiGibson: Ultimately it seems to better serve readers and writers of code to disallow use of `[UnscopedRef]` in cases where it is not having any effect. We are already doing this in several scenarios with latest LangVersion/TFM, where applying the attribute isn't defined to do something meaningful.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -1579,7 +1579,10 @@ protected override void DecodeWellKnownAttributeImpl(ref DecodeWellKnownAttribut
                 }
                 else
                 {
-                    diagnostics.Add(ErrorCode.ERR_UnscopedRefAttributeUnsupportedMemberTarget, arguments.AttributeSyntaxOpt.Location);
+                    var code = this.GetMethod?.UseUpdatedEscapeRules == false || this.SetMethod?.UseUpdatedEscapeRules == false
```
RikkiGibson, line 1582: These are source accessors, so they just delegate to the containing module. Perhaps it would be simpler to `this.ContainingModule.UseUpdatedEscapeRules` instead.

src/Compilers/CSharp/Portable/Symbols/Source/ThisParameterSymbol.cs
```diff
@@ -179,7 +179,8 @@ internal override ScopedKind EffectiveScope
             {
                 var scope = _containingType.IsStructType() ? ScopedKind.ScopedRef : ScopedKind.None;
                 if (scope != ScopedKind.None &&
-                    HasUnscopedRefAttribute)
+                    HasUnscopedRefAttribute &&
+                    UseUpdatedEscapeRules)
```
RikkiGibson, line 183: I am seeing this pair of checks very frequently in this PR. I am wondering when we would *not* want to do this check, i.e. when would we behave as if `HasUnscopedRefAttribute` is true, even when old escape rules are being used.

Essentially, should we replace `HasUnscopedRefAttribute` itself with some similar helper which already implies that new escape rules are being used.

jjonescz: > when would we behave as if `HasUnscopedRefAttribute` is true, even when old escape rules are being used.

We still look at `HasUnscopedRefAttribute` without checking `UseUpdatedEscapeRules` when synthesizing delegates for example - the synthesized delegate might have `[UnscopedRef]` even if it's synthesized in an old-rules assembly - if it points to a method from an updated-rules assembly. See tests UnscopedRefAttribute_RefSafetyRules_Reference_SynthesizedDelegate_*.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
jaredpar: The other part I'm interested in though is where do we document how attributes related to language features should be treated? This is a topic that's come up several times in the last few releases. Want us to find a place to codify what the default approach should be for such attributes. It also seems like an item that should be added to the test plan.

src/Compilers/CSharp/Test/Semantic/Semantics/RefFieldTests.cs
```diff
@@ -22681,6 +22681,82 @@ static void validate(ModuleSymbol module)
...
+                .VerifyDiagnostics().EmitToImageReference();
+
+            var source2 = """
+                int x = 1;
+                R r = default;
+                C.M(ref x, ref r);
+                var d = C.M;
+                d(ref x, ref r);
+                """;
+            // Should the delegate invocation be an error as well? https://github.com/dotnet/roslyn/issues/76087
```
RikkiGibson, line 22713: So the cause of the missing error is the wrong safety rules being used for the delegate invocation?

RikkiGibson (approved):

src/Compilers/CSharp/Test/Semantic/Semantics/RefFieldTests.cs
jjonescz: Yes, the delegate lives in the "old rules" compilation, so "old rules" are used for it. But it is assigned a method from "new rules". Presumably that assignment should be disallowed. Note that as the test below demonstrates, this is a pre-existing issue unrelated to UnscopedRefAttribute.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
jjonescz: Opened https://github.com/dotnet/roslyn/pull/76112.

Wanted to use `CallerArgumentExpressionAttribute` as an example, but turns out we don't ignore it in old LangVersions (probably not that big of an issue for that particular attribute).

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -8017,4 +8017,10 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="ERR_IteratorRefLikeElementType" xml:space="preserve">
     <value>Element type of an iterator may not be a ref struct or a type parameter allowing ref structs</value>
   </data>
+  <data name="WRN_UnscopedRefAttributeOldRules" xml:space="preserve">
+    <value>UnscopedRefAttribute does not have any effect in C# 10 or earlier with net6.0 or earlier.</value>
```
jaredpar, line 8021: Not following the `net6.0` part. The attribute should have effect if lang version is 11 or higher. net6.0 becomes relevant because at net7.0 or higher the new rules take effect irrespective of language version.

jjonescz: The "updated" ref safety rules apply either in C# 11 or when referencing net7.0. The [speclet](https://github.com/dotnet/csharplang/blob/85612a1abd8b5c66c7c7be20814c79b89ed98808/proposals/csharp-11.0/low-level-struct-improvements.md#refsafetyrulesattribute) says:

> If the module containing the method declaration references System.Runtime { ver: 7.0 }, the method call is analyzed with C#11 rules. This rule is a temporary mitigation for modules compiled with earlier previews of C#11 / .NET 7 and will be removed later.

jaredpar: > The "updated" ref safety rules apply either in C# 11 or when referencing net7.0. 

Agreed. But the error says 

> UnscopedRefAttribute does not have any effect in C# 10 or earlier with net6.0 or earlier.

Given the `or` this can be read as 

> UnscopedRefAttribute does not have any with net6.0 or earlier.

I get what the message is trying to say but not sure the wording is 100% clear. Think it would be better if we turned this into a positive declaration

> UnscopedRefAttribute is only valid in C# 11 or later or when targeting net7.0 or later.

jjonescz: I see, I will update it, thanks.

jjonescz:
@cston for another look, thanks

src/Compilers/CSharp/Portable/Symbols/Source/SourceComplexParameterSymbol.cs
```diff
@@ -867,7 +868,10 @@ protected override void DecodeWellKnownAttributeImpl(ref DecodeWellKnownAttribut
             {
                 if (!this.IsValidUnscopedRefAttributeTarget())
                 {
-                    diagnostics.Add(ErrorCode.ERR_UnscopedRefAttributeUnsupportedTarget, arguments.AttributeSyntaxOpt.Location);
+                    var code = !UseUpdatedEscapeRules
+                        ? ErrorCode.WRN_UnscopedRefAttributeOldRules
```
cston, line 872: >WRN_UnscopedRefAttributeOldRules [](http://example.com/codeflow?start=36&length=32)

Are we reducing an existing error to a warning here, in cases where the previous error applied regardless of which rules were used? For instance, when compiling the following with C# 10 and .NET 6:
```csharp
ref struct R<T> { }
class C
{
    ref T F1<T>([UnscopedRef] R<T> r) => throw null;
    ref T F2<T>([UnscopedRef] ref T r) => throw null;
    ref T F3<T>([UnscopedRef] T r) => throw null;
}
```

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -625,7 +625,10 @@ private void DecodeWellKnownAttributeAppliedToMethod(ref DecodeWellKnownAttribut
                 }
                 else
                 {
-                    diagnostics.Add(ErrorCode.ERR_UnscopedRefAttributeUnsupportedMemberTarget, arguments.AttributeSyntaxOpt.Location);
+                    var code = !this.UseUpdatedEscapeRules
+                        ? ErrorCode.WRN_UnscopedRefAttributeOldRules
```
cston, line 629: >WRN_UnscopedRefAttributeOldRules [](http://example.com/codeflow?start=36&length=32)

Are we reducing an existing error to a warning here, in cases where the previous error applied regardless of which rules were used? For instance, when compiling the following with C# 10 and .NET 6:
```csharp
class C
{
    [UnscopedRef] object F() => null;
}
```

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -1579,7 +1579,10 @@ protected override void DecodeWellKnownAttributeImpl(ref DecodeWellKnownAttribut
                 }
                 else
                 {
-                    diagnostics.Add(ErrorCode.ERR_UnscopedRefAttributeUnsupportedMemberTarget, arguments.AttributeSyntaxOpt.Location);
+                    var code = !this.ContainingModule.UseUpdatedEscapeRules
+                        ? ErrorCode.WRN_UnscopedRefAttributeOldRules
```
cston, line 1583: >WRN_UnscopedRefAttributeOldRules [](http://example.com/codeflow?start=36&length=32)

Are we reducing an existing error to a warning here, in cases where the previous error applied regardless of which rules were used? For instance, when compiling the following with C# 10 and .NET 6:
```csharp
struct S
{
    [UnscopedRef] object P { get; init; }
}
```

src/Compilers/CSharp/Portable/Symbols/Source/SourceComplexParameterSymbol.cs
jjonescz: Yes, I will change that, thanks.

cston (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourceComplexParameterSymbol.cs
```diff
@@ -865,9 +866,13 @@ protected override void DecodeWellKnownAttributeImpl(ref DecodeWellKnownAttribut
             }
             else if (attribute.IsTargetAttribute(AttributeDescription.UnscopedRefAttribute))
             {
-                if (!this.IsValidUnscopedRefAttributeTarget())
+                bool validTarget = this.IsValidUnscopedRefAttributeTarget();
+                if (!validTarget || !this.UseUpdatedEscapeRules)
```
RikkiGibson, line 870: nit: In my opinion, it would be less "error prone" (no pun intended) to simply report the warning for this attribute in presence of old rules, regardless of value of `validTarget`. Then there is less risk of a small adjustment to this branching code, causing a warning to be reported instead of an error.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -614,7 +614,8 @@ private void DecodeWellKnownAttributeAppliedToMethod(ref DecodeWellKnownAttribut
             }
             else if (attribute.IsTargetAttribute(AttributeDescription.UnscopedRefAttribute))
             {
-                if (this.IsValidUnscopedRefAttributeTarget())
+                bool validTarget = this.IsValidUnscopedRefAttributeTarget();
+                if (validTarget && this.UseUpdatedEscapeRules)
```
RikkiGibson, line 618: ditto here. Just have a standalone if-block which says if old rules are in use, then report the warning, then continue on to the regular/existing logic.

src/Compilers/CSharp/Test/Semantic/Semantics/RefFieldTests.cs
```diff
@@ -9747,9 +9747,9 @@ static ref S<T> F2<T>(S<T> x2)
             if (languageVersion == LanguageVersion.CSharp10)
             {
                 comp.VerifyEmitDiagnostics(
-                    // (8,28): error CS9063: UnscopedRefAttribute cannot be applied to this parameter because it is unscoped by default.
+                    // (8,28): warning CS9269: UnscopedRefAttribute is only valid in C# 11 or later or when targeting net7.0 or later.
```
RikkiGibson, line 9750: It feels like there is no need to stop reporting the original error as part of reporting this warning. I do not think it serves the user experience, for example, since if user updates TFM and recompiles, it's inconvenient to wait all the way thru a second build just to be presented with errors which could have been reported the first time. It's ok to just tell the user everything that we know to be wrong.

RikkiGibson (approved):
LGTM but had a general suggestion related to branches which report either a warning or an error.

src/Compilers/CSharp/Test/Semantic/Semantics/RefFieldTests.cs
jjonescz: Some tests changed as you suggest due to the change to decouple the error and warning reporting.

However, this particular test stays the same, but I think that's correct. As you say, "if user updates TFM and recompiles, it's inconvenient to wait all the way thru a second build just to be presented with errors which could have been reported the first time" - but here on TFM/LangVersion update, no new errors for this signature are reported (as can be seen in the `else` branch below).

RikkiGibson (approved):

jjonescz:
@cston for another look, thanks

cston (approved):

PR #75941, opened 2024-11-16
VB Overload Resolution Priority - handle operators and respect language version

AlekseyTs:


AlekseyTs:
@333fred, @cston, @dotnet/roslyn-compiler Please review

src/Compilers/CSharp/Test/Emit3/OverloadResolutionPriorityTests.cs
```diff
@@ -2706,4 +2706,94 @@ static void Main()
...
+
+    static void M1(long x) => System.Console.Write(3);
+}
+""";
+
+        CompileAndVerify([source, OverloadResolutionPriorityAttributeDefinition], expectedOutput: "3").VerifyDiagnostics();
+    }
+
+    [Fact]
+    public void LiftedOperator_01()
```
333fred, line 2768: Is this meaningfully different than `UnaryOperator_Allowed`?

AlekseyTs: Probably not, but it is clear what this test is about and, therefore, it is less likely that the test will be modified in a way that the lifted/un-lifted aspect is not covered

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
```diff
@@ -983,5 +1257,114 @@ BC30526: Property 'M' is 'ReadOnly'.
...
+        Public Sub ConversionOperator_01()
+            Dim compilationDef = "
+Structure S
+
+    public shared widening operator CType(x As Integer) As S
+        System.Console.Write(1)
+        return Nothing
+    End Operator
+
+    <System.Runtime.CompilerServices.OverloadResolutionPriority(1)>
```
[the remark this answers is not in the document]
AlekseyTs: > Are we planning to report an error here? The equivalent C# is disallowed.

There is an open design question about that in the speclet

333fred (approved):

AlekseyTs:

AlekseyTs:

cston (approved):

PR #75944, opened 2024-11-16
Add a couple common type sizes to SegmentedArrayHelper's size/shift/offsetmask methods so those values can be inlined.

ToddGrun:
This isn't super significant, but about 175ms of CPU are accounted for these methods in the scrolling speedometer test in our OOP. I believe those should go away completely as these are the only common values that weren't already being handled when I manually debugged the product.

*** relevant speedometer CPU usage before this change ***
image


ToddGrun:
test insertion based on commit 1:

https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/results?buildId=10570748&view=results

sharwell (approved):

PR #75946, opened 2024-11-17
Include list of processes that lock file in `can't write file` error message

tmat:
Use Win32 APIs to get the list of processes that have the file open. Based on msbuild implementation: https://github.com/dotnet/msbuild/blob/6cd445d84e59a36c7fbb6f50b7a5a62767a6da51/src/Utilities/LockCheck.cs

Related: https://github.com/dotnet/runtime/issues/109927

### DIFF
--- a/src/Compilers/Core/Portable/CommandLine/CommonCompiler.CompilerEmitStreamProvider.cs
+++ b/src/Compilers/Core/Portable/CommandLine/CommonCompiler.CompilerEmitStreamProvider.cs
@@ -118,8 +118,11 @@ namespace Microsoft.CodeAnalysis
                 var lockedBy = FileLockCheck.TryGetLockingProcessInfos(_filePath);
 
                 var message = lockedBy.IsEmpty
-                    ? e.Message
-                    : string.Format(CodeAnalysisResources.ExceptionMessage_FileMayBeLockedBy,
+                    ? (object)e.Message
+                    : new LocalizableResourceString(
+                        nameof(CodeAnalysisResources.ExceptionMessage_FileMayBeLockedBy),
+                        CodeAnalysisResources.ResourceManager,
+                        typeof(CodeAnalysisResources),
                         e.Message,
                         string.Join(", ", lockedBy.Select(info => $"'{info.applicationName}' ({info.processId})")));
 
--- a/src/Compilers/Core/Portable/CommandLine/CommonCompiler.CompilerEmitStreamProvider.cs
+++ b/src/Compilers/Core/Portable/CommandLine/CommonCompiler.CompilerEmitStreamProvider.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using Roslyn.Utilities;
 
 namespace Microsoft.CodeAnalysis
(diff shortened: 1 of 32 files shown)


tmat:
@rainersigwald fyi

jjonescz (approved):

src/Compilers/Core/Portable/CommandLine/CommonCompiler.CompilerEmitStreamProvider.cs
```diff
@@ -114,7 +115,15 @@ private Stream OpenFileStream()
...
             {
                 var messageProvider = _compiler.MessageProvider;
-                diagnostics.Add(messageProvider.CreateDiagnostic(messageProvider.ERR_CantOpenFileWrite, Location.None, _filePath, e.Message));
+                var lockedBy = FileLockCheck.TryGetLockingProcessInfos(_filePath);
+
+                var message = lockedBy.IsEmpty
+                    ? e.Message
+                    : string.Format(CodeAnalysisResources.ExceptionMessage_FileMayBeLockedBy,
+                        e.Message,
+                        string.Join(", ", lockedBy.Select(info => $"'{info.applicationName}' ({info.processId})")));
```
AlekseyTs, line 124: This is likely to create a localization problem. I think creating a `DiagnosticInfo` instead would work.

tmat: Note that msbuild does this the same way: https://github.com/dotnet/msbuild/blob/6cd445d84e59a36c7fbb6f50b7a5a62767a6da51/src/Utilities/LockCheck.cs#L274

AlekseyTs: >What kind of loc problem?

Given the code, different parts of the message can use different language. ```CodeAnalysisResources.ExceptionMessage_FileMayBeLockedBy``` uses language that is current at the time this code is executed, not the language requested when ```Message``` is requested. Strings should not be pulled manually from resources at the time a DiagnosticInfo is created. They should be pulled when one requests the message string and the process should respect the language requested at that time. Basically, the same DiagnosticInfo instance should be able to give a message string for any supported localization language.  

>Not sure what do you suggest to do with DiagnosticInfo.

We support a ```DiagnosticInfo``` instance as an argument for another ```DiagnosticInfo```. I am suggesting to create a ```DiagnosticInfo``` for the "file may be locked by" part instead of accessing resource strings manually.

AlekseyTs: We actually have another helper type that can be used in this specific situation - ```LocalizableResourceString```

tmat: This is already the case though. We take the exception message from the runtime and include it in the info as an argument.

tmat: >  I am suggesting to create a DiagnosticInfo for the "file may be locked by" part instead of accessing resource strings manually.

I'll try to do that.

tmat: Done.

AlekseyTs (approved):
LGTM (commit 2)

PR #75952, opened 2024-11-18
Handle `using var` in redundant assignment removal

### LINKED ISSUE #72829: "Remove redundant assignment" removes unrelated code
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Remove-redundant-assignment-removes-un/10627248)._

---
[severity:It's more difficult to complete my work]
If I use the Quick Action IDE0059 "Remove redundant assignement" with a particular setup, it will erroneously remove unrelated code and break logic. This is the problematic code setup:
```
DynamicBuffer<Entity> items = default;
if (SimWorldReader.Common_TryGetInventoryBuffer(..., out items))
{
    ...
}

using var _ = ListPool<DisplayedItemData>.Get(out var displayedInventory);
```
If I use IDE0059 on the variable ``items``, it will correctly remove ``= default``, but it will also erroneously remove ``using var _ =`` a few lines below, which breaks the code logic. ``using var _`` is not a redundant assignment and it is needed to dispose of the ``ListPool<DisplayedItemData>.Get(..)`` return value.

This is a dangerous bug because it alters the logic of the code without telling you.

---
#### Feedback Bot on 3/28/2024, 10:19 PM: 

---
### Original Solutions
(no solutions)

obonn1:
- Adds a test to ensure redundant assignment removal respects the `using var` statement.
- Fixes #72829.

On changing an unused variable to discard `CSharpReplaceDiscardDeclarationsWithAssignmentsService.ReplaceAsync()` ensures that instead of local declarations, simple discards are used. Without an exception for using statements, this end up changing `using var _` within the scope to be replaced with `_` regardless of any relation to the variable targeted by the suggestion. This change adds a filter to the method, preserving using statements and the disposal they provide.

### DIFF
--- a/src/Analyzers/CSharp/Tests/RemoveUnusedParametersAndValues/RemoveUnusedValueAssignmentTests.cs
+++ b/src/Analyzers/CSharp/Tests/RemoveUnusedParametersAndValues/RemoveUnusedValueAssignmentTests.cs
@@ -10072,4 +10072,33 @@ parseOptions: CSharpParseOptions.Default.WithLanguageVersion(LanguageVersion.CSh
             ReferenceAssemblies = ReferenceAssemblies.Net.Net80,
         }.RunAsync();
     }
+
+    [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/72829")]
+    public async Task RemoveRedundantAssignment_PreservesUsingVar()
+    {
+        await TestInRegularAndScriptAsync(
+            """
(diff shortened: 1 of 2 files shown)


obonn1:
@dotnet-policy-service agree

CyrusNajmabadi (approved):

CyrusNajmabadi:
Thanks!

PR #71173, opened 2023-12-08
Ensure the editorconfig for code-style analysis ("IDExxxx") rules is included

joegoldman2:
Given that the code-style analysis ("IDExxxx") rules are not present in the editorconfig files in C:\Program Files\dotnet\sdk\\*\Sdks\Microsoft.NET.Sdk\analyzers\build\config, the editorconfig for the code-style analysis rules must be included even if `AnalysisLevelStyle` or `AnalysisModeStyle` are equal to the core `AnalysisStyle` and `AnalysisMode` properties.

Considering:

```xml
<PropertyGroup>
  <OutputType>Exe</OutputType>
  <TargetFramework>net8.0</TargetFramework>
  <AnalysisLevel>latest</AnalysisLevel>
  <AnalysisMode>All</AnalysisMode>
  <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
</PropertyGroup>
```
```cs
using System;
Console.WriteLine();

// 👇 IDE0040 'Accessibility modifiers required' should be raised here
class Test
{
}
```

There is no IDE0040 warning raised in this example while the `AnalysisMode` is set to `All`.
This is because the editorconfig (`$(_GlobalAnalyzerConfigFile_MicrosoftCodeAnalysis{language}CodeStyle)`) is not included due to this condition:

https://github.com/dotnet/roslyn/blob/96b7bb2d85175c526bc25b53ebc859af0704896e/src/CodeStyle/Tools/Program.cs#L263

`$(AnalysisLevelStyle)` is equal to `$(AnalysisLevel)` (`latest` in this case) and `$(AnalysisModeStyle)` is equal to `$(AnalysisMode)` (`all` in this case).

This condition does not seem correct here. The severity of IDExxxx rules will not be overridden by `$(_GlobalAnalyzerConfigFile_MicrosoftCodeAnalysis{language}CodeStyle)` editorconfig anyway as those rules are not present in the editorconfig for CAxxx rules. So, if the file exists, it should be included.

This PR fixes this behavior.

cc @mavasani

PR #75978, opened 2024-11-19
Postpone respecting the code-style analysis config rule until NET10

arunchndr:
Related to #71173

Until we can connect the dots better for the user that has analysislevel set to latest-all (which currently is a lot of users), that this is the exact action and result that caused these errors in the IDE and cmdline, lets postpone enabling this.


arunchndr:
LGTM! @JoeRobich feel free to hit merge when ready. We may need a rebase or cherry pick to 17.12.

dibarbet (approved):

PR #76002, opened 2024-11-21
Expose FAR to cohosting

davidwengier:
Roslyn side of https://github.com/dotnet/razor/issues/11237

### DIFF
--- a/src/Tools/ExternalAccess/Razor/Cohost/Handlers/FindAllReferences.cs
+++ b/src/Tools/ExternalAccess/Razor/Cohost/Handlers/FindAllReferences.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.LanguageServer.Handler;
+using Microsoft.CodeAnalysis.MetadataAsSource;
+using Microsoft.CodeAnalysis.Options;
+using Microsoft.CodeAnalysis.Shared.TestHooks;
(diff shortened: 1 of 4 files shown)


ryzngard (approved):

src/Tools/ExternalAccess/Razor/Cohost/Handlers/FindAllReferences.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.CodeAnalysis.Text;
+using Roslyn.LanguageServer.Protocol;
+
+namespace Microsoft.CodeAnalysis.ExternalAccess.Razor.Cohost.Handlers;
+
+using Location = Roslyn.LanguageServer.Protocol.Location;
+
+internal static class FindAllReferences
+{
+    public static async Task<SumType<VSInternalReferenceItem, Location>[]?> FindReferencesAsync(Workspace workspace, Document document, LinePosition linePosition, bool supportsVSExtensions, CancellationToken cancellationToken)
```
dibarbet, line 20: You might consider having a progress parameter here and passing it in from Razor - otherwise you could be waiting a while for anything to stream in.  I don't think you need all the references before you can do anything with them right?

dibarbet (commented):
lgtm, apologies for not getting to this sooner - def ping me directly on teams if its taking too long.

src/Tools/ExternalAccess/Razor/Cohost/Handlers/FindAllReferences.cs
davidwengier: Right now Razor doesn't support streaming at all, because the LSP client middle layer doesn't. Cohosting removes that limitation, but none of our shared code on the Razor file knows how to deal with it, plus we'd have to work out how to stream things from OOP back to devenv to send to the language client, so I just decided to punt all of that until later :)

PR #76008, opened 2024-11-21
VBOverloadResolutionPriority: break attribute binding cycles, clone a bunch of C# tests

AlekseyTs:


AlekseyTs:
@cston, @333fred, @dotnet/roslyn-compiler Please review

333fred (approved):

AlekseyTs:

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
```diff
@@ -1560,5 +1562,1613 @@ End Module
...
+    System.Console.WriteLine(1)
+End Sub
+"
+
+            Dim i2Source = "
+public Shared Sub M(x As I2)
+    throw DirectCast(Nothing, System.Exception)
+End Sub
+"
+
```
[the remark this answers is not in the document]
AlekseyTs: > It looks like these are not used.

Looks this way. Will remove.

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
```diff
@@ -1560,5 +1562,1613 @@ End Module
...
+    Shared Sub Main
+        Dim d = new Derived()
+        d.M(""test"")
+    End Sub
+End Class
+"
+
+            Dim compilation = CreateCompilation({source, OverloadResolutionPriorityAttributeDefinitionVB}, options:=TestOptions.DebugExe)
+
+            ' Note, VB silently ignores attributes at wrong location
```
cston, line 2066: >Note, VB silently ignores attributes at wrong location [](http://example.com/codeflow?start=14&length=54)

Do we need a test for C# consuming such cases?

cston: It looks like there is such a test already: `Overrides_ChangePriorityInMetadata_Methods`.

src/Compilers/VisualBasic/Test/Semantic/Semantics/OverloadResolutionPriorityTests.vb
```diff
@@ -1560,5 +1562,1613 @@ End Module
...
+
+    Sub Main
+        Dim i3 As I3 = Nothing
+        i3.M()
+    End Sub
+End Module
+"
+
+            Dim compilation = CreateCompilation(source, options:=TestOptions.DebugExe)
+            compilation.AssertTheseEmitDiagnostics(
```
cston, line 2588: >compilation.AssertTheseEmitDiagnostics [](http://example.com/codeflow?start=12&length=38)

The equivalent C# test avoids verifying diagnostics, presumably to increase the chance of a cycle in attribute binding when using the semantic model. Is that not useful here?

AlekseyTs: >Is that not useful here?

It is done below. The compilation is recreated after we check diagnostics

cston (approved):

PR #76009, opened 2024-11-21
Avoid reusing temps whose refs might be captured

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/67435.

### DIFF
--- a/src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
+++ b/src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
@@ -1,0 +1,158 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Immutable;
+using System.Diagnostics;
+using Microsoft.CodeAnalysis.CSharp.Symbols;
+
+namespace Microsoft.CodeAnalysis.CSharp.CodeGen;
+
+internal partial class CodeGenerator
+{
+    private static bool MightEscapeTemporaryRefs(BoundCall node, bool used, AddressKind? receiverAddressKind)
+    {
+        return MightEscapeTemporaryRefs(
+            used: used,
+            returnType: node.Type,
+            returnRefKind: node.Method.RefKind,
+            receiverType: !node.Method.RequiresInstanceReceiver ? null : node.ReceiverOpt?.Type,
+            receiverScope: node.Method.TryGetThisParameter(out var thisParameter) ? thisParameter?.EffectiveScope : null,
+            receiverAddressKind: receiverAddressKind,
+            isReceiverReadOnly: node.Method.IsEffectivelyReadOnly,
+            parameters: node.Method.Parameters,
+            arguments: node.Arguments,
+            argsToParamsOpt: node.ArgsToParamsOpt,
+            expanded: node.Expanded);
+    }
+
+    private static bool MightEscapeTemporaryRefs(BoundObjectCreationExpression node, bool used)
+    {
+        return MightEscapeTemporaryRefs(
+            used: used,
+            returnType: node.Type,
+            returnRefKind: RefKind.None,
+            receiverType: null,
+            receiverScope: null,
+            receiverAddressKind: null,
+            isReceiverReadOnly: false,
+            parameters: node.Constructor.Parameters,
+            arguments: node.Arguments,
+            argsToParamsOpt: node.ArgsToParamsOpt,
+            expanded: node.Expanded);
+    }
+
+    private static bool MightEscapeTemporaryRefs(BoundFunctionPointerInvocation node, bool used)
+    {
+        FunctionPointerMethodSymbol method = node.FunctionPointer.Signature;
+        return MightEscapeTemporaryRefs(
+            used: used,
+            returnType: node.Type,
+            returnRefKind: method.RefKind,
+            receiverType: null,
+            receiverScope: null,
+            receiverAddressKind: null,
+            isReceiverReadOnly: false,
+            parameters: method.Parameters,
+            arguments: node.Arguments,
+            argsToParamsOpt: default,
+            expanded: false);
+    }
+
+    private static bool MightEscapeTemporaryRefs(
+        bool used,
+        TypeSymbol returnType,
+        RefKind returnRefKind,
+        TypeSymbol? receiverType,
+        ScopedKind? receiverScope,
+        AddressKind? receiverAddressKind,
+        bool isReceiverReadOnly,
+        ImmutableArray<ParameterSymbol> parameters,
+        ImmutableArray<BoundExpression> arguments,
+        ImmutableArray<int> argsToParamsOpt,
+        bool expanded)
+    {
+        Debug.Assert(receiverAddressKind is null || receiverType is not null);
+
+        int writableRefs = 0;
+        int readonlyRefs = 0;
+
+        if (used && (returnRefKind != RefKind.None || returnType.IsRefLikeOrAllowsRefLikeType()))
+        {
+            writableRefs++;
+        }
+
+        if (receiverType is not null)
+        {
+            receiverScope ??= ScopedKind.None;
+            if (receiverAddressKind is { } a && !IsAnyReadOnly(a) && receiverScope == ScopedKind.None)
+            {
+                writableRefs++;
+            }
+            else if (receiverType.IsRefLikeOrAllowsRefLikeType() && receiverScope != ScopedKind.ScopedValue)
+            {
+                if (isReceiverReadOnly || receiverType.IsReadOnly)
+                {
+                    readonlyRefs++;
+                }
+                else
+                {
+                    writableRefs++;
+                }
+            }
+            else if (receiverAddressKind != null && receiverScope == ScopedKind.None)
+            {
+                readonlyRefs++;
+            }
+        }
+
+        if (shouldReturnTrue(writableRefs, readonlyRefs))
+        {
+            return true;
+        }
+
+        for (var arg = 0; arg < arguments.Length; arg++)
+        {
+            var parameter = Binder.GetCorrespondingParameter(
+                arg,
+                parameters,
+                argsToParamsOpt,
+                expanded);
+
+            if (parameter is not null)
+            {
+                if (parameter.RefKind.IsWritableReference() && parameter.EffectiveScope == ScopedKind.None)
+                {
+                    writableRefs++;
+                }
+                else if (parameter.Type.IsRefLikeOrAllowsRefLikeType() && parameter.EffectiveScope != ScopedKind.ScopedValue)
+                {
+                    if (parameter.Type.IsReadOnly || !parameter.RefKind.IsWritableReference())
+                    {
+                        readonlyRefs++;
+                    }
+                    else
+                    {
+                        writableRefs++;
+                    }
+                }
+                else if (parameter.RefKind != RefKind.None && parameter.EffectiveScope == ScopedKind.None)
+                {
+                    readonlyRefs++;
+                }
+            }
+
+            if (shouldReturnTrue(writableRefs, readonlyRefs))
+            {
+                return true;
+            }
+        }
+
+        return false;
+
+        static bool shouldReturnTrue(int writableRefs, int readonlyRefs)
+        {
+            return writableRefs > 0 && (writableRefs + readonlyRefs) > 1;
+        }
+    }
+}
--- a/src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
+++ b/src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
@@ -943,7 +943,7 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
             }
         }
 
-        private void EmitArguments(ImmutableArray<BoundExpression> arguments, ImmutableArray<ParameterSymbol> parameters, ImmutableArray<RefKind> argRefKindsOpt)
+        private void EmitArguments(ImmutableArray<BoundExpression> arguments, ImmutableArray<ParameterSymbol> parameters, ImmutableArray<RefKind> argRefKindsOpt, bool mightEscapeTemporaryRefs)
         {
             // We might have an extra argument for the __arglist() of a varargs method.
             Debug.Assert(arguments.Length == parameters.Length ||
@@ -953,11 +953,19 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
             Debug.Assert(argRefKindsOpt.IsDefault || argRefKindsOpt.Length == arguments.Length ||
                 (argRefKindsOpt.Length == arguments.Length - 1 && arguments is [.., BoundArgListOperator]), "if we have argRefKinds, we should have one for each argument");
 
+            var oldTempRefsMightEscape = _tempRefsMightEscape;
+            if (mightEscapeTemporaryRefs)
+            {
+                _tempRefsMightEscape = true;
+            }
+
             for (int i = 0; i < arguments.Length; i++)
             {
                 RefKind argRefKind = GetArgumentRefKind(arguments, parameters, argRefKindsOpt, i);
                 EmitArgument(arguments[i], argRefKind);
             }
+
+            _tempRefsMightEscape = oldTempRefsMightEscape;
         }
 
         /// <summary>
@@ -1660,7 +1668,11 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
 
             Debug.Assert(method.IsStatic);
 
-            EmitArguments(arguments, method.Parameters, call.ArgumentRefKindsOpt);
+            EmitArguments(
+                arguments,
+                method.Parameters,
+                call.ArgumentRefKindsOpt,
+                mightEscapeTemporaryRefs: MightEscapeTemporaryRefs(call, used: useKind != UseKind.Unused, receiverAddressKind: null));
             int stackBehavior = GetCallStackBehavior(method, arguments);
 
             if (method.IsAbstract || method.IsVirtual)
@@ -1689,6 +1701,7 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
             AddressKind? addressKind;
             bool box;
             LocalDefinition tempOpt;
+            bool mightEscapeTemporaryRefs;
 
             if (receiverIsInstanceCall(call, out BoundCall nested))
             {
@@ -1704,7 +1717,7 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
                 }
 
                 callKind = determineEmitReceiverStrategy(call, out addressKind, out box);
-                emitReceiver(call, callKind, addressKind, box, out tempOpt);
+                emitReceiver(call, callKind, addressKind, box, out tempOpt, out mightEscapeTemporaryRefs);
 
                 while (calls.Count != 0)
                 {
@@ -1754,7 +1767,11 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
                         }
                     }
 
-                    emitArgumentsAndCallEpilogue(call, callKind, receiverUseKind);
+                    emitArgumentsAndCallEpilogue(call, callKind, receiverUseKind,
+                        mightEscapeTemporaryRefs: MightEscapeTemporaryRefs(
+                            call,
+                            used: useKind != UseKind.Unused,
+                            receiverAddressKind: receiverUseKind != UseKind.UsedAsAddress ? null : addressKind));
                     FreeOptTemp(tempOpt);
                     tempOpt = null;
 
@@ -1811,10 +1828,10 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
             else
             {
                 callKind = determineEmitReceiverStrategy(call, out addressKind, out box);
-                emitReceiver(call, callKind, addressKind, box, out tempOpt);
+                emitReceiver(call, callKind, addressKind, box, out tempOpt, out mightEscapeTemporaryRefs);
             }
 
-            emitArgumentsAndCallEpilogue(call, callKind, useKind);
+            emitArgumentsAndCallEpilogue(call, callKind, useKind, mightEscapeTemporaryRefs);
             FreeOptTemp(tempOpt);
 
             return;
@@ -1911,11 +1928,12 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
             }
 
             [MethodImpl(MethodImplOptions.NoInlining)]
-            void emitReceiver(BoundCall call, CallKind callKind, AddressKind? addressKind, bool box, out LocalDefinition tempOpt)
+            void emitReceiver(BoundCall call, CallKind callKind, AddressKind? addressKind, bool box, out LocalDefinition tempOpt, out bool mightEscapeTemporaryRefs)
             {
                 var receiver = call.ReceiverOpt;
                 var receiverType = receiver.Type;
                 tempOpt = null;
+                mightEscapeTemporaryRefs = MightEscapeTemporaryRefs(call, used: useKind != UseKind.Unused, receiverAddressKind: addressKind);
 
                 if (addressKind is null)
                 {
@@ -1932,12 +1950,18 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
                     Debug.Assert(!receiverType.IsVerifierReference());
                     tempOpt = EmitReceiverRef(receiver, addressKind.GetValueOrDefault());
 
+                    if (mightEscapeTemporaryRefs)
+                    {
+                        AddBlockTemp(tempOpt);
+                        tempOpt = null;
+                    }
+
                     emitGenericReceiverCloneIfNecessary(call, callKind, ref tempOpt);
                 }
             }
 
             [MethodImpl(MethodImplOptions.NoInlining)]
-            void emitArgumentsAndCallEpilogue(BoundCall call, CallKind callKind, UseKind useKind)
+            void emitArgumentsAndCallEpilogue(BoundCall call, CallKind callKind, UseKind useKind, bool mightEscapeTemporaryRefs)
             {
                 var method = call.Method;
                 var receiver = call.ReceiverOpt;
@@ -1991,7 +2015,11 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
                 }
 
                 var arguments = call.Arguments;
-                EmitArguments(arguments, method.Parameters, call.ArgumentRefKindsOpt);
+                EmitArguments(
+                    arguments,
+                    method.Parameters,
+                    call.ArgumentRefKindsOpt,
+                    mightEscapeTemporaryRefs: mightEscapeTemporaryRefs);
                 int stackBehavior = GetCallStackBehavior(method, arguments);
                 switch (callKind)
                 {
@@ -2446,7 +2474,11 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
                 }
 
                 // none of the above cases, so just create an instance
-                EmitArguments(expression.Arguments, constructor.Parameters, expression.ArgumentRefKindsOpt);
+                EmitArguments(
+                    expression.Arguments,
+                    constructor.Parameters,
+                    expression.ArgumentRefKindsOpt,
+                    mightEscapeTemporaryRefs: MightEscapeTemporaryRefs(expression, used));
 
                 var stackAdjustment = GetObjCreationStackBehavior(expression);
                 _builder.EmitOpCode(ILOpCode.Newobj, stackAdjustment);
@@ -2704,7 +2736,11 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
             Debug.Assert(temp == null, "in-place ctor target should not create temps");
 
             var constructor = objCreation.Constructor;
-            EmitArguments(objCreation.Arguments, constructor.Parameters, objCreation.ArgumentRefKindsOpt);
+            EmitArguments(
+                objCreation.Arguments,
+                constructor.Parameters,
+                objCreation.ArgumentRefKindsOpt,
+                mightEscapeTemporaryRefs: MightEscapeTemporaryRefs(objCreation, used));
             // -2 to adjust for consumed target address and not produced value.
             var stackAdjustment = GetObjCreationStackBehavior(objCreation) - 2;
             _builder.EmitOpCode(ILOpCode.Call, stackAdjustment);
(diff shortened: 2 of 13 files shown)


src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
```diff
@@ -64,6 +64,23 @@ public override bool Equals(object? obj)
...
+        //         locals: LoweringTemp.1
+        //         ...
+        //       sequence:
+        //         locals: LoweringTemp.1
+        //         ...
+        //
+        // At the same time, these expression locals might be lifted to the containing block
+        // (to avoid reusing them if they might be captured by a ref struct).
+        // Then we use this map to keep track of the redeclared locals.
+        private Dictionary<LocalDefinition, LocalDefinition>? _redeclaredLocals;
```
AlekseyTs, line 82: >private Dictionary<LocalDefinition, LocalDefinition>? _redeclaredLocals; [](http://example.com/codeflow?start=8&length=72)

It is not obvious what do map to what here. Could you, please, elaborate and add a comment?

jjonescz: It maps from the old LocalDefinition to the new LocalDefinition (which is redeclaring/shadowing the old one). Does that make sense?

I will try to explain it in the comment. Thanks.

src/Compilers/CSharp/Portable/Lowering/SyntheticBoundNodeFactory.cs
```diff
@@ -887,7 +887,7 @@ public BoundCall Call(BoundExpression? receiver, MethodSymbol method, ImmutableA
             return new BoundCall(
                 Syntax, receiver, initialBindingReceiverIsSubjectToCloning: ThreeState.Unknown, method, args,
                 argumentNamesOpt: default(ImmutableArray<String?>), argumentRefKindsOpt: refKinds, isDelegateCall: false, expanded: false, invokedAsExtensionMethod: false,
-                argsToParamsOpt: ImmutableArray<int>.Empty, defaultArguments: default(BitVector), resultKind: LookupResultKind.Viable, type: method.ReturnType)
+                argsToParamsOpt: default, defaultArguments: default(BitVector), resultKind: LookupResultKind.Viable, type: method.ReturnType)
```
AlekseyTs, line 890: >argsToParamsOpt: default, [](http://example.com/codeflow?start=16&length=25)

Why is this change necessary?

src/Compilers/CSharp/Portable/CodeGen/EmitStatement.cs
```diff
@@ -673,6 +673,8 @@ private void EmitBlock(BoundBlock block)
             {
                 EmitUninstrumentedBlock(block);
             }
+
+            ReleaseBlockTemps();
```
AlekseyTs, line 677: >ReleaseBlockTemps(); [](http://example.com/codeflow?start=12&length=20)

This feels fragile because we cannot assume that this block necessary maps to one syntactically present in source, which was used for "ref safety analysis already checks refs to rvalues cannot escape blocks". We still might not be done with that block.

src/Compilers/CSharp/Portable/Lowering/SyntheticBoundNodeFactory.cs
jjonescz: Because the utility `Binder.GetCorrespondingParameter` that's used in this PR as part of `CodeGenerator.MightEscapeTemporaryRefs` would fail on an assert - it expects `argsToParamsOpt` to be either default or matching the number of arguments:

https://github.com/dotnet/roslyn/blob/14244f6fcdf77663d7e866ebcdf45e6881a51195/src/Compilers/CSharp/Portable/Binder/Binder_Invocation.cs#L1447

src/Compilers/CSharp/Portable/CodeGen/EmitStatement.cs
jjonescz: > we cannot assume that this block necessary maps to one syntactically present in source

Why cannot we assume that? To me it seems we don't synthesize BoundBlocks.

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
```diff
@@ -64,6 +64,23 @@ public override bool Equals(object? obj)
         // maps local identities to locals.
         private Dictionary<ILocalSymbolInternal, LocalDefinition>? _localMap;
 
+        // The lowered tree might define the same local symbol
+        // in multiple sequences that are part of one expression, for example:
```
AlekseyTs, line 68: >in multiple sequences that are part of one expression, for example: [](http://example.com/codeflow?start=11&length=67)

This doesn't sound right.  I think we should fix this, a local symbol should belong to exactly one scope in the bound tree. I assume this is a pre-existing condition, i.e. it is not introduced by this change. If so, I would prefer the fix to go into a separate PR.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        if (shouldReturnTrue(writableRefs, readonlyRefs))
+        {
+            return true;
+        }
+
+        for (var arg = 0; arg < arguments.Length; arg++)
+        {
+            var parameter = Binder.GetCorrespondingParameter(
```
AlekseyTs, line 116: >Binder.GetCorrespondingParameter( [](http://example.com/codeflow?start=28&length=33)

I think post lowering phase all the parameters must be in the right order, no one is going to reorder them (lowering does that). Therefore, we probably do not need to use this helper.

src/Compilers/CSharp/Portable/CodeGen/EmitStatement.cs
AlekseyTs: > Why cannot we assume that? To me it seems we don't synthesize BoundBlocks.

Because we shouldn't be assuming that. There is nothing wrong in introducing a bound block.

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
jjonescz: Sounds good, I will look into that.

Just to confirm: I think this "invalid" tree shape is introduced somewhere in lowering simply by reusing the same node twice, e.g., something like `_factory.Call(arguments: [node, node])`, and that `node` happens to contain a BoundSequence which then causes the local to be declared twice in the tree. So are you saying we should never do that (reuse the same node) and if we do it somewhere, it's a bug (hopefully a rare one)?

src/Compilers/CSharp/Portable/CodeGen/EmitStatement.cs
jjonescz: > There is nothing wrong in introducing a bound block.

This comment suggests otherwise:

https://github.com/dotnet/roslyn/blob/14244f6fcdf77663d7e866ebcdf45e6881a51195/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml#L1030

AlekseyTs: > This comment suggests otherwise: ...

I am failing to guess what part of the comment you find relevant. Could you quote that part, etc.? 

For example, ```SyntheticBoundNodeFactory``` has a bunch of helpers to synthesize blocks and they are used for one reason or the other. There is also a way to synthesize a block manually. So, even if you haven't found an example, it doesn't mean there isn't one already, or that one won't be introduced in the future.

jjonescz: Sorry, I was referring to this line: "BoundBlock specify SCOPE (visibility) of a variable" but that doesn't imply it's wrong to synthesize blocks, you're right.

Do you have suggestions how to best solve this?

I'm thinking if we can assume blocks are never lost during lowering, we could perhaps make sure synthesized blocks are marked as such (e.g., via WasCompilerGenerated) and then we could call ReleaseBlockTemps only for non-compiler-generated blocks.

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
AlekseyTs: In general, we can reuse bound nodes as long as that doesn't violate some invariants. For example, we reuse some primitive nodes: for constants, for local references (BoundLocal), etc. I consider "a local symbol used in a tree should belong to exactly one scope in the the tree" as one of the invariants that is violated in the specific scenario.

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
```diff
@@ -1660,7 +1668,11 @@ private void EmitStaticCallExpression(BoundCall call, UseKind useKind)
 
             Debug.Assert(method.IsStatic);
 
-            EmitArguments(arguments, method.Parameters, call.ArgumentRefKindsOpt);
+            EmitArguments(
+                arguments,
+                method.Parameters,
+                call.ArgumentRefKindsOpt,
+                mightEscapeTemporaryRefs: MightEscapeTemporaryRefs(call, used: useKind != UseKind.Unused, receiverAddressKind: null));
```
AlekseyTs, line 1675: >mightEscapeTemporaryRefs: MightEscapeTemporaryRefs(call, used: useKind != UseKind.Unused, receiverAddressKind: null) [](http://example.com/codeflow?start=16&length=116)

Instead of threading an extra parameter through many layers, consider simply truncating ```_expressionTemps``` after the ```EmitArguments``` call as ```EmitAssignmentValue``` does.

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
```diff
@@ -1754,7 +1767,11 @@ private void EmitInstanceCallExpression(BoundCall call, UseKind useKind)
                         }
                     }
 
-                    emitArgumentsAndCallEpilogue(call, callKind, receiverUseKind);
+                    emitArgumentsAndCallEpilogue(call, callKind, receiverUseKind,
```
AlekseyTs, line 1770: >emitArgumentsAndCallEpilogue [](http://example.com/codeflow?start=20&length=28)

Similar suggestion about truncating the list vs. passing a parameter.

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
```diff
@@ -2446,7 +2474,11 @@ private void EmitObjectCreationExpression(BoundObjectCreationExpression expressi
                 }
 
                 // none of the above cases, so just create an instance
-                EmitArguments(expression.Arguments, constructor.Parameters, expression.ArgumentRefKindsOpt);
+                EmitArguments(
+                    expression.Arguments,
+                    constructor.Parameters,
+                    expression.ArgumentRefKindsOpt,
+                    mightEscapeTemporaryRefs: MightEscapeTemporaryRefs(expression, used));
```
AlekseyTs, line 2481: >mightEscapeTemporaryRefs: MightEscapeTemporaryRefs(expression, used) [](http://example.com/codeflow?start=20&length=68)

Similar suggestion about truncating the list vs. passing a parameter.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,158 @@
...
+        TypeSymbol returnType,
+        RefKind returnRefKind,
+        TypeSymbol? receiverType,
+        ScopedKind? receiverScope,
+        AddressKind? receiverAddressKind,
+        bool isReceiverReadOnly,
+        ImmutableArray<ParameterSymbol> parameters,
+        ImmutableArray<BoundExpression> arguments,
+        ImmutableArray<int> argsToParamsOpt,
+        bool expanded)
```
AlekseyTs, line 73: The last two parameters should not be relevant post lowering.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,158 @@
...
+                    else
+                    {
+                        writableRefs++;
+                    }
+                }
+                else if (parameter.RefKind != RefKind.None && parameter.EffectiveScope == ScopedKind.None)
+                {
+                    readonlyRefs++;
+                }
+            }
```
AlekseyTs, line 143: >} [](http://example.com/codeflow?start=12&length=1)

What are the scenarios when we can fail to get a parameter and why they are not interesting?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_Call.cs
```diff
@@ -1640,13 +1639,11 @@ private void RewriteArgumentsForComCall(
 
                 actualArguments[argIndex] = new BoundSequence(
                     argument.Syntax,
-                    locals: ImmutableArray<LocalSymbol>.Empty,
-                    sideEffects: ImmutableArray.Create<BoundExpression>(boundAssignmentToTemp),
+                    locals: [boundTemp.LocalSymbol],
```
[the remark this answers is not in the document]
jjonescz: > It looks like we are going to pass a reference to a local that goes out of scope.

True, but even without the change, the temp might live too short - see test `RefTemp_Escapes_ComCall`. Moving the temp into the sequence inside the call ensures the emit layer does not free the temp after it detects the call might escape temps. If the temp is declared outside the call, the emit layer does not see that it is a part of a "might escape" call and that its freeing should be avoided.

It turns out this change is what causes the locals being declared twice in one lowered tree. But I imagine this can be fixed in the callers of this method.

> But that is likely to lead to a worse IL

The mentioned test verifies IL and at least in that scenario it does not look worse.

src/Compilers/CSharp/Portable/Lowering/SyntheticBoundNodeFactory.cs
AlekseyTs: Is this change still necessary? I.e. is anything going to fail if it is reverted? If nothing is going to fail, I suggest unbundling it from the fix this PR implements.

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
```diff
@@ -246,7 +252,54 @@ private LocalDefinition DeclareLocalImpl(
         internal void FreeSlot(LocalDefinition slot)
         {
             Debug.Assert(slot.Name == null);
-            FreeSlots.Push(new LocalSignature(slot.Type, slot.Constraints), slot);
+
+            if (_nonReusableLocals?.Contains(slot) != true)
```
AlekseyTs, line 256: >_nonReusableLocals?.Contains(slot) != true [](http://example.com/codeflow?start=16&length=42)

Would it make sense to remove the entry from the hash table? Do we expect more than one attempt to free a local?

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
```diff
@@ -246,7 +252,54 @@ private LocalDefinition DeclareLocalImpl(
...
+                FreeSlots.Push(new LocalSignature(slot.Type, slot.Constraints), slot);
+            }
+        }
+
+        internal int? StartScopeOfTrackingAddressedLocals()
+        {
+            if (_addressedLocals is null)
+            {
+                _addressedLocals = ArrayBuilder<LocalDefinition>.GetInstance();
+                return null;
```
AlekseyTs, line 267: >null [](http://example.com/codeflow?start=23&length=4)

Would `0` work here as return? What advantage do we get from using nullable result?

AlekseyTs: I see, `null` is used to indicate that we can free the builder. This feels subtle. An explicit "ref count" might be clearer. I assume we expect ```StartScopeOfTrackingAddressedLocals```/```EndScopeOfTrackingAddressedLocals``` calls come in pairs.

AlekseyTs:

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
jjonescz: > Do we expect more than one attempt to free a local?

No, as far as I can tell by investigating the callers of FreeSlot, that should never happen.

> Would it make sense to remove the entry from the hash table?

So I will do that. Thanks.

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
```diff
@@ -246,7 +254,53 @@ private LocalDefinition DeclareLocalImpl(
...
+            }
+
+            _addressedLocalScopes--;
+            if (_addressedLocalScopes > 0)
+            {
+                _addressedLocals.Count = countBefore;
+            }
+            else
+            {
+                Debug.Assert(_addressedLocalScopes == 0);
```
AlekseyTs, line 300: >Debug.Assert(_addressedLocalScopes == 0); [](http://example.com/codeflow?start=16&length=41)

Consider also asserting that ```countBefore``` is 0.

AlekseyTs (approved):
LGTM (commit 9). However, this should not be considered as a sign-off on ```MightEscapeTemporaryRefs``` method. I think we should get a sign-off for its logic from two "ref safety" experts, at least.

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
```diff
@@ -246,7 +254,53 @@ private LocalDefinition DeclareLocalImpl(
...
+            _addressedLocals ??= ArrayBuilder<LocalDefinition>.GetInstance();
+            _addressedLocalScopes++;
+            return _addressedLocals.Count;
+        }
+
+        internal void AddAddressedLocal(LocalDefinition localDef)
+        {
+            if (localDef != null)
+            {
+                _addressedLocals?.Add(localDef);
```
AlekseyTs, line 276: >_addressedLocals?.Add(localDef); [](http://example.com/codeflow?start=16&length=32)

Should user defined locals be subject for this treatment? Presumably ref safety for them was verified during binding. Let's add a unit test with a scenario affecting a user local.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,158 @@
...
+            receiverScope: null,
+            receiverAddressKind: null,
+            isReceiverReadOnly: false,
+            parameters: method.Parameters,
+            arguments: node.Arguments,
+            argsToParamsOpt: default,
+            expanded: false);
+    }
+
+    private static bool MightEscapeTemporaryRefs(
```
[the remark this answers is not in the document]
jaredpar: Agree having a strategy comment would help. Not intuitive at a glance why we're using the approach of counting refs here.

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
jaredpar: my intuition was that this should impact temps only so share the concern about user defined locals.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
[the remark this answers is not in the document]
jjonescz: The ref counters have comments and then they are used to compute the return value, also commented in the `shouldReturnTrue` local function. But I will add another higher-level comment.

src/Compilers/Core/Portable/CodeGen/LocalSlotManager.cs
jjonescz: This does not have any effect on user-defined locals - they are ignored by `CodeGenerator.FreeLocal`, i.e., they are never reused anyway, regardless of whether we add them to the `_addressedLocals` and `_nonReusableLocals` lists. But I could filter them here as well, i.e., avoid adding them to those lists in the first place.

> Let's add a unit test with a scenario affecting a user local.

Not sure which exact scenario you are looking for. There is for example `RefEscapingTests.RefTemp_NoTemp` added as part of this PR.

AlekseyTs: > But I could filter them here as well, i.e., avoid adding them to those lists in the first place if you think that would be better.

Yes, I think it would be better to be explicit about handling of user-defined locals.

jjonescz:

jjonescz:

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,131 @@
...
+        bool used,
+        TypeSymbol returnType,
+        RefKind returnRefKind,
+        TypeSymbol? receiverType,
+        ScopedKind? receiverScope,
+        AddressKind? receiverAddressKind,
+        bool isReceiverReadOnly,
+        ImmutableArray<ParameterSymbol> parameters)
+    {
+        Debug.Assert(receiverAddressKind is null || receiverType is not null);
```
cston, line 63: >|| [](http://example.com/codeflow?start=49&length=2)

Could this be strengthened to `==`?

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,131 @@
...
+            receiverScope ??= ScopedKind.None;
+            if (receiverType.IsRefLikeOrAllowsRefLikeType() && receiverScope != ScopedKind.ScopedValue)
+            {
+                refSources++;
+                if (!isReceiverReadOnly && !receiverType.IsReadOnly)
+                {
+                    refTargets++;
+                }
+            }
+            else if (receiverAddressKind != null && receiverScope == ScopedKind.None)
```
cston, line 91: >receiverAddressKind != null [](http://example.com/codeflow?start=21&length=27)

Do we ever have a receiver and `recevierAddressKind == null`?

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,131 @@
...
+        Debug.Assert(receiverAddressKind is null || receiverType is not null);
+
+        // We check the signature of the method, counting potential `ref` sources and destinations
+        // to determine whether a `ref` can be captured by the method.
+        // The emit layer then uses this information to avoid reusing temporaries that are passed by ref to such methods.
+
+        // number of outputs that can capture `ref`s
+        int refTargets = 0;
+        // number of inputs that can contain `ref`s
+        int refSources = 0;
```
cston, line 72: Just curious: Why count ref inputs and outputs rather than simply checking if there are any of each?

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        if (used && (returnRefKind != RefKind.None || returnType.IsRefLikeOrAllowsRefLikeType()))
+        {
+            // If returning by ref or returning a ref struct, the result might capture `ref`s.
+            refTargets++;
+        }
+
+        if (receiverType is not null)
+        {
+            receiverScope ??= ScopedKind.None;
```
cston, line 82: >??= [](http://example.com/codeflow?start=26&length=3)

Do we ever have a receiver and `receiverScope == null`?

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
jjonescz: The logic used to be more complicated before I realized silly cyclic assignment needs to be handled and then simplified this, so yeah, you're right now booleans should be enough, I will change that, thanks.

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
```diff
@@ -1755,6 +1763,14 @@ private void EmitInstanceCallExpression(BoundCall call, UseKind useKind)
                     }
 
                     emitArgumentsAndCallEpilogue(call, callKind, receiverUseKind);
+
+                    _builder.LocalSlotManager.EndScopeOfTrackingAddressedLocals(countBefore, MightEscapeTemporaryRefs(
+                        call,
+                        used: useKind != UseKind.Unused,
```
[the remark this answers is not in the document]
jjonescz: Good catch, thanks. I think passing `used: true` here is the correct thing to do because it's a nested call, i.e., the receiver of another call, i.e., it is always used.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
jjonescz: Failing CI shows there are actually cases like that:

```
Microsoft.CodeAnalysis.CSharp.UnitTests.Semantics.NativeIntegerTests.BinaryOperators
Microsoft.CodeAnalysis.CSharp.UnitTests.Semantics.NativeIntegerTests.ReadOnlyField_VirtualMethods
Microsoft.CodeAnalysis.CSharp.UnitTests.Semantics.NativeIntegerTests.Retargeting_06
Microsoft.CodeAnalysis.CSharp.UnitTests.Semantics.NativeIntegerTests.SwitchStatement_02
Microsoft.CodeAnalysis.CSharp.UnitTests.Semantics.NativeIntegerTests.SwitchStatement_01
Microsoft.CodeAnalysis.CSharp.UnitTests.Semantics.NativeIntegerTests.MultipleTypeRefs_01
```

It fails on `nint.ToString` which is a `NativeIntegerMethodSymbol` that does not implement TryGetThisParameter. It seems that the correct fix is overriding that, so I will do that.

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
```diff
@@ -1755,6 +1763,14 @@ private void EmitInstanceCallExpression(BoundCall call, UseKind useKind)
                     }
 
                     emitArgumentsAndCallEpilogue(call, callKind, receiverUseKind);
+
+                    _builder.LocalSlotManager.EndScopeOfTrackingAddressedLocals(countBefore, MightEscapeTemporaryRefs(
+                        call,
+                        used: true,
```
cston, line 1769: >used: true [](http://example.com/codeflow?start=24&length=10)

Is there a relatively simple test we could add to test this change?

cston (approved):

src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs
jjonescz: Actually yes, after giving it another thought I came up with a test hitting this. Thanks.

cston (approved):

RikkiGibson (commented):
These comments are quite old, but Jan suggested to just publish them so that he can sort out which ones are still of use.

RikkiGibson:
A more general thought I had when I was reviewing is that it seems there are several places which are doing the equivalent of a "best common type", except for safe-contexts. This, as well as "method arguments must match" checks, and likely other places. Possibly a shared implementation should be used for these.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,131 @@
...
+
+internal partial class CodeGenerator
+{
+    private static bool MightEscapeTemporaryRefs(BoundCall node, bool used, AddressKind? receiverAddressKind)
+    {
+        return MightEscapeTemporaryRefs(
+            used: used,
+            returnType: node.Type,
+            returnRefKind: node.Method.RefKind,
+            receiverType: !node.Method.RequiresInstanceReceiver ? null : node.ReceiverOpt?.Type,
```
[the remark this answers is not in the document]
jjonescz: Thanks, that allows me to simplify and share the logic between the receiver and other parameter nicely.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,131 @@
...
+        ImmutableArray<ParameterSymbol> parameters)
+    {
+        Debug.Assert(receiverAddressKind is null || receiverType is not null);
+
+        // We check the signature of the method, counting potential `ref` sources and destinations
+        // to determine whether a `ref` can be captured by the method.
+        // The emit layer then uses this information to avoid reusing temporaries that are passed by ref to such methods.
+
+        // number of outputs that can capture `ref`s
+        int refTargets = 0;
```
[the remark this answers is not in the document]
jjonescz: Both of these should be already addressed.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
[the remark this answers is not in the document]
jjonescz: > I'm concerned that we may have to re-implement too much of the logic of "method arguments must match" here, i.e. the process of searching for the possible sources and destinations of escapes.

True, although I think the logic is sufficiently different given here we work on lowered trees and are interested in only a heuristic (we are fine with some false positives).

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            return true;
+        }
+
+        foreach (var parameter in parameters)
+        {
+            if (parameter.Type.IsRefLikeOrAllowsRefLikeType() && parameter.EffectiveScope != ScopedKind.ScopedValue)
+            {
+                refSources++;
+                if (!parameter.Type.IsReadOnly && parameter.RefKind.IsWritableReference())
```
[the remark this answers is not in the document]
jjonescz: Good catch, thanks.

jjonescz:
@RikkiGibson when you have time again, can you take another look? Thanks.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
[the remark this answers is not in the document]
RikkiGibson: I'm not fully convinced this rationale applies, because this check is only concerning itself with the method we are calling, lowered trees are not involved.

However, the implementation is sufficiently simple that I am OK with keeping it this way.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// </remarks>
+    private static bool MightEscapeTemporaryRefs(
+        bool used,
+        TypeSymbol returnType,
+        RefKind returnRefKind,
+        ParameterSymbol? thisParameterSymbol,
+        AddressKind? receiverAddressKind,
+        ImmutableArray<ParameterSymbol> parameters)
+    {
+        Debug.Assert(receiverAddressKind is null || thisParameterSymbol is not null);
```
RikkiGibson, line 73: It looks like the `receiverAddressKind` is now used only in an assertion, should we eliminate this and the same from the other overloads?

RikkiGibson (commented):
MightEscapeTemporaryRefs looks good to me, but, I would like some time to get thru the tests.

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator_RefSafety.cs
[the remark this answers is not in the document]
jjonescz: > because this check is only concerning itself with the method we are calling, lowered trees are not involved.

You're right, that's currently true. We could check arguments as well though if we wanted to improve precision of the check. For example, we could check for discards (currently I think it is not possible to recognize them in lowered trees, but we could preserve some flag info during lowering). Similarly, the binding-time ref safety analysis does check arguments. I don't think it would be possible to reuse the logic without some significant refactoring.

RikkiGibson:
I will try to get to this this week

RikkiGibson (approved):

PR #76010, opened 2024-11-21
[Backport] RedirectFeaturesAnalyzers test fixes

JoeRobich:
Backports integration test fixes from https://github.com/dotnet/roslyn/pull/75672

### DIFF
--- a/src/VisualStudio/IntegrationTest/New.IntegrationTests/CSharp/CSharpRedirectFeaturesAnalyzers.cs
+++ b/src/VisualStudio/IntegrationTest/New.IntegrationTests/CSharp/CSharpRedirectFeaturesAnalyzers.cs
@@ -145,7 +145,7 @@ public class CSharpRedirectFeaturesAnalyzers : AbstractEditorTest
                 FeatureAttribute.ErrorList
              ],
              cancellationToken);
-        return await TestServices.ErrorList.GetErrorsAsync(ErrorSource.Other, __VSERRORCATEGORY.EC_MESSAGE, cancellationToken);
+        return await TestServices.ErrorList.GetErrorsAsync(ErrorSource.Other, __VSERRORCATEGORY.EC_WARNING, cancellationToken);
     }
 
     private async Task WaitForCodeActionListToPopulateAsync(CancellationToken cancellationToken)


arunchndr (approved):

src/VisualStudio/IntegrationTest/New.IntegrationTests/CSharp/CSharpRedirectFeaturesAnalyzers.cs
```diff
@@ -143,14 +145,16 @@ await TestServices.Workspace.WaitForAllAsyncOperationsAsync(
...
-        return await TestServices.ErrorList.GetErrorsAsync(cancellationToken);
+        return await TestServices.ErrorList.GetErrorsAsync(ErrorSource.Other, __VSERRORCATEGORY.EC_WARNING, cancellationToken);
     }
 
     private async Task WaitForCodeActionListToPopulateAsync(CancellationToken cancellationToken)
     {
         await TestServices.Editor.ActivateAsync(cancellationToken);
         await TestServices.Editor.PlaceCaretAsync("void M()", charsOffset: -1, cancellationToken);
 
+        await Task.Delay(TimeSpan.FromSeconds(1));
```
sharwell, line 156: 😬 Why this?

src/VisualStudio/IntegrationTest/New.IntegrationTests/CSharp/CSharpRedirectFeaturesAnalyzers.cs
```diff
@@ -143,14 +145,16 @@ await TestServices.Workspace.WaitForAllAsyncOperationsAsync(
                 FeatureAttribute.ErrorList
              ],
              cancellationToken);
-        return await TestServices.ErrorList.GetErrorsAsync(cancellationToken);
+        return await TestServices.ErrorList.GetErrorsAsync(ErrorSource.Other, __VSERRORCATEGORY.EC_WARNING, cancellationToken);
```
sharwell, line 148: ❔ Why this change? None of the tests build the project, so `ErrorSource.Build` (which this change explictly omits) is guaranteed to be empty. The default behavior of this call is a better match to the default and most common IDE configuration for users.

src/VisualStudio/IntegrationTest/New.IntegrationTests/CSharp/CSharpRedirectFeaturesAnalyzers.cs
sharwell: Note that `InvokeCodeActionListAsync` invokes `WaitForLightBulbAsync`, so this delay will not improve stability.

PR #76036, opened 2024-11-22
Emit opted-in string literals into data section as UTF8

jjonescz:
Spec: https://github.com/dotnet/roslyn/pull/76139

When this opt-in feature flag is enabled by users, string literals with length over the configured threshold are emitted into data section of the PE file, hence allowing users to overcome the "error CS8103: Combined length of user strings used by the program exceeds allowed limit."

VB support is not implemented in this PR, could be added in the future if there's demand.

I've manually tried this on our unit tests (e.g., Emit2 compiled standalone resulted in significantly smaller file size due to the use of UTF8 instead of UTF16; run time was the same; combining Emit2 and Emit3 into one assembly worked without emit errors). It was also tested on aspnetcore benchmarks and that didn't reveal any significant runtime/build perf changes.

Main motivation for this feature are Razor projects which can accumulate lots of string literals over time (due to how `.razor` file compilation works) and then customers are one day hit with this error without any prior warnings (imagine they are fixing a small bug and suddenly they cannot compile anymore without refactoring their codebase into multiple projects).


docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,18 @@
...
+
+This opt-in Roslyn feature allows changing how string literals in C# programs are emitted into PE files (`.dll`/`.exe`).
+By default, string literals are emitted to the UserString heap which is limited to [2^24 bytes](https://github.com/dotnet/roslyn/issues/9852).
+When the limit is reached, the following compiler error is reported by Roslyn:
+
+```
+error CS8103: Combined length of user strings used by the program exceeds allowed limit. Try to decrease use of string literals.
+```
+
+By turning on the feature flag `utf8-string-encoding`, string literals (where possible) are instead emitted as UTF-8 data into a different section of the PE file
```
AlekseyTs, line 11: >utf8-string-encoding [](http://example.com/codeflow?start=32&length=20)

"utf8-string-**literal**-encoding"?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -229,31 +231,14 @@ internal Cci.IFieldReference CreateArrayCachingField(ImmutableArray<ConstantValu
...
-        /// .pack value isn't required by anything else in the compiler (the compiler always aligns RVA fields at 8-byte
-        /// boundaries, which accomodates any element type that's relevant), it is necessary for IL rewriters. Such rewriters
-        /// also need to ensure an appropriate alignment is maintained for the RVA field, and while they could also simplify
-        /// by choosing a worst-case alignment as does the compiler, they may instead use the .pack value as the alignment
-        /// to use for that field, since it's an opaque blob with no other indication as to what kind of data is
-        /// stored and what alignment might be required.
-        /// </param>
-        /// <returns>The field. This may have been newly created or may be an existing field previously created for the same data and alignment.</returns>
-        internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort alignment)
+        internal Cci.ITypeReference GetOrAddProxyType(int length, ushort alignment)
```
AlekseyTs, line 234: >GetOrAddProxyType [](http://example.com/codeflow?start=36&length=17)

The name looks meaningless to me. "GetOrAddDataFieldType"?

AlekseyTs: Consider renaming ```_proxyTypes``` as well

src/Compilers/Core/Portable/CodeGen/DataStringHolder.cs
```diff
@@ -0,0 +1,424 @@
...
+
+    public override Cci.INamespaceTypeReference AsNamespaceTypeReference => this;
+
+    public Cci.IFieldReference CreateDataField(ImmutableArray<byte> data)
+    {
+        Cci.ITypeReference type = _privateImplementationDetails.GetOrAddProxyType(data.Length, alignment: 1);
+
+        Debug.Assert(_mappedFields.IsEmpty, "We need to generate unique field names if we are synthesizing more than one.");
+
+        (_, StringField stringField) = _mappedFields.GetOrAdd((data, Alignment: 1), key =>
```
AlekseyTs, line 82: >_mappedFields [](http://example.com/codeflow?start=39&length=13)

Why do we need a ```ConcurrentDictionary``` if we create at most one field?

AlekseyTs:
@jjonescz Please add details how do we achieve the goal: what artifacts are getting generated, what APIs are needed and how they are used, etc. Basically, I think we need something like an implementation speclet for the compiler feature.

src/Compilers/Core/Portable/CodeGen/DataStringHolder.cs
```diff
@@ -0,0 +1,424 @@
...
+
+        foreach (var (mappedField, stringField) in _mappedFields.Values)
+        {
+            fieldsBuilder.Add(mappedField);
+            fieldsBuilder.Add(stringField);
+        }
+
+        fieldsBuilder.Sort(PrivateImplementationDetails.FieldComparer.Instance);
+
+        _orderedSynthesizedFields = fieldsBuilder.ToImmutableAndFree();
```
AlekseyTs, line 112: >_orderedSynthesizedFields = fieldsBuilder.ToImmutableAndFree(); [](http://example.com/codeflow?start=8&length=63)

It looks like we always synthesize two fields. It feels like all the code around ```fieldsBuilder``` is unnecessary complicated.

src/Compilers/Core/Portable/CodeGen/DataStringHolder.cs
```diff
@@ -0,0 +1,424 @@
...
+            ilBuilder.EmitOpCode(ILOpCode.Ldsflda);
+            ilBuilder.EmitToken(stringField.MappedField, null, diagnostics);
+
+            // Push the byte size.
+            ilBuilder.EmitIntConstant(stringField.MappedField.MappedData.Length);
+
+            // Call `<PrivateImplementationDetails>.BytesToString(byte*, int)`.
+            bytesToStringHelper ??= GetOrSynthesizeBytesToStringHelper(diagnostics);
+            ilBuilder.EmitOpCode(ILOpCode.Call, -1);
+            ilBuilder.EmitToken(bytesToStringHelper, null, diagnostics);
```
AlekseyTs, line 214: >bytesToStringHelper [](http://example.com/codeflow?start=32&length=19)

Given that we only ever have one field to initialize, it doesn't feel like we need this helper method. We can simply inline what it is doing

src/Compilers/Core/Portable/CodeGen/DataStringHolder.cs
```diff
@@ -0,0 +1,424 @@
...
+using System.Threading;
+using Microsoft.CodeAnalysis.Debugging;
+using Microsoft.CodeAnalysis.Emit;
+using Microsoft.CodeAnalysis.PooledObjects;
+using Microsoft.CodeAnalysis.Symbols;
+using Roslyn.Utilities;
+
+namespace Microsoft.CodeAnalysis.CodeGen;
+
+internal sealed class DataStringHolder : DefaultTypeDef, Cci.INamespaceTypeDefinition
```
AlekseyTs, line 21: >DataStringHolder [](http://example.com/codeflow?start=22&length=16)

I could be missing something, but it feels like we don't really need this type. Everything what it is doing can be done by PrivateImplementationDetail

AlekseyTs:
I stopped reviewing the PR because I think first we need to agree on the implementation strategy and have a speclet to review against.

src/Compilers/Core/Portable/CodeGen/DataStringHolder.cs
jjonescz: We could have an alternative implementation strategy that has more than one field in one class. But I will remove it for now.

src/Compilers/Core/Portable/CodeGen/DataStringHolder.cs
jjonescz: I think you're right, that seems possible, thanks. Then it shouldn't be necessary to re-implement some of the thread safety and determinism code.

jjonescz: Can you elaborate?

I just noticed there is `PrivateImplementationDetails.TryAddSynthesizedType(Cci.INamespaceTypeDefinition type)` which we could definitely use, hopefully simplifying some thread safety / determinism code.

But we still need a `Cci.INamespaceTypeDefinition` representing the generated `<S>` classes.

jjonescz:
@AlekseyTs thanks for the review. Sounds like a good idea to get a spec reviewed first. Opened a separate PR for that: https://github.com/dotnet/roslyn/pull/76139

docs/features/string-literals-data-section.md
jjonescz: In the spec, I've renamed it to `data-section-string-literals` which IMO best reflects what the feature is about.

AlekseyTs:
@jjonescz The current PR title: "Add feature flag for emitting string literals into data section as UTF8" sounds like the only thing this PR is doing is adding a flag. I suggest to adjust the title to reflect the main goal/purpose of the PR, which I think is to store values of string literals as Utf-8 encoded bytes in a blob, etc. By comparison to that, the feature flag is a minor detail, which, I think, can be dropped from the title if its length becomes too big.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -292,8 +319,131 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
                 };
 
-                return new MappedField(name, this, type, key.Data);
-            });
+                return new MappedField(name, @this, type, data);
+            },
+            this);
+        }
+
+        internal DataSectionStringType? GetOrCreateDataSectionStringLiteralType(
```
AlekseyTs, line 327: >DataSectionStringType? [](http://example.com/codeflow?start=17&length=22)

Does this function ever return `null` value? The name of the function implies that it doesn't.

src/Compilers/Core/Portable/CodeGen/ITokenDeferral.cs
```diff
@@ -21,6 +21,8 @@ internal interface ITokenDeferral
 
         Cci.IFieldReference GetArrayCachingFieldForConstants(ImmutableArray<ConstantValue> constants, Cci.IArrayTypeReference arrayType, SyntaxNode syntaxNode, DiagnosticBag diagnostics);
 
+        Cci.IFieldReference? GetFieldForDataString(string text, ImmutableArray<byte> data, SyntaxNode syntaxNode, DiagnosticBag diagnostics);
```
AlekseyTs, line 24: >GetFieldForDataString [](http://example.com/codeflow?start=29&length=21)

It feels like the "GetFieldForDataString" name is somewhat confusing. What is a "data string" for which we are requesting a field? It looks like we are requesting a static field that is supposed to store the value of the ```text``` parameter during at runtime. Therefore, "GetFieldForStringValue" feels better.

AlekseyTs: Also, If we expect this function to fail to produce a field under some conditions, consider adding "Try" prefix to the name of the function.

AlekseyTs, line 24: >ImmutableArray<byte> data [](http://example.com/codeflow?start=64&length=25)

It looks like the ```data``` parameter is redundant. Its value is defined by content of the ```text``` parameter. Passing redundant data is a possible source of bugs because it becomes responsibility of each consumer to keep data in sync. Not to mention the fact that each consumer is responsible to do more work.

AlekseyTs, line 24: >GetFieldForDataString [](http://example.com/codeflow?start=29&length=21)

It looks like this API is not consumed from any shared code. Therefore, it looks unnecessary to have it in this interface.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -175,8 +191,9 @@ internal void Freeze()
             // Sort top-level types.
             _orderedTopLevelTypes = _synthesizedTopLevelTypes.OrderBy(kvp => kvp.Key).Select(kvp => (Cci.INamespaceTypeDefinition)kvp.Value).AsImmutable();
 
-            // Sort proxy types.
-            _orderedProxyTypes = _proxyTypes.OrderBy(kvp => kvp.Key.Size).ThenBy(kvp => kvp.Key.Alignment).Select(kvp => kvp.Value).AsImmutable();
+            // Sort nested types.
+            _orderedNestedTypes = _dataFieldTypes.OrderBy(kvp => kvp.Key.Size).ThenBy(kvp => kvp.Key.Alignment).Select(kvp => kvp.Value).OfType<ExplicitSizeStruct>()
```
AlekseyTs, line 195: >.OfType<ExplicitSizeStruct>() [](http://example.com/codeflow?start=136&length=29)

The purpose of ```OfType``` API is to filter the input by type. I do not think that we intend to do that here. If our goal is to do a simple type cast, there is a ```Cast``` helper, but, I think it would be more efficient to perform the cast in the lambda passed to the preceding ```Select``` call.

AlekseyTs: Would it make sense to strengthen the type of values stored in ```_dataFieldTypes``` to ```ExplicitSizeStruct``` or to ```Cci.INestedTypeDefinition```?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -275,15 +278,39 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+        /// the data, and that type needs its .pack set to the alignment required for the underlying data. While that
+        /// .pack value isn't required by anything else in the compiler (the compiler always aligns RVA fields at 8-byte
+        /// boundaries, which accomodates any element type that's relevant), it is necessary for IL rewriters. Such rewriters
+        /// also need to ensure an appropriate alignment is maintained for the RVA field, and while they could also simplify
+        /// by choosing a worst-case alignment as does the compiler, they may instead use the .pack value as the alignment
+        /// to use for that field, since it's an opaque blob with no other indication as to what kind of data is
+        /// stored and what alignment might be required.
+        /// </param>
+        /// <returns>The field. This may have been newly created or may be an existing field previously created for the same data and alignment.</returns>
+        internal MappedField CreateDataField(ImmutableArray<byte> data, ushort alignment)
```
AlekseyTs, line 300: >MappedField [](http://example.com/codeflow?start=17&length=11)

Is there a specific reason for changing return type of this function?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -612,6 +770,101 @@ public override void Dispatch(Cci.MetadataVisitor visitor)
...
+            Cci.ITypeReference systemObject,
+            Cci.ITypeReference systemString,
+            Cci.INamespaceTypeDefinition containingType,
+            MappedField dataField,
+            Cci.IMethodDefinition bytesToStringHelper,
+            DiagnosticBag diagnostics)
+        {
+            _name = name;
+            _systemObject = systemObject;
+            _containingType = containingType;
```
AlekseyTs, line 799: >_containingType [](http://example.com/codeflow?start=12&length=15)

Would it make sense to use ```PrivateImplementationDetails``` for this field and get rid of the ```_systemObject``` field? Also we could avoid the need to pass ```systemString``` and ```bytesToStringHelper``` explicitly.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -292,8 +319,131 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+            static Cci.IMethodDefinition synthesizeBytesToStringHelper(
+                PrivateImplementationDetails @this,
+                Compilation compilation,
+                SyntaxNode syntaxNode,
+                DiagnosticBag diagnostics)
+            {
+                if (!tryGetWellKnownTypeMember(compilation, syntaxNode, diagnostics, WellKnownMember.System_Text_Encoding__get_UTF8, out var encodingUtf8) |
+                    !tryGetWellKnownTypeMember(compilation, syntaxNode, diagnostics, WellKnownMember.System_Text_Encoding__GetString, out var encodingGetString))
+                {
+                    return new DefaultMethodDef(
```
AlekseyTs, line 383: >return new DefaultMethodDef( [](http://example.com/codeflow?start=20&length=28)

I am not sure what is going on here. Are we creating a "bogus" method definition in order to recover from an error? I think it would be preferred to not do that.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -911,4 +1158,192 @@ public sealed override int GetHashCode()
...
+            Cci.TypeMemberVisibility.Private,
+            containingType,
+            maxStack,
+            il)
+    {
+        public override bool IsRuntimeSpecial => true;
+        public override bool IsSpecialName => true;
+    }
+
+    file class DefaultMethodDef(
```
AlekseyTs, line 1215: >DefaultMethodDef [](http://example.com/codeflow?start=15&length=16)

I do not think we have a concept of "default method definition" and I think it would be better to not introduce one

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -292,8 +319,131 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+                        name: SynthesizedBytesToStringFunctionName,
+                        visibility: Cci.TypeMemberVisibility.Private,
+                        containingType: @this,
+                        maxStack: 8,
+                        il: []);
+                }
+
+                Debug.Assert(encodingUtf8 is not null && encodingGetString is not null);
+
+                var ilBuilder = new ILBuilder(
```
AlekseyTs, line 393: >var ilBuilder = new ILBuilder( [](http://example.com/codeflow?start=16&length=30)

I think it would be better to encapsulate the IL emit logic in ```BytesToStringHelper``` type.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -292,8 +319,131 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+            }
+
+            static bool tryGetWellKnownTypeMember(
+                Compilation compilation,
+                SyntaxNode syntaxNode,
+                DiagnosticBag diagnostics,
+                WellKnownMember member,
+                [NotNullWhen(returnValue: true)] out Cci.IMethodReference? result)
+            {
+                if (compilation.CommonGetWellKnownTypeMember(member) is { } symbol)
```
AlekseyTs, line 431: >if (compilation.CommonGetWellKnownTypeMember(member) is { } symbol) [](http://example.com/codeflow?start=16&length=67)

In general, I think there is no guarantee that the symbol returned by ```CommonGetWellKnownTypeMember``` is good. We should find a way to do what ```Binder.GetWellKnownTypeMember``` is doing.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -449,32 +599,40 @@ public Cci.IUnitReference GetUnit(EmitContext context)
 
         public string NamespaceName => string.Empty;
 
-        private static string DataToHex(ImmutableArray<byte> data)
+        internal static string DataToHex(ImmutableArray<byte> data)
```
AlekseyTs, line 602: >internal [](http://example.com/codeflow?start=8&length=8)

Is the accessibility change necessary?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -449,32 +599,40 @@ public Cci.IUnitReference GetUnit(EmitContext context)
...
         {
             ImmutableArray<byte> hash = CryptographicHashProvider.ComputeSourceHash(data);
+            return HashToHex(hash.AsSpan());
+        }
+
+        internal static string DataToHexViaXxHash128(ImmutableArray<byte> data)
+        {
+            Span<byte> hash = stackalloc byte[sizeof(ulong) * 2];
+            var bytesWritten = XxHash128.Hash(data.AsSpan(), hash);
```
AlekseyTs, line 611: >var [](http://example.com/codeflow?start=12&length=3)

Consider using an explicit type

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -484,7 +642,7 @@ static void toHex(ImmutableArray<byte> source, Span<char> destination)
             static char hexchar(int x) => (char)((x <= 9) ? (x + '0') : (x + ('A' - 10)));
         }
 
-        private sealed class FieldComparer : IComparer<SynthesizedStaticField>
+        internal sealed class FieldComparer : IComparer<SynthesizedStaticField>
```
AlekseyTs, line 645: >internal [](http://example.com/codeflow?start=8&length=8)

Is the accessibility change necessary?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -502,7 +660,7 @@ public int Compare(SynthesizedStaticField? x, SynthesizedStaticField? y)
             }
         }
 
-        private sealed class DataAndUShortEqualityComparer : EqualityComparer<(ImmutableArray<byte> Data, ushort Value)>
+        internal sealed class DataAndUShortEqualityComparer : EqualityComparer<(ImmutableArray<byte> Data, ushort Value)>
```
AlekseyTs, line 663: >internal [](http://example.com/codeflow?start=8&length=8)

Is the accessibility change necessary?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -612,6 +770,101 @@ public override void Dispatch(Cci.MetadataVisitor visitor)
         public override Cci.INestedTypeReference AsNestedTypeReference => this;
     }
 
+    /// <summary>
+    /// A type synthesized for each eligible string literal to hold the lazily-initialized string.
+    ///
+    /// https://github.com/dotnet/roslyn/blob/main/docs/features/string-literals-data-section.md
+    /// </summary>
+    internal sealed class DataSectionStringType : DefaultTypeDef, Cci.INestedTypeDefinition
```
AlekseyTs, line 778: >DefaultTypeDef [](http://example.com/codeflow?start=50&length=14)

I think it would be good to extract a common abstract base "NestedTypeDefinition" to share across ```ExplicitSizeStruct``` and ```DataSectionStringType```

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -612,6 +770,101 @@ public override void Dispatch(Cci.MetadataVisitor visitor)
...
+            _name = name;
+            _systemObject = systemObject;
+            _containingType = containingType;
+
+            var stringField = new DataSectionStringField("s", this, systemString);
+
+            var staticConstructor = synthesizeStaticConstructor(module, containingType, dataField, stringField, bytesToStringHelper, diagnostics);
+
+            _field = stringField;
+            _fields = [_field];
```
AlekseyTs, line 806: >_field [](http://example.com/codeflow?start=23&length=6)

It looks like the ```_field``` is redundant, we can get its value from ```_fields```

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
AlekseyTs, line 327: >GetOrCreateDataSectionStringLiteralType [](http://example.com/codeflow?start=40&length=39)

It doesn't look like anyone is interested in the type. The only consumer is interested in a field. I think we need to adjust this API accordingly.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
AlekseyTs: It looks like we already have ```RootModuleStaticConstructor``` that implements ```Cci.IMethodDefinition, Cci.IMethodBody```. Let's extract a common abstract base "MethodDefinition", rename ```RootModuleStaticConstructor``` to ```StaticConstructor``` and reuse the new base across ```StaticConstructor``` and ```BytesToStringHelper```.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -911,4 +1158,192 @@ public sealed override int GetHashCode()
             throw Roslyn.Utilities.ExceptionUtilities.Unreachable();
         }
     }
+
+    /// <summary>
+    /// A helper method used in static constructor of <see cref="DataSectionStringType"/>.
```
AlekseyTs, line 1163: >A helper method used in static constructor of <see cref="DataSectionStringType"/>. [](http://example.com/codeflow?start=8&length=82)

I think it would be good to include information what signature this method is supposed to have.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -911,4 +1158,192 @@ public sealed override int GetHashCode()
...
+        string name,
+        Cci.PlatformType type)
+        : ParameterDefinitionBase(index, name)
+    {
+        private readonly Cci.PlatformType _type = type;
+
+        public override Cci.ITypeReference GetType(EmitContext context) => context.Module.GetPlatformType(_type, context);
+    }
+
+    file abstract class ParameterDefinitionBase(
```
AlekseyTs, line 1320: >ParameterDefinitionBase [](http://example.com/codeflow?start=24&length=23)

Consider whether there is a benefit in extracting and sharing a common base with ```ReturnValueParameter```.

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
```diff
@@ -1088,6 +1095,7 @@ internal PrivateImplementationDetails GetPrivateImplClass(TSyntaxNode syntaxNode
                         this.GetSpecialType(SpecialType.System_Int16, syntaxNodeOpt, diagnostics),
                         this.GetSpecialType(SpecialType.System_Int32, syntaxNodeOpt, diagnostics),
                         this.GetSpecialType(SpecialType.System_Int64, syntaxNodeOpt, diagnostics),
+                        this.GetSpecialType(SpecialType.System_String, syntaxNodeOpt, diagnostics),
```
AlekseyTs, line 1098: >this.GetSpecialType(SpecialType.System_String, syntaxNodeOpt, diagnostics), [](http://example.com/codeflow?start=24&length=75)

Let's say we have a program that utilizes a private implementation detail class, but doesn't use string type, and the string type is not defined. Is this code going to break such a program?

src/Compilers/Core/Portable/PEWriter/Members.cs
```diff
@@ -415,7 +415,7 @@ ImmutableArray<ExceptionHandlerRegion> ExceptionRegions
         /// <summary>
         /// Debugging information associated with a MoveNext method of a state machine.
         /// </summary>
-        StateMachineMoveNextBodyDebugInfo MoveNextBodyInfo { get; }
+        StateMachineMoveNextBodyDebugInfo? MoveNextBodyInfo { get; }
```
AlekseyTs, line 418: >StateMachineMoveNextBodyDebugInfo? MoveNextBodyInfo { get; } [](http://example.com/codeflow?start=8&length=60)

Are changes in this file primarily related to the goal of the PR?

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator.cs
```diff
@@ -101,6 +104,7 @@ public CodeGenerator(
             _builder = builder;
             _module = moduleBuilder;
             _diagnostics = diagnostics;
+            _dataSectionStringLiteralThreshold = compilation.DataSectionStringLiteralThreshold;
```
AlekseyTs, line 107: >_dataSectionStringLiteralThreshold = compilation.DataSectionStringLiteralThreshold; [](http://example.com/codeflow?start=12&length=83)

Would it make sense to cache this information on the Compilation instead?

src/Compilers/CSharp/Portable/CodeGen/CodeGenerator.cs
```diff
@@ -87,7 +89,8 @@ public CodeGenerator(
             PEModuleBuilder moduleBuilder,
             BindingDiagnosticBag diagnostics,
             OptimizationLevel optimizations,
-            bool emittingPdb)
+            bool emittingPdb,
+            Compilation compilation)
```
AlekseyTs, line 93: >Compilation compilation [](http://example.com/codeflow?start=12&length=23)

It looks like Compilation is available from PEModuleBuilder

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_Conversion.cs
```diff
@@ -120,20 +121,37 @@ private BoundExpression MakeUtf8Span(BoundExpression node, IReadOnlyList<byte>?
             return result;
         }
 
-        private byte[]? GetUtf8ByteRepresentation(BoundUtf8String node)
+        internal static bool TryGetUtf8ByteRepresentation(
+            string s,
+            [NotNullWhen(returnValue: true)] out byte[]? result,
+            [NotNullWhen(returnValue: false)] out string? error)
         {
-            var utf8 = new System.Text.UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
```
AlekseyTs, line 125: >var utf8 = new System.Text.UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true); [](http://example.com/codeflow?start=12&length=107)

Is there a specific reason behind caching this value? Would it make sense to initialize the cache on demand rather than using a field initializer?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
AlekseyTs: We can make it a responsibility of the only consumer of this functionality, the ```TryEmitStringLiteralAsUtf8Encoded```, to check the presence/goodness of required APIs and to report appropriate diagnostics. In this file, we could assume that relevant APIs won't be called unless the helpers are good. I think if we do that, the logic in this file can be simplified, changes to ```CommonMessageProvider``` can be reverted, etc.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
AlekseyTs: While extracting common bases, I recommend extract/copy/paste code from pre-existing types, without changing relative order of members. That is going to make code review much easier.

AlekseyTs:
Done with review pass (commit 2), tests are not looked at.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
jjonescz: `_dataFieldTypes` might contain System.Int* types, not just ExplicitSizeStructs. The filtering via OfType was always performed - it has just moved from here: [https://github.com/dotnet/roslyn/blob/98ea496177bc8e607dbf454abd6b5a5e4678aed2/src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs#L414](https://github.com/dotnet/roslyn/blob/98ea496177bc8e607dbf454abd6b5a5e4678aed2/src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs#L414)

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
jjonescz: Yes, its MappedData.Length is accessed inside the synthesizeStaticConstructor local function.

src/Compilers/Core/Portable/PEWriter/Members.cs
jjonescz: I don't remember changing this file, might be wrong merge/rebase, will revert.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_Conversion.cs
jjonescz: > Is there a specific reason behind caching this value?

No specific reason other than avoiding unnecessary allocations.

> Would it make sense to initialize the cache on demand rather than using a field initializer?

Yes, thanks.

src/Compilers/Core/Portable/PEWriter/Members.cs
jjonescz: No but they allow the new code to implement these interfaces without nullability warnings.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -292,8 +311,77 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+        }
+
+        /// <summary>
+        /// Gets the field of <see cref="DataSectionStringType"/> or creates one
+        /// if the type does not exist yet for the given <paramref name="text"/>.
+        /// </summary>
+        internal Cci.IFieldReference GetOrCreateFieldForStringValue(
+            string text,
+            ImmutableArray<byte> data,
+            Compilation compilation,
```
AlekseyTs, line 326: >Compilation compilation [](http://example.com/codeflow?start=12&length=23)

It looks like compilation is available from ```_moduleBuilder```

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -292,8 +311,77 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+            Cci.ITypeReference systemStringType,
```
AlekseyTs, line 327: >Cci.ITypeReference systemStringType [](http://example.com/codeflow?start=12&length=35)

It looks like string type can be requested from ```moduleBuilder```

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -292,8 +311,77 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+            this);
+        }
+
+        /// <summary>
+        /// Gets the field of <see cref="DataSectionStringType"/> or creates one
+        /// if the type does not exist yet for the given <paramref name="text"/>.
+        /// </summary>
+        internal Cci.IFieldReference GetOrCreateFieldForStringValue(
+            string text,
+            ImmutableArray<byte> data,
```
AlekseyTs, line 325: >ImmutableArray<byte> data, [](http://example.com/codeflow?start=12&length=26)

We still keep passing redundant data. I understand that we want to avoid creation of private implementation detail type when a string cannot be encoded. This can be achieved by making this a static method and passing CommonPEModuleBuilder in. The method can attempt the encoding, request the instance of the type from the builder and complete the work.

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
```diff
@@ -1044,6 +1045,25 @@ public Cci.IFieldReference GetArrayCachingFieldForConstants(ImmutableArray<Const
...
+        /// <see cref="PrivateImplementationDetails.GetOrCreateFieldForStringValue"/>
+        /// </summary>
+        public Cci.IFieldReference TryGetOrCreateFieldForStringValue(string text, TSyntaxNode syntaxNode, DiagnosticBag diagnostics)
+        {
+            if (text.TryGetUtf8ByteRepresentation(out byte[] data, out _))
+            {
+                var privateImpl = GetPrivateImplClass(syntaxNode, diagnostics);
+                return privateImpl.GetOrCreateFieldForStringValue(
+                    text,
+                    data.ToImmutableArray(),
```
AlekseyTs, line 1058: >data.ToImmutableArray() [](http://example.com/codeflow?start=20&length=23)

Do we have to allocate an immutable array here?

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
AlekseyTs: Please make sure we have a test for a scenario like that.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -911,4 +1095,108 @@ public sealed override int GetHashCode()
...
+            {
+                return _encodingGetString.GetParameters(context)[0].GetType(context);
+            }
+        }
+    }
+
+    file sealed class ParameterDefinition(
+        ushort index,
+        string name,
+        Cci.PlatformType type)
```
AlekseyTs, line 1191: >Cci.PlatformType type [](http://example.com/codeflow?start=8&length=21)

It looks like we always pass ```Cci.PlatformType.SystemInt32```, similar with name, it looks like it is always the same. Let's get rid of the parameters if that is the case

AlekseyTs: ```index``` as well?

src/Compilers/Core/Portable/Compilation/Compilation.cs
```diff
@@ -3519,6 +3519,34 @@ internal static bool SerializePeToStream(
         /// </summary>
         internal bool CatchAnalyzerExceptions => Feature("debug-analyzers") == null;
 
+        internal int? DataSectionStringLiteralThreshold
```
AlekseyTs, line 3522: >DataSectionStringLiteralThreshold [](http://example.com/codeflow?start=22&length=33)

Would it make sense to cache the result?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -590,26 +683,100 @@ public override string ToString()
...
+        private readonly ImmutableArray<Cci.IFieldDefinition> _fields;
+        private readonly ImmutableArray<Cci.IMethodDefinition> _methods;
+
+        public DataSectionStringType(
+            ITokenDeferral module,
+            string name,
+            PrivateImplementationDetails containingType,
+            MappedField dataField,
+            Cci.IMethodDefinition bytesToStringHelper,
+            Cci.ITypeReference systemStringType,
```
AlekseyTs, line 715: >Cci.ITypeReference systemStringType [](http://example.com/codeflow?start=12&length=35)

I think we should be able to remove the need to thread string type through many layers of method calls. For example, string is a platform type, it can be obtained from a module builder, a module builder is available to ```PrivateImplementationDetails```, which is passed here.

AlekseyTs: Also, it looks like we are passing module builder here as ```ITokenDeferral```. This is another redundancy which I would prefer to get rid of

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
jjonescz: There is a pre-existing test that has this scenario - `PatternTests.MissingExceptions_01`. You can see I reverted changes to it in commit [https://github.com/dotnet/roslyn/pull/76036/commits/fc88a212d00097028a5457a0718d189bfb1524c4](https://github.com/dotnet/roslyn/pull/76036/commits/fc88a212d00097028a5457a0718d189bfb1524c4).

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
jjonescz: Passing CommonPEModuleBuilder won't work because GetPrivateImplClass and GetSpecialType are defined only on PEModuleBuilder (and passing that in directly would require lot of generic type parameters). But I will try solving this by passing in a lambda.

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
jjonescz: We need an immutable array because we use `data` as a key of an entry in `_mappedFields` which has immutable array type.

I think we can use `ImmutableCollectionsMarshal.AsImmutableArray` instead of `ToImmutableArray` here, so I will do that.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -292,8 +311,81 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+            var @this = factory(arg);
+            return @this._dataSectionStringLiteralTypes.GetOrAdd(text, static (key, arg) =>
+            {
+                var (@this, data, diagnostics) = arg;
+
+                string name = "<S>" + DataToHexViaXxHash128(data);
+
+                MappedField dataField = @this.CreateDataField(data, alignment: 1);
+
+                Cci.IMethodDefinition bytesToStringHelper = @this.GetOrSynthesizeBytesToStringHelper(@this.ModuleBuilder.CommonCompilation, diagnostics);
```
AlekseyTs, line 344: >@this.ModuleBuilder.CommonCompilatio [](http://example.com/codeflow?start=101&length=36)

It looks like we are calling an instance method and explicitly passing an information that is directly available from the instance.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -590,46 +687,146 @@ public override string ToString()
...
+            : SynthesizedStaticFieldBase(name, containingType)
+        {
+            private readonly PrivateImplementationDetails _privateImplDetails = privateImplDetails;
+
+            public override ImmutableArray<byte> MappedData => default;
+            public override bool IsReadOnly => true;
+
+            public override Cci.ITypeReference GetType(EmitContext context)
+            {
+                return _privateImplDetails.ModuleBuilder.GetPlatformType(Cci.PlatformType.SystemString, context);
```
AlekseyTs, line 786: >_privateImplDetails.ModuleBuilder [](http://example.com/codeflow?start=23&length=33)

It looks like the module builder is available from ```EmitContext```. Therefore, ```_privateImplDetails``` field looks redundant.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -590,46 +687,146 @@ public override string ToString()
...
             RoslynDebug.Assert(containingType != null);
-            RoslynDebug.Assert(type != null);
 
             _containingType = containingType;
-            _type = type;
             _name = name;
         }
 
-        public override string ToString() => $"{(object?)_type.GetInternalSymbol() ?? _type} {(object?)_containingType.GetInternalSymbol() ?? _containingType}.{this.Name}";
+        public abstract override string ToString();
```
AlekseyTs, line 829: >public abstract override string ToString(); [](http://example.com/codeflow?start=8&length=43)

It feels unnecessary to require a derived type to override `ToString`. It is, obviously, a choice to override the method, but `ToString` behavior is not important for the purpose of the type.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -911,4 +1123,104 @@ public sealed override int GetHashCode()
...
+            {
+                return _encodingGetString.GetParameters(context)[0].GetType(context);
+            }
+        }
+
+        private sealed class LengthParameterDefinition : Cci.ParameterDefinitionBase
+        {
+            private LengthParameterDefinition() { }
+
+            public static LengthParameterDefinition Instance { get; } = new LengthParameterDefinition();
```
AlekseyTs, line 1219: > public static LengthParameterDefinition Instance { get; } = new LengthParameterDefinition(); [](http://example.com/codeflow?start=11&length=93)

In cases like this (when we are not dealing with a public API), we, usually, prefer readonly fields instead of get-only properties.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -911,4 +1123,104 @@ public sealed override int GetHashCode()
...
+
+            public override ushort Index => 0;
+            public override string Name => "length";
```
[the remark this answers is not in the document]
AlekseyTs: Scratch that, probably not worth it.

src/Compilers/Core/Portable/Compilation/Compilation.cs
```diff
@@ -70,6 +70,8 @@ public abstract partial class Compilation
         // Protected for access in CSharpCompilation.WithAdditionalFeatures
         protected readonly IReadOnlyDictionary<string, string> _features;
 
+        private Optional<int?> _lazyDataSectionStringLiteralThreshold;
```
AlekseyTs, line 73: >Optional<int?> [](http://example.com/codeflow?start=16&length=14)

I am not sure what advantage we are getting from using ```Optional``` structure here. `int?` already has an ability to represent an unknown value.

src/Compilers/Core/Portable/Compilation/Compilation.cs
```diff
@@ -3519,6 +3521,44 @@ internal static bool SerializePeToStream(
         /// </summary>
         internal bool CatchAnalyzerExceptions => Feature("debug-analyzers") == null;
 
+        internal int? DataSectionStringLiteralThreshold
+        {
+            get
+            {
+                if (!_lazyDataSectionStringLiteralThreshold.HasValue)
+                {
+                    _lazyDataSectionStringLiteralThreshold = compute();
```
AlekseyTs, line 3530: >_lazyDataSectionStringLiteralThreshold = compute(); [](http://example.com/codeflow?start=20&length=51)

```Optional<T>``` is a struct, therefore, this assignment is not guaranteed to be atomic. Therefore, the code is not thread safe.

File: src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs (hunk 38 above)
Comment (jjonescz) [reply src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs:829]: It is displayed in IL snapshots in tests, so it is kind of important, but I can see that's perhaps not for the purpose of the type, will remove.

File: src/Compilers/Core/Portable/Compilation/Compilation.cs (hunk 41 above)
Comment (jjonescz) [reply src/Compilers/Core/Portable/Compilation/Compilation.cs:73]: These are different states:
- `Optional.None` = nothing is cached
- `(int?)null` = the value is cached - and it is `null`

Perhaps we could use negative `int` values for some of these but that seems less clear to me.

File: src/Compilers/Core/Portable/Compilation/Compilation.cs (hunk 42 above)
Comment (jjonescz) [reply src/Compilers/Core/Portable/Compilation/Compilation.cs]: I wasn't sure how to make this thread safe, so I changed it to Lazy, but let me know if you think there is a better way.

File: src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs [hunk 43]
```diff
@@ -1072,7 +1083,7 @@ public ArrayMethods ArrayMethods
 
 #nullable enable
 
-        internal PrivateImplementationDetails GetPrivateImplClass(TSyntaxNode syntaxNodeOpt, DiagnosticBag diagnostics)
+        internal PrivateImplementationDetails GetPrivateImplClass(TSyntaxNode? syntaxNode, DiagnosticBag diagnostics)
```
AlekseyTs, line 1086: >syntaxNode [](http://example.com/codeflow?start=79&length=10)

It doesn't look like the rename is necessary. In fact, I think the original name is more descriptive

AlekseyTs, line 1086: >TSyntaxNode? [](http://example.com/codeflow?start=66&length=12)

Given the override below, it feels like having an overload with this parameter type adds no type safety benefit.

AlekseyTs: That isn't a big problem I guess.

jjonescz: By adding the nullable annotation, `syntaxNodeOpt` triggers `RS0046: Avoid the 'Opt' suffix in nullable-enable code` - it's not a build warning, just a hint, but looks like our own rule we want to follow?

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
AlekseyTs: That looks like an error scenario. I am thinking about a success scenario.

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
jjonescz: FWIW, this rule is part of the Roslyn.Diagnostics.Analyzers (it has the RS prefix in the id) which are analyzers specific to dotnet/roslyn repository per [the docs](https://github.com/dotnet/roslyn-analyzers?tab=readme-ov-file#roslyndiagnosticsanalyzers). That's what made me think it's our own rule. But either way works for me, will revert the name change.

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
```diff
@@ -12,6 +12,7 @@
 using System.Security.Cryptography;
 using System.Threading;
 using Microsoft.CodeAnalysis.CodeGen;
+using Microsoft.CodeAnalysis.Collections;
```
AlekseyTs, line 15: Is this using necessary?

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -2936,5 +2936,329 @@ public class Child : Parent, IParent
...
+                // (2,22): error CS0656: Missing compiler required member 'System.Text.Encoding.GetString'
+                // System.Console.Write("bb");
+                Diagnostic(ErrorCode.ERR_MissingPredefinedMember, @"""bb""").WithArguments("System.Text.Encoding", "GetString").WithLocation(2, 22),
+                // (3,22): error CS0656: Missing compiler required member 'System.Text.Encoding.GetString'
+                // System.Console.Write("ccc");
+                Diagnostic(ErrorCode.ERR_MissingPredefinedMember, @"""ccc""").WithArguments("System.Text.Encoding", "GetString").WithLocation(3, 22));
+
+            comp = CreateCompilation(source, parseOptions: TestOptions.Regular.WithFeature("experimental-data-section-string-literals", "0"));
+            comp.MakeMemberMissing(WellKnownMember.System_Text_Encoding__get_UTF8);
+            comp.MakeMemberMissing(WellKnownMember.System_Text_Encoding__GetString);
```
AlekseyTs, line 2977: Consider testing scenarios when only one helper is missing.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -2936,5 +2936,329 @@ public class Child : Parent, IParent
...
+                System.Console.Write("bb");
+                System.Console.Write("ccc");
+                """;
+
+            var expectedOutput = "abbccc";
+
+            foreach (var feature in new[] { "3", "1000", "off" })
+            {
+                CompileAndVerify(source,
+                    parseOptions: TestOptions.Regular.WithFeature("experimental-data-section-string-literals", feature),
```
AlekseyTs, line 3035: >parseOptions: TestOptions.Regular.WithFeature("experimental-data-section-string-literals", feature), [](http://example.com/codeflow?start=20&length=100)

Consider covering scenario when the feature is not set. For completeness.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
jjonescz: That is tested above, on line 2950 and 2963.

AlekseyTs: > That is tested above, on line 2950 and 2963.

Indeed. My mistake

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
AlekseyTs: I guess we need to cover at least these scenarios:
- an empty value
- negative number
- not something that can be parsed as a number
- null
- No feature entry (no ```WithFeature``` call)

AlekseyTs: Let's throw in a number that doesn't fit into an int

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -2936,5 +2936,329 @@ public class Child : Parent, IParent
...
+            var source = """
+                System.Console.WriteLine("Hello");
+                """;
+            CompileAndVerify(source,
+                emitOptions: EmitOptions.Default.WithEmitMetadataOnly(true),
+                parseOptions: TestOptions.Regular.WithFeature("experimental-data-section-string-literals", feature),
+                verify: Verification.FailsPEVerify with
+                {
+                    PEVerifyMessage = """
+                        Bad token as entry point in CLR header.
```
AlekseyTs, line 3249: >Bad token as entry point in CLR header. [](http://example.com/codeflow?start=24&length=39)

What does this complain about?

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -2936,5 +2936,329 @@ public class Child : Parent, IParent
...
+                {
+                    AssertEx.AssertEqualToleratingWhitespaceDifferences("""
+                        <Module>
+                        EmbeddedAttribute
+                        RefSafetyRulesAttribute
+                        Program
+                        """, module.TypeNames.Join("\n"));
+                })
+                .VerifyDiagnostics();
+        }
```
AlekseyTs, line 3262: >} [](http://example.com/codeflow?start=8&length=1)

I think we need to test the following aspects:
- sharing of an "RVA" field type for strings of the same length but different in value
- sharing of an "RVA" field type with an array initializer of the same length but different value
- sharing of an "RVA" field by different literals of equal value
- sharing of an "RVA" field with an array initializer
- sharing of an "<S>..." type by different literals of equal value

AlekseyTs:
Done with review pass (commit 18)

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
jjonescz: I'm not sure how to get more details about the error, but it's caused by the metadata-only emit (`EmitOptions.Default.WithEmitMetadataOnly(true)`) so I suspect some non-metadata parts of the DLL are missing and PEVerify doesn't like that.

src/Compilers/Core/Portable/Emit/CommonPEModuleBuilder.cs
jjonescz: Adding `EmitMetadataTests.PrivateImplDetailsWithoutString`.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
AlekseyTs: Is this error reported in scenarios that do not utilize the feature? For example, are there existing tests observing the same failure? Would it be reported when explicit `Main` is used and we emit EXE or DLL? It is quite possible that the feature has nothing to do with the error, but the error might still be something to be concerned about.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -2936,5 +2936,546 @@ public class Child : Parent, IParent
...
+                  IL_000f:  call       "void System.Console.Write(string)"
+                  IL_0014:  ldstr      "ccc"
+                  IL_0019:  call       "void System.Console.Write(string)"
+                  IL_001e:  ret
+                }
+                """;
+
+            CompileAndVerify(source, expectedOutput: expectedOutput)
+                .VerifyDiagnostics()
+                .VerifyIL("<top-level-statements-entry-point>", expectedIl);
```
AlekseyTs, line 3048: >VerifyIL [](http://example.com/codeflow?start=17&length=8)

Unfortunately, here and in the loop below, verifying IL doesn't feel sufficient enough. For example, if, for whatever reason, the ```DataSectionStringLiteralThreshold``` would be set to >=3, we wouldn't be able to observe this fact. What our ultimate goal here is to confirm whether the feature gets "activated" and, if so, with what threshold. I suggest explicitly asserting expected value of ```DataSectionStringLiteralThreshold``` for each of these scenarios.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
jjonescz: > Is this error reported in scenarios that do not utilize the feature?

Yes, this test verifies the scenario with both the feature on and off.

> Would it be reported when an explicit `Main` is used and we emit EXE or DLL?

That seems to be the reason, I will update the test. Also filed https://github.com/dotnet/roslyn/issues/76707.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -2936,5 +2936,546 @@ public class Child : Parent, IParent
...
+                    var privateImplDetails = module.GlobalNamespace.GetTypeMember("<PrivateImplementationDetails>");
+
+                    // Data field types
+                    AssertEx.AssertEqualToleratingWhitespaceDifferences("""
+                        <PrivateImplementationDetails>.__StaticArrayInitTypeSize=3
+                        <PrivateImplementationDetails>.__StaticArrayInitTypeSize=3
+                        System.Byte
+                        System.Byte
+                        """,
+                        privateImplDetails.GetMembers().OfType<FieldSymbol>().Select(f => f.Type.ToTestDisplayString()).Order().Join("\n"));
```
AlekseyTs, line 3296: >f.Type.ToTestDisplayString() [](http://example.com/codeflow?start=90&length=28)

Consider including field names to make the baseline more informative

AlekseyTs: It looks like this suggestion is applicable to other similar checks

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -2936,5 +2936,546 @@ public class Child : Parent, IParent
...
+                        Int32
+                        Int64
+                        InvalidOperationException
+                        <PrivateImplementationDetails>
+                        """, module.TypeNames.Join("\n"));
+                })
+                .VerifyDiagnostics(
+                    // warning CS8021: No value for RuntimeMetadataVersion found. No assembly containing System.Object was found nor was a value for RuntimeMetadataVersion specified through options.
+                    Diagnostic(ErrorCode.WRN_NoRuntimeMetadataVersion).WithLocation(1, 1),
+                    // (3,24): warning CS8509: The switch expression does not handle all possible values of its input type (it is not exhaustive). For example, the pattern '0' is not covered.
```
AlekseyTs, line 3476: >// (3,24): warning CS8509: The switch expression does not handle all possible values of its input type (it is not exhaustive). For example, the pattern '0' is not covered. [](http://example.com/codeflow?start=20&length=171)

Consider suppressing this warning with a pragma

AlekseyTs:
Done with review pass (commit 20)

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -2936,5 +2936,576 @@ public class Child : Parent, IParent
...
+            Assert.Null(verifier.Compilation.DataSectionStringLiteralThreshold);
+
+            foreach (var feature in new[] { "off", null })
+            {
+                verifier = CompileAndVerify(source,
+                    parseOptions: TestOptions.Regular.WithFeature("experimental-data-section-string-literals", null),
+                    expectedOutput: expectedOutput)
+                    .VerifyDiagnostics()
+                    .VerifyIL("<top-level-statements-entry-point>", expectedIl);
+                Assert.Null(verifier.Compilation.DataSectionStringLiteralThreshold);
```
AlekseyTs, line 3058: >Assert.Null(verifier.Compilation.DataSectionStringLiteralThreshold); [](http://example.com/codeflow?start=16&length=68)

It feels somewhat surprising that `null` is not treated as an unrecognized input, defaulting to threshold == 100

AlekseyTs:
Done with review pass (commit 21)

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
jjonescz: `null` is commonly treated as "not specified" by feature flags - see:

https://github.com/dotnet/roslyn/blob/f68a47ab0e9137b21db4fc8409e542a314cc0272/src/Compilers/Core/Portable/Compilation/Compilation.cs#L3506-L3510

I think it would be confusing if this feature behaved differently.

AlekseyTs: Consider explicitly stating this behavior in the spec

AlekseyTs (approved):
LGTM (commit 21)

jaredpar:
@cston PTAL

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -275,15 +270,39 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+        /// Gets a field that can be used to to store data directly in an RVA field.
+        /// </summary>
+        /// <param name="data">The data for the field.</param>
+        /// <param name="alignment">
+        /// The alignment value is the necessary alignment for addresses for the underlying element type of the array.
+        /// The data is stored by using a type whose size is equal to the total size of the blob. If a built-in system
+        /// type has an appropriate size and .pack, it can be used. Otherwise, a type is generated of the same size as
+        /// the data, and that type needs its .pack set to the alignment required for the underlying data. While that
+        /// .pack value isn't required by anything else in the compiler (the compiler always aligns RVA fields at 8-byte
+        /// boundaries, which accomodates any element type that's relevant), it is necessary for IL rewriters. Such rewriters
```
cston, line 285: >accomodates [](http://example.com/codeflow?start=30&length=11)

"accommodates"?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -275,15 +270,39 @@ internal Cci.IFieldReference CreateDataField(ImmutableArray<byte> data, ushort a
...
+        /// also need to ensure an appropriate alignment is maintained for the RVA field, and while they could also simplify
+        /// by choosing a worst-case alignment as does the compiler, they may instead use the .pack value as the alignment
+        /// to use for that field, since it's an opaque blob with no other indication as to what kind of data is
+        /// stored and what alignment might be required.
+        /// </param>
+        /// <returns>The field. This may have been newly created or may be an existing field previously created for the same data and alignment.</returns>
+        internal MappedField CreateDataField(ImmutableArray<byte> data, ushort alignment)
```
cston, line 292: >CreateDataField [](http://example.com/codeflow?start=29&length=15)

Consider calling this `GetOrAddDataField()`.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -590,47 +688,143 @@ public override string ToString()
...
+        {
+            public override ImmutableArray<byte> MappedData => default;
+            public override bool IsReadOnly => true;
+
+            public override Cci.ITypeReference GetType(EmitContext context)
+            {
+                return context.Module.GetPlatformType(Cci.PlatformType.SystemString, context);
+            }
+
+            public override string ToString()
```
cston, line 788: >ToString [](http://example.com/codeflow?start=35&length=8)

Where is `ToString()` used? Consider removing if not needed.

src/Compilers/Core/Portable/InternalUtilities/StringExtensions.cs
```diff
@@ -275,5 +277,25 @@ public static int GetCaseSensitivePrefixLength(this string string1, string strin
...
         }
+
+        internal static bool TryGetUtf8ByteRepresentation(
+            this string s,
+            [NotNullWhen(returnValue: true)] out byte[]? result,
+            [NotNullWhen(returnValue: false)] out string? error)
+        {
+            try
+            {
+                s_lazyUtf8 ??= new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
```
cston, line 288: >s_lazyUtf8 ??= new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true); [](http://example.com/codeflow?start=16&length=99)

Consider moving this assignment before the `try { }` block.

src/Compilers/Core/Portable/PEWriter/StaticConstructor.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    internal sealed class StaticConstructor(
+        ITypeDefinition containingTypeDefinition, ushort maxStack, ImmutableArray<byte> il)
+        : MethodDefinitionBase(containingTypeDefinition, maxStack, il)
+    {
+        public override string Name => WellKnownMemberNames.StaticConstructorName;
+        public override TypeMemberVisibility Visibility => TypeMemberVisibility.Private;
+        public override bool IsRuntimeSpecial => true;
+        public override bool IsSpecialName => true;
+    }
```
cston, line 18: >} [](http://example.com/codeflow?start=4&length=1)

Do we need `override bool IsConstructor => true:`?

src/Compilers/Core/Portable/PEWriter/MethodDefinitionBase.cs
```diff
@@ -0,0 +1,182 @@
...
+
+    public bool IsExternal => false;
+
+    public bool IsHiddenBySignature => true;
+
+    public bool IsNewSlot => false;
+
+    public bool IsPlatformInvoke => false;
+
+    public virtual bool IsRuntimeSpecial => false;
```
cston, line 54: >virtual [](http://example.com/codeflow?start=11&length=7)

It looks like there are only two distinct implementations of `MethodDefinitionBase`. Given that, consider using `abstract` members in this base type and implementing in both derived types.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
jjonescz: It's part of the decompiled IL in tests apparently.

src/Compilers/Core/Portable/PEWriter/MethodDefinitionBase.cs
jjonescz: I would prefer to not do that. I imagine in the future there could be more derived types and they probably won't need to override IsRuntimeSpecial and IsSpecialName. The purpose of this base class is to share code.

src/Compilers/Core/Portable/PEWriter/StaticConstructor.cs
jjonescz: Probably yes, thanks. Although the `RootModuleStaticConstructor` that this was extracted from had `IsConstructor => false`.

PR #76045, opened 2024-11-23
Update Copilot rename telemetry

AmadeusW:
Update Copilot rename telemetry with the following information:
- CorrelationID for correlation with Editor and Copilot telemetry. Defaults to empty GUID if not available (it will be available in 17.13 preview 2)
- Time it took to get context
- Whether context was used and whether there were any errors
image


src/EditorFeatures/Core.Wpf/Lightup/ISmartRenameSessionWrapper.cs
```diff
@@ -93,6 +95,7 @@ private ISmartRenameSessionWrapper(object instance)
     public string StatusMessage => s_statusMessageAccessor(_instance);
     public bool StatusMessageVisibility => s_statusMessageVisibilityAccessor(_instance);
     public IReadOnlyList<string> SuggestedNames => s_suggestedNamesAccessor(_instance);
+    public Guid CorrelationId => s_correlationIdAccessor(_instance);
```
Cosifne, line 98: Looks like we need to insert editor change first

Cosifne:
Change looks good, but we might want to wait 17.13 P2 to merge this in

AmadeusW:
Thanks for the review. That's in line with my expectations, let's merge in December.

Cosifne (approved):

PR #76050, opened 2024-11-23
Update Razor DynamicFile Provider

ryzngard:
Roslyn side of https://github.com/dotnet/vscode-csharp/pull/7826


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
```diff
@@ -3,60 +3,61 @@
...
 using Roslyn.Utilities;
 
 namespace Microsoft.CodeAnalysis.LanguageServer.HostWorkspace;
 
-[Export(typeof(IDynamicFileInfoProvider)), Shared]
+[Shared]
+[Export(typeof(IDynamicFileInfoProvider))]
 [ExportMetadata("Extensions", new string[] { "cshtml", "razor", })]
-internal class RazorDynamicFileInfoProvider : IDynamicFileInfoProvider
+[ExportCSharpVisualBasicStatelessLspService(typeof(RazorDynamicFileInfoProvider))]
```
davidwengier, line 20: David can confirm, but I think having this be exported as a LSP service and a regular one will cause lifetime issues. Or we'll end up with two instances, which might be problematic if you were hoping to have shared state? Either way I think separating into two classes would be clearer.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
```diff
@@ -67,26 +68,44 @@ public RazorDynamicFileInfoProvider(Lazy<RazorWorkspaceListenerInitializer> razo
...
+
+        // Since we only sent one file over, we should get either zero or one URI back
+        var responseUri = response.CSharpDocument.Uri;
+        var dynamicFileInfoFilePath = ProtocolConversions.GetDocumentFilePathFromUri(responseUri);
+
+        if (response.Edits is not null)
         {
-            var dynamicFileInfoFilePath = ProtocolConversions.GetDocumentFilePathFromUri(responseUri);
-            return new DynamicFileInfo(dynamicFileInfoFilePath, SourceCodeKind.Regular, EmptyStringTextLoader.Instance, designTimeOnly: true, documentServiceProvider: null);
+            var workspaceManager = LanguageServerHost.Instance.GetRequiredLspService<LspWorkspaceManager>();
```
davidwengier, line 85: If this is causing a hang, perhaps the easiest thing to try is to put this logic in a custom text loader, so the retrieval is deferred?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
```diff
@@ -67,26 +68,44 @@ public RazorDynamicFileInfoProvider(Lazy<RazorWorkspaceListenerInitializer> razo
...
+            var (_, _1, document) = await workspaceManager.GetLspDocumentInfoAsync(response.CSharpDocument, cancellationToken);
```
dibarbet, line 86: this is not allowed, GetLspDocumentInfo can only be used inside the queue

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
```diff
@@ -3,60 +3,61 @@
...
+            TimeSpan.FromMilliseconds(200),
+            UpdateAsync,
+            listenerProvider.GetListener(nameof(RazorDynamicFileInfoProvider)),
+            CancellationToken.None);
+        _workspaceFactory = workspaceFactory;
     }
 
-    private const string RemoveRazorDynamicFileInfoMethodName = "razor/removeDynamicFileInfo";
+    public bool MutatesSolutionState => false;
+    public bool RequiresLSPSolution => false;
```
davidwengier, line 47: Can be deleted now

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
```diff
@@ -101,15 +122,42 @@ public Task RemoveDynamicFileInfoAsync(ProjectId projectId, string? projectFileP
...
         {
             return Task.FromResult(TextAndVersion.Create(SourceText.From(""), VersionStamp.Default));
         }
     }
+
+    private sealed class TextChangesTextLoader(Document? document, IEnumerable<TextChange> changes) : TextLoader
+    {
+        public override async Task<TextAndVersion> LoadTextAndVersionAsync(LoadTextOptions options, CancellationToken cancellationToken)
+        {
+            var sourceText = document is null
```
davidwengier, line 150: If document is null, but changes isn't, then aren't we dropping edits on the floor? 

For a new document, it makes sense that the document would be null until after the `GetDynamicFileInfo` method has finished executing, but doesn't that mean that when this method is called, its not null any more? Perhaps the look up of the document should be in this method, not when the request is made?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
ryzngard: Woops yup

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
ryzngard: I guess if document is null we should be able to assume changes is just the whole document. We could add a check for that? What edits would be dropped without that? 

This is definitely only intended for a new document so anything that can better reflect that is good.

davidwengier: Oh whoops, yes, I missed that if document was null its just as though we're starting with a blank document. That makes complete sense, I retract my comment.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
```diff
@@ -67,26 +68,46 @@ public RazorDynamicFileInfoProvider(Lazy<RazorWorkspaceListenerInitializer> razo
...
+        var responseUri = response.CSharpDocument.Uri;
+        var dynamicFileInfoFilePath = ProtocolConversions.GetDocumentFilePathFromUri(responseUri);
+
+        if (response.Edits is not null)
         {
-            var dynamicFileInfoFilePath = ProtocolConversions.GetDocumentFilePathFromUri(responseUri);
-            return new DynamicFileInfo(dynamicFileInfoFilePath, SourceCodeKind.Regular, EmptyStringTextLoader.Instance, designTimeOnly: true, documentServiceProvider: null);
+            var project = _workspaceFactory.Workspace.CurrentSolution.GetRequiredProject(projectId);
+            var document = project.Documents.FirstOrDefault(
+                d => d.FilePath is not null && PathUtilities.PathsEqual(d.FilePath, dynamicFileInfoFilePath));
```
dibarbet, line 87: you should just be able to use `https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/Protocol/Extensions/Extensions.cs#L103` on the solution

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
```diff
@@ -67,26 +68,46 @@ public RazorDynamicFileInfoProvider(Lazy<RazorWorkspaceListenerInitializer> razo
...
+
+            var textChanges = response.Edits.Select(e => new TextChange(e.Span.ToTextSpan(), e.NewText));
+            var textLoader = new TextChangesTextLoader(document, textChanges);
```
dibarbet, line 90: how often does this happen?  Wondering if we need to instantiate a new loader each time or there was one TextLoader that we updated with the latest edits

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorProvideDynamicFileParams.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Text.Json.Serialization;
+using Roslyn.LanguageServer.Protocol;
+
+namespace Microsoft.CodeAnalysis.LanguageServer.HostWorkspace;
```
dibarbet, line 8: mind moving these typedefs into a razor subfolder?  could also put the actual provider there too as well

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
```diff
@@ -67,26 +68,46 @@ public RazorDynamicFileInfoProvider(Lazy<RazorWorkspaceListenerInitializer> razo
         Contract.ThrowIfNull(LanguageServerHost.Instance, "We don't have an LSP channel yet to send this request through.");
         var clientLanguageServerManager = LanguageServerHost.Instance.GetRequiredLspService<IClientLanguageServerManager>();
 
-        var response = await clientLanguageServerManager.SendRequestAsync<ProvideDynamicFileParams, ProvideDynamicFileResponse>(
+        var response = await clientLanguageServerManager.SendRequestAsync<RazorProvideDynamicFileParams, RazorProvideDynamicFileResponse>(
```
dibarbet, line 71: does Razor also use source texts on the other side?  Wondering if it would be worht sending over the Roslyn base document source text checksum over so Razor could verify its returning edits using the same source text as a base

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
ryzngard: For closed documents it shouldn't be too often. It would be based on how often we have a diff when we generate. The most immediate cause I can think of would be if project info changed (going from unknown project to known project).

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs
ryzngard: @davidwengier I think we do but want to confirm

davidwengier: In theory, yes, we use SourceText on the Razor LSP server side. I'm not if the checksum algorithm is in sync, though. Roslyn would presumably use whatever is configured for the project, Razor might not know that. Or maybe the Razor compiler needs it, so it is synced? Worth testing 🤷‍♂️

ryzngard: ❓ since this is pure text, is there a reason we can't use something like a sha256 checksum for the text on both sides?

davidwengier: Other than losing some benefit from Roslyn having presumably already calculated the checksum, no, and I've no idea if that is expensive. I guess the follow up question is what do we do if it fails? Maybe we need to experiment here and see if we have a reliability issue, or if version number is sufficient, or if we need to do something else. I don't think I have a handle on how well this is going to work, to understand how much effort we should put into battening down the hatches

ryzngard: My plan was to do the following: 

1. Send version/checksum/whatever in request
2. Check version in client side: 
   * If matches, send diff edits.
   * If doesn't match. Send back full content.
   * Somehow know if it matches and catch in telemetry

dibarbet: If we're not sure that the source text is created in the same way (same encoding, same checksum algorithm) there is definitely potential for false negatives.  

However if the source texts have different encodings it seems like the text diff might not be useful anyway (so maybe not a true false negative).  At the very least we'd know when stuff is going off the rails (whereas currently we'd get mangled documents instead)

ryzngard: It looks like `RazorCodeDocument.Source.Text` is a `SourceText` that should match given we make sure the encoding and checksum algo are the same. What if we send that data from razor so when we create a sourcetext it's the same? If it's not the same encoding, checksum algo, AND checksum, then we send all content and log telemetry so we know we hit a case that is unexpected. 

That might also cover if a document is opened then closed again, but I'm not sure tbh

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (document is null)
+            {
+                var text = SourceText.From("").WithChanges(changes);
+                return TextAndVersion.Create(text, VersionStamp.Default.GetNewerVersion());
+            }
+
+            var sourceText = await document.GetTextAsync(cancellationToken).ConfigureAwait(false);
+
+            // Validate the checksum information so the edits are known to be correct
+            if (IsSourceTextMatching(sourceText))
```
ryzngard, line 33: @davidwengier @dibarbet this is the logic I was thinking of on how to handle checksums. Telemetry would probably be done on the razor client side unless there's a preference for something else?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        private bool IsSourceTextMatching(SourceText sourceText)
+        {
+            if (sourceText.ChecksumAlgorithm != checksumAlgorithm)
+            {
+                return false;
+            }
+
+            if (sourceText.Encoding?.WebName.Equals(encoding) == false)
```
davidwengier, line 50: This feels redundant, in that the checksums wouldn't match if the encodings were different, but I don't know enough about it to be sure.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+        byte[] checksum,
+        SourceHashAlgorithm checksumAlgorithm,
+        string? encoding,
+        Uri razorUri) : TextLoader
+    {
+        public override async Task<TextAndVersion> LoadTextAndVersionAsync(LoadTextOptions options, CancellationToken cancellationToken)
+        {
+            if (document is null)
+            {
+                var text = SourceText.From("").WithChanges(changes);
```
davidwengier, line 26: Does this need to specify the checksum algo and encoding we got from the request?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+                        Uri = razorUri,
+                    },
+                    FullText = true
+                },
+                cancellationToken);
+
+            RoslynDebug.AssertNotNull(response.Edits);
+
+            var textChanges = response.Edits.Select(e => new TextChange(e.Span.ToTextSpan(), e.NewText));
+            var text = SourceText.From("").WithChanges(textChanges);
```
davidwengier, line 83: Same question as above, re: specifying encoding etc.

davidwengier:
Love the idea that this gives us a "get out of jail free card", in being able to get the whole document contents. Will be curious what the telemetry looks like for sure

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
ryzngard: yes it does

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
ryzngard: I think of it as `Encoding = how to read bytes for content, ChecksumAlgorithm = how to read bytes to produce a checksum` 

They're explicitly different because some encodings allow for things like different preambles but if stored as a UTF-16 string will be the same (or displayed to a user I suppose). 

I should have written this as `sourceText.Encoding?.WebName == encoding` though and idk why I was weird about it

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
dibarbet: We should definitely track it, fine with doing on the Razor side.   However if we're doing the checks here (and reporting in Razor) we're going to lose the reason why they were different in telemetry, which might be key to figuring out where things are going wrong.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            // Validate the checksum information so the edits are known to be correct
+            if (IsSourceTextMatching(sourceText))
+            {
+                var version = await document.GetTextVersionAsync(cancellationToken).ConfigureAwait(false);
+                var newText = sourceText.WithChanges(changes);
+                return TextAndVersion.Create(newText, version.GetNewerVersion());
+            }
+
+            return await GetFullDocumentFromServerAsync(razorUri, cancellationToken).ConfigureAwait(false);
```
dibarbet, line 47: I would say though we should evaluate this in a month or so - if we never get non-matching texts this should be changed into an assert, and if we do, we should track down where

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        private bool IsSourceTextMatching(SourceText sourceText)
+        {
+            if (sourceText.ChecksumAlgorithm != checksumAlgorithm)
+            {
+                return false;
+            }
+
+            if (sourceText.Encoding?.CodePage == codePage)
```
dibarbet, line 57: a couple things
1.  might be missing something - shouldn't this be `!=` (if the encodings are not equal we don't have a match)?
2.  i think the Encoding should override equals - https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Text/Encoding.cs#L1052
3.  the source text does include the encoding in the checksum - https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/Text/SourceTextStream.cs#L121

dibarbet: nvm the other type isn't an `Encoding`, ignore 2

dibarbet (approved):

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
ryzngard: hmm... I interpreted 3 slightly different: https://github.com/dotnet/roslyn/blob/ad6795a50273e90af90fa534498f856abed0b85e/src/Compilers/Core/Portable/Text/SourceText.cs#L584 

1. is definitely right. I'll add unit tests now that this seems the way forward.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
ryzngard: 💯 should revisit to make sure this number is not high because that indicates perf is suboptimal

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorDynamicFileInfoProvider.TextChangesTextLoader.cs
ryzngard: 💯 should revisit to make sure this number is not high because that indicates perf is suboptimal

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorProvideDynamicFileResponse.cs
```diff
@@ -0,0 +1,27 @@
...
+internal class RazorProvideDynamicFileResponse
+{
+    [JsonPropertyName("csharpDocument")]
+    public required TextDocumentIdentifier CSharpDocument { get; set; }
+
+    [JsonPropertyName("edits")]
+    public ServerTextChange[]? Edits { get; set; }
+
+    [JsonPropertyName("checksum")]
+    public required byte[] Checksum { get; set; }
```
davidwengier, line 20: Did this need to change to `string`? or did you work that out?

davidwengier (approved):

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/Razor/RazorProvideDynamicFileResponse.cs
ryzngard: oh woops. Have it local and didn't push

PR #76070, opened 2024-11-25
Fix LangVersion computation on non-English locales

jjonescz:
Fixes failing Spanish leg in roslyn-CI after https://github.com/dotnet/roslyn/pull/75795.

Given `$(_TargetFrameworkVersionWithoutV)` = `5.0`, the subtraction `$([MSBuild]::Subtract($(_TargetFrameworkVersionWithoutV), 5))` evaluates to `45` (I guess because the period in `5.0` is not interpreted as decimal separator but as thousands separator instead). Adding `.Split('.')[0]` makes sure only the major version is taken and the subtraction works.

### DIFF
--- a/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
+++ b/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
@@ -20,7 +20,7 @@
         to determine the correct language version.
     -->
     <_MaxSupportedLangVersion Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND
-                                         '$(_MaxSupportedLangVersion)' == ''">$([MSBuild]::Add(9, $([MSBuild]::Subtract($(_TargetFrameworkVersionWithoutV), 5)))).0</_MaxSupportedLangVersion>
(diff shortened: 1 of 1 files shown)


Youssef1313 (commented):
Isn't this related to the bug on MSBuild side?

https://github.com/dotnet/msbuild/issues/9757

src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets
```diff
@@ -20,7 +20,7 @@
         to determine the correct language version.
     -->
     <_MaxSupportedLangVersion Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND
-                                         '$(_MaxSupportedLangVersion)' == ''">$([MSBuild]::Add(9, $([MSBuild]::Subtract($(_TargetFrameworkVersionWithoutV), 5)))).0</_MaxSupportedLangVersion>
+                                         '$(_MaxSupportedLangVersion)' == ''">$([MSBuild]::Add(9, $([MSBuild]::Subtract($(_TargetFrameworkVersionWithoutV.Split('.')[0]), 5)))).0</_MaxSupportedLangVersion>
```
jaredpar, line 23: Can we add a comment that points to the msbuild issue about localization?

jaredpar (approved):

PR #76073, opened 2024-11-25
Remove unnecessary EOL package references

ViktorHofer:
Microsoft.IO.Redist, System.Bufferes, System.Memory and more recently shipped from the dotnet/maintenance-packages repository.

**This change might be considered a breaking change as packages like System.Memory are being updated which are considered highly impactful and might also require coordination with VS.**

Tries to unblock https://github.com/dotnet/sdk/pull/45042


ViktorHofer:
@jaredpar can someone on your team please take a look at the `MissingMethodException` failure. This looks related to the `System.Memory` version bump.

eng/Directory.Packages.props
```diff
@@ -194,18 +194,15 @@
     <PackageVersion Include="System.IO.Hashing" Version="$(SystemIOHashingVersion)" />
     <PackageVersion Include="System.IO.Pipelines" Version="$(SystemIOPipelinesVersion)" />
     <PackageVersion Include="System.IO.Pipes.AccessControl" Version="5.0.0" />
-    <PackageVersion Include="System.Memory" Version="4.5.5" />
-    <PackageVersion Include="System.Numerics.Vectors" Version="4.5.0" />
-    <PackageVersion Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
-    <PackageVersion Include="System.Runtime.Loader" Version="4.3.0" />
+    <PackageVersion Include="System.Memory" Version="4.6.0" />
```
jjonescz, line 197: > can someone on your team please take a look at the `MissingMethodException` failure. This looks related to the `System.Memory` version bump.

I think the MissingMethodException happens because MSBuild loads our custom build task (SemanticSearch.BuildTask.dll) which now depends on System.Memory 4.6.0, but MSBuild itself depends on an older version of System.Memory, so there's a mismatch.

As is linked in the other PR, the solution might be https://github.com/dotnet/msbuild/pull/11038.

ViktorHofer: Yes that's my standing theory but we should still temporarily be able to consume a newer version by adding the assembly app-local. It seems like this isn't happening here.

ViktorHofer:
Now that https://github.com/dotnet/roslyn/pull/76076 is merged, I rebased on top of it. The changes here remove package references and package versions that aren't necessary anymore because the assembly got moved inbox. The xunit version used in roslyn now targets netstandard2.0 as well so workaround can be removed for that as well.

eng/Directory.Packages.props
jaredpar: How would app local help here though? The version in msbuild will be loaded first, it will win and the one local to the task will not be considered.

PR #76082, opened 2024-11-26
Reduce allocations from NamedTypeSymbol.AddOperators usage

ToddGrun:
This method is internal and returned an ImmutableArray. Instead, modify the callers to pass in an ArrayBuilder, as all of them simply enumerate over the returned contents, and don't need to store or create an immutable array from the contents.

This shows up as about 1.5% of total allocations in the codeanalysis process in the scrolling speedometer test.

This feels familiar, so maybe I tried something against this code and didn't end up checking it in earlier, but I'm not really seeing any downside with proceeding here.

Test insertion (commit 1) started here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10617892&view=results

*** relevant prior allocations from the scrolling speedometer ***
image


cston (approved):

jcouv (approved):
LGTM Thanks (iteration 1)

PR #76107, opened 2024-11-26
Flatten SourceMethodSymbol hierarchy

333fred:
Once upon a time, there were source method symbols that didn't have attributes, and thus we had `SourceMethodSymbol` and `SourceMethodSymbolWithAttributes`, to simplify/save on symbol size for such methods. Today, there is no such difference, and the only direct implementor of `SourceMethodSymbol` is `SourceMethodSymbolWithAttributes`. Therefore, I've cleaned up the hierarchy by removing the dead intermediate class. Highly suggest reviewing commit-by-commit, so that the rename-with-no-changes is correctly displayed. I will _not_ be squash merging this PR to preserve that rename in history as well.

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
@@ -24,14 +24,14 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
     /// <summary>
     /// A source method that can have attributes, including a member method, accessor, or local function.
     /// </summary>
-    internal abstract class SourceMethodSymbolWithAttributes : SourceMethodSymbol, IAttributeTargetSymbol
+    internal abstract partial class SourceMethodSymbol : IAttributeTargetSymbol
     {
         private CustomAttributesBag<CSharpAttributeData> _lazyCustomAttributesBag;
         private CustomAttributesBag<CSharpAttributeData> _lazyReturnTypeCustomAttributesBag;
 
         // some symbols may not have a syntax (e.g. lambdas, synthesized event accessors)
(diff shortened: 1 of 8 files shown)


src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -1746,5 +1746,78 @@ internal override System.Reflection.MethodImplAttributes ImplementationAttribute
 
         internal override int TryGetOverloadResolutionPriority()
             => GetEarlyDecodedWellKnownAttributeData()?.OverloadResolutionPriority ?? 0;
+
```
333fred: These methods were copy/pasted from `SourceMethodSymbol.cs` with no changes to their content.

cston (approved):

jcouv (approved):
LGTM Thanks (iteration 3)

333fred:
@cston @jcouv, I've taken advice from Aleksey and restructured the approach to keep everything in the same location; instead, I now just make SourceMethodSymbol a `partial` type and leave the existing files alone. Apologies for the force push, but it was the cleanest way to approach the restructure.

cston (approved):

AlekseyTs (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
jjonescz, line 1: Should we also rename the file?

jjonescz: Git would likely detect it as rename, so there would be no noise

PR #76133, opened 2024-11-28
avoid ITextSnapshotLine.GetText()

olegtk:
Avoid allocating the whole text of the line in ITextSnapshotLine.GetFirstNonWhitespacePosition() extension method.
It allocates 7Mb in typing scenario of ScrollingSpeedometer.ScrollingAndTypingInLargeFile (roughly 1% of allocations)
image

### DIFF
--- a/src/EditorFeatures/Text/Shared/Extensions/ITextSnapshotLineExtensions.cs
+++ b/src/EditorFeatures/Text/Shared/Extensions/ITextSnapshotLineExtensions.cs
@@ -19,19 +19,11 @@ namespace Microsoft.CodeAnalysis.Text.Shared.Extensions
         /// </summary>
         public static int? GetFirstNonWhitespacePosition(this ITextSnapshotLine line)
         {
-            Contract.ThrowIfNull(line);
+            var firstNonWhitespaceOffset = line.GetFirstNonWhitespaceOffset();
 
-            var text = line.GetText();
-
-            for (var i = 0; i < text.Length; i++)
-            {
-                if (!char.IsWhiteSpace(text[i]))
-                {
-                    return line.Start + i;
-                }
-            }
-
-            return null;
+            return firstNonWhitespaceOffset.HasValue
+                ? firstNonWhitespaceOffset + line.Start
+                : null;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/EditorFeatures/Text/Shared/Extensions/ITextSnapshotLineExtensions.cs
```diff
@@ -43,11 +35,10 @@ internal static class ITextSnapshotLineExtensions
         {
             Contract.ThrowIfNull(line);
 
-            var text = line.GetText();
-
-            for (var i = 0; i < text.Length; i++)
+            var snapshot = line.Snapshot;
+            for (var i = 0; i < line.Length; i++)
```
ToddGrun, line 39: >for (var i = 0; i < line.Length; i++) [](http://example.com/codeflow?start=12&length=37)

nit: might be a bit simpler to iterate over the line extents. eg:

```
var snapshot = line.Snapshot;

for (var i = line.Start.Position; i < line.End.Position; i++)
{
    if (!char.IsWhiteSpace(snapshot[i]))
    {
        return i;
    }
}
```

olegtk: good idea, fixed

jasonmalinowski: @olegtk @ToddGrun I think this broke the code -- it's now returning the position in the snapshot, not the offset from the start of the line.

olegtk: yeah, tests are failing, fixing it now

olegtk: sorry @ToddGrun , I reverted the change, I think this way is OK too

ToddGrun: Oh yeah, it should have been switched to return "i - line.Start.Position" as it returns an offset. The current code is probably better than the modified version of what I proposed then, as I didn't notice it was an offset method.

ToddGrun (approved):
:shipit:

olegtk:
@ToddGrun, @jasonmalinowski - can you please merge this? I don't have permissions.

PR #76139, opened 2024-11-28
Add spec for the data section string literals feature

jjonescz:
Implementation: https://github.com/dotnet/roslyn/pull/76036
Link to rendered markdown: https://github.com/jjonescz/roslyn/blob/DataSectionStringLiterals-spec/docs/features/string-literals-data-section.md

### DIFF
--- a/docs/features/string-literals-data-section.md
+++ b/docs/features/string-literals-data-section.md
@@ -19,9 +19,9 @@ The feature is currently implemented only for C#, not VB.
 ## Configuration
 
 The feature flag can take a non-negative integer threshold.
-Only string literals whose length (number of characters, not bytes) is over the threshold are emitted using the utf8 encoding strategy.
+Only string literals whose length (number of characters, not bytes) is greater than the threshold are emitted using the utf8 encoding strategy.
 If the flag is set, but no value is specified, the threshold defaults to 100.
-Specifying 0 means all string literals are considered for the feature.
+Specifying 0 means all non-empty string literals are considered for the feature.
 Specifying `off` as the value turns the feature off (this is the default).
 
 The feature flag can be specified on the command line like `/features:experimental-data-section-string-literals` or `/features:experimental-data-section-string-literals=20`,
@@ -45,13 +45,28 @@ The utf8 string literal encoding emit strategy emits `ldsfld` of a field in a ge
 
 For every string literal, a unique internal static class is generated which:
 - has name composed of `<S>` followed by a hex-encoded XXH128 hash of the string,
-- lives in the global namespace (for consistency with other compiler-generated types),
-- has one `.data` field which is emitted the same way as for [explicit u8 string literals][u8-literals],
-- has one `string` field which is initialized in a static constructor of the class.
+- is nested in the `<PrivateImplementationDetails>` type to avoid polluting the global namespace
+  (although [some other compiler-generated types are synthesized there](https://github.com/dotnet/roslyn/blob/e3954f0cee4085436a590a14a01ec2f1b8a2571f/src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs#L89))
+  and to avoid having to enforce name uniqueness across modules,
+- has one internal static readonly `string` field which is initialized in a static constructor of the class,
+- is marked `beforefieldinit` so the static constructor can be called eagerly if deemed better by the runtime for some reason.
+
+There is also an internal static readonly `.data` field generated into `<PrivateImplementationDetails>` containing the actual bytes,
+similar to [u8 string literals][u8-literals] and [constant array initializers][constant-array-init].
+These other scenarios might also reuse the data field, e.g., the following statements could all reuse the same data field:
+
+```cs
+ReadOnlySpan<byte> a = new byte[6] { 72, 101, 108, 108, 111, 46 };
+ReadOnlySpan<byte> b = stackalloc byte[6] { 72, 101, 108, 108, 111, 46 };
+ReadOnlySpan<byte> c = "Hello."u8;
+string d = "Hello."; // assuming this string literal is eligible for the `ldsfld` emit strategy
+```
 
 The initialization calls `<PrivateImplementationDetails>.BytesToString` helper which in turn calls `Encoding.UTF8.GetBytes`.
 This is an optimization so each of the generated class static constructors is slightly smaller in IL size.
 These size savings can add up since one class is generated per one eligible string literal.
+A compile-time error is reported if the `Encoding.UTF8.GetBytes` API is not available and needs to be used
+(the user can then either ensure the API is available or turn off the feature flag).
 
 The following example demonstrates the code generated for string literal `"Hello."`.
 
@@ -61,24 +76,23 @@ using System.Runtime.InteropServices;
 using System.Text;
 
 [CompilerGenerated]
-internal static class <S>F6D59F18520336E6E57244B792249482
-{
-    internal static readonly <PrivateImplementationDetails>.__StaticArrayInitTypeSize=6 f = /* IL: data(48 65 6C 6C 6F 2E) */;
-
-    internal static readonly string s;
-
-    unsafe static <S>F6D59F18520336E6E57244B792249482()
-    {
-        s = <PrivateImplementationDetails>.BytesToString((byte*)Unsafe.AsPointer(ref f), 6);
-    }
-}
-
-[CompilerGenerated]
 internal static class <PrivateImplementationDetails>
 {
+    internal static readonly __StaticArrayInitTypeSize=6 2BBCE396D68A9BBD2517F6B7064666C772694CD92010887BA379E10E7CDDA960 = /* IL: data(48 65 6C 6C 6F 2E) */;
+
     [StructLayout(LayoutKind.Explicit, Pack = 1, Size = 6)]
     internal struct __StaticArrayInitTypeSize=6
     {
+    }
+    
+    internal static class <S>F6D59F18520336E6E57244B792249482
+    {
+        internal static readonly string s;
+
+        unsafe static <S>F6D59F18520336E6E57244B792249482()
+        {
+            s = <PrivateImplementationDetails>.BytesToString((byte*)Unsafe.AsPointer(ref <PrivateImplementationDetails>.2BBCE396D68A9BBD2517F6B7064666C772694CD92010887BA379E10E7CDDA960), 6);
+        }
     }
 
     internal unsafe static string BytesToString(byte* bytes, int length)
@@ -92,6 +106,7 @@ internal static class <PrivateImplementationDetails>
 
 The actual generated code does not call `(byte*)Unsafe.AsPointer`.
 It simply emits `ldsflda` of the data field.
+The same approach is taken by other scenarios like [u8 string literals][u8-literals] and [constant array initializers][constant-array-init].
 That works but does not pass IL/PE verification.
 
 ## Runtime
@@ -150,10 +165,14 @@ Ahead-of-time compilation tools would need to be updated to recognize this new p
 
 Need to confirm that no additional work is needed for ref assemblies.
 
-### Configuration/emit alternatives
+### Configuration/emit granularity
 
 Instead of one global feature flag, the emit strategy could be controlled using compiler-recognized attributes (applicable to assemblies or classes).
 Furthermore, we could emit more than one string per one class. That could be configurable as well.
+
+### Weak references
+
+To avoid rooting the `string` references forever, we could turn the fields into `WeakReference<string>`s. This could be configurable as well.
 
 ### Automatic threshold
 
@@ -173,6 +192,16 @@ hence synthesizing the utf8 string classes in the former should be possible and 
 
 The compiler could emit an info diagnostic with useful statistics for customers to determine what threshold to set.
 For example, a histogram of string literal lengths encountered while emitting of the program.
+
+### Single blob
+
+We could merge all UTF-8 byte sequences for affected strings into a single blob - a virtual User-String heap.
+The sequences can be deduped as we dedupe entries in User-String heap.
+We can even attempt to do some vary basic compression,
+for example, in the name heap that we generate (contains type and member names, etc.), names with shared content can overlap.
+We would generate a single `__StaticArrayInitTypeSize=*` structure for the entire virtual heap blob and
+add a single `.data` field to `<PrivateImplementationDetails>` that points to the blob.
+At runtime, we would do an offset to where the required data reside in the blob and decode the required length from UTF-8 to UTF-16.
 
 ## Alternatives
 
@@ -199,3 +228,4 @@ However, that would likely result in worse machine code due to more branches and
 
 <!-- links -->
 [u8-literals]: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/utf8-string-literals
+[constant-array-init]: https://github.com/dotnet/roslyn/pull/24621
--- a/docs/features/string-literals-data-section.md
+++ b/docs/features/string-literals-data-section.md
@@ -145,10 +145,6 @@ e.g., startup performance and the overhead of keeping track of these types.
 The generated types are returned from reflection like `Assembly.GetTypes()`
 which might impact the performance of Dependency Injection and similar systems.
 
-> [!NOTE]
-> In practice, there might not be many generated types, it depends on the kind of the program
-> (whether it has lots of short strings or a few large strings) and how [the threshold](#configuration) is configured.
-
 ## Implementation
 
 `CodeGenerator` obtains [configuration of the feature flag](#configuration) from `Compilation` passed to its constructor.
--- a/docs/features/string-literals-data-section.md
+++ b/docs/features/string-literals-data-section.md
@@ -27,6 +27,11 @@ Specifying `off` as the value turns the feature off (this is the default).
 The feature flag can be specified on the command line like `/features:experimental-data-section-string-literals` or `/features:experimental-data-section-string-literals=20`,
 or in a project file in a `<PropertyGroup>` like `<Features>$(Features);experimental-data-section-string-literals</Features>` or `<Features>$(Features);experimental-data-section-string-literals=20</Features>`.
 
+> [!NOTE]
+> This configuration is useful for experimenting (to see the impact of the threshold on runtime performance and IL size)
+> and also allowing users to avoid a potential overhead of many type definitions generated if their program has many short string literals
+> which do not need this new emit strategy because they can fit into the size-limited UserString heap.
+
 ## Eligible string literals
 
 A string literal is emitted with the utf8 encoding if and only if all of these are true:
@@ -44,7 +49,8 @@ That puts the string literal on the UserString heap of the PE file.
 The utf8 string literal encoding emit strategy emits `ldsfld` of a field in a generated class instead.
 
 For every string literal, a unique internal static class is generated which:
-- has name composed of `<S>` followed by a hex-encoded XXH128 hash of the string,
+- has name composed of `<S>` followed by a hex-encoded XXH128 hash of the string
+  (when there is a conflict, a compile-time error is reported),
 - is nested in the `<PrivateImplementationDetails>` type to avoid polluting the global namespace
   and to avoid having to enforce name uniqueness across modules,
 - has one internal static readonly `string` field which is initialized in a static constructor of the class,
@@ -62,7 +68,7 @@ string d = "Hello."; // assuming this string literal is eligible for the `ldsfld
 ```
 
 The initialization calls `<PrivateImplementationDetails>.BytesToString` helper which in turn calls `Encoding.UTF8.GetBytes`.
-This is an optimization so each of the generated class static constructors is slightly smaller in IL size.
+This is an optimization so each of the generated static constructors is slightly smaller in IL size.
 These size savings can add up since one class is generated per one eligible string literal.
 A compile-time error is reported if the `Encoding.UTF8.GetBytes` API is not available and needs to be used
 (the user can then either ensure the API is available or turn off the feature flag).
@@ -107,6 +113,19 @@ The actual generated code does not call `(byte*)Unsafe.AsPointer`.
 It simply emits `ldsflda` of the data field.
 The same approach is taken by other scenarios like [u8 string literals][u8-literals] and [constant array initializers][constant-array-init].
 That works but does not pass IL/PE verification.
+
+### Ref assemblies
+
+The synthesized types are not part of ref assemblies. That makes them smaller
+and incremental compilation is faster because it does not need to recompile dependent projects when only string literal contents change.
+
+This is automatically implemented, because during metadata-only compilation, method bodies are not emitted,
+and the synthesized types used by this feature are only emitted lazily as part of generating method body instructions.
+
+## Diagnostics
+
+The error CS8103 (Combined length of user strings used by the program exceeds allowed limit) is updated to suggest the feature flag,
+albeit with a disclaimer during the experimental phase of the feature.
 
 ## Runtime
 
@@ -160,10 +179,6 @@ That can be implemented in the future.
 
 Ahead-of-time compilation tools would need to be updated to recognize this new pattern of string literal use in place of `ldstr`.
 
-### Ref assemblies
-
-Need to confirm that no additional work is needed for ref assemblies.
-
 ### Automatic threshold
 
 The threshold could be determined automatically with some objective, for example,
@@ -181,7 +196,8 @@ hence synthesizing the utf8 string classes in the former should be possible and 
 ### Statistics
 
 The compiler could emit an info diagnostic with useful statistics for customers to determine what threshold to set.
-For example, a histogram of string literal lengths encountered while emitting of the program.
+For example, a histogram of string literal lengths encountered while emitting the program.
+Tools like ILSpy can already be used to obtain some related data, e.g., count, length, and contents of string literals in the UserString heap.
 
 ### Single blob
 
@@ -200,6 +216,9 @@ At runtime, we would do an offset to where the required data reside in the blob 
 Instead of one global feature flag, the emit strategy could be controlled using compiler-recognized attributes (applicable to assemblies or classes).
 Furthermore, we could emit more than one string per one class. That could be configurable as well.
 
+One interesting strategy would be grouping strings which occur in one user-defined type into one compiler-generated `<S>` type.
+The idea is that strings from one class are likely used "together" so there is no performance impact from initializing them all at once and we save on metadata.
+
 ### GC
 
 To avoid rooting the `string` references forever, we could turn the fields into `WeakReference<string>`s.
@@ -211,6 +230,7 @@ This could be configurable as well.
 We could emit the strings to data fields similarly but we would not synthesize the `string` fields and static constructors.
 Instead, at the place of `ldstr`, we would directly call a runtime-provided helper, passing a pointer to the data field.
 That runtime helper would be an intrinsic recognized by the JIT (with a fallback calling `Encoding.UTF8.GetString`).
+The runtime would then be responsible for loading and caching the strings, perhaps similarly to how `ldstr` works (using a frozen GC heap).
 
 ### Custom lazy initialization
 
--- a/docs/features/string-literals-data-section.md
+++ b/docs/features/string-literals-data-section.md
@@ -57,14 +57,14 @@ using System.Text;
 [CompilerGenerated]
 internal static class <S>2D8BD7D9BB5F85BA643F0110D50CB506A1FE439E769A22503193EA6046BB87F7
 {
-	internal static readonly <PrivateImplementationDetails>.__StaticArrayInitTypeSize=6 f = /* IL: data(48 65 6C 6C 6F 2E) */;
+    internal static readonly <PrivateImplementationDetails>.__StaticArrayInitTypeSize=6 f = /* IL: data(48 65 6C 6C 6F 2E) */;
 
-	internal static readonly string s;
+    internal static readonly string s;
 
-	unsafe static <S>2D8BD7D9BB5F85BA643F0110D50CB506A1FE439E769A22503193EA6046BB87F7()
-	{
-		s = <PrivateImplementationDetails>.BytesToString((byte*)Unsafe.AsPointer(ref f), 6);
-	}
+    unsafe static <S>2D8BD7D9BB5F85BA643F0110D50CB506A1FE439E769A22503193EA6046BB87F7()
+    {
+        s = <PrivateImplementationDetails>.BytesToString((byte*)Unsafe.AsPointer(ref f), 6);
+    }
 }
 
 [CompilerGenerated]
--- a/docs/features/string-literals-data-section.md
+++ b/docs/features/string-literals-data-section.md
@@ -8,11 +8,11 @@ When the limit is reached, the following compiler error is reported by Roslyn:
 error CS8103: Combined length of user strings used by the program exceeds allowed limit. Try to decrease use of string literals.
 ```
 
-By turning on the feature flag `utf8-string-encoding`, string literals (where possible) are instead emitted as UTF-8 data into a different section of the PE file
+By turning on the feature flag `utf8-string-literal-encoding`, string literals (where possible) are instead emitted as UTF-8 data into a different section of the PE file
 which does not have the same limit. The emit format is similar to [explicit u8 string literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/utf8-string-literals).
 
 The feature flag can take a non-negative integer threshold. Only string literals whose length is over the threshold are emitted in the new way described above.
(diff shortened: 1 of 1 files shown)


docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,162 @@
...
+
+By turning on [the feature flag](#configuration), string literals ([where possible](#eligible-string-literals)) are instead emitted as UTF-8 data into a different section of the PE file
+which does not have the same limit. The emit format is similar to [explicit u8 string literals][u8-literals].
+
+The feature is currently implemented only for C#, not VB.
+
+## Configuration
+
+The feature flag can take a non-negative integer threshold. Only string literals whose length is over the threshold are emitted using the utf8 encoding strategy.
+By default, the threshold is 100. Specifying 0 means all string literals are considered for the feature. Specifying `off` turns off the feature (this is the default).
```
jcouv, line 19: There're two mentions of "default" here, which is a bit confusing. Consider something like:
```suggestion
If the flag is set, but no value is specified, the threshold defaults to 100. Specifying 0 means all string literals are considered for the feature. Specifying `off` as the value turns the feature off (this is the default).
```

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,162 @@
...
+
+The following example demonstrates the code generated for string literal `"Hello."`.
+
+```cs
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Text;
+
+[CompilerGenerated]
+internal static class <S>2D8BD7D9BB5F85BA643F0110D50CB506A1FE439E769A22503193EA6046BB87F7
```
jcouv, line 58: Would it be possible to avoid having one type per string?
We can have all the members for a given string (the data field, the `string` field (lazily initialized) and possibly a helper method to handle the lazy initialization) in `<PrivateImplementationDetails>`? Those three members can include the hash in their name to avoid conflicts.

jcouv: Also, by caching strings in a static field, are we creating a problem for the GC? Would using `WeakReference` solve that problem?

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,162 @@
...
+When the limit is reached, the following compiler error is reported by Roslyn:
+
+```
+error CS8103: Combined length of user strings used by the program exceeds allowed limit. Try to decrease use of string literals.
+```
+
+By turning on [the feature flag](#configuration), string literals ([where possible](#eligible-string-literals)) are instead emitted as UTF-8 data into a different section of the PE file
+which does not have the same limit. The emit format is similar to [explicit u8 string literals][u8-literals].
+
+The feature is currently implemented only for C#, not VB.
```
jaredpar, line 14: Want us to add a very specific line here about our support level. Want to clearly indicate that this feature is currently experimental and can be changed or removed at any time.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,162 @@
+# String literals in data section
+
+This opt-in Roslyn feature allows changing how string literals in C# programs are emitted into PE files (`.dll`/`.exe`).
```
jaredpar, line 3:
```suggestion
This opt-in **experimental** Roslyn feature allows changing how string literals in C# programs are emitted into PE files (`.dll`/`.exe`).
```

docs/features/string-literals-data-section.md
jaredpar: Should specify that 100 means 100 characters not 100 bytes (50 chars)

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,162 @@
...
+which does not have the same limit. The emit format is similar to [explicit u8 string literals][u8-literals].
+
+The feature is currently implemented only for C#, not VB.
+
+## Configuration
+
+The feature flag can take a non-negative integer threshold. Only string literals whose length is over the threshold are emitted using the utf8 encoding strategy.
+By default, the threshold is 100. Specifying 0 means all string literals are considered for the feature. Specifying `off` turns off the feature (this is the default).
+
+The feature flag can be specified on the command line like `/features:data-section-string-literals` or `/features:data-section-string-literals=20`,
```
jaredpar, line 21: For now lets put `experimental` in the feature flag name. That reduces the chance that someone can take a dependency on this and not realize that it's experimental in nature.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,162 @@
...
+- its `Length` is greater than the threshold specified by [the feature flag](#configuration).
+
+## Emit
+
+The standard string literal emit strategy emits `ldstr "string"`.
+That puts the string literal on the UserString heap of the PE file.
+
+The utf8 string literal encoding emit strategy emits `ldsfld` of a field in a generated class instead.
+
+For every string literal, a unique internal static class is generated which
```
jaredpar, line 40:
```suggestion
For every string literal, a unique internal static class is generated which:

```

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,162 @@
...
+- has name composed of `<S>` followed by a hex-encoded SHA-256 hash of the string,
```
jaredpar, line 41: Insead of SHA-256 let's use a non-crypto hash like XXH128. Every use of a crypto hash requires future risk as we have to go back and change the implementations when SHA-256 is considered broken.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,162 @@
...
+- lives in the global namespace,
```
jaredpar, line 42: Is there a reason that you chose to put it in the global namespace? Or is it just following other types.

jjonescz: Just following other types.

docs/features/string-literals-data-section.md
jjonescz: > Would it be possible to avoid having one type per string?
> We can have all the members for a given string (the data field, the `string` field (lazily initialized) and possibly a helper method to handle the lazy initialization) in `<PrivateImplementationDetails>`? Those three members can include the hash in their name to avoid conflicts.

That's a possible alternative, I can mention it.

As @davidwrighton said elsewhere, the static constructor approach emits a single mov instruction and no branches and calls.

> Also, by caching strings in a static field, are we creating a problem for the GC?

It's not any different than current `string` literal behavior I think.

docs/features/string-literals-data-section.md
jjonescz: Ok. Note that I chose SHA-256 because that's being used for other similar situations in `<PrivateImplementationDetails>`.

docs/features/string-literals-data-section.md
jcouv: Thanks. I just caught up on a Teams discussion you forwarded me today. It looks like the trade-off was discussed and the runtime team is not too worried about the proliferation of types, they could do some optimization and we're engaged on benchmarking.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,201 @@
...
+
+> [!WARNING]
+> This feature is currently experimental and can be changed or removed at any time.
+
+## Configuration
+
+The feature flag can take a non-negative integer threshold.
+Only string literals whose length (number of characters, not bytes) is over the threshold are emitted using the utf8 encoding strategy.
+If the flag is set, but no value is specified, the threshold defaults to 100.
+Specifying 0 means all string literals are considered for the feature.
```
AlekseyTs, line 24: Should empty string literals be affected by the setting at all?

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,201 @@
...
+## Emit
+
+The standard string literal emit strategy emits `ldstr "string"`.
+That puts the string literal on the UserString heap of the PE file.
+
+The utf8 string literal encoding emit strategy emits `ldsfld` of a field in a generated class instead.
+
+For every string literal, a unique internal static class is generated which:
+- has name composed of `<S>` followed by a hex-encoded XXH128 hash of the string,
+- lives in the global namespace (for consistency with other compiler-generated types),
```
AlekseyTs, line 48: To what other compiler generated types does this refer to?
- Attributes? Those have predefined names we do not choose what name they have and what namespace they belong to.
- Structure types generated for RVA fields are nested into ```'<PrivateImplementationDetails>'```.
- If I remember correctly. closures and state-machine types are nested.

I think it would be preferred to not pollute global namespace with implementation artifacts. For example, when I open an assembly in ildasm, I wouldn't want to be distracted with a bunch of strange types at the top of the hierarchy. Another disadvantage of having the types at the top level is the fact that we would have to enforce uniqueness of their name across modules. It is illegal for multiple modules in the same assembly to declare types with conflicting names.

Therefore, I suggest that we put all types that we synthesize as implementation artifacts into ```'<PrivateImplementationDetails>'```.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,201 @@
...
+- has one `.data` field which is emitted the same way as for [explicit u8 string literals][u8-literals],
```
AlekseyTs, line 49: I think it would be good to explicitly state that the field is going to be static.

AlekseyTs, line 49: Is there a reason to put this field into this type? Wouldn't that prevent us from sharing it with fields used for other blobs that compiler generates? For example, if we have an explicit u8 string literal or an array initializer and a "regular" string literal that results in the same byte sequence, would it make sense to share a blob entry and the RVA field across all of them?

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,201 @@
...
+- has one `string` field which is initialized in a static constructor of the class.
```
AlekseyTs, line 50: Are we going to apply `beforefieldinit` flag to the class?

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,201 @@
...
+
+The initialization calls `<PrivateImplementationDetails>.BytesToString` helper which in turn calls `Encoding.UTF8.GetBytes`.
```
AlekseyTs, line 52: What is the expected behavior when ```Encoding.UTF8.GetBytes``` API is not available?

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,201 @@
...
+    {
+        return Encoding.UTF8.GetString(bytes, length);
+    }
+}
+```
+
+### ILVerify and PEVerify
+
+The actual generated code does not call `(byte*)Unsafe.AsPointer`.
+It simply emits `ldsflda` of the data field.
```
AlekseyTs, line 94: Do we already use this instruction to get a pointer to a blob in other scenarios?

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,201 @@
...
+That works but does not pass IL/PE verification.
+
+## Runtime
+
+Throughput of `ldstr` vs `ldsfld` is very similar (both result in one or two move instructions).
+
+In the `ldsfld` emit strategy, the `string` instances won't ever be collected by the GC once the generated class is initialized.
```
AlekseyTs, line 101: I would speculate that users that hit User-String heap limit might prefer a behavior that allows these strings to be collected. This a trade-off that is often made, use less memory at the price of a small slowdown. For example, in our test assemblies, each unit-test is executed once and the rate of sharing of big literals between unit-tests is probably low. Therefore, it might be beneficial to not hold onto the strings for the entire duration of a test-assembly run, thus avoiding potential out of memory situations.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,201 @@
...
+(it's also how the string literals are emitted into the UserString heap).
+It is called from `SerializeMethodBodies` which precedes `PopulateSystemTables` call,
+hence synthesizing the utf8 string classes in the former should be possible and they would be emitted in the latter.
+
+### Statistics
+
+The compiler could emit an info diagnostic with useful statistics for customers to determine what threshold to set.
+For example, a histogram of string literal lengths encountered while emitting of the program.
+
+## Alternatives
```
AlekseyTs, line 177: It might be worth adding another alternative. We might merge all UTF8 byte sequences for affected strings into a single blob - a virtual User-String heap. The sequences can be deduped as we dedupe entries in User-String heap. We even can attempt to do some vary basic compression. For example, if I remember correctly, in the name heap that we generate (contains type and member names, etc.), names with shared content can overlap. We would generate a single ```'__StaticArrayInitTypeSize=*'``` structure for the entire virtual heap blob and add a single RVA field to ```'<PrivateImplementationDetails>'``` that points to the blob. At runtime, we would do an offset to where the required data reside in the blob and decode the required length from Utf-8 to Unicode.

AlekseyTs:

docs/features/string-literals-data-section.md
jjonescz: They are not, only string literals with length *greater than* (not equal to) the threshold are affected. I will clarify that.

docs/features/string-literals-data-section.md
jjonescz: I don't think there is an observable difference between applying it or not (with current runtime implementations). Applying the flag can allow runtimes to call the static constructor at any time before the static string field is accessed, otherwise it would be called exactly before the string field is accessed for the first time. So I think applying it is better because it can allow runtimes to optimize if necessary.

docs/features/string-literals-data-section.md
jjonescz: I will mention this as future work. There are some trade-offs (the string loading is a more complicated IL than just `ldsfld`), so it's not clear this should be implemented as part of the first iteration.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,231 @@
...
+
+The standard string literal emit strategy emits `ldstr "string"`.
+That puts the string literal on the UserString heap of the PE file.
+
+The utf8 string literal encoding emit strategy emits `ldsfld` of a field in a generated class instead.
+
+For every string literal, a unique internal static class is generated which:
+- has name composed of `<S>` followed by a hex-encoded XXH128 hash of the string,
+- is nested in the `<PrivateImplementationDetails>` type to avoid polluting the global namespace
+  (although [some other compiler-generated types are synthesized there](https://github.com/dotnet/roslyn/blob/e3954f0cee4085436a590a14a01ec2f1b8a2571f/src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs#L89))
```
AlekseyTs, line 49: I am not sure if this is worth mentioning. For example, due to #76299.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,231 @@
...
+- its `Length` is greater than the threshold specified by [the feature flag](#configuration).
+
+## Emit
+
+The standard string literal emit strategy emits `ldstr "string"`.
+That puts the string literal on the UserString heap of the PE file.
+
+The utf8 string literal encoding emit strategy emits `ldsfld` of a field in a generated class instead.
+
+For every string literal, a unique internal static class is generated which:
```
AlekseyTs, line 46: Is there an advantage to use a class vs. a struct?

docs/features/string-literals-data-section.md
AlekseyTs: > I will mention this as future work. 

I do not see this mentioned. If you were referring to "Weak references" section, I do not think this is equivalent because use of weak references is not required to make strings collectable. Also, I think this is more an alternative rather than a "future work".

AlekseyTs:

docs/features/string-literals-data-section.md
jjonescz: I can't think of any difference between the two for this use case.

AlekseyTs (approved):
LGTM (commit 9)

docs/features/string-literals-data-section.md
jaredpar: @davidwrighton who suggested this code pattern of using `class`

docs/features/string-literals-data-section.md
jcouv: > [by caching strings in a static field, are we creating a problem for the GC?] It's not any different than current string literal behavior I think.

Let's confirm with runtime folks. I think we're changing the current string literal behavior.
Today, if I use `M("hello")` we do a `ldstr` with a metadata token to the string. This allocates for the string (unless the string is re-used/interned then we can reference existing allocated string) and later on that can be garbage collected.
With a static field, I don't think it can be collected unless the type is unloaded.

jcouv: I see you already have a follow-up tracked below in "GC" section. Thanks

jcouv (commented):
Done with review pass (iteration 9)

docs/features/string-literals-data-section.md
jaredpar: I think that is mostly an accident of history. Back then it was more acceptable to use a cryto hash and hashes like XXH128 weren't as readily available. Going forward when possible we should avoid crypto hashes unless they're specifically for crypto purposes.

AlekseyTs:
@jjonescz Consider updating PR's title. The "feature flag" is just part of the feature, it is not the feature.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,252 @@
...
+It simply emits `ldsflda` of the data field.
+The same approach is taken by other scenarios like [u8 string literals][u8-literals] and [constant array initializers][constant-array-init].
+That works but does not pass IL/PE verification.
+
+### Ref assemblies
+
+The synthesized types are not part of ref assemblies. That makes them smaller
+and incremental compilation is faster because it does not need to recompile dependent projects when only string literal contents change.
+
+This is automatically implemented, because during metadata-only compilation, method bodies are not emitted,
```
AlekseyTs, line 122: Has this been confirmed? This sounds like stating a fact. However, Jared mentioned that some IL emit artifacts are possibly making their way into ref assemblies.

AlekseyTs:
Done with review pass (commit 10)

docs/features/string-literals-data-section.md
jjonescz: Yes, I have confirmed this. With metadata-only (e.g., using `EmitOptions.Default.WithEmitMetadataOnly(true)`), the `MethodCompiler` is not involved at all, instead a `SynthesizedMetadataCompiler` is used.

AlekseyTs (approved):
LGTM (commit 11)

jcouv (approved):
LGTM Thanks (iteration 11)

teo-tsirpanis:
Would be interesting to extend this to encoding constant strings being passed to `Convert.FromBase64String`, as raw binary data.

The protobuf compiler for example [generates such strings](https://github.com/googleapis/google-cloud-dotnet/blob/cd7c4a21176ae7da80669defa9da0550c7bf5470/apis/Google.Cloud.Storage.V2/Google.Cloud.Storage.V2/Storage.g.cs#L25-L496). And BTW if I replace `string.Concat` with `+`s the compiler gets stuck.

AlekseyTs:
@teo-tsirpanis 

> Would be interesting to extend this to encoding constant strings being passed to `Convert.FromBase64String`, as raw binary data.
> 

Could you please elaborate what would be an advantage for doing this? Also, how does this align with the specific goal that we are trying to achieve here, which is not a goal of supporting different encodings for string literals.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,257 @@
...
+- its `Length` is greater than the threshold specified by [the feature flag](#configuration).
+
+## Emit
+
+The standard string literal emit strategy emits `ldstr "string"`.
+That puts the string literal on the UserString heap of the PE file.
+
+The utf8 string literal encoding emit strategy emits `ldsfld` of a field in a generated class instead.
+
+For every string literal, a unique internal static class is generated which:
```
cston, line 55: Should this be "For every **unique** string literal, an internal static class ..."?

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,257 @@
...
+
+Generating new types by the compiler means more type loads and hence runtime impact,
+e.g., startup performance and the overhead of keeping track of these types.
+
+The generated types are returned from reflection like `Assembly.GetTypes()`
+which might impact the performance of Dependency Injection and similar systems.
+
+> [!NOTE]
+> In practice, there might not be many generated types, it depends on the kind of the program
+> (whether it has lots of short strings or a few large strings) and how [the threshold](#configuration) is configured.
```
cston, line 150: It looks like the release build of Microsoft.CodeAnalysis.CSharp.dll has roughly 200 strings for file paths, presumably from calls to methods such as `ExceptionUtilities.Unreachable()` that use `[CallerFilePath]`.

Given that, won't many assemblies contain a fair number of strings at 100 characters from paths alone? If so, let's consider removing this `NOTE`.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,257 @@
...
+### AOT
+
+Ahead-of-time compilation tools would need to be updated to recognize this new pattern of string literal use in place of `ldstr`.
+
+### Automatic threshold
+
+The threshold could be determined automatically with some objective, for example,
+use the utf8 encoding emit strategy for the lowest number of string literals necessary to avoid overflowing the UserString heap.
+
+The set of string literals is not know up front in the compiler, it is discovered lazily (and in parallel) by the emit layer.
```
cston, line 191: Should this be "not known"?

cston (approved):

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,253 @@
...
+## Emit
+
+The standard string literal emit strategy emits `ldstr "string"`.
+That puts the string literal on the UserString heap of the PE file.
+
+The utf8 string literal encoding emit strategy emits `ldsfld` of a field in a generated class instead.
+
+For every unique string literal, a unique internal static class is generated which:
+- has name composed of `<S>` followed by a hex-encoded XXH128 hash of the string
+  (collisions [should not happen][xxh128] with XXH128 and so they aren't currently detected or reported, the behavior in that case is undefined),
```
jkotas, line 57: Is this a security bug that the collision is undetected?

The collisions can happen with some very small but non-zero probability. If a program is unlucky to hit the collision, the string literals will be messed up and the behavior of the program will be impacted in interesting ways. This can be used by adversaries to contribute an innocently looking string to an open-source project where the string is carefully crafted to cause a collision with some other critical string.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,253 @@
...
+
+Instead of one global feature flag, the emit strategy could be controlled using compiler-recognized attributes (applicable to assemblies or classes).
+Furthermore, we could emit more than one string per one class. That could be configurable as well.
+
+One interesting strategy would be grouping strings which occur in one user-defined type into one compiler-generated `<S>` type.
+The idea is that strings from one class are likely used "together" so there is no performance impact from initializing them all at once and we save on metadata.
+
+### GC
+
+To avoid rooting the `string` references forever, we could turn the fields into `WeakReference<string>`s.
```
jkotas, line 224: This would provide no benefit and it would be quite expensive - for both direct overhead and indirectly for the GC (longer GC pause times). You can scratch this idea.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,253 @@
...
+Or we could avoid the caching altogether (each eligible `ldstr` would be replaced with a direct call to `Encoding.UTF8.GetString`).
+This could be configurable as well.
+
+### Runtime support
+
+We could emit the strings to data fields similarly but we would not synthesize the `string` fields and static constructors.
```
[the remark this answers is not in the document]
jaredpar: The plan is to use the current state of implementation, compiler only trick, to get feedback on the scenarios. We did a number of benchmarks on the work and didn't see meaningful differences that justified preemptively doing runtime work. Decided that we would ship it out as experimental, get real world feedback from teams like Bing, and then let that drive what we wanted the final state of the feature to be before we remove experimental lable

@davidwrighton, @DamianEdwards

docs/features/string-literals-data-section.md
jaredpar: There are a few places in the compiler where we use hashes and collisions could lead to program behavior changes. These are part of our security reviews in the past and this will be included in our next one. Thus far nothing has risen to the level of forcing extra checking. There is an increase in non-cryptographic hashing support though and I will make sure that is included in the next security review. @RikkiGibson

jaredpar: Understand the concern here. But you can create similar collisions today at much lower level places in the compiler by creating collisions with file contents. The use of non-crypto hashes was approved at that layer too (was in fact suggested by security team). The general reasoning is that once you can control code being submitted to the project you have lots of avenues for subverting expectations.

jkotas: > The general reasoning is that once you can control code being submitted to the project you have lots of avenues for subverting expectations.

In the open-source world, it is important that the project maintainers are able to review the changes with confidence.

For example, let's say that this feature gets enabled for some part of OpenTelemetry project and somebody submits a PR that includes a code like this:

```
void DiagnosticThatIsTurnedOffByDefault()
{
   // Random unique string to help us identify blah
   DoStuff("ruroiodffdshofps09rw403492049358302849052peldfjskdfshgfdshiofofjsekljhkdkjafhkds3");
}
```

How can the maintainer of the OpenTelemetry project tell that this change is actually planting a backdoor into an unrelated code due to a hash collision?

RikkiGibson: xxHash128 is also used to identify source documents in the IDE currently, as well as to identify interceptable calls in source code. We consider it collision-safe for these purposes. See also the Examples at the bottom of this doc: https://web.archive.org/web/20240223124645/https://fossies.org/linux/xxHash/tests/collisions/README.md

@CyrusNajmabadi in case you have further thoughts.

jkotas: > xxHash128 is also used to identify source documents in the IDE currently

I would not be worried about the IDE case from the security angle. If somebody crafts a file name with a hash collision and it makes the IDE to misbehave or crash, it is not a big deal. (You may be scratching your head for a while if you are tasked with investigating the crash report.)

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,253 @@
...
+## Diagnostics
+
+The error CS8103 (Combined length of user strings used by the program exceeds allowed limit) is updated to suggest the feature flag,
+albeit with a disclaimer during the experimental phase of the feature.
+
+## Runtime
+
+Throughput of `ldstr` vs `ldsfld` is very similar (both result in one or two move instructions).
+
+In the `ldsfld` emit strategy, the `string` instances won't ever be collected by the GC once the generated class is initialized.
```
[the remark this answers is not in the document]
jjonescz: That's what I figured and the sentence "`ldstr` has similar behavior" below was supposed to say that, although I could clarify that.

docs/features/string-literals-data-section.md
jjonescz: FWIW, I think it should be straightforward to add collision detection, it just wasn't straightforward to test that.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,253 @@
...
+`ldstr` has similar behavior, but there are some optimizations in the runtime around `ldstr`,
+e.g., they are loaded into a different frozen heap so machine codegen can be more efficient (no need to worry about pointer moves).
+
+Generating new types by the compiler means more type loads and hence runtime impact,
+e.g., startup performance and the overhead of keeping track of these types.
```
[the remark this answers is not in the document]
jjonescz: Using that benchmark I'm getting:

| feature enabled | feature disabled |
| --- | --- |
| 2228224 | 2568192 |
| Approximate Cost per string literal: 222 | Approximate Cost per string literal: 256 |

(I have tried several times, the results are always roughly similar)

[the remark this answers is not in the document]
jjonescz: Btw, if you want to experiment more, simply add this to the project file:

```csproj
  <PropertyGroup>
    <Features>$(Features);experimental-data-section-string-literals=0</Features>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="4.14.0-2.25104.16">
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
  </ItemGroup>
```

and `dotnet-tools` feed to nuget.config:

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
  </packageSources>
</configuration>
```

To turn the feature off again, replace `0` with `off`. in the `<Features>` property.

docs/features/string-literals-data-section.md
```diff
@@ -0,0 +1,253 @@
...
+
+The feature flag can take a non-negative integer threshold.
+Only string literals whose length (number of characters, not bytes) is greater than the threshold are emitted using the utf8 encoding strategy.
+If the flag is set, but no value or an unrecognized value is specified, the threshold defaults to 100.
+Specifying 0 means all non-empty string literals are considered for the feature.
+
+The feature is turned off if
+- the feature flag is not specified, or
+- the string `off` is provided as the feature flag value, or
+- `null` is provided as the feature flag value (usually only possible using Roslyn APIs like `ParseOptions.WithFeatures`).
```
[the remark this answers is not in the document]
jjonescz: Oh, interesting, let me just call out in the doc that this is not supported for now.

docs/features/string-literals-data-section.md
jjonescz: https://github.com/dotnet/roslyn/pull/77061

docs/features/string-literals-data-section.md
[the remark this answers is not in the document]
jkotas: I had a bug in the test methodology. Here is updated perf test: https://github.com/jkotas/stringliteralperf

Feature on:
```
C:\stringliteralperf\driver>dotnet run -c release
Cost per string literal (bytes): 1037
Cost per string literal (microseconds): 20.3
```
Feature off:
```
C:\stringliteralperf\driver>dotnet run -c release
Cost per string literal (bytes): 550
Cost per string literal (microseconds): 3.1
```

The runtime overhead of this feature per 100 char string literal is ~500 bytes of working set memory (~2x cost of regular string literal) and ~17 microseconds of startup time (~7x cost of regular string literal).

The startup time overhead does depend on the length of the string literal. It is cost of the type loads and JITing the static constructor.

The working set has two components: private working set (r/w pages) and non-private working set (r/o pages backed by the binary). The private working set overhead (~600 bytes) does not depend on the length of the string literal. Again, it is the cost the type loads and the static constructor code. Non-private working set is reduced by this feature since the binary is smaller. Once the string literal is about 600 characters, the private working set overhead and non-private working set improvement will break even.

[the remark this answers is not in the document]
jjonescz: Thanks! I'm going to quote you in the spec if you don't mind.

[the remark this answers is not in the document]
jjonescz: > The private working set overhead (~600 bytes) does not depend on the length of the string literal. Again, it is the cost the type loads and the static constructor code.

This is unclear to me - that overhead should depend on the length of the string literal, no? Also why "~600" bytes?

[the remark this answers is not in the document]
jjonescz: > The private working set overhead (~600 bytes) does not depend on the length of the string literal.

Should this be "~500 bytes"?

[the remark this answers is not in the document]
jkotas: When the feature is off, 550 bytes cost of 100 char string literal is composed from:
- The string in the binary (~200 bytes)
- The string allocated on the GC heap (~200 bytes)
- Fixed overheads: metadata encoding, runtime hashtable of all allocated string literals, code that reference the string in the benchmark (~150 bytes)

When the feature is on, ~1050 bytes (1037 rounded up) cost of 100 char string literal is composed from:
- The string in the binary (~100 bytes)
- The string allocated on the GC heap (~200 bytes)
- Fixed overheads: metadata encoding, the extra types, code that reference the string in the benchmark (~750 bytes)

750 - 150 = 600. Vast majority of it are the extra types.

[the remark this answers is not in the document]
jkotas: A bit of the extra fixed overheads when the feature is probably in the non-private working set. It is pain to measure it since there is no managed API to get private vs. non-private working set. It does not impact the estimate of the break-even point for the total working set.

docs/features/string-literals-data-section.md
jjonescz: Actually I think hash collisions might not be a problem - https://github.com/dotnet/roslyn/pull/77061#issuecomment-2645748907

PR #76196, opened 2024-12-01
Update xunit

Youssef1313:


Youssef1313:
@jaredpar The updated versions are not present in the internal feeds. Would you consider adding them and update?

My goal here is to catch up early with updates just in case there are new analyzer warnings that need to be addressed and keeping up with newer xunit versions to avoid having a future "large" jump in versions. Smaller jumps are usually a lot easier.

Note that when xunit v3 is stable, it may be really good to update to it, and even switching to Microsoft.Testing.Platform instead of VSTest.

jaredpar:
@azp run

jaredpar:
@Youssef1313 looks like real build ambiguities

src/Tools/TestDiscoveryWorker/Program.cs
```diff
@@ -65,7 +65,7 @@
         await Console.Out.WriteLineAsync($"Discovering tests in {testDescriptor}...").ConfigureAwait(false);
 
         using var xunit = new XunitFrontController(AppDomainSupport.IfAvailable, assemblyFileName, shadowCopy: false);
-        var configuration = ConfigReader.Load(assemblyFileName);
+        var configuration = ConfigReader.Load(assemblyFileName, configFileName: null);
```
Youssef1313, line 68: For reference: A new overload was added that also has optional parameter, causing ambiguity.

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -28,7 +29,7 @@ public static TheoryData<LanguageVersion> LangVersions()
     }
 
     private sealed class CombinatorialLangVersions()
-        : CombinatorialValuesAttribute(LangVersions().Select(d => d.Single()).ToArray());
+        : CombinatorialValuesAttribute(((IEnumerable<object[]>)LangVersions()).Select(d => d.Single()).ToArray());
```
Youssef1313, line 32: For reference: Previously `TheoryData<T>` was an `IEnumerable<object[]>`, but now it's both `IEnumerable<object[]>` and `IEnumerable<T>`, causing Linq's Select to not be resolved properly without the cast.

Youssef1313:
Thanks @jaredpar. Let's see now maybe.

Youssef1313:
@jaredpar Are you interested in fixing https://xunit.net/xunit.analyzers/rules/xUnit2031, or do you think it should be suppressed instead? There is not much violations.

Youssef1313:
I pushed a fix for now. If you prefer to disable xUnit2031, let me know and I'll revert the changes and disable the analyzer.

jasonmalinowski (approved):

jaredpar (approved):

jaredpar:
I'm fine with a fix here.

jcouv (approved):
LGTM Thanks (iteration 4)

jcouv:
> Note that when xunit v3 is stable, it may be really good to update to it

@Youssef1313 I'm looking forward to xunit v3. I think it adds a "PrintMaxStringLength" setting which lets us change how to xunit truncates actual and expected strings in its output. The default is 50 characters, which is often annoying...

Youssef1313:
@jcouv The current main issue for upgrading to xunit.v3 is the integration test harness. I started some work in https://github.com/microsoft/vs-extension-testing/pull/179 but so far not yet finished. The migration is much more involving that I thought, as you see by the size of that unfinished PR.

Youssef1313:
@jcouv Yes. It's vice-versa. The first step is updating to xUnit v3, then the second step is to update to MTP. But if the update to xUnit v3 happened, I'd strongly recommend that you update to MTP after. (We can have a chat about that internally if you want)

Issue #75759, opened 2024-11-05
Theoretical race condition in PETypeParameterSymbol (and possibly other locations)


333fred:
There is a theoretical race condition in how we initialize data in `PETypeParameterSymbol` (currently just C#, but soon to be VB as well since it's copying the same code style). This race condition is practically impossible to reproduce, to be clear; on x86/64, the memory model should make it impossible. On ARM, it is technically possible to hit due to the looser memory model, but would also be extremely unlikely to happen in reality. However, we should fix this so that we're not relying on undefined behavior for this scenario.

### The race

The original file is [here](https://github.com/dotnet/roslyn/blob/33e986c4c814f374e853ddc145dcb73a23a15d2f/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PETypeParameterSymbol.cs). I'm going to copy over a simplified version to this issue for demonstration purposes.

```cs
internal sealed class PETypeParameterSymbol
    : TypeParameterSymbol
{
    private ThreeState _lazyHasIsUnmanagedConstraint; // This is only valid when _lazyDeclaredConstaintTypes is non-default
    private ImmutableArray<TypeWithAnnotations> _lazyDeclaredConstraintTypes; // This is a sentinel value

    private ImmutableArray<TypeWithAnnotations> GetDeclaredConstraintTypes(ConsList<PETypeParameterSymbol> inProgress)
    {
        if (_lazyDeclaredConstraintTypes.IsDefault)
        {
            // ...
            _lazyHasIsUnmanagedConstraint = hasUnmanagedModreqPattern.ToThreeState();
            // This is a full read-write barrier: as long as we guarantee that _lazyDeclaredConstraintTypes is read before
            // _lazyHasIsUnmanagedConstraint, we aren't in a torn state: either _lazyDeclaredConstraintTypes is non-default,
            // and _lazyHasIsUnmanagedConstraint is valid, or _lazyDeclaredConstraintTypes is default, and we'll (re)do the
            // initialization work on this thread before observing _lazyHasIsUnmanagedConstraint
            ImmutableInterlocked.InterlockedInitialize(ref _lazyDeclaredConstraintTypes, declaredConstraintTypes);
        }
    }

    public override bool HasUnmanagedTypeConstraint
    {
        get
        {
            GetDeclaredConstraintTypes(ConsList<PETypeParameterSymbol>.Empty);
            // There is no read barrier between these locations: the JIT and/or CPU is allowed to speculatively read
[...]

PR #76230, opened 2024-12-02
Threading fixes

333fred:
We have a few theoretical races in some of our threading code, as detailed in https://github.com/dotnet/roslyn/issues/75759. I've audited our uses of `ImmutableArray` as backing fields for this pattern, fixing it up where I find issues, as well as a couple of other threading issues that I found while looking at things named `lazy` in our codebase.

### DIFF
--- a/src/Compilers/CSharp/Portable/Emitter/Model/PEModuleBuilder.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/PEModuleBuilder.cs
@@ -613,9 +613,9 @@ namespace Microsoft.CodeAnalysis.CSharp.Emit
 
             if (_lazyExportedTypes.IsDefault)
             {
-                _lazyExportedTypes = CalculateExportedTypes();
+                var initialized = ImmutableInterlocked.InterlockedInitialize(ref _lazyExportedTypes, CalculateExportedTypes());
 
-                if (_lazyExportedTypes.Length > 0)
+                if (initialized && _lazyExportedTypes.Length > 0)
                 {
                     ReportExportedTypeNameCollisions(_lazyExportedTypes, diagnostics);
                 }
--- a/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEMethodSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEMethodSymbol.cs
@@ -131,24 +131,24 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols.Metadata.PE
             public bool IsExplicitFinalizerOverride => (_bits & IsExplicitFinalizerOverrideBit) != 0;
             public bool IsExplicitClassOverride => (_bits & IsExplicitClassOverrideBit) != 0;
             public bool IsExplicitOverrideIsPopulated => (_bits & IsExplicitOverrideIsPopulatedBit) != 0;
-            public bool IsObsoleteAttributePopulated => (_bits & IsObsoleteAttributePopulatedBit) != 0;
-            public bool IsCustomAttributesPopulated => (_bits & IsCustomAttributesPopulatedBit) != 0;
-            public bool IsUseSiteDiagnosticPopulated => (_bits & IsUseSiteDiagnosticPopulatedBit) != 0;
-            public bool IsConditionalPopulated => (_bits & IsConditionalPopulatedBit) != 0;
-            public bool IsOverriddenOrHiddenMembersPopulated => (_bits & IsOverriddenOrHiddenMembersPopulatedBit) != 0;
+            public bool IsObsoleteAttributePopulated => (Volatile.Read(ref _bits) & IsObsoleteAttributePopulatedBit) != 0;
+            public bool IsCustomAttributesPopulated => (Volatile.Read(ref _bits) & IsCustomAttributesPopulatedBit) != 0;
+            public bool IsUseSiteDiagnosticPopulated => (Volatile.Read(ref _bits) & IsUseSiteDiagnosticPopulatedBit) != 0;
+            public bool IsConditionalPopulated => (Volatile.Read(ref _bits) & IsConditionalPopulatedBit) != 0;
+            public bool IsOverriddenOrHiddenMembersPopulated => (Volatile.Read(ref _bits) & IsOverriddenOrHiddenMembersPopulatedBit) != 0;
             public bool IsReadOnly => (_bits & IsReadOnlyBit) != 0;
             public bool IsReadOnlyPopulated => (_bits & IsReadOnlyPopulatedBit) != 0;
             public bool DoesNotReturn => (_bits & DoesNotReturnBit) != 0;
             public bool IsDoesNotReturnPopulated => (_bits & IsDoesNotReturnPopulatedBit) != 0;
-            public bool IsMemberNotNullPopulated => (_bits & IsMemberNotNullPopulatedBit) != 0;
+            public bool IsMemberNotNullPopulated => (Volatile.Read(ref _bits) & IsMemberNotNullPopulatedBit) != 0;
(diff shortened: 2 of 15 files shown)


src/Compilers/CSharp/Portable/Emitter/Model/PEModuleBuilder.cs
```diff
@@ -613,9 +613,9 @@ private static void GetExportedTypes(NamespaceOrTypeSymbol symbol, int parentInd
 
             if (_lazyExportedTypes.IsDefault)
             {
-                _lazyExportedTypes = CalculateExportedTypes();
+                var initialized = ImmutableInterlocked.InterlockedInitialize(ref _lazyExportedTypes, CalculateExportedTypes());
```
333fred, line 616: This is not a case of immutable arrays being used as a gate for other logic fields. Rather, it is a simple threading issue that I discovered while auditing all of our lazy immutable array fields. It could result in diagnostics being double reported depending on calling patterns, based on simple thread timing.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PEMethodSymbol.cs
```diff
@@ -131,24 +131,24 @@ public MethodKind MethodKind
             public bool IsExplicitFinalizerOverride => (_bits & IsExplicitFinalizerOverrideBit) != 0;
             public bool IsExplicitClassOverride => (_bits & IsExplicitClassOverrideBit) != 0;
             public bool IsExplicitOverrideIsPopulated => (_bits & IsExplicitOverrideIsPopulatedBit) != 0;
-            public bool IsObsoleteAttributePopulated => (_bits & IsObsoleteAttributePopulatedBit) != 0;
-            public bool IsCustomAttributesPopulated => (_bits & IsCustomAttributesPopulatedBit) != 0;
-            public bool IsUseSiteDiagnosticPopulated => (_bits & IsUseSiteDiagnosticPopulatedBit) != 0;
-            public bool IsConditionalPopulated => (_bits & IsConditionalPopulatedBit) != 0;
-            public bool IsOverriddenOrHiddenMembersPopulated => (_bits & IsOverriddenOrHiddenMembersPopulatedBit) != 0;
+            public bool IsObsoleteAttributePopulated => (Volatile.Read(ref _bits) & IsObsoleteAttributePopulatedBit) != 0;
```
333fred, line 134: For these types of flags enums, where we read a bit from packed flags, then look at a separate structure for the actual value, nothing was preventing us from looking at the separate structure first due to speculation or reordering. I did not use `Volatile.Read` everywhere because, for the other flags, both the "is populated" and "result" sections are both bits on the same integer, and we do not have to worry about out-of-order reads of the same field.

src/Compilers/CSharp/Portable/Symbols/Metadata/PE/PENestedNamespaceSymbol.cs
```diff
@@ -125,11 +125,14 @@ internal override ModuleSymbol ContainingModule
...
+            var typesByNS = Volatile.Read(ref _typesByNS);
 
-            if (lazyTypes == null || lazyNamespaces == null)
+            if (typesByNS == null)
+            {
+                Debug.Assert(lazyNamespaces != null && lazyTypes != null);
+            }
+            else
             {
-                System.Diagnostics.Debug.Assert(typesByNS != null);
```
333fred, line 132: This was a simple thread-timing issue waiting to happen. Rather than using `lazyTypes` and `lazyNamespaces` as proxies to the real sentinel, we should just use `_typesByNS`, which is the actual sentinel that work was completed.

src/Compilers/CSharp/Portable/Symbols/Source/LocalFunctionSymbol.cs
```diff
@@ -221,7 +222,7 @@ private void ComputeParameters()
                 _declarationDependencies.AddAll(diagnostics.DependenciesBag);
                 diagnostics.Free();
                 _lazyIsVarArg = isVararg;
-                _lazyParameters = parameters;
+                RoslynImmutableInterlocked.VolatileWrite(ref _lazyParameters, parameters);
```
333fred, line 225: Nothing prevents writes _within_ a lock from being observed in alternate orders by other threads, when those other threads don't use a `lock` to check themselves. Since `_lazyParameters` is checked for `Default` outside of a lock as the sentinel, we therefore need to make sure that it's the last write inside this `lock` or we could still race.

src/Compilers/Core/Portable/Syntax/SyntaxTree.cs
```diff
@@ -369,11 +370,11 @@ internal int GetDisplayLineNumber(TextSpan span)
         /// </summary>
         internal Cci.DebugSourceInfo GetDebugSourceInfo()
         {
-            if (_lazyChecksum.IsDefault)
+            if (RoslynImmutableInterlocked.VolatileRead(ref _lazyChecksum).IsDefault)
```
333fred, line 373: Stumbling on this was frankly a bit shocking. My only conclusion is that `GetDebugSourceInfo` can never be called from threaded code, because no speculation or reordering is required to race here. It just sets the fields in the wrong order!

AlekseyTs (approved):
LGTM (commit 5)

333fred:

333fred:

src/Compilers/VisualBasic/Portable/Symbols/Attributes/PEAttributeData.vb
```diff
@@ -120,7 +120,7 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols.Metadata.PE
 
         Private Sub EnsureLazyMembersAreLoaded()
 
-            If _lazyConstructorArguments Is Nothing Then
+            If Volatile.Read(_lazyConstructorArguments) Is Nothing Then
```
cston, line 123: >Nothing [](http://example.com/codeflow?start=59&length=7)

Do we need `OrElse Volatile.Read(_lazyNamedArguments) Is Nothing` here as well, or `Volatile.Read(_lazyNamedArguments) Is Nothing` in `CommonNamedArguments`?

src/Compilers/VisualBasic/Portable/Symbols/Metadata/PE/PENestedNamespaceSymbol.vb
```diff
@@ -118,10 +118,11 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols.Metadata.PE
         End Property
 
         Protected Overrides Sub EnsureAllMembersLoaded()
-            Dim typesByNS = _typesByNS
+            Dim typesByNS = Volatile.Read(_typesByNS)
 
-            If m_lazyTypes Is Nothing OrElse m_lazyMembers Is Nothing Then
-                Debug.Assert(typesByNS IsNot Nothing)
+            If typesByNS Is Nothing Then
+                Debug.Assert(m_lazyMembers IsNot Nothing AndAlso m_lazyMembers IsNot Nothing)
```
cston, line 124: >m_lazyMembers [](http://example.com/codeflow?start=65&length=13)

`m_lazyTypes`?

src/Compilers/VisualBasic/Portable/Symbols/Attributes/PEAttributeData.vb
333fred: I don't see anything in this type that is gated on it. Since `EnsureLazyMembers` uses `_lazyConstructorArguments` as the gate for whether to initialize everything, and assigns that last, the access in `CommonNamedArguments` should be safe: either `_lazyNamedArguments` was non-default, and the second read is fine, or it wasn't null, and this `Volatile.Read` will ensure that the second read of `_lazyNamedArguments` isn't reordered before a potential initialization.

cston (approved):

PR #76233, opened 2024-12-03
Add Color Scheme support for VS 2022 tinted themes.

### LINKED ISSUE #76232: C# Methods are not colored
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/C-Methods-are-not-colored/10519205)._

---
[severity:It bothers me. A fix would be nice] [regression] [worked-in:17.8]
Hi, since update 17.9.0 Preview 1.0, methods are not colored as they were before. It is not something groundbreaking, but I already got used to it and now it is little bit more difficult to navigate. Thank you.

---
#### Feedback Bot on 11/19/2023, 08:35 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

#### Anonymized GDPR on 11/19/2023, 11:08 PM: 

<p>Sorry for the problem you encountered, in order to better investigate the problem, could you provide detailed repro steps and a demo project which can repro this problem? Could you describe in detail what your expected and actual results were? It would be better if you can provide error screenshots.</p>
<p>We look forward to hearing from you!</p>

#### Mihok, Marcel on 11/19/2023, 11:28 PM: 

<p>Hi,<br>
I was gonna show you screenshots of VS 17.8 on my machine, where the methods are not colored versus VS 17.7 on AVD, where the methods are colored, but I just noticed, that the cause of this is somewhere else - When the VS theme is set to System Settings, then the methods are not colored, but when you set it to Dark or Light theme, the methods color is ok.<br>
Here is a screenshot of VS and how methods are not colored when using system theme:<br>
</p>
<p>Here is a screenshot when using Dark or Light theme:<br>
<br>
</p>
<p>Here are screenshots from other themes (no method color):<br>
<br>
</p>
<p>So yeah, the culprit must be the “Use system setting” and perhaps how the other themes are implemented or something. I guess we found a workaround then?</p>
<p>Oh and i did not enclose any demo project or wrote repro steps explicitly, I think there is no need for it, because it is clear what is going on.</p>
<p>Thank you</p>

#### Anonymized GDPR on 11/22/2023, 01:43 AM: 

<p>Thanks for your sharing the detail steps, I can reproduce this issue through your shared steps and this issue has been escalated for further investigation, if there is any process, I will inform you immediately.</p>

#### Lazie Alcione Wouters on 2/15/2024, 00:32 PM: 

<p>I am having the same issue after updating from 17.8.4 to 17.9.0<br>
Theme’s colors are not being applied to method names.</p>

#### Павло Прийма on 2/20/2024, 08:38 AM: 

<p>the same issue</p>

#### Martin Janus on 2/27/2024, 08:31 AM: 

<p>same on 17.9.1 I have VS Pro 2022 at work and I tried to export the settings, applying it to CE at home applies everything, the font and colors are set but are not shows, as said above for example a method has the same color as an identifier, it also bothers me because I got used to the color scheme and I could identify items quicker</p>

#### Nick Dick on 3/7/2024, 04:31 AM: 

<p>I have same issue on 17.9.2</p>

#### Christopher Ormsby on 3/8/2024, 10:35 AM: 

<p>Same issue with 17.9.2 as well. Issue extends to custom theme packs like the Visual Studio Theme Pack. I use Dark+, exact same issue.</p>
<p>It’s a huge problem when you’re really cooking and you have to break your flow to mouse over and double check your method is actually a method.</p>

#### U_1310234 on 3/13/2024, 00:22 PM: 

<p>The issue is still present in 17.9.3.</p>

#### Femi Alogba on 3/14/2024, 05:59 AM: 

<p>Same issue here but with Dark+ theme by vs devlabs. The default themes work okay I think. First noticed after updating to vs version: 17.9.3.</p>

#### U_1310234 on 3/27/2024, 08:13 AM: 

<p>Confirmed the workaround in 17.9.3.</p>

#### Lazie Alcione Wouters on 3/28/2024, 07:14 AM: 

<p>Thanks <a target='_blank' href="/user/1308496">Christopher Ormsby</a></p>

---
#### Christopher Ormsby solved on 3/27/2024, 06:27 AM, undefined votes: 

<p>Found the following workaround on a github comment string:<br>
</p>
<p>Following the instructions in the users screenshot will temporarily fix the issue until MS can get itself together and fix the issue. The source linked in the SS:<br>
https://developercommunity.visualstudio.com/t/Identifier-has-too-high-priority-in-sy/10547451</p>

JoeRobich:
So, there are really two new things happening in VS.

1. The platform team separated the VS Theme into a Shell Theme and an Editor Theme.
2. They introduced new "tinted" Shell Themes which then specify an Editor Theme from the set of known themes to be their default.

By removing the known theme check, we are always applying the color scheme even when the theme is unknown to us. Which is the case for these new "tinted" themes as they have new Guids. However, with the color scheme in place, the Editor colors are able to be picked up because the Editor theme is using one of the known themes which we provide colors for.

Changes:
- Removes the KnownTheme check from the ColorSchemeApplier. This check was unnecessary because the editor colors we provide are tied to the known themes and if a user was using a custom theme it would not interfere with their configuration.
- Removes the IsThemeCustomized check from ColorSchemeApplier. This check was used when determining the default scheme to apply when the color scheme setting had no value. The goal was that we would use VS2017 colors for existing users who had already tweaked their colors so they would not be impacted by the new semantic color defaults. This is no longer necessary as I would imagine most existing user have used VS2019 or VS2022 at this point and a color scheme setting has been applied. For those users coming straight from VS2017, it is an easy fix to update the Color Scheme setting.
- Removes the ClassificationVerifier as we no longer check whether the user has customized their theme.
- Removes the warnings from the Advanced Options page as we no longer disable the color scheme combobox when not in a known theme.

Fixes #76232

https://github.com/user-attachments/assets/17973c11-80b9-4729-9671-b0e38e912577


Cosifne:
I think there is an analyzer warning

JoeRobich:
> I think there is an analyzer warning

@Cosifne Thanks! Updated.

Cosifne (approved):

Issue #75802, opened 2024-11-07
Collection initializers can leak refs


jjonescz:
**Version Used**: 4.13.0-2.24555.1 ([9dcb2a3](https://github.com/dotnet/roslyn/commit/9dcb2a31699f7d09974c9b5af571b102b291ee44))

**Steps to Reproduce**:

```cs
using System;
using System.Collections;
using System.Diagnostics.CodeAnalysis;

R r = M();
Console.WriteLine(r.F);

static R M()
{
    var local = 1;
    return new R() { local };
}

ref struct R : IEnumerable
{
    public ref readonly int F;

    public void Add([UnscopedRef] in int x)
    {
        F = ref x;
    }

    IEnumerator IEnumerable.GetEnumerator() => throw null;
}
```

**Expected Behavior**: Error on `return new R() { local };` like there would be if explicitly written as `var r = new R(); r.Add(local); return r;`.

**Actual Behavior**: No errors. The code accesses a dead reference when executing.

Issue #63306, opened 2022-08-10
Escape analysis ignores local passed by reference to interpolated string handler `AppendFormatted()` method


cston:
```csharp
using System;
using System.Runtime.CompilerServices;

[InterpolatedStringHandler]
ref struct CustomHandler
{
    private ref readonly int _i;
    public CustomHandler(int literalLength, int formattedCount) { }
    public void AppendFormatted(in int i) { _i = ref i; }
}

class Program
{
    static CustomHandler F1()
    {
        int i = 1;
        return $"{i}"; // error?
    }

    static CustomHandler F2()
    {
        CustomHandler h = $"{2}"; // error?
        return h;
    }
}
```
Expected:
Escape analysis should report errors for the conversions from interpolated string to `CustomHandler` since a reference to the local is potentially returned to the calling method.

Actual:
No errors.

RikkiGibson:
In this bit, it seems like we weren't precisely correct in our assumptions on how the calls within interpolated string conversions could be formed. An unscoped in-parameter could come in here which we don't handle properly: https://github.com/dotnet/roslyn/blob/3a17b7bca601565f3629ac62e964e02312183eb7/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs#L4131-L4135

I think we could avoid entirely making assumptions about the shape of the call here, and instead just pass it through the appropriate general purpose helper for calls (GetInvocationArgumentsForEscape or GetInvocationArgumentsForEscapeWithUpdatedRules in #62886). Another layer may be synthesizing the call within some limited set of constraints, but it doesn't matter to the escape analysis layer.

jaredpar:
This is no longer a bug because the `in` here is RTSE of _return only_ so it can't be captured. A very related scenario is if a `ref struct` is passed by value that is tracked by our current code. That is failing and there is a test case so closing out. 

[...]

PR #76237, opened 2024-12-03
Check ref safety of arg mixing in collection initializers

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/75802.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
@@ -4454,12 +4454,52 @@ namespace Microsoft.CodeAnalysis.CSharp
                     return GetValEscapeOfObjectInitializer(initExpr, scopeOfTheContainingExpression);
 
                 case BoundKind.CollectionInitializerExpression:
-                    var colExpr = (BoundCollectionInitializerExpression)expr;
-                    return GetValEscape(colExpr.Initializers, scopeOfTheContainingExpression);
+                    {
+                        var colExpr = (BoundCollectionInitializerExpression)expr;
 
-                case BoundKind.CollectionElementInitializer:
-                    var colElement = (BoundCollectionElementInitializer)expr;
-                    return GetValEscape(colElement.Arguments, scopeOfTheContainingExpression);
+                        //    var c = new C() { element };
+                        //
+                        // is equivalent to
+                        //
+                        //    var c = new C();
+                        //    c.Add(element);
+                        //
+                        // So we check arg mixing of `(receiverPlaceholder).Add(element)` calls,
+                        // and make the result "scoped" if any call could cause the elements to escape.
+
+                        using var _ = new PlaceholderRegion(this, [(colExpr.Placeholder, SafeContext.CallingMethod)]) { ForceRemoveOnDispose = true };
+                        foreach (var initializer in colExpr.Initializers)
+                        {
+                            switch (initializer)
+                            {
+                                case BoundCollectionElementInitializer init:
+                                    if (!CheckInvocationArgMixing(
+                                            init.Syntax,
+                                            MethodInfo.Create(init.AddMethod),
+                                            receiverOpt: colExpr.Placeholder,
+                                            receiverIsSubjectToCloning: init.InitialBindingReceiverIsSubjectToCloning,
+                                            parameters: init.AddMethod.Parameters,
+                                            argsOpt: init.Arguments,
+                                            argRefKindsOpt: default,
+                                            argsToParamsOpt: init.ArgsToParamsOpt,
+                                            scopeOfTheContainingExpression: scopeOfTheContainingExpression,
+                                            BindingDiagnosticBag.Discarded))
+                                    {
+                                        return scopeOfTheContainingExpression;
+                                    }
+                                    break;
+
+                                case BoundDynamicCollectionElementInitializer dynamicInit:
+                                    break;
+
+                                default:
+                                    Debug.Fail($"{initializer.Kind} expression of {initializer.Type} type");
+                                    break;
+                            }
+                        }
+
+                        return SafeContext.CallingMethod;
+                    }
 
                 case BoundKind.ObjectInitializerMember:
                     // this node generally makes no sense outside of the context of containing initializer
@@ -4468,9 +4508,16 @@ namespace Microsoft.CodeAnalysis.CSharp
                     return scopeOfTheContainingExpression;
 
                 case BoundKind.ImplicitReceiver:
-                case BoundKind.ObjectOrCollectionValuePlaceholder:
                     // binder uses this as a placeholder when binding members inside an object initializer
                     // just say it does not escape anywhere, so that we do not get false errors.
+                    return scopeOfTheContainingExpression;
+
+                case BoundKind.ObjectOrCollectionValuePlaceholder:
+                    if (_placeholderScopes?.TryGetValue((BoundObjectOrCollectionValuePlaceholder)expr, out var scope) == true)
+                    {
+                        return scope;
+                    }
+
                     return scopeOfTheContainingExpression;
 
                 case BoundKind.InterpolatedStringHandlerPlaceholder:
@@ -4773,6 +4820,18 @@ namespace Microsoft.CodeAnalysis.CSharp
                         return inUnsafeRegion;
                     }
                     return true;
+
+                case BoundKind.ObjectOrCollectionValuePlaceholder:
+                    if (_placeholderScopes?.TryGetValue((BoundObjectOrCollectionValuePlaceholder)expr, out var scope) == true)
+                    {
+                        if (!scope.IsConvertibleTo(escapeTo))
+                        {
+                            Error(diagnostics, inUnsafeRegion ? ErrorCode.WRN_EscapeVariable : ErrorCode.ERR_EscapeVariable, node, expr.Syntax);
+                            return inUnsafeRegion;
+                        }
+                        return true;
+                    }
+                    goto default;
 
                 case BoundKind.Local:
                     var localSymbol = ((BoundLocal)expr).LocalSymbol;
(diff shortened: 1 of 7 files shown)


src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -4454,12 +4454,52 @@ internal SafeContext GetValEscape(BoundExpression expr, SafeContext scopeOfTheCo
...
+                        //
+                        // is equivalent to
+                        //
+                        //    var c = new C();
+                        //    c.Add(element);
+                        //
+                        // So we check arg mixing of `(receiverPlaceholder).Add(element)` calls,
+                        // and make the result "scoped" if any call could cause the elements to escape.
+
+                        using var _ = new PlaceholderRegion(this, [(colExpr.Placeholder, SafeContext.CallingMethod)]) { ForceRemoveOnDispose = true };
```
jjonescz, line 4470: Added `ForceRemoveOnDispose = true` because otherwise it can happen that the placeholder is added twice causing an assert to fail.

jjonescz: This might need to be extended to also handle the assert in the Add part, like I'm doing in a similar PR: https://github.com/dotnet/roslyn/pull/76263/commits/b64ee878d237bcf28fad525a9d7c3b79eb737aca

But the bigger question is if this is a viable approach in general - using PlaceholderRegion temporarily during the check.

jaredpar:
@RikkiGibson, @cston FYI

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -4313,6 +4313,180 @@ static void Main()
...
+                using System.Diagnostics.CodeAnalysis;
+
+                class C
+                {
+                    R M()
+                    {
+                        var local = 1;
+                        var r = new R() { local };
+                        return r;
+                    }
```
cston, line 4361: >} [](http://example.com/codeflow?start=20&length=1)

Consider testing a `scoped R` as well.
```csharp
static void M2()
{
    var local = 2;
    scoped R r;
    r = new R() { local }; // no error
}
```

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -4313,6 +4313,180 @@ static void Main()
...
+                    IEnumerator IEnumerable.GetEnumerator() => throw null;
+                }
+                """;
+            CreateCompilation([source, UnscopedRefAttributeDefinition]).VerifyDiagnostics(
+                // (8,16): error CS9203: A collection expression of type 'R' cannot be used in this context because it may be exposed outside of the current scope.
+                //         return [local];
+                Diagnostic(ErrorCode.ERR_CollectionExpressionEscape, "[local]").WithArguments("R").WithLocation(8, 16));
+        }
+
+        [Fact, WorkItem("https://github.com/dotnet/roslyn/issues/75802")]
```
cston, line 4454: >https://github.com/dotnet/roslyn/issues/75802 [](http://example.com/codeflow?start=25&length=45)

Is this test related to this issue or change? (It looks like this is expected behavior because the `ReadOnlySpan<int>` argument to `Builder.Create()` is not `scoped`.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -4313,6 +4313,180 @@ static void Main()
...
+                    public void Add(in int x) { }
+                    IEnumerator IEnumerable.GetEnumerator() => throw null;
+                }
+                """;
+            CreateCompilation(source).VerifyDiagnostics();
+        }
+
+        [Theory, WorkItem("https://github.com/dotnet/roslyn/issues/75802")]
+        [InlineData("[System.Diagnostics.CodeAnalysis.UnscopedRef]")]
+        [InlineData("")]
```
cston, line 4427: >[InlineData("")] [](http://example.com/codeflow?start=8&length=16)

It looks like the error reported when there is no `[UnscopedRef]` attribute may be unrelated to this change. Instead, it looks like `RefSafetyAnalysis.HasLocalScope` is not expecting a `ref struct` to implement `IEnumerable`, other than in error cases.

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
cston: It looks like adding a PlaceholderRegion temporarily here works, and it is unfortunate that PlaceholderRegions are typically added during the Visit methods and not removed. I'm curious though: was there a reason it was necessary to add the PlaceholderRegion temporarily in this particular case rather than adding it once (and not removing), in the calling `VisitObjectCreationExpressionBase()`?

cston (approved):

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -4774,6 +4782,18 @@ internal bool CheckValEscape(SyntaxNode node, BoundExpression expr, SafeContext
                     }
                     return true;
 
+                case BoundKind.ObjectOrCollectionValuePlaceholder:
+                    if (_placeholderScopes?.TryGetValue((BoundObjectOrCollectionValuePlaceholder)expr, out var scope) == true)
+                    {
+                        if (!scope.IsConvertibleTo(escapeTo))
+                        {
+                            Error(diagnostics, inUnsafeRegion ? ErrorCode.WRN_EscapeVariable : ErrorCode.ERR_EscapeVariable, node, expr.Syntax);
```
cston, line 4790: >inUnsafeRegion ? ErrorCode.WRN_EscapeVariable [](http://example.com/codeflow?start=47&length=45)

Are we testing the `unsafe { }` region case?

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
jjonescz: > was there a reason it was necessary to add the PlaceholderRegion temporarily in this particular case rather than adding it once (and not removing), in the calling `VisitObjectCreationExpressionBase()`?

Yes. When GetValEscape is called on a collection initializer, I need to decide whether it can be returnable or must be scoped to current block. So I temporarily mark the receiver with CallingMethod scope to see if that is ref safe - but it might not be ref safe (and depending on that, I determine the ValEscape of the collection initializer). Put another way, the receiver scope is just speculative and hence I don't think it should be set permanently in the visitor.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
jjonescz: Similarly to `CollectionExpression_Builder`, I was just verifying that references cannot be captured with collection expressions. Haven't realized the error is unexpected. Should we file an issue for that?

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
jjonescz: Added `unsafe` test. I'm not sure the `WRN_EscapeVariable` here is reachable (at least in the tested scenarios, this is used with discarded diagnostics just for the return value), but at least it's consistent with other similar places (and I imagine it might be reachable with more complex nested scenarios).

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
cston: > Haven't realized the error is unexpected. Should we file an issue for that?

Yes, we should file an issue for the error reported when the `Add(in int x)` parameter is not marked `[UnscopedRef]`.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -4313,6 +4313,247 @@ static void Main()
...
+                        var local = 1;
+                        R r;
+                        r = new R() { local }; // 2
+                    }
+
+                    unsafe R M4()
+                    {
+                        var local = 1;
+                        var r = new R() { local };
+                        return r;
```
cston, line 4390: Shouldn't we report a warning for this case?

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -4313,6 +4313,247 @@ static void Main()
...
+            var source = """
+                using System.Collections;
+                using System.Diagnostics.CodeAnalysis;
+
+                class C
+                {
+                    R M1()
+                    {
+                        var local = 1;
+                        return new R() { local }; // 1
```
RikkiGibson, line 4328: What happens with a collection initializer like `new R() { 1 }`? I would expect the same error to be reported. Please add a test.

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -4455,11 +4455,12 @@ internal SafeContext GetValEscape(BoundExpression expr, SafeContext scopeOfTheCo
...
                 case BoundKind.CollectionInitializerExpression:
                     var colExpr = (BoundCollectionInitializerExpression)expr;
-                    return GetValEscape(colExpr.Initializers, scopeOfTheContainingExpression);
-
-                case BoundKind.CollectionElementInitializer:
-                    var colElement = (BoundCollectionElementInitializer)expr;
-                    return GetValEscape(colElement.Arguments, scopeOfTheContainingExpression);
+                    // If arg mixing fails when the receiver has calling-method scope (i.e., some arguments could escape into the receiver), make the value scoped.
+                    return
+                        !scopeOfTheContainingExpression.IsConvertibleTo(SafeContext.CallingMethod) &&
```
RikkiGibson, line 4460: Why should we predicate this only on convertibility to `CallingMethod` scope? Consider the following:

```cs
RS M(ref int param)
{
    int local = 1;
    return new RS(ref param) { local }; // error expected
}

// ref struct RS, Add([UnscopedRef] in int item), etc.
```

In this case, `scopeOfTheContainingExpression` is `ReturnOnly` (right?), and not convertible to `CallingMethod`. But don't we still need to do a mixing check here?

RikkiGibson: I could also imagine scenarios where `new RS(...)` has some local lifetime, but a collection initializer argument has an even narrower lifetime.

jjonescz: The check `!scopeOfTheContainingExpression.IsConvertibleTo(SafeContext.CallingMethod)` is just an optimization - if it's false, it actually means `scopeOfTheContainingExpression == SafeContext.`, and we can return `SafeContext.CallingMethod` immediately.

I will refactor the code to make that clearer.

> In this case, `scopeOfTheContainingExpression` is `ReturnOnly` (right?), and not convertible to `CallingMethod`. But don't we still need to do a mixing check here?

We do a mixing check there. The `!scopeOfTheContainingExpression.IsConvertibleTo(SafeContext.CallingMethod)` condition is `true` in that case.

However, I'm not sure how that example is different from currently tested scenarios - you added a `ref param` to the ref struct's constructor, but that already goes through different code paths, in this PR I've only changed handling of the collection initializer, i.e., the part after the constructor - ` { local }`.

> I could also imagine scenarios where `new RS(...)` has some local lifetime, but a collection initializer argument has an even narrower lifetime.

I don't follow. Here we are computing ValEscape for the expression `new R() { ... }`, so it does not yet have any lifetime. And we either determine the expression to have ValEscape of `scopeOfTheContainingExpression` or `CallingMethod`. I don't think the collection initializer argument can have narrower lifetime than `scopeOfTheContainingExpression` (that's the narrowest lifetime we are currently at).

cston (approved):

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -4454,12 +4454,22 @@ internal SafeContext GetValEscape(BoundExpression expr, SafeContext scopeOfTheCo
                     return GetValEscapeOfObjectInitializer(initExpr, scopeOfTheContainingExpression);
 
                 case BoundKind.CollectionInitializerExpression:
-                    var colExpr = (BoundCollectionInitializerExpression)expr;
-                    return GetValEscape(colExpr.Initializers, scopeOfTheContainingExpression);
+                    {
+                        if (scopeOfTheContainingExpression == SafeContext.CallingMethod)
```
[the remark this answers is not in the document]
jjonescz: Actually never, will remove, thanks.

src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -4313,6 +4313,341 @@ static void Main()
...
+
+                    public void Add([UnscopedRef] in int x) { }
+                    IEnumerator IEnumerable.GetEnumerator() => throw null;
+                }
+                """;
+
+            // The `new R() { param }` expression could be inferred to have ReturnOnly scope in M1 (then there would be no errors),
+            // however the ref safety analysis is currently more conservative and only infers either CallingMethod or local scope for the expression.
+            // It is equivalent to what would happen for `var r = new R(); r.Add(param);` where the scope of `r` is also either CallingMethod or local (if `scoped` keyword is used).
+            // Similarly `new R(param) { param }` in M2 is equivalent to `var r = new R(param); r.Add(param);` so the scope of `r` is ReturnOnly.
```
jjonescz, line 4507: Made a change to make it work as described in this comment. An offline discussion with @RikkiGibson led to me to add this test case and I *think* this is how it should behave?

jjonescz: Talked with Rikki offline, I will think about a more robust approach. Basically now we are doing arg mixing check inside GetValEscape which is sort of a hack (we normally never to CheckValEscape inside GetValEscape) but it's simple to implement and probably can be considered good enough for this edge case scenario. The alternative approach would be to extract some sort of GetInvocationEscape utility that would analyze the invocation arguments and tell us what can escape into the receiver (now similar utilities tell us only what escapes into the return value). We could then hopefully use this also for similar bugs like https://github.com/dotnet/roslyn/issues/63306 and it wouldn't feel so hacky.

RikkiGibson (commented):
Per offline discussion I think it is worth a spike to see if we can allow the initializer arguments to contribute more precisely to the resulting safe-context of the object-creation. We think the solution to this may generalize to some interpolated string bugs, etc.

PR #76253, opened 2024-12-04
Fix reproducibility of local builds

jaredpar:
The `test-determinism.cmd` run was failing on my local machine. After a lot of debugging I was able to track it down to a bug in [xliff][xliff].

The full details are in the bug. The summary though is there is a race condition with how the manifest resource name is generated for RESX files when doing local builds in arcade. To work around this I'm just adding an explicit manifest resource name so that the race doesn't come into play.

The race only occurs when the RESX is outside the directory cone of the project that is including it hence only needed to update this file.

[xliff]: https://github.com/dotnet/arcade/issues/15296

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ UnitTestResults.html
 *.nuget.props
 *.nuget.targets
 project.lock.json
-msbuild.binlog
+*.binlog
 *.project.lock.json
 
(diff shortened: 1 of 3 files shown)


jaredpar:

JoeRobich (approved):

PR #76257, opened 2024-12-04
Dictionary expressions: initial binding and lowering support

cston:
See [proposals/dictionary-expressions.md](https://github.com/dotnet/csharplang/blob/main/proposals/dictionary-expressions.md)

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
@@ -193,17 +193,46 @@ namespace Microsoft.CodeAnalysis.CSharp
                     return Conversion.NoConversion;
                 }
 
-                if (elements.Length > 0 &&
+                if ((!ReferenceEquals(elementType.OriginalDefinition, Compilation.GetWellKnownType(WellKnownType.System_Collections_Generic_KeyValuePair_KV)) ||
+                    !_binder.HasCollectionExpressionApplicableIndexer(syntax, targetType, elementType, out _, BindingDiagnosticBag.Discarded)) &&
+                    elements.Length > 0 &&
                     !_binder.HasCollectionExpressionApplicableAddMethod(syntax, targetType, addMethods: out _, BindingDiagnosticBag.Discarded))
                 {
                     return Conversion.NoConversion;
                 }
             }
 
+            TypeWithAnnotations keyType;
+            TypeWithAnnotations valueType;
+            bool usesKeyValuePairs = IsKeyValuePairType(Compilation, elementType, WellKnownType.System_Collections_Generic_KeyValuePair_KV, out keyType, out valueType);
+
             var builder = ArrayBuilder<Conversion>.GetInstance(elements.Length);
             foreach (var element in elements)
             {
-                Conversion elementConversion = convertElement(element, elementType, ref useSiteInfo);
+                Conversion elementConversion;
+                if (usesKeyValuePairs)
+                {
+                    // PROTOTYPE: This change in behavior should only be enabled in C#14. Test both
+                    // dictionaries and other collections of KeyValuePair<,> with earlier language version.
+                    Debug.Assert(keyType.Type is { });
+                    Debug.Assert(valueType.Type is { });
+                    elementConversion = element switch
+                    {
+                        BoundCollectionExpressionSpreadElement spreadElement => GetSpreadElementAsKeyValuePairConversion(spreadElement, keyType.Type, valueType.Type, ref useSiteInfo),
+                        BoundKeyValuePairElement keyValuePairElement => GetKeyValuePairElementConversion(keyValuePairElement, keyType.Type, valueType.Type, ref useSiteInfo),
+                        _ => GetExpressionElementAsKeyValuePairConversion((BoundExpression)element, keyType.Type, valueType.Type, ref useSiteInfo),
+                    };
+                }
+                else
+                {
+                    elementConversion = element switch
+                    {
+                        BoundCollectionExpressionSpreadElement spreadElement => GetCollectionExpressionSpreadElementConversion(spreadElement, elementType, ref useSiteInfo),
+                        BoundKeyValuePairElement keyValuePairElement => Conversion.NoConversion, // PROTOTYPE: Test this case.
+                        _ => ClassifyImplicitConversionFromExpression((BoundExpression)element, elementType, ref useSiteInfo),
+                    };
+                }
+
                 if (!elementConversion.Exists)
                 {
                     builder.Free();
@@ -214,15 +243,6 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
 
             return Conversion.CreateCollectionExpressionConversion(collectionTypeKind, elementType, constructor, isExpanded, builder.ToImmutableAndFree());
-
-            Conversion convertElement(BoundNode element, TypeSymbol elementType, ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
-            {
-                return element switch
-                {
-                    BoundCollectionExpressionSpreadElement spreadElement => GetCollectionExpressionSpreadElementConversion(spreadElement, elementType, ref useSiteInfo),
-                    _ => ClassifyImplicitConversionFromExpression((BoundExpression)element, elementType, ref useSiteInfo),
-                };
-            }
         }
 
         internal Conversion GetCollectionExpressionSpreadElementConversion(
@@ -235,10 +255,64 @@ namespace Microsoft.CodeAnalysis.CSharp
             {
                 return Conversion.NoConversion;
             }
+            // PROTOTYPE: This should be conversion from type rather than conversion
+            // from expression. The difference is in handling of dynamic.
             return ClassifyImplicitConversionFromExpression(
                 new BoundValuePlaceholder(element.Syntax, enumeratorInfo.ElementType),
                 targetType,
                 ref useSiteInfo);
+        }
+
+        private Conversion GetExpressionElementAsKeyValuePairConversion(
+            BoundExpression element,
+            TypeSymbol keyType,
+            TypeSymbol valueType,
+            ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+        {
+            // PROTOTYPE: Test expression element with no type.
+            if (IsKeyValuePairType(Compilation, element.Type, WellKnownType.System_Collections_Generic_KeyValuePair_KV, out var elementKeyType, out var elementValueType))
+            {
+                var key = ClassifyImplicitConversionFromType(elementKeyType.Type, keyType, ref useSiteInfo);
+                var value = ClassifyImplicitConversionFromType(elementValueType.Type, valueType, ref useSiteInfo);
+                if (key.Exists && value.Exists)
+                {
+                    return new Conversion(ConversionKind.KeyValuePair, [key, value]);
+                }
+            }
+            return Conversion.NoConversion;
+        }
+
+        private Conversion GetSpreadElementAsKeyValuePairConversion(
+            BoundCollectionExpressionSpreadElement element,
+            TypeSymbol keyType,
+            TypeSymbol valueType,
+            ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+        {
+            var enumeratorInfo = element.EnumeratorInfoOpt;
+            if (enumeratorInfo is null)
+            {
+                return Conversion.NoConversion;
+            }
+            return GetExpressionElementAsKeyValuePairConversion(
+                new BoundValuePlaceholder(element.Syntax, enumeratorInfo.ElementType),
+                keyType,
+                valueType,
+                ref useSiteInfo);
+        }
+
+        private Conversion GetKeyValuePairElementConversion(
+            BoundKeyValuePairElement element,
+            TypeSymbol keyType,
+            TypeSymbol valueType,
+            ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+        {
+            var key = ClassifyImplicitConversionFromExpression(element.Key, keyType, ref useSiteInfo);
+            var value = ClassifyImplicitConversionFromExpression(element.Value, valueType, ref useSiteInfo);
+            if (key.Exists && value.Exists)
+            {
+                return new Conversion(ConversionKind.KeyValuePair, [key, value]);
+            }
+            return Conversion.NoConversion;
         }
 #nullable disable
 
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
@@ -1693,6 +1693,10 @@ namespace Microsoft.CodeAnalysis.CSharp
             {
                 return CollectionExpressionTypeKind.ArrayInterface;
             }
+            else if (isDictionaryInterface(compilation, destination, out elementType))
+            {
+                return CollectionExpressionTypeKind.DictionaryInterface;
+            }
 
             elementType = default;
             return CollectionExpressionTypeKind.None;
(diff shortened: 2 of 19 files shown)


src/Compilers/Core/Portable/WellKnownMembers.cs
```diff
@@ -5170,6 +5170,36 @@ static WellKnownMembers()
...
+
+                // System_Collections_Generic_Dictionary_KV__ctor
+                (byte)MemberFlags.Constructor,                                                                                               // Flags
+                (byte)WellKnownType.ExtSentinel, (byte)(WellKnownType.System_Collections_Generic_Dictionary_KV - WellKnownType.ExtSentinel),  // DeclaringTypeId
+                0,                                                                                                          // Arity
+                    0,                                                                                                      // Method Signature
+                    (byte)SignatureTypeCode.TypeHandle, (byte)SpecialType.System_Void,
+
+                // System_Collections_Generic_Dictionary_KV__Add
+                (byte)MemberFlags.Method,                                                           // Flags
```
jcouv, line 5182: >// Flags [](http://example.com/codeflow?start=100&length=8)

nit: can we improve the indentation just a bit?

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs
```diff
@@ -1716,6 +1734,35 @@ internal static bool IsSpanOrListType(CSharpCompilation compilation, TypeSymbol
...
+                keyType = typeArguments[0];
+                valueType = typeArguments[1];
+                return true;
+            }
+            keyType = default;
+            valueType = default;
+            return false;
+        }
+
+        internal static bool CollectionUsesKeyValuePairs(CSharpCompilation compilation, CollectionExpressionTypeKind collectionTypeKind, TypeSymbol elementType, out TypeWithAnnotations keyType, out TypeWithAnnotations valueType)
```
jcouv, line 1753: >CollectionUsesKeyValuePairs [](http://example.com/codeflow?start=29&length=27)

Since a dictionary type is defined as a collection type whose element type is a KVP, we could name this helper `IsDictionaryType`

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
```diff
@@ -200,29 +200,53 @@ protected override Conversion GetCollectionExpressionConversion(
...
             var builder = ArrayBuilder<Conversion>.GetInstance(elements.Length);
             foreach (var element in elements)
             {
-                Conversion elementConversion = convertElement(element, elementType, ref useSiteInfo);
-                if (!elementConversion.Exists)
+                switch (element)
                 {
-                    builder.Free();
-                    return Conversion.NoConversion;
+                    case BoundCollectionExpressionSpreadElement spreadElement:
```
jcouv, line 209: I expected both the spread case and the expression element case to also be updated with new rules (including for C# 13?) since the element type is a KVP.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -450,6 +452,23 @@ SpecialType.System_Collections_Generic_IReadOnlyCollection_T or
...
+            Debug.Assert(node.Type is NamedTypeSymbol);
+            Debug.Assert(node.CollectionCreation is null);
+            Debug.Assert(node.Placeholder is null);
+
+            var collectionType = (NamedTypeSymbol)node.Type;
+            var typeArguments = collectionType.TypeArgumentsWithAnnotationsNoUseSiteDiagnostics;
+            var elements = node.Elements;
+            // PROTOTYPE: Create a custom interface implementation for IReadOnlyDictionary<K, V> to enforce immutability?
+            var collection = CreateAndPopulateDictionary(node, typeArguments, elements);
+            return _factory.Convert(collectionType, collection);
```
jcouv, line 469: Would this handle the situation where I use a custom `Dictionary` type that doesn't implement `IDictionary` interface? We may not need to address if we're planning to eliminate usage of `Dictionary` though

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
```diff
@@ -200,29 +200,53 @@ protected override Conversion GetCollectionExpressionConversion(
...
+                        {
+                            var elementConversion = ClassifyImplicitConversionFromExpression((BoundExpression)element, elementType, ref useSiteInfo);
+                            if (elementConversion.Exists)
+                            {
+                                builder.Add(elementConversion);
+                                continue;
+                            }
+                        }
+                        break;
                 }
```
jcouv, line 244: nit: `CreateCollectionExpressionConversion` used below takes "elementConversions", but we're storing key and value conversions in there too. Consider renaming the parameter to something more generic, such as "nestedConversions". This would likely flow into `CollectionExpressionUncommonData` too.

src/Compilers/CSharp/Test/Emit3/Semantics/DictionaryExpressionTests.cs
```diff
@@ -0,0 +1,616 @@
...
+                Diagnostic(ErrorCode.WRN_ConvertingLock, "x").WithLocation(9, 23),
+                // (10,42): warning CS9216: A value of type 'System.Threading.Lock' converted to a different type will use likely unintended monitor-based locking in 'lock' statement.
+                //         IDictionary<object, object> d = [x:1, 2:y];
+                Diagnostic(ErrorCode.WRN_ConvertingLock, "x").WithLocation(10, 42),
+                // (10,49): warning CS9216: A value of type 'System.Threading.Lock' converted to a different type will use likely unintended monitor-based locking in 'lock' statement.
+                //         IDictionary<object, object> d = [x:1, 2:y];
+                Diagnostic(ErrorCode.WRN_ConvertingLock, "y").WithLocation(10, 49));
+        }
+    }
+}
```
jcouv, line 616: From offline discussion, some scenarios that'll be interesting (although don't have to be handled in this PR):
- `IDictionary<object, object> d = [kvpOfIntInt, ..spreadOfIntInt, intKey: intValue];`
- `IEnumerable<MyKVP<int, int>> c = [new KeyValuePair<int, int>(), kvpOfIntInt, ..spreadOfIntInt];` where there is an implicit conversion from `KeyValuePair<int, int>` to `MyKeyValuePair<int, int>`.
- `IDictionary<int, int> d = [new MyKeyValuePair<int, int>()];` (I assume this one will just fail, by design)
- `IDictionary<int, int> d = [1: 42, 1: 43];` I'm unclear from spec whether we expect this to throw (current implementation) or overwrite. Did we settle that? That will affect whether we should use `Dictionary.Add` or the indexer on `Dictionary`

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -1156,6 +1175,106 @@ private BoundExpression CreateAndPopulateList(BoundCollectionExpression node, Ty
...
+            var constructor = ((MethodSymbol)_factory.WellKnownMember(WellKnownMember.System_Collections_Generic_Dictionary_KV__ctor)).AsMember(collectionType);
+            BoundObjectCreationExpression rewrittenReceiver = _factory.New(constructor, ImmutableArray<BoundExpression>.Empty);
+
+            // Create a temp for the dictionary.
+            BoundAssignmentOperator assignmentToTemp;
+            BoundLocal dictionaryTemp = _factory.StoreToTemp(rewrittenReceiver, out assignmentToTemp);
+            localsBuilder.Add(dictionaryTemp);
+            sideEffects.Add(assignmentToTemp);
+
+            var addMethod = _factory.WellKnownMethod(WellKnownMember.System_Collections_Generic_Dictionary_KV__Add).AsMember(collectionType);
```
jcouv, line 1201: It's not obvious whether we should use the `Add` method or the indexer.
The "Construction" section of the spec says: "If Eᵢ is a key value pair element Kᵢ:Vᵢ, first Kᵢ is evaluated, then Vᵢ is evaluated, and the applicable indexer is invoked on the dictionary instance with the converted values of Kᵢ and Vᵢ."
Also, this affects the behavior for duplicate entries, as `Add` throws, but the indexer overwrites.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -1156,6 +1175,106 @@ private BoundExpression CreateAndPopulateList(BoundCollectionExpression node, Ty
...
+                throw ExceptionUtilities.UnexpectedValue(expression.Type);
+            }
+
+            BoundAssignmentOperator assignmentToTemp;
+            BoundLocal expressionTemp = _factory.StoreToTemp(VisitExpression(expression), out assignmentToTemp);
+            localsBuilder.Add(expressionTemp);
+            sideEffects.Add(assignmentToTemp);
+            var getKeyMethod = ((MethodSymbol)_factory.WellKnownMember(WellKnownMember.System_Collections_Generic_KeyValuePair_KV__get_Key)).AsMember(sourceType);
+            var getValueMethod = ((MethodSymbol)_factory.WellKnownMember(WellKnownMember.System_Collections_Generic_KeyValuePair_KV__get_Value)).AsMember(sourceType);
+            return (_factory.Convert(addMethod.Parameters[0].Type, _factory.Call(expressionTemp, getKeyMethod)),
```
jcouv, line 1274: This seems wrong. We're making brand new conversions here and they are not the same that we checked and calculated during initial binding (those were attempts to convert the whole element, but here we're converting the key and the value separately)

src/Compilers/CSharp/Test/Emit3/Semantics/DictionaryExpressionTests.cs
```diff
@@ -0,0 +1,616 @@
...
+                    }
+                    builder.Append("]");
+                    Console.Write(builder.ToString());
+                }
+            }
+            """;
+
+        [Theory]
+        [InlineData(LanguageVersion.CSharp13)]
+        [InlineData(LanguageVersion.Preview)]
```
jcouv, line 44: Please add `Next` (which represents CSharp14 before it is declared)

src/Compilers/CSharp/Test/Emit3/Semantics/DictionaryExpressionTests.cs
```diff
@@ -0,0 +1,616 @@
...
+                d = [1:"one"];
+                d = [new KeyValuePair<int, string>(2, "two")];
+                d = [.. new KeyValuePair<int, string>[] { new(3, "three") }];
+                """;
+
+            var comp = CreateCompilation(source);
+            comp.VerifyEmitDiagnostics();
+
+            comp = CreateCompilation(source);
+            comp.MakeMemberMissing(WellKnownMember.System_Collections_Generic_KeyValuePair_KV__get_Key);
```
jcouv, line 309: Do we have a test where the KeyValuePair type is missing?

src/Compilers/CSharp/Test/Emit3/Semantics/DictionaryExpressionTests.cs
```diff
@@ -0,0 +1,616 @@
...
+                }
+                """);
+        }
+
+        [Fact]
+        public void InferredType_ExpressionElement()
+        {
+            string source = """
+                using System.Collections.Generic;
+                IDictionary<string, int> d = [default, new()];
```
jcouv, line 475: What are the inferred types? Same question below (in `InferredType_KeyValueElement`)
Consider executing and reporting the contents
Also, it's not blocking for this PR, but at some point we'll want to verify the semantic model (types and conversions) on the elements in these new scenario

src/Compilers/CSharp/Test/Emit3/Semantics/DictionaryExpressionTests.cs
```diff
@@ -0,0 +1,616 @@
...
+
+        [Theory]
+        [InlineData("IDictionary<int, string>")]
+        [InlineData("IReadOnlyDictionary<int, string>")]
+        public void KeyValuePair_MissingMember(string typeName)
+        {
+            string source = $$"""
+                using System.Collections.Generic;
+                {{typeName}} d;
+                d = [];
```
jcouv, line 299: Do we have a successful test with empty collection expression?

jcouv (commented):
Done with review pass (iteration 11)

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
cston: Implicit item-wise conversions from expression of `KeyValuePair<K1, V1>` type to distinct `KeyValuePair<K2, V2>` type is not included in this PR.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_CollectionExpression.cs
```diff
@@ -1156,6 +1175,106 @@ private BoundExpression CreateAndPopulateList(BoundCollectionExpression node, Ty
...
+        /// Create and populate a dictionary from a collection expression.
+        /// The collection may or may not have a known length.
+        /// </summary>
+        private BoundExpression CreateAndPopulateDictionary(BoundCollectionExpression node, ImmutableArray<TypeWithAnnotations> typeArguments, ImmutableArray<BoundNode> elements)
+        {
+            Debug.Assert(!_inExpressionLambda);
+
+            var collectionType = _factory.WellKnownType(WellKnownType.System_Collections_Generic_Dictionary_KV).Construct(typeArguments);
+
+            var localsBuilder = ArrayBuilder<BoundLocal>.GetInstance();
```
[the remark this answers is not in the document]
cston: I'll leave this as `BoundLocal` since it matches other methods in this file that track, and it's likely we'll need `BoundLocal` instances when we start relying on helper methods such as `RewriteCollectionExpressionElementsIntoTemporaries`.

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
cston: I'll leave this as "elementConversions" for now since we might need to handle key-value pair conversions as grouped `Conversion` instances.

src/Compilers/CSharp/Test/Emit3/Semantics/DictionaryExpressionTests.cs
jcouv: Was this resolved?

jcouv (approved):
LGTM Thanks (iteration 14)

src/Compilers/CSharp/Test/Emit3/Semantics/DictionaryExpressionTests.cs
cston: Yes, added test `DictionaryInterface_01`.

jcouv: Ah, I searched for `= []` so missed it. Thanks

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
```diff
@@ -235,6 +259,8 @@ internal Conversion GetCollectionExpressionSpreadElementConversion(
             {
                 return Conversion.NoConversion;
             }
+            // This should be conversion from type rather than conversion
```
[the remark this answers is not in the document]
cston: Using conversion from expression rather than conversion from type is a bug. The one case that I'm aware of where this is observable is when the spread element type is `dynamic`. Unfortunately, fixing this would be a breaking change for that case.

For instance, the following is currently allowed but would be an error when using conversion from type:
```csharp
dynamic[] x = [1, 2, 3];
int[] y = [..x];
```

I could log a bug and reference that here if that seems useful.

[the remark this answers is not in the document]
cston: I've updated the comment to include the example.

src/Compilers/CSharp/Test/Emit3/Semantics/DictionaryExpressionTests.cs
```diff
@@ -0,0 +1,1027 @@
...
+                }
+                """;
+            var comp = CreateCompilation([source, s_dictionaryExtensions]);
+            comp.VerifyEmitDiagnostics(
+                // (12,16): error CS9215: Collection expression type 'Dictionary<int, string>' must have an instance or extension method 'Add' that can be called with a single argument.
+                //         return [1:"one", x, ..y];
+                Diagnostic(ErrorCode.ERR_CollectionExpressionMissingAdd, @"[1:""one"", x, ..y]").WithArguments("System.Collections.Generic.Dictionary<int, string>").WithLocation(12, 16),
+                // (12,17): error CS9268: Collection expression type 'Dictionary<int, string>' does not support key-value pair elements.
+                //         return [1:"one", x, ..y];
+                Diagnostic(ErrorCode.ERR_CollectionExpressionKeyValuePairNotSupported, @"1:""one""").WithArguments("System.Collections.Generic.Dictionary<int, string>").WithLocation(12, 17));
```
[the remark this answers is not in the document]
cston: This PR implements support for *dictionary interface* target types only. It does not include support for concrete types such as `Dictionary<K, V>`.

PR #76259, opened 2024-12-04
Use named pipes to connect to the BuildHost process.

JoeRobich:
Suggested in https://github.com/dotnet/roslyn/issues/75967#issuecomment-2515889705


src/Workspaces/Core/MSBuild/MSBuild/BuildHostProcessManager.cs
```diff
@@ -350,6 +355,11 @@ public enum BuildHostProcessKind
 
     private sealed class BuildHostProcess : IAsyncDisposable
     {
+        /// <summary>
+        /// The time to wait for a named pipe connection to complete for a newly started server
+        /// </summary>
+        internal const int TimeOutMsNewProcess = 20_000;
```
dibarbet, line 361: Might be worth making this a bit longer, maybe a minute.  We've seen issues with 30s on the VSCode side, where it takes just a tiny bit longer (especially on slow machines, like CI).

dibarbet (approved):
nice - thanks for this!

Issue #76100, opened 2024-11-26
UnscopedRef can be different in interface implementation, leading to ref safety holes


jjonescz:
**Version Used**: 20241126.6 ([b23da04](https://github.com/dotnet/roslyn/commit/b23da0417d19a84b2032f886481ba0bb3c5def38))

**Steps to Reproduce**:

```cs
using System;
using System.Diagnostics.CodeAnalysis;

var r = F();
Console.WriteLine(r.F.ToString());

static R F()
{
    var r = new R();
    ((I)new C()).M(ref r);
    return r;
}

interface I
{
    void M(ref R r);
}

class C : I
{
    public void M([UnscopedRef] ref R r)
    {
        r.F = ref r.I;
    }
}

ref struct R
{
    public int I;
    public ref int F;
}
```

**Expected Behavior**: An error at `((I)new C()).M(ref r);` - an error would be reported there if replaced with `new C().M(ref r);`.

**Actual Behavior**: No errors. The program accesses an invalid reference at runtime.

Sergio0694:
Just so I understand, the problem is just because the implementation is narrowing the escape scope, right? Ie. if you had:

```csharp
interface I
{
    void M([UnscopedRef] ref R r);
}

class C : I
{
    public void M(ref R r)
    {
    }
}
```

That should be fine, right? Would that still be allowed?

jjonescz:
> That should be fine, right? Would that still be allowed?

Yes, that should be fine. Per [the spec](https://github.com/dotnet/csharplang/blob/56dae0a6e0efd3479fe55cd8eb6d6c2c0a53f4aa/proposals/csharp-11.0/low-level-struct-improvements.md#parameter-scope-variance), removing `[UnscopedRef]` in the interface implementation is allowed.

jjonescz:
Also this section of the spec needs to be updated - it does not consider the "silly cyclic self-assignment":

> The compiler will report a diagnostic for _unsafe scoped mismatches_ across overrides, interface implementations, and delegate conversions when:
> - The method returns a `ref struct` or returns a `ref` or `ref readonly`, or the method has a `ref` or `out` parameter of `ref struct` type, and
> - The method has at least one additional `ref`, `in`, or `out` parameter, or a parameter of `ref struct` type.

jjonescz:
Reopening as the PR was reverted.

PR #76261, opened 2024-12-04
Consider silly cyclic assignment in scoped variance

jjonescz:
Speclet update: https://github.com/dotnet/csharplang/pull/8766
Fixes https://github.com/dotnet/roslyn/issues/76100.

### DIFF
--- a/src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
+++ b/src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
@@ -10358,5 +10358,278 @@ public struct Vec4
                 """;
             CreateCompilation(source, targetFramework: TargetFramework.Net70).VerifyDiagnostics();
         }
+
+        [Theory, CombinatorialData, WorkItem("https://github.com/dotnet/roslyn/issues/76100")]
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Test/Semantic/Semantics/RefEscapingTests.cs
```diff
@@ -10358,5 +10358,278 @@ public static void M([UnscopedRef] ref S s)
...
+                delegate void D({{scoped1}} {{modifier}} R r);
+
+                static class C
+                {
+                    public static void M({{scoped2}} {{modifier}} R r) { }
+                }
+
+                ref struct R;
+                """;
+            CreateCompilation(source).VerifyDiagnostics();
```
jaredpar, line 10632: Do we have tests like this? 

```csharp
interface I1 {
  ref int M(); 
}
struct S {
  int field;
  [UnscopedRef] ref int M() => ref field;
}
```

jjonescz: Yes, `RefFieldTests.UnscopedRefAttribute_InterfaceImplementation_01`.

cston (approved):

jaredpar (approved):

PR #76273, opened 2024-12-05
Update SystemTextJsonVersion

WeihanLi:
fix vulnerability nuget audit

image

msbuild already using the `8.0.5` version

image

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -61,7 +61,7 @@
     <SystemSecurityPermissionsVersion>8.0.0</SystemSecurityPermissionsVersion>
     <!-- Note: When updating SystemTextJsonVersion ensure that the version is no higher than what is used by MSBuild. -->
     <SystemTextEncodingsWebVersion>8.0.0</SystemTextEncodingsWebVersion>
-    <SystemTextJsonVersion>8.0.4</SystemTextJsonVersion>
+    <SystemTextJsonVersion>8.0.5</SystemTextJsonVersion>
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -61,7 +61,7 @@
     <SystemSecurityPermissionsVersion>8.0.0</SystemSecurityPermissionsVersion>
     <!-- Note: When updating SystemTextJsonVersion ensure that the version is no higher than what is used by MSBuild. -->
```
jjonescz, line 62: > msbuild already using the 8.0.5 version

msbuild `main` is using STJ 8.0.5, but isn't this comment trying to say that the msbuild which *roslyn is referencing* should be using STJ 8.0.5? (and I'm not sure if that's the case)

WeihanLi: Nice catch, I'm not sure either.

While from the NuGet package dependencies, seems there's no dependency for `Microsoft.Build` in the package I referenced `Microsoft.CodeAnalysis.CSharp.Features`
Think maybe it's related to the MSBuild version in VS/VSCode, and from the latest `Microsoft.Build` version released last month, it's also using the `8.0.5` System.Text.Json`

JoeRobich: If I recall, it has to do with VS insertions. At one point we tried to insert with a version higher than MSBuild and it caused NGen errors. We should be fine to take this change.

JoeRobich (approved):

PR #76276, opened 2024-12-05
Cache the MEF composition in the Roslyn LSP.

### LINKED ISSUE #68568: Cache vs-mef composition to improve startup perf
We can cache the VS-MEF composition to improve startup perf of the language server in some instances. 

We will need to check to make sure that the inputs have not changed before we use the cache.  It is not enough to check things like extension version as we have settings and 3rd party extensions to the server which can change the mef cache.

Some things to check
1.  The main Roslyn server dlls (could change if the server path changes for example)
2.  The additional first party components change (xaml tools, Razor sg, devkit components).  These can also change via a vscode setting.
3.  Third party component extensions (e.g. compiler dev SDK).  These can change based on settings, extensions, etc etc.
4.  .NET runtime changes (generally it should mainly be the major version, eg. `System.Runtime, 8.0.0.0` vs `System.Runtime, 9.0.0.0`)

Perhaps the easiest way is to cache some kind of assembly checksum of all assemblies that contribute + the .net runtime info.

JoeRobich:
To improve startup time of the LSP we can cache the MEF composition and load it from file instead of building up a new composition.

The cache is invalidated when:
1. The location of the Microsoft.CodeAnalysis.LanguageServer changes (such as installing an updated C# extension or `dotnet.server.path` is updated).
2. Changing major .NET runtime version.
3. The set of assembly paths that make up the composition changes.
4. The last write time of any assembly that is part of the composition changes.

Resolves #68568

### DIFF
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/ExportProviderBuilderTests.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/ExportProviderBuilderTests.cs
@@ -14,9 +14,9 @@ namespace Microsoft.CodeAnalysis.LanguageServer.UnitTests
         {
             await using var testServer = await CreateLanguageServerAsync(includeDevKitComponents: false);
 
-            var mefCompositions = Directory.EnumerateFiles(MefCacheDirectory.Path, "*.mef-composition", SearchOption.AllDirectories);
+            await AssertCacheWriteWasAttemptedAsync();
 
-            Assert.Single(mefCompositions);
+            AssertCachedCompositionCountEquals(expectedCount: 1);
         }
 
         [Fact]
@@ -24,12 +24,14 @@ namespace Microsoft.CodeAnalysis.LanguageServer.UnitTests
         {
             await using var testServer = await CreateLanguageServerAsync(includeDevKitComponents: false);
 
+            await AssertCacheWriteWasAttemptedAsync();
+
             // Second test server with the same set of assemblies.
             await using var testServer2 = await CreateLanguageServerAsync(includeDevKitComponents: false);
 
-            var mefCompositions = Directory.EnumerateFiles(MefCacheDirectory.Path, "*.mef-composition", SearchOption.AllDirectories);
+            AssertNoCacheWriteWasAttempted();
 
-            Assert.Single(mefCompositions);
+            AssertCachedCompositionCountEquals(expectedCount: 1);
         }
 
         [Fact]
@@ -37,12 +39,35 @@ namespace Microsoft.CodeAnalysis.LanguageServer.UnitTests
         {
             await using var testServer = await CreateLanguageServerAsync(includeDevKitComponents: false);
 
+            await AssertCacheWriteWasAttemptedAsync();
+
             // Second test server with a different set of assemblies.
             await using var testServer2 = await CreateLanguageServerAsync(includeDevKitComponents: true);
 
+            await AssertCacheWriteWasAttemptedAsync();
+
+            AssertCachedCompositionCountEquals(expectedCount: 2);
+        }
+
+        private async Task AssertCacheWriteWasAttemptedAsync()
+        {
+            var cacheWriteTask = ExportProviderBuilder.TestAccessor.GetCacheWriteTask();
+            Assert.NotNull(cacheWriteTask);
+
+            await cacheWriteTask;
+        }
+
+        private void AssertNoCacheWriteWasAttempted()
+        {
+            var cacheWriteTask2 = ExportProviderBuilder.TestAccessor.GetCacheWriteTask();
+            Assert.Null(cacheWriteTask2);
+        }
+
+        private void AssertCachedCompositionCountEquals(int expectedCount)
+        {
             var mefCompositions = Directory.EnumerateFiles(MefCacheDirectory.Path, "*.mef-composition", SearchOption.AllDirectories);
 
-            Assert.Equal(2, mefCompositions.Count());
+            Assert.Equal(expectedCount, mefCompositions.Count());
         }
     }
 }
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/ExportProviderBuilderTests.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/ExportProviderBuilderTests.cs
@@ -1,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Xunit.Abstractions;
+
+namespace Microsoft.CodeAnalysis.LanguageServer.UnitTests
+{
+    public sealed class ExportProviderBuilderTests(ITestOutputHelper testOutputHelper)
+        : AbstractLanguageServerHostTests(testOutputHelper)
+    {
+        [Fact]
+        public async Task MefCompositionIsCached()
+        {
+            await using var testServer = await CreateLanguageServerAsync(includeDevKitComponents: false);
+
+            var mefCompositions = Directory.EnumerateFiles(MefCacheDirectory.Path, "*.mef-composition", SearchOption.AllDirectories);
+
+            Assert.Single(mefCompositions);
+        }
+
+        [Fact]
+        public async Task MefCompositionIsReused()
+        {
+            await using var testServer = await CreateLanguageServerAsync(includeDevKitComponents: false);
+
+            // Second test server with the same set of assemblies.
+            await using var testServer2 = await CreateLanguageServerAsync(includeDevKitComponents: false);
+
+            var mefCompositions = Directory.EnumerateFiles(MefCacheDirectory.Path, "*.mef-composition", SearchOption.AllDirectories);
+
+            Assert.Single(mefCompositions);
+        }
+
+        [Fact]
+        public async Task MultipleMefCompositionsAreCached()
+        {
+            await using var testServer = await CreateLanguageServerAsync(includeDevKitComponents: false);
+
+            // Second test server with a different set of assemblies.
+            await using var testServer2 = await CreateLanguageServerAsync(includeDevKitComponents: true);
+
+            var mefCompositions = Directory.EnumerateFiles(MefCacheDirectory.Path, "*.mef-composition", SearchOption.AllDirectories);
+
+            Assert.Equal(2, mefCompositions.Count());
+        }
+    }
+}
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/Utilities/LanguageServerTestComposition.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/Utilities/LanguageServerTestComposition.cs
@@ -39,6 +39,6 @@ internal sealed class LanguageServerTestComposition
             ExtensionLogDirectory: string.Empty);
         var extensionManager = ExtensionAssemblyManager.Create(serverConfiguration, loggerFactory);
         assemblyLoader = new CustomExportAssemblyLoader(extensionManager, loggerFactory);
-        return ExportProviderBuilder.CreateExportProviderAsync(extensionManager, assemblyLoader, devKitDependencyPath, loggerFactory);
+        return ExportProviderBuilder.CreateExportProviderAsync(extensionManager, assemblyLoader, devKitDependencyPath, cacheDirectory: null, loggerFactory);
     }
 }
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/Utilities/LanguageServerTestComposition.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/Utilities/LanguageServerTestComposition.cs
@@ -23,6 +23,7 @@ internal sealed class LanguageServerTestComposition
     public static Task<ExportProvider> CreateExportProviderAsync(
         ILoggerFactory loggerFactory,
         bool includeDevKitComponents,
+        string cacheDirectory,
         out ServerConfiguration serverConfiguration,
         out IAssemblyLoader assemblyLoader)
     {
@@ -39,6 +40,6 @@ internal sealed class LanguageServerTestComposition
             ExtensionLogDirectory: string.Empty);
         var extensionManager = ExtensionAssemblyManager.Create(serverConfiguration, loggerFactory);
         assemblyLoader = new CustomExportAssemblyLoader(extensionManager, loggerFactory);
-        return ExportProviderBuilder.CreateExportProviderAsync(extensionManager, assemblyLoader, devKitDependencyPath, cacheDirectory: null, loggerFactory);
+        return ExportProviderBuilder.CreateExportProviderAsync(extensionManager, assemblyLoader, devKitDependencyPath, cacheDirectory, loggerFactory);
     }
 }
--- a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
+++ b/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
@@ -120,9 +120,12 @@ internal sealed class ExportProviderBuilder
 
     private static string GetCompositionCacheFilePath(string cacheDirectory, ImmutableArray<string> assemblyPaths)
     {
-        // Include the .NET runtime version in the cache path so that running on a newer
-        // runtime causes the cache to be rebuilt.
+        // This should vary based on .NET runtime major version so that as some of our processes switch between our target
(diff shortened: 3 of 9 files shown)


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -59,20 +105,52 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+    {
+        // Include the .NET runtime version in the cache path so that running on a newer
+        // runtime causes the cache to be rebuilt.
+        var cacheSubdirectory = $".NET {Environment.Version.Major}";
+        return Path.Combine(cacheDirectory, cacheSubdirectory, $"c#-languageserver.{ComputeAssemblyHash(assemblyPaths)}.mef-composition");
+
+        static string ComputeAssemblyHash(ImmutableArray<string> assemblyPaths)
+        {
+            var assemblies = new StringBuilder();
+            foreach (var assemblyPath in assemblyPaths)
```
ToddGrun, line 140: >assemblyPaths [](http://example.com/codeflow?start=41&length=13)

Is the input array is guaranteed to come through in the same order? Would it make sense to sort them to guarantee this?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -59,20 +105,52 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+            {
+                // Include assembly path in the hash so that changes to the set of included
+                // assemblies cause the composition to be rebuilt.
+                assemblies.Append(assemblyPath);
+                // Include the last write time in the hash so that newer assemblies written
+                // to the same location cause the composition to be rebuilt.
+                assemblies.Append(File.GetLastWriteTimeUtc(assemblyPath).ToString("F"));
+            }
+
+            var hash = SHA256.HashData(Encoding.UTF8.GetBytes(assemblies.ToString()));
```
ToddGrun, line 150: >SHA256 [](http://example.com/codeflow?start=23&length=6)

I don't know much about hashing, pinging @CyrusNajmabadi 

I know that some other code that doesn't need cryptographically strong hashes just uses XxHash128

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -59,20 +105,52 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+            // Convert to filename safe base64 string.
+            return Convert.ToBase64String(hash).Replace('+', '-').Replace('/', '_').TrimEnd('=');
```
ToddGrun, line 152: >Replace [](http://example.com/codeflow?start=48&length=7)

we're pretty confident that these are the only invalid chars? If not, would it be safer to use Path.GetInvalidPathChars and convert those to underscores?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -38,17 +41,60 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+    private static async Task<IExportProviderFactory> GetCompositionConfigurationAsync(
+        ImmutableArray<string> assemblyPaths,
+        IAssemblyLoader assemblyLoader,
+        string? cacheDirectory,
+        ILogger logger)
+    {
         // Create a MEF resolver that can resolve assemblies in the extension contexts.
         var resolver = new Resolver(assemblyLoader);
 
+        string? compositionCacheFile = cacheDirectory is not null
```
dibarbet, line 69: think we should make the cacheDirectory non-nullable - seems like its only tests that pass null atm?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -38,17 +41,60 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+
+                CachedComposition cachedComposition = new();
+                using FileStream cacheStream = new(compositionCacheFile, FileMode.Open, FileAccess.Read, FileShare.Read, 4096, useAsync: true);
+                return await cachedComposition.LoadExportProviderFactoryAsync(cacheStream, resolver);
+            }
+        }
+        catch (Exception ex)
+        {
+            // Log the error, and move on to recover by recreating the MEF composition.
+            logger.LogError($"Loading cached MEF composition failed: {ex}");
```
dibarbet, line 88: we should add telemetry events for all the cases here (used existing cache, existing cache failed, no cache)

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -59,20 +105,52 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+            {
+                Directory.CreateDirectory(directory);
+            }
 
-        // Create an export provider, which represents a unique container of values.
-        // You can create as many of these as you want, but typically an app needs just one.
-        var exportProvider = exportProviderFactory.CreateExportProvider();
+            CachedComposition cachedComposition = new();
+            var tempFilePath = Path.Combine(Path.GetTempPath(), Path.GetTempFileName());
+            using (FileStream cacheStream = new(tempFilePath, FileMode.Create, FileAccess.Write, FileShare.None, 4096, useAsync: true))
```
dibarbet, line 118: named param for 4096

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/Utilities/LanguageServerTestComposition.cs
```diff
@@ -39,6 +39,6 @@ public static Task<ExportProvider> CreateExportProviderAsync(
             ExtensionLogDirectory: string.Empty);
         var extensionManager = ExtensionAssemblyManager.Create(serverConfiguration, loggerFactory);
         assemblyLoader = new CustomExportAssemblyLoader(extensionManager, loggerFactory);
-        return ExportProviderBuilder.CreateExportProviderAsync(extensionManager, assemblyLoader, devKitDependencyPath, loggerFactory);
+        return ExportProviderBuilder.CreateExportProviderAsync(extensionManager, assemblyLoader, devKitDependencyPath, cacheDirectory: null, loggerFactory);
```
dibarbet, line 42: would be good to add some tests here to verify a cache was written, and ideally re-used (might need a test accessor for it)

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
JoeRobich: According to the base64 spec those are the trouble characters. We could bring in the Base64Url by adding a package reference ([docs](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.text.base64url?view=net-9.0). Once we move to .net 9.0 we can drop the package as it will be built-in.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -59,20 +105,52 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+            {
+                await cachedComposition.SaveAsync(config, cacheStream);
+            }
 
-        // Immediately set the logger factory, so that way it'll be available for the rest of the composition
-        exportProvider.GetExportedValue<ServerLoggerFactory>().SetFactory(loggerFactory);
+            File.Move(tempFilePath, compositionCacheFile, overwrite: true);
```
dibarbet, line 123: should the save also be done in a try catch?  We probably don't want to block the language server starting on a failure to save the cache file (should definitely error though)

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
ToddGrun: Interesting, I hadn't realized that + and / were the only non-alphanumeric values in base64. Out of curiosity, which OS has + as a problematic file path char?

JoeRobich: I am not sure. I was trying to mimic [RC 3548](https://www.rfc-editor.org/rfc/rfc3548#page-6) which is why I made those two substitutions.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
JoeRobich: I am sure that will work just fine for this. Thanks!

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -59,20 +105,52 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+        // Prepare an ExportProvider factory based on this graph.
+        return config.CreateExportProviderFactory();
+    }
 
-        return exportProvider;
+    private static string GetCompositionCacheFilePath(string cacheDirectory, ImmutableArray<string> assemblyPaths)
+    {
+        // Include the .NET runtime version in the cache path so that running on a newer
+        // runtime causes the cache to be rebuilt.
+        var cacheSubdirectory = $".NET {Environment.Version.Major}";
```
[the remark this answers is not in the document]
JoeRobich: I am not 100% sure. I took this bit from DevKit and have updated the comment to match theirs. I know we have changed target runtime from 8.0.9 to 8.0.10 and now 8.0.11 seemingly without incident.

JoeRobich:
@ToddGrun @dibarbet Ready for another review.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -59,20 +109,69 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+        }
+    }
+
+    private static async Task WriteCompositionCacheAsync(string compositionCacheFile, CompositionConfiguration config, ILogger logger)
+    {
+        try
+        {
+            if (Path.GetDirectoryName(compositionCacheFile) is string directory)
+            {
+                Directory.CreateDirectory(directory);
```
ToddGrun, line 159: >CreateDirectory [](http://example.com/codeflow?start=26&length=15)

Feels like we should be hopping off the calling thread at some point before we attempt to do any IO

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -38,17 +42,63 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
...
+        {
+            if (File.Exists(compositionCacheFile))
+            {
+                logger.LogTrace($"Loading cached MEF catalog: {compositionCacheFile}");
+
+                CachedComposition cachedComposition = new();
+                using FileStream cacheStream = new(compositionCacheFile, FileMode.Open, FileAccess.Read, FileShare.Read, bufferSize: 4096, useAsync: true);
+                var exportProviderFactory = await cachedComposition.LoadExportProviderFactoryAsync(cacheStream, resolver);
+
+                RoslynLog.Logger.Log(RoslynLog.FunctionId.LSP_MEF_Cache_Load_Success);
```
dibarbet, line 83: talked offline, but this actually doesn't work as-is (since the telemetry session is instantiated after mef creation)

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
```diff
@@ -59,20 +109,69 @@ public static async Task<ExportProvider> CreateExportProviderAsync(
         // Verify we only have expected errors.
         ThrowOnUnexpectedErrors(config, catalog, logger);
 
+        RoslynLog.Logger.Log(RoslynLog.FunctionId.LSP_MEF_Cache_Built);
+
+        // Try to cache the composition.
+        _ = WriteCompositionCacheAsync(compositionCacheFile, config, logger);
```
dibarbet, line 115: should probably have ReportNonFatalError tracking on this async call

dibarbet (approved):
generally looks good - assuming the telemetry is fixed/removed and the async call to save the cache is tracked

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
ToddGrun: Maybe WriteCompositionCacheAsync should only catch the ioexceptions (perhaps via IOUtilities) and if anything crazy happens that is outside the bounds of normal IO exceptions, then the ReportNonFatalError can come through.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/ExportProviderBuilder.cs
JoeRobich: Removed telemetry.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/ExportProviderBuilderTests.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    public sealed class ExportProviderBuilderTests(ITestOutputHelper testOutputHelper)
+        : AbstractLanguageServerHostTests(testOutputHelper)
+    {
+        [Fact]
+        public async Task MefCompositionIsCached()
+        {
+            await using var testServer = await CreateLanguageServerAsync(includeDevKitComponents: false);
+
+            var mefCompositions = Directory.EnumerateFiles(MefCacheDirectory.Path, "*.mef-composition", SearchOption.AllDirectories);
```
ToddGrun, line 17: >var mefCompositions = Directory.EnumerateFiles(MefCacheDirectory.Path, "*.mef-composition", SearchOption.AllDirectories); [](http://example.com/codeflow?start=12&length=121)

Sorry, I'm probably being dense here. Since the saving is asynchronous, how do you know that it's completed before we check for it here?

JoeRobich: ah, thanks. Will fix up.

JoeRobich: @ToddGrun Updated.

ToddGrun: Out of curiosity, I thought the IAsynchronousOperationListener was the way that we usually exposed async operations to tests. Was that not an option?

JoeRobich: I will admit I have not tested a lot of these async operations. Looking through the source it seems the IAsynchronousOperationListenerProvider is typically MEF imported but this code is building the MEF composition, so we couldn't follow that pattern. Nothing is stopping us from creating our own listener instance and making it available through the TestAccessor, but I am not sure that has any benefits over using a Task. If you know of any reasons, I am happy to rework this bit in a follow up.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Program.cs
```diff
@@ -86,7 +86,9 @@ static async Task RunAsync(ServerConfiguration serverConfiguration, Cancellation
     var assemblyLoader = new CustomExportAssemblyLoader(extensionManager, loggerFactory);
     var typeRefResolver = new ExtensionTypeRefResolver(assemblyLoader, loggerFactory);
 
-    using var exportProvider = await ExportProviderBuilder.CreateExportProviderAsync(extensionManager, assemblyLoader, serverConfiguration.DevKitDependencyPath, loggerFactory);
+    var cacheDirectory = Path.Combine(Path.GetDirectoryName(typeof(Program).Assembly.Location)!, "cache");
```
nickodei, line 89: Hey, I just wanted to mention that this breaks read-only filesystem systems like NixOS where this is throwing and Exception because it cant create the folder in the nix-store. It's a bit unfortunate that this path could not be configured:

image

JoeRobich: Thanks @nickodei! I opened https://github.com/dotnet/roslyn/issues/76892 to track this.

PR #76289, opened 2024-12-05
Fix broken strings in Unified Setting page

Cosifne:
Clearly strings are going to be broken if this is merged.

https://github.com/dotnet/roslyn/commit/dfc660c23aecaf5aae4ad2249d8eead228db43d5#diff-a16de2a58a10f14a7fc8512f47bbd467869339a6135046bb0274d2fce1151f93R1901

This PR brings back to track.

Current: 
image
After:
image

This should be covered by unit test. I am preparing it in https://github.com/Cosifne/roslyn/tree/dev/shech/UnifiedSettingsAdvPage, will cover these strings will that is ready

### DIFF
--- a/src/VisualStudio/CSharp/Impl/Options/IntelliSenseOptionPageStrings.cs
+++ b/src/VisualStudio/CSharp/Impl/Options/IntelliSenseOptionPageStrings.cs
@@ -49,7 +49,7 @@ namespace Microsoft.VisualStudio.LanguageServices.CSharp.Options
             => ServicesVSResources.Never_add_new_line_on_enter;
 
         public static string Option_Only_add_new_line_on_enter_with_whole_word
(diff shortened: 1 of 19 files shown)


src/VisualStudio/CSharp/Impl/Options/IntelliSenseOptionPageStrings.cs
```diff
@@ -49,7 +49,7 @@ public static string Option_Never_add_new_line_on_enter
             => ServicesVSResources.Never_add_new_line_on_enter;
 
         public static string Option_Only_add_new_line_on_enter_with_whole_word
-            => ServicesVSResources.Only_add_new_line_on_enter_after_end_of_fully_typed_word;
+            => ServicesVSResources._Only_add_new_line_on_enter_after_end_of_fully_typed_word;
```
Cosifne, line 52: This is for the accelerator  `_`

dibarbet (approved):

Issue #56765, opened 2021-09-27
[LSP] Goto/Peek definition on string literal does not go to definition


dibarbet:
Original bug - https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1327376/

Going to definition on a string should show the string metadata file.  However in the LSP editor, no definition is found.
![def_string](https://user-images.githubusercontent.com/5749229/134971602-d9ab29f4-036b-4c1a-9f5a-809d5102f504.gif)
.

Issue is that no navigable source location is found from the IFindDefinitionService, so we hit this line
https://sourceroslyn.io/#Microsoft.CodeAnalysis.LanguageServer.Protocol/Handler/Definitions/AbstractGoToDefinitionHandler.cs,64

However, `SymbolFinder.FindSymbolAtPositionAsync` eventually calls `semanticInfo.GetAnySymbol(includeType: false)` which excludes the string symbol.

dibarbet:
moving over to @davidwengier to be looked at for gtd parity in 17.3

PR #76291, opened 2024-12-06
Enable LSP go to definition on string literals

dibarbet:
Resolves https://github.com/dotnet/roslyn/issues/56765

gtd_string


JoeRobich:
I feel like this change should be made to `Go to Type Definition` instead. Literals are define at their location and this is instead going to the literal's type.

dibarbet:
> I feel like this change should be made to `Go to Type Definition` instead. Literals are defined at their location and this is instead going to the literal's type.

go to type definition does work for that already (no change needed).  Happy to say someone should use go to type definition instead, but I also don't mind if go to definition also supported this (also makes it work in VS Razor which doesn't have go to type def)

CyrusNajmabadi (approved):

PR #75853, opened 2024-11-11
Prefer array to ReadOnlySpan over Span

jjonescz:
Test plan: https://github.com/dotnet/roslyn/issues/73445
Corresponding open question: https://github.com/dotnet/csharplang/blob/main/proposals/first-class-span-types.md#prefer-converting-from-t-to-readonlyspant-over-spant-in-overload-resolution

PR #76300, opened 2024-12-06
Prefer ReadOnlySpan over Span as better conversion target

jjonescz:
Follow up on https://github.com/dotnet/roslyn/pull/75853 from LDM 2024-12-04.
Speclet change: https://github.com/dotnet/csharplang/pull/8779

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
@@ -2980,19 +2980,6 @@ outerDefault:
             switch ((conv1.Kind, conv2.Kind))
             {
                 case (ConversionKind.ImplicitSpan, ConversionKind.ImplicitSpan):
-                    // If the expression is of an array type, prefer ReadOnlySpan over Span (to avoid ArrayTypeMismatchExceptions).
-                    if (node.Type is ArrayTypeSymbol)
-                    {
-                        if (t1.IsReadOnlySpan() && t2.IsSpan())
-                        {
-                            return BetterResult.Left;
-                        }
-
-                        if (t1.IsSpan() && t2.IsReadOnlySpan())
-                        {
-                            return BetterResult.Right;
-                        }
-                    }
                     break;
                 case (_, ConversionKind.ImplicitSpan):
                     return BetterResult.Right;
@@ -3454,6 +3441,19 @@ outerDefault:
                 return BetterResult.Neither;
             }
 
+            // SPEC: T₁ is System.ReadOnlySpan<E₁>, T₂ is System.Span<E₂>, and an identity conversion from E₁ to E₂ exists
+            if (Compilation.IsFeatureEnabled(MessageID.IDS_FeatureFirstClassSpan))
+            {
+                if (isBetterSpanConversionTarget(type1, type2))
+                {
+                    return BetterResult.Left;
+                }
+                else if (isBetterSpanConversionTarget(type2, type1))
+                {
+                    return BetterResult.Right;
+                }
+            }
+
             // Given two different types T1 and T2, T1 is a better conversion target than T2 if no implicit conversion from T2 to T1 exists,
             // and at least one of the following holds:
             bool type1ToType2 = Conversions.ClassifyImplicitConversionFromType(type1, type2, ref useSiteInfo).IsImplicit;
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -8062,7 +8062,121 @@ static class C
             }
             """;
         var comp = CreateCompilationWithSpanAndMemoryExtensions(source);
-        CompileAndVerify(comp, expectedOutput: "2212").VerifyDiagnostics();
+        CompileAndVerify(comp, expectedOutput: "2112").VerifyDiagnostics();
+    }
```
cston, line 8066: >} [](http://example.com/codeflow?start=4&length=1)

Please add a test for the identity conversion between type arguments. Perhaps: `ReadOnlySpan<(int, int)>` and `Span<(int X, int Y)>`, or `ReadOnlySpan<dynamic>` and `Span<object>`.

cston (approved):

src/Compilers/CSharp/Test/Emit3/FirstClassSpanTests.cs
```diff
@@ -8061,8 +8054,235 @@ static class C
...
+            {
+                public static implicit operator Span<string>(D d) => default;
+                public static implicit operator ReadOnlySpan<object>(D d) => default;
+            }
+            """;
+        CreateCompilationWithSpanAndMemoryExtensions(source, parseOptions: TestOptions.Regular.WithLanguageVersion(langVersion)).VerifyDiagnostics(
+            // (3,3): error CS0121: The call is ambiguous between the following methods or properties: 'C.M(Span<string>)' and 'C.M(ReadOnlySpan<object>)'
+            // C.M(new D());
+            Diagnostic(ErrorCode.ERR_AmbigCall, "M").WithArguments("C.M(System.Span<string>)", "C.M(System.ReadOnlySpan<object>)").WithLocation(3, 3));
+    }
```
cston, line 8216: >} [](http://example.com/codeflow?start=4&length=1)

Should we also test `Span<object>` and `ReadOnlySpan<object>`:
```csharp
C.M(new D());
            
static class C
{
    public static void M(Span<object> arg) => Console.Write(1);
    public static void M(ReadOnlySpan<object> arg) => Console.Write(2);
}
            
class D
{
    public static implicit operator Span<object>(D d) => default;
    public static implicit operator ReadOnlySpan<object>(D d) => default;
}
```

And `Span<object>` and `ReadOnlySpan<string>`:
```csharp
C.M(new D());
            
static class C
{
    public static void M(Span<object> arg) => Console.Write(1);
    public static void M(ReadOnlySpan<string> arg) => Console.Write(2);
}
            
class D
{
    public static implicit operator Span<object>(D d) => default;
    public static implicit operator ReadOnlySpan<string>(D d) => default;
}
```

src/Compilers/CSharp/Portable/Binder/Semantics/OverloadResolution/OverloadResolution.cs
```diff
@@ -3454,6 +3441,25 @@ private BetterResult BetterConversionTargetCore(
...
+                {
+                    return BetterResult.Left;
+                }
+                else if (isBetterSpanConversionTarget(type2, type1))
+                {
+                    return BetterResult.Right;
+                }
+                // The next case (a type is a better target if an implicit conversion from it to the other type exists)
+                // does not apply to span conversions except the covariant ReadOnlySpan->ReadOnlySpan conversion.
+                else if (type1.IsSpan() || type2.IsSpan())
```
cston, line 3457: >else if (type1.IsSpan() || type2.IsSpan()) [](http://example.com/codeflow?start=16&length=42)

What if one type is `Span<E1>` and the other is not a span type? For instance:
```csharp
using System;
            
C.M(new E());
            
static class C
{
    public static void M(Span<object> arg) => Console.Write(1);
    public static void M(D arg) => Console.Write(2);
}
            
class D
{
    public static implicit operator Span<object>(D d) => default;
}

class E : D
{
}
```

cston (approved):

jjonescz:
@333fred @dotnet/roslyn-compiler for a second review, thanks

jjonescz:
@333fred or @dotnet/roslyn-compiler for the second review, thanks (would like to merge this today before the snap so it goes into 17.13)

jaredpar:
@333fred PTAL

333fred (approved):

Issue #72855, opened 2024-04-02
Ensure existing usages of InterceptsLocationAttribute are onboarded to the new checksum-based approach


RikkiGibson:
Prerequisite: #72133

This issue tracks removing support for the filepath-line-column-based interceptors from the compiler.

Once we ship the checksum-based interceptors pattern we will want to check in with all the partners who have onboarded to the .NET 8 experimental feature and work with them to get them moved to the new checksum-based approach.

Partner issues:
dotnet/runtime#101079
dotnet/aspnetcore#55152

RikkiGibson:
Before we remove the path-line-column support, let's report a warning when that attribute signature is used, e.g. late in the .NET 9 cycle. then later we can fully remove.

RikkiGibson:
We need to wait till Nov 2026, when .NET 8 is [out of support](https://devblogs.microsoft.com/dotnet/net-core-releases-and-support/), before fully removing path-line-column support.

Issue #72133, opened 2024-02-15
Add public API to obtain a location specifier for a call


RikkiGibson:
Continuation of #72093

## Need for generators to obtain a location specifier for a call

We are making adjustments to the interceptors design in order to improve the following aspects:
- Portability of generated code. (i.e. the ability to generate code on one machine and compile the generated code on another machine.)
- Simplicity for generator authors, by introducing public API accessible for generators, in order to get the correct argument(s) for `[InterceptsLocation]`, with minimal guesswork.
- Evolvability of the representation of locations, transparent to generator authors, by introducing an `[InterceptsLocation(int, string)]` constructor overload, the encoding of which can be updated over time.

### Proposed API

We plan to add the following well-known constructor to `InterceptsLocationAttribute`:
```diff
 namespace System.Runtime.CompilerServices;

 public class InterceptsLocationAttribute
 {
     public InterceptsLocationAttribute(string filePath, int line, int character) { } // plan to drop support prior to stable release of interceptors feature
+    public InterceptsLocationAttribute(int version, string data) { }
 }
```

We propose adding the following public API to Roslyn:
```diff
 namespace Microsoft.CodeAnalysis
 {
     public static class CSharpExtensions
     {
+        /// <summary>
+        /// If 'node' cannot be intercepted, returns null. Otherwise, returns an InterceptableLocation instance which can be used to intercept the call denoted by 'node'.
+        /// </summary>
+        [Experimental(RoslynExperiments.Interceptors)]
+        public InterceptableLocation? GetInterceptableLocation(this SemanticModel model, InvocationExpressionSyntax node);
     }
 
     namespace CSharp
     {
+        /// <summary>Location data which can be used to intercept a call.</summary>
+        [Experimental(RoslynExperiments.Interceptors)]
+        public sealed class InterceptableLocation : IEquatable<InterceptableLocation>
+        {
+            /// <summary>The version of the InterceptsLocationAttribute data encoding. Used as an argument for the parameter 'version' of InterceptsLocationAttribute.</summary>
+            public int Version { get; }
+
+            /// <summary>Opaque data which references an interceptable call. Used as an argument for the parameter 'data' of InterceptsLocationAttribute.</summary>
+            public string Data { get; }
+
+            /// <summary>Gets a human-readable location description for the interceptable call.</summary>
+            public string GetDisplayLocation();
+
[...]

PR #76312, opened 2024-12-06
Mark interceptors feature as stable and deprecate file path based attributes

RikkiGibson:
Related to #72855

- Update feature doc to indicate stability of the feature.
- Remove `[Experimental(RoslynExperiments.Interceptors, ...)]` attributes from public APIs.
- Add a warning when `[InterceptsLocation(path, line, column)]` is used in user code.
    - After we have had a long enough notice period, we will completely rip out support for this.
- Migrate tests to use `[InterceptsLocation(version, data)]` where possible.
    - I considered several ways to not have to churn tons of tests but eventually decided to just bite the bullet right here. This brings us way closer to being able to rip out the path-based attribute support anyway.
    - For several tests where this didn't make sense, I simply added the new warnings to the test baselines. Such tests will likely be deleted when support is removed.
    - Filed #76341 to follow up on an issue I observed when porting the tests.


RikkiGibson:
@jcouv @cston does our policy require two reviews on the test changes in this PR?

jcouv:
Yes, this requires two sign-offs. From quick glance, I see at least one product code change `diagnostics.Add(ErrorCode.WRN_InterceptsLocationAttributeUnsupportedSignature, attributeLocation);`

src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs
```diff
@@ -14467,7 +14467,18 @@ public InterceptsLocationAttribute(string filePath, int line, int character)
                 """;
             var generator = new SingleFileTestGenerator(generatedSource, "Generated.cs");
 
-            VerifyOutput(dir, src, includeCurrentAssemblyAsAnalyzerReference: false, additionalFlags: ["/langversion:preview", "/out:embed.exe", "/features:InterceptorsNamespaces=Generated"], generators: [generator], analyzers: null);
+            // Future note: it would be good to have end-to-end tests using InterceptableLocation APIs.
```
cston, line 14470: >Future note [](http://example.com/codeflow?start=15&length=11)

Would it make sense to reference [https://github.com/dotnet/roslyn/issues/72855](https://github.com/dotnet/roslyn/issues/72855) here, so that we consider adding end-to-end tests before closing that issue?

cston: Is `EndToEndTests.Interceptors()` that test?

RikkiGibson: `EndToEndTests.Interceptors` is more of a load test, while these are more like integration tests (does configuration flow properly thru the command line, into CommonCompiler, into generators, and allow producing a valid final compilation).

The final version of the tests in this file would likely look a little more like real world versions of the "Usage Examples" section of #72133.

RikkiGibson: Also, we will be prompted to revisit these tests when support for the path-based attributes is dropped, since the tests will stop working completely. Happy to add the link, but I am also not worried about forgetting about them.

cston:
>         // note: it seems a little surprising that this warning is reported, but it's not important enough to fix.

Don't we normally report the diagnostics from any `CompilationReference` instances?

---
Refers to: src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs:591 in 99a9290. [](commit_id = 99a92907409915fbefab813f19d03bbf110b3aac, deletion_comment = False)

cston:
>         var compilation = CreateCompilation([source, interceptors, s_attributesSource], parseOptions: RegularWithInterceptors);

Should `source` be named "Program.cs"? (It looks like the `ERR_InterceptorPathNotInCompilation` errors are unrelated to the original test.)

---
Refers to: src/Compilers/CSharp/Test/Semantic/Semantics/InterceptorsTests.cs:1472 in 99a9290. [](commit_id = 99a92907409915fbefab813f19d03bbf110b3aac, deletion_comment = False)

cston (approved):

cston (approved):

docs/features/interceptors.md
```diff
@@ -3,7 +3,7 @@
 ## Summary
 [summary]: #summary
 
-*Interceptors* are an experimental compiler feature planned to ship in .NET 8 (with support for C# only). The feature may be subject to breaking changes or removal in a future release.
+*Interceptors* is a C# compiler feature, first shipped experimentally in .NET 8.
```
jcouv, line 6: nit: Consider recording the version where it'll be "shipped"

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -1171,6 +1170,8 @@ private void DecodeInterceptsLocationAttributeExperimentalCompat(
             var attributeSyntax = arguments.AttributeSyntaxOpt;
             Debug.Assert(attributeSyntax is object);
             var attributeLocation = attributeSyntax.Location;
+            diagnostics.Add(ErrorCode.WRN_InterceptsLocationAttributeUnsupportedSignature, attributeLocation);
```
jcouv, line 1173: We've added a warning, but it doesn't look like support was actually removed. Is that intended? Should we keep the tracking issue to remove the logic below at some point?

jcouv (approved):
LGTM Thanks (iteration 15)

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
RikkiGibson: That's correct, the warning is here to give folks notice that we will fully eliminate support here. At which point the warning will likely be upgraded to error.

docs/features/interceptors.md
RikkiGibson: I am expecting the "stable" designation to arrive in 9.0.2xx, so that's what I'll record here.

Issue #76277, opened 2024-12-05
Crash in the compiler when trying random things out with the "Implicit index access" feature syntax


hamarb123:
**Version Used**: sharplab.io (`4.13.0-ci (<developer build>)`), godbolt (`4.12.0-3.24523.4 (f3348c2a)` AND `4.12.0-3.24473.3 (5ef52ae3)`)

**Steps to Reproduce**:

Put in this code:
```csharp
using System;

class Program
{
    static void M1(ref S s)
    {
        s = new S()
        {
            sp =
            {
                [^1] = 5,
            },
        };
    }
}

ref struct S
{
    public Span<int> sp;
}
```

https://sharplab.io/#v2:EYLgtghglgdgNAFxFANgHwAICYCMBYAKEOwAIMcB2Qgb0JPrJwDYyAWEgWRwAoAnAUwBmJAMokAzgEo6DWgQYKJJALwkY/AO6ju0+Yvpz9+8QAcVMo7IuXFAbQB6OALoqSAVjjXLAX09793gDc1t6EoUQEAsLiCLwArgDGCKI01hgAzKImEDAAPLAIAHwSJsEE4UA===

Also reproduces on godbolt with both .NET 9 & main.

**Diagnostic Id**:

N/A

**Expected Behavior**:

Roslyn does not crash.

**Actual Behavior**:

Crashes. Sharplab stack trace:
```
System.InvalidCastException: Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.BoundImplicitIndexerAccess' to type 'Microsoft.CodeAnalysis.CSharp.BoundObjectInitializerMember'.
   at Microsoft.CodeAnalysis.CSharp.RefSafetyAnalysis.GetValEscapeOfObjectMemberInitializer(BoundExpression expr, SafeContext scopeOfTheContainingExpression) in /_/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs:line 4602
   at Microsoft.CodeAnalysis.CSharp.RefSafetyAnalysis.GetValEscapeOfObjectInitializer(BoundObjectInitializerExpression initExpr, SafeContext scopeOfTheContainingExpression) in /_/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs:line 4585
[...]

PR #76339, opened 2024-12-09
Handle other object initializer kinds in ref safety analysis

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/76277.

I do not think the other bound nodes (like implicit indexer access) can affect ref safety, so I didn't add any special handling for them.

### DIFF
--- a/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
+++ b/src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
@@ -2016,6 +2016,7 @@
     <Field Name="Initializers" Type="ImmutableArray&lt;BoundExpression&gt;"/>
   </AbstractNode>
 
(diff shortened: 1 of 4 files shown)


src/Compilers/CSharp/Portable/BoundTree/BoundNodes.xml
```diff
@@ -2016,6 +2016,7 @@
     <Field Name="Initializers" Type="ImmutableArray&lt;BoundExpression&gt;"/>
   </AbstractNode>
 
+  <!-- Initializers of this are BoundAssignmentOperators whose Left is BadBoundExpression or BoundObjectInitializerMember or BoundDynamicObjectInitializerMember or BoundImplicitIndexerAccess or BoundArrayAccess or BoundPointerElementAccess. -->
```
jaredpar, line 2019: Nit: use hard line breaks in this file for comments (seems to be the standard)

jaredpar (approved):

333fred (approved):

AlekseyTs:
@jjonescz Consider adjusting the PR title to be more specific. In my opinion, "Handle other initializer kinds in ref safety analysis" says pretty much nothing about what initializers the change is going to affect.

Issue #72357, opened 2024-03-01
Cloning a .NET 8 record deriving from a .NET 7 record leads to loss of members


Watno:
**Version Used**: 
Compiler  "4.8.0-7.24067.24 (26357112)"

**Steps to Reproduce**:

1. Create a project targeting .NET 7 and define a record class
```
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <TargetFramework>net7.0</TargetFramework>
        <ImplicitUsings>enable</ImplicitUsings>
        <Nullable>enable</Nullable>
    </PropertyGroup>

</Project>
```

```
public record BaseRecord
{
    public required object Object1 { get; init; }
    public required object Object2 { get; init; }
}
```
2. Create a project targeting .NET 8 and define a derived record class there

```
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <TargetFramework>net8.0</TargetFramework>
        <OutputType>Exe</OutputType>
    </PropertyGroup>

    <ItemGroup>
      <ProjectReference Include="..\DemoBase\DemoBase.csproj" />
    </ItemGroup>

</Project>
```

```
public record DerivedRecord : BaseRecord
{
}
```

3. Instantiate the derived record and clone it using `with`

```
using System.Diagnostics;

var sut = new DerivedRecord
{
    Object1 = new object(),
    Object2 = new object()
};

var broken = sut with { Object2 = new object()}; 

Debug.Assert(broken.Object1 != null);
```

[Demo.zip](https://github.com/dotnet/roslyn/files/14463754/Demo.zip)
https://gist.github.com/Watno/f6d78b49a15530e69a2a1cc1ce8d1e40

**Expected Behavior**: Members of the base record get cloned to the copy

**Actual Behavior**: Members of the base record disappear

**Additional information**: The IL code of the constructor of `DerivedRecord` seems to be broken:
```
  .method family hidebysig specialname rtspecialname instance void
    .ctor(
      class DerivedRecord original
    ) cil managed
  {
    .custom instance void [System.Runtime]System.Runtime.CompilerServices.CompilerGeneratedAttribute::.ctor()
      = (01 00 00 00 )
    .custom instance void [System.Runtime]System.Diagnostics.CodeAnalysis.SetsRequiredMembersAttribute::.ctor()
      = (01 00 00 00 )
    .maxstack 8

    // [1 15 - 1 28]
    IL_0000: nop
    IL_0001: ret

[...]

PR #76347, opened 2024-12-10
Do not bail generating base type initializer in the presence of use site warnings

333fred:
We were returning early from generating a call to the base record copy constructor initializer if use site diagnostics reported any kind of diagnostic, including warnings. This isn't good in general, but is particularly bad for warnings like CS1701, which are typically suppressed by the SDK and results in what seems like a completely clean csc invocation skipping calling the base constructor. Fixes https://github.com/dotnet/roslyn/issues/72357.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
@@ -7197,7 +7197,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             return succeededConsideringAccessibility;
         }
 
-        internal static bool ReportConstructorUseSiteDiagnostics(Location errorLocation, BindingDiagnosticBag diagnostics, bool suppressUnsupportedRequiredMembersError, in CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+        internal static void ReportConstructorUseSiteDiagnostics(Location errorLocation, BindingDiagnosticBag diagnostics, bool suppressUnsupportedRequiredMembersError, in CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
         {
             if (suppressUnsupportedRequiredMembersError && useSiteInfo.AccumulatesDiagnostics && useSiteInfo.Diagnostics is { Count: not 0 })
             {
@@ -7213,12 +7213,10 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                     diagnostics.ReportUseSiteDiagnostic(diagnostic, errorLocation);
                 }
-
-                return true;
             }
             else
             {
-                return diagnostics.Add(errorLocation, useSiteInfo);
+                diagnostics.Add(errorLocation, useSiteInfo);
             }
         }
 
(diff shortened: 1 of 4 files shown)


333fred:
@dotnet/roslyn-compiler for reviews

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -7197,7 +7197,7 @@ internal bool TryPerformConstructorOverloadResolution(
             return succeededConsideringAccessibility;
         }
 
-        internal static bool ReportConstructorUseSiteDiagnostics(Location errorLocation, BindingDiagnosticBag diagnostics, bool suppressUnsupportedRequiredMembersError, in CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
+        internal static void ReportConstructorUseSiteDiagnostics(Location errorLocation, BindingDiagnosticBag diagnostics, bool suppressUnsupportedRequiredMembersError, in CompoundUseSiteInfo<AssemblySymbol> useSiteInfo)
```
333fred, line 7200: There are lots of uses of this method, but this was the only one that did something conditional. That seemed like a bad thing to keep around, so I just made the method return `void` instead to prevent something like this from happening again.

jcouv (approved):
LGTM Thanks (iteration 1)

src/Compilers/CSharp/Test/Emit3/Semantics/RecordTests.cs
```diff
@@ -14073,6 +14073,51 @@ public static void Main()
...
+                    Object2 = new object()
+                };
+
+                var broken = sut with { Object2 = new object()};
+                System.Console.Write(broken.Object1 is null);
+
+                public record DerivedRecord : BaseRecord;
+                """, assemblyName: "Derived", references: [comp1.EmitToImageReference()], targetFramework: TargetFramework.Net80);
+
+            var verifier = CompileAndVerify(comp2, expectedOutput: "False");
```
333fred, line 14100:
```suggestion
            var verifier = CompileAndVerify(comp2, expectedOutput: "False", verify: Verification.FailsPEVerify);
```

AlekseyTs:

src/Compilers/CSharp/Test/Emit3/Semantics/RecordTests.cs
```diff
@@ -14073,6 +14073,51 @@ public static void Main()
...
+                    Object2 = new object()
+                };
+
+                var broken = sut with { Object2 = new object()};
+                System.Console.Write(broken.Object1 is null);
+
+                public record DerivedRecord : BaseRecord;
+                """, assemblyName: "Derived", references: [comp1.EmitToImageReference()], targetFramework: TargetFramework.Net80);
+
+            var verifier = CompileAndVerify(comp2, expectedOutput: ExecutionConditionUtil.IsCoreClr ? "False" : null, verify: Verification.FailsPEVerify);
```
AlekseyTs, line 14100: >IsCoreClr [](http://example.com/codeflow?start=90&length=9)

I think we usually use ```IsMonoOrCoreClr``` in situations like this.

AlekseyTs (approved):
LGTM (commit 3)

src/Compilers/CSharp/Test/Emit3/Semantics/RecordTests.cs
333fred: Since this is specifically .NET 8, I don't think all mono environments would work, so I only did `CoreClr`.

PR #76350, opened 2024-12-10
Implement editorconfig serializer for naming style preferences

tmat:
The serializer is required when saving naming style preferences specified in Tools > Options to solution fallback options. The option was throwing NotSupportedException, which was reported via NFW but for some reason does not appear in telemetry  data (TBD why).

In order to preserve ordering of the naming style rules as specified in Tools > Options settings we introduce a new editor option `dotnet_naming_rule.{rule-name}.priority`. The highest priority is 0, which is the default. When saving VS options we generate priorities 0..N-1 where N is the number of rules. This causes VS option order to be preserved when the preferences are deserialized from fallback options.

Note that if any naming style is set in .editorconfig file, all naming style settings in VS options are ignored. This behavior is consistent with VS 2019.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2297536

### DIFF
--- a/src/Features/TestUtilities/Options/NamingStyleTestUtilities.cs
+++ b/src/Features/TestUtilities/Options/NamingStyleTestUtilities.cs
@@ -1,0 +1,55 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Xml.Linq;
+using Microsoft.CodeAnalysis.Diagnostics.Analyzers.NamingStyles;
+
+namespace Microsoft.CodeAnalysis.Test.Utilities;
+
+internal static class NamingStyleTestUtilities
+{
+    public static string Inspect(this NamingStylePreferences preferences, string[]? excludeNodes = null)
+    {
+        var xml = preferences.CreateXElement();
+
+        // filter out insignificant elements:
+        var elementsToRemove = new List<XElement>();
+        foreach (var element in xml.DescendantsAndSelf())
+        {
+            if (excludeNodes != null && excludeNodes.Contains(element.Name.LocalName))
+            {
+                elementsToRemove.Add(element);
+            }
+        }
+
+        foreach (var element in elementsToRemove)
+        {
+            element.Remove();
+        }
+
+        // replaces GUIDs with unique deterministic numbers:
+        var ordinal = 0;
+        var guidMap = new Dictionary<Guid, int>();
+        foreach (var element in xml.DescendantsAndSelf())
+        {
+            foreach (var attribute in element.Attributes())
+            {
+                if (Guid.TryParse(attribute.Value, out var guid))
+                {
+                    if (!guidMap.TryGetValue(guid, out var existingOrdinal))
+                    {
+                        existingOrdinal = ordinal++;
+                    }
+
+                    attribute.Value = existingOrdinal.ToString();
+                }
+            }
+        }
+
+        return xml.ToString();
+    }
+}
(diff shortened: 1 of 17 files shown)


tmat:
@sharwell ptal

src/Features/TestUtilities/Options/NamingStyleTestUtilities.cs
```diff
@@ -0,0 +1,55 @@
...
+        var guidMap = new Dictionary<Guid, int>();
+        foreach (var element in xml.DescendantsAndSelf())
+        {
+            foreach (var attribute in element.Attributes())
+            {
+                if (Guid.TryParse(attribute.Value, out var guid))
+                {
+                    if (!guidMap.TryGetValue(guid, out var existingOrdinal))
+                    {
+                        existingOrdinal = ordinal++;
```
JoeRobich, line 45: Should this new ordinal be added to the guidMap?

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/NamingStyles/EditorConfig/EditorConfigNamingStyleParser.cs
```diff
@@ -82,11 +66,12 @@ public static NamingStylePreferences ParseDictionary(AnalyzerConfigOptions allRa
         // which a user has trouble ordering, the intersection of the two rules can be broken out into a new rule
         // will always match earlier than the broader rules it was derived from.
         var orderedRules = preferences.Rules.NamingRules
-            .OrderBy(rule => rule, NamingRuleModifierListComparer.Instance)
+            .OrderByDescending(rule => ruleNamesAndPriorities[(rule.SymbolSpecification.ID, rule.NamingStyle.ID, rule.EnforcementLevel)].priority)
```
sharwell, line 69: This should be ascending right?

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/NamingStyles/Serialization/NamingStylePreferencesEditorConfigSerializer.cs
```diff
@@ -39,35 +39,47 @@ public static void AppendNamingStylePreferencesToEditorConfig(
             serializableNamingRules,
             language,
             entryWriter: (name, value) => builder.AppendLine($"{name} = {value}"),
-            triviaWriter: trivia => builder.AppendLine(trivia));
+            triviaWriter: trivia => builder.AppendLine(trivia),
+            priority: null);
```
sharwell, line 43: Why no priority here?

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/NamingStyles/EditorConfig/EditorConfigNamingStyleParser.cs
tmat: Ah, yes

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/NamingStyles/Serialization/NamingStylePreferencesEditorConfigSerializer.cs
tmat: This is used for generating physical editorconfig file from settings. I didn't want to change the existing behavior.

tmat: I think the default priority is fine there

arunchndr (approved):

sharwell (approved):

tmat:
Closing in favor of 17.12 change, which will flow to main.

PR #76353, opened 2024-12-10
Disable signing validation for the PR Val pipeline.

JoeRobich:
This will allow us to use test signed builds for PR validation.


azure-pipelines-pr-validation.yml
```diff
@@ -338,6 +338,7 @@ extends:
         publishingInfraVersion: 3
         # Symbol validation is not entirely reliable as of yet, so should be turned off until
         # https://github.com/dotnet/arcade/issues/2871 is resolved.
+        enableSigningValidation: false
```
jjonescz, line 341: The comment above probably does not apply to this line.

JoeRobich: Thanks!

jjonescz (approved):

PR #76357, opened 2024-12-10
Razor synchronize loader

chsienki:
This updates the way the razor assembly loader installs its handler to ensure we can synchronize correctly between Razor and Roslyn. The approach is to track which assemblies have been loaded and only install the filter if the requested assembly has not yet been loaded. 

 Most of the work is on the Razor side, the Roslyn side is deliberately simple with nothing hard coded so we aren't baking in any Razor specific logic or dependencies. A WIP doc explaining the process from both sides is available at https://gist.github.com/chsienki/f3b3b19bf6cf6d1f8e047e59242db648.

Also adds a method to retrieve only the documents from a specific source generator. Again, we don't bake in the razor generator, so that we can change that without needing to change roslyn.

### DIFF
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -45,12 +45,14 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
 
         public Assembly? ResolveAssembly(AssemblyName assemblyName)
         {
+            Func<AssemblyName, Assembly?>? resolver = null;
             lock (s_resolverLock)
             {
                 s_assembliesRequested.Add(assemblyName);
+                resolver = s_assemblyResolver;
             }
 
-            return s_assemblyResolver?.Invoke(assemblyName);
+            return resolver?.Invoke(assemblyName);
         }
     }
 }
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -3,49 +3,53 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.Composition;
 using System.Diagnostics;
 using System.Reflection;
 using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.PooledObjects;
 using Microsoft.VisualStudio.Composition;
 
 namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
 {
     [Export(typeof(IAnalyzerAssemblyResolver)), Shared]
-    internal class RazorAnalyzerAssemblyResolver : IAnalyzerAssemblyResolver
+    [method: ImportingConstructor]
+    [method: Obsolete(MefConstruction.ImportingConstructorMessage, error: true)]
+    internal class RazorAnalyzerAssemblyResolver() : IAnalyzerAssemblyResolver
     {
         private static Func<AssemblyName, Assembly?>? s_assemblyResolver;
 
+        private static readonly List<AssemblyName>? s_assembliesRequested = [];
+
+        private static readonly object s_resolverLock = new();
+
         /// <summary>
-        /// We use this as a heuristic to catch a case where we set the resolver too 
-        /// late and the resolver has already been asked to resolve a razor assembly.
-        /// 
-        /// Note this isn't perfectly accurate but is only used to trigger an assert
-        /// in debug builds. 
+        /// Attempts to set the assembly resolver. Will only succeed if the <paramref name="canaryAssembly"/> has not already been requested to load.
         /// </summary>
-        private static bool s_razorRequested = false;
-
-        [ImportingConstructor]
-        [Obsolete(MefConstruction.ImportingConstructorMessage, error: true)]
-        public RazorAnalyzerAssemblyResolver()
+        /// <param name="resolver">The resolver function to set.</param>
+        /// <param name="canaryAssembly">The name of an assembly that is checked to see if it has been requested to load. Setting the resolver will fail if this has already been requested.</param>
+        /// <returns><c>true</c> when the resolver was successfully set.</returns>
+        internal static bool TrySetAssemblyResolver(Func<AssemblyName, Assembly?> resolver, AssemblyName canaryAssembly)
         {
-        }
-
-        internal static Func<AssemblyName, Assembly?>? AssemblyResolver
-        {
-            get => s_assemblyResolver;
-            set
+            lock (s_resolverLock)
             {
-                Debug.Assert(s_assemblyResolver == null, "Assembly resolver should not be set multiple times.");
-                Debug.Assert(!s_razorRequested, "A razor assembly has already been requested before setting the resolver.");
-                s_assemblyResolver = value;
+                if (s_assemblyResolver is not null && s_assembliesRequested is not null && !s_assembliesRequested.Contains(canaryAssembly))
+                {
+                    s_assemblyResolver = resolver;
+                    return true;
+                }
+                return false;
             }
         }
 
         public Assembly? ResolveAssembly(AssemblyName assemblyName)
         {
-            s_razorRequested |= assemblyName.FullName.Contains("Razor");
-            return s_assemblyResolver?.Invoke(assemblyName);
+            lock (s_resolverLock)
+            {
+                s_assembliesRequested?.Add(assemblyName);
+                return s_assemblyResolver?.Invoke(assemblyName);
+            }
         }
     }
 }
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -34,7 +34,7 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
         {
             lock (s_resolverLock)
             {
-                if (s_assemblyResolver is not null && !s_assembliesRequested.Contains(canaryAssembly))
+                if (s_assemblyResolver is null && !s_assembliesRequested.Contains(canaryAssembly))
                 {
                     s_assemblyResolver = resolver;
                     return true;
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -20,7 +20,7 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
     {
         private static Func<AssemblyName, Assembly?>? s_assemblyResolver;
 
-        private static readonly List<AssemblyName> s_assembliesRequested = [];
+        private static readonly HashSet<AssemblyName> s_assembliesRequested = [];
 
         private static readonly object s_resolverLock = new();
 
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -20,7 +20,7 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
     {
         private static Func<AssemblyName, Assembly?>? s_assemblyResolver;
 
-        private static readonly List<AssemblyName>? s_assembliesRequested = [];
+        private static readonly List<AssemblyName> s_assembliesRequested = [];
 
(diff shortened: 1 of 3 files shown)


src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -3,49 +3,53 @@
...
 
-        /// <summary>
-        /// We use this as a heuristic to catch a case where we set the resolver too 
-        /// late and the resolver has already been asked to resolve a razor assembly.
-        /// 
-        /// Note this isn't perfectly accurate but is only used to trigger an assert
-        /// in debug builds. 
-        /// </summary>
-        private static bool s_razorRequested = false;
+        private static readonly List<AssemblyName>? s_assembliesRequested = [];
```
chsienki, line 23: All access to this is done under a lock, and there's only a single instance so didn't seem worth using an immutable variant or pooling it.

src/Tools/ExternalAccess/Razor/RazorProjectExtensions.cs
```diff
@@ -17,5 +17,14 @@ internal static class RazorProjectExtensions
         {
             return project.GetSourceGeneratorRunResultAsync(cancellationToken);
         }
+
+        internal static async ValueTask<IEnumerable<SourceGeneratedDocument>> GetSourceGeneratedDocumentsForGeneratorAsync(this Project project, string assemblyName, string assemblyPath, Version assemblyVersion, string typeName, CancellationToken cancellationToken)
+        {
+            var results = await project.GetSourceGeneratedDocumentsAsync(cancellationToken).ConfigureAwait(false);
+
+            var generatorIdentity = new SourceGeneratorIdentity(assemblyName, assemblyPath, assemblyVersion, typeName);
```
chsienki, line 25: `SourceGeneratorIdentity` isn't public, so have to pass through the requisite parts. We could make an IVT copy but doesn't seem worth it.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
jaredpar: Why is this marked as nullable? Are you trying to account for a `static` constructor race here?

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -3,49 +3,53 @@
...
+        internal static bool TrySetAssemblyResolver(Func<AssemblyName, Assembly?> resolver, AssemblyName canaryAssembly)
         {
-            get => s_assemblyResolver;
-            set
+            lock (s_resolverLock)
             {
-                Debug.Assert(s_assemblyResolver == null, "Assembly resolver should not be set multiple times.");
-                Debug.Assert(!s_razorRequested, "A razor assembly has already been requested before setting the resolver.");
-                s_assemblyResolver = value;
+                if (s_assemblyResolver is not null && s_assembliesRequested is not null && !s_assembliesRequested.Contains(canaryAssembly))
```
jaredpar, line 37: Not sure how `s_assembliesRequested` could be `null` here. Even in a `static` ctor race `s_resolverLock` would also be `null` so we'd throw before this part. 

Shouldn't the check be `s_assemblyResolver is null`? The field starts off as `null` and this is the only place that it is written to so as stated this will never be set (I believe).

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -3,49 +3,53 @@
...
         }
 
         public Assembly? ResolveAssembly(AssemblyName assemblyName)
         {
-            s_razorRequested |= assemblyName.FullName.Contains("Razor");
-            return s_assemblyResolver?.Invoke(assemblyName);
+            lock (s_resolverLock)
+            {
+                s_assembliesRequested?.Add(assemblyName);
+                return s_assemblyResolver?.Invoke(assemblyName);
```
jaredpar, line 51: This is doing a callback inside of a `lock`. That usually gives me a bit of pause. Is there a specific reason it needs to be inside the `lock`? Did you consider say saving the field to a local and then invoking that outside the `lock`?

RikkiGibson:
Could you please link to related issues/PRs so that I can get the full context on the scenario this PR is addressing? Thanks!

chsienki:
@RikkiGibson There isn't really a good issue/PR to address this, but I linked a doc in the description which should give you the context you need.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
RikkiGibson, line 37: I am confused as to how the `s_assemblyResolver` can ever become non-null, if entering the block that assigns it requires it to already be non-null.

chsienki: Duh. It was supposed to be nulled out if the resolver is installed, but apparently in my refactoring I missed that off 🤦

RikkiGibson:
> @RikkiGibson There isn't really a good issue/PR to address this, but I linked a doc in the description which should give you the context you need.

I took a look and things are mostly making sense. Just trying to get a full sense of the impact here. It sounds like we already had some logic to coordinate this generator load across roslyn and razor, it's just that there were reliability issues due to lack of synchronization. Is that right?

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
chsienki: Just a hangover from an earlier version pre-refactoring that I hadn't cleaned up. Fixed.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
chsienki: Hmm, we need to retrieve the resolver under the lock, but once we've done that I can't think of any reason why we need to actually execute it under the lock.

RikkiGibson: There's no concern about the resolver being invoked on different threads without synchronization, right? Just a need for the `s_assembliesRequested` update and `s_assemblyResolver` read to be synchronized?

chsienki: Thinking about it, we don't even need to retrieve it under the lock. Before we enter the lock we either 1) have the resolver set or 2) we don't:

In the case of 1) the locking doesn't actually have any effect. 

In the case of 2) if the assembly being requested is the canary assembly, then once we exit the lock the resolver can never be set, so it's also safe to just use it. 

In the case of 2) where it's _not_ the canary assembly being requested, we actually have a race condition. We exit the lock having added the load to the list, but the state of the resolver can still change. If the install call comes in in the time between exiting the lock and executing the resolver we'll use the resolver, if it doesn't we wont. Now, for our code this actually doesn't matter, really, we only want to execute the resolver for the canary assembly anyway (but we can't do that because we don't know ahead of time what it is). 

We can make the last part deterministic by retrieving the loader under the lock and executing against the local value, but don't need to for it to be correct for our usage. I'm leaning towards making it deterministic just for ease-of-understanding the code though.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
chsienki: Just a hangover from an earlier version pre-refactoring that I hadn't cleaned up. Fixed.

chsienki:
@RikkiGibson 
>it's just that there were reliability issues due to lack of synchronization. Is that right?

We were previously assuming the Razor tooling would come up before Roslyn loaded the source generator, but it's actually a non-fixable race between which set of tooling gets loaded first, due to the way the service hub process comes up (and a bunch of other stuff like what files a user has open). It's completely legal for either razor or Roslyn to start first, so we have to have this synchronization to ensure we only load a single copy.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
chsienki: Ended up just making this deterministic in all cases so its easier to reason about.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -3,49 +3,56 @@
...
+        internal static bool TrySetAssemblyResolver(Func<AssemblyName, Assembly?> resolver, AssemblyName canaryAssembly)
         {
-            get => s_assemblyResolver;
-            set
+            lock (s_resolverLock)
             {
-                Debug.Assert(s_assemblyResolver == null, "Assembly resolver should not be set multiple times.");
-                Debug.Assert(!s_razorRequested, "A razor assembly has already been requested before setting the resolver.");
-                s_assemblyResolver = value;
+                if (s_assemblyResolver is not null && !s_assembliesRequested.Contains(canaryAssembly))
```
jaredpar, line 37:
```suggestion
                if (s_assemblyResolver is null && !s_assembliesRequested.Contains(canaryAssembly))
```

Otherwise `s_assemblyResolver` is never set to a value.

chsienki: *sigh* 🤦

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -3,49 +3,56 @@
...
-        [ImportingConstructor]
-        [Obsolete(MefConstruction.ImportingConstructorMessage, error: true)]
-        public RazorAnalyzerAssemblyResolver()
-        {
-        }
+        private static readonly object s_resolverLock = new();
 
-        internal static Func<AssemblyName, Assembly?>? AssemblyResolver
+        /// <summary>
+        /// Attempts to set the assembly resolver. Will only succeed if the <paramref name="canaryAssembly"/> has not already been requested to load.
```
jjonescz, line 28:
```suggestion
        /// Attempts to set the assembly resolver. Will only succeed if the <paramref name="canaryAssembly"/> has not already been requested to load and the resolver has not been set previously.
```

jjonescz (approved):

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -3,49 +3,56 @@
...
+                    s_assemblyResolver = resolver;
+                    return true;
+                }
+                return false;
             }
         }
 
         public Assembly? ResolveAssembly(AssemblyName assemblyName)
         {
-            s_razorRequested |= assemblyName.FullName.Contains("Razor");
```
RikkiGibson, line 47: Are there any potential problems with razor being out of sync with this EA? i.e. razor is older than roslyn, and razor expects the old assembly hook pattern to be used, or roslyn is older than razor, and razor expects the new hook pattern.

Maybe we just have to take care to dual insert and we'll be ok?

RikkiGibson (approved):

src/Tools/ExternalAccess/Razor/RazorProjectExtensions.cs
```diff
@@ -17,5 +17,14 @@ internal static class RazorProjectExtensions
...
+            return results.Where(s => s.Identity.Generator == generatorIdentity);
+
```
DustinCampbell, line 27: nit: extra blank line

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -3,49 +3,56 @@
...
 
-        /// <summary>
-        /// We use this as a heuristic to catch a case where we set the resolver too 
-        /// late and the resolver has already been asked to resolve a razor assembly.
-        /// 
-        /// Note this isn't perfectly accurate but is only used to trigger an assert
-        /// in debug builds. 
-        /// </summary>
-        private static bool s_razorRequested = false;
+        private static readonly List<AssemblyName> s_assembliesRequested = [];
```
DustinCampbell, line 23: Since all of the code here just adds items to this collection and checks to see if items are present, should this be a `HashSet<AssemblyName>` rather than a `List<AssemblyName>`?

DustinCampbell (approved):

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
chsienki: Good idea, thanks.

PR #76361, opened 2024-12-10
Reduce allocations from doing repeated adds into an ImmutableDictionary in SolutionCompilationState.ComputeFrozenSnapshot

ToddGrun:
Instead, use an ImmutableSegmentedDictionaryBuilder to do the adds and then do a single update to the SolutionCompilationState._projectIdToTrackerMap after all new compilation trackers are created.

Note that there is a potential negative aspect to this change as it increases the time window during which a compilation state can be queried for in the SolutionCompilationState and not found. However, I don't believe this will cause functional issues, and wasn't hitting that concern during local testing.

*** allocations before change from typing period in csharpediting speedometer ***
image

### DIFF
--- a/src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
+++ b/src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
@@ -1454,49 +1454,7 @@ internal sealed partial class SolutionCompilationState
 
     private SolutionCompilationState ComputeFrozenSnapshot(CancellationToken cancellationToken)
     {
-        var newIdToProjectStateMapBuilder = this.SolutionState.ProjectStates.ToBuilder();
-        var newIdToTrackerMapBuilder = _projectIdToTrackerMap.ToBuilder();
-
-        // Used to track any new compilation trackers created in the loop. This is done to avoid allocations
-        // from individually adding to the _projectIdToTrackerMap ImmutableSegmentedDictionary .
-        var updatedIdToTrackerMapBuilder = _projectIdToTrackerMap.ToBuilder();
-
-        foreach (var projectId in this.SolutionState.ProjectIds)
-        {
-            cancellationToken.ThrowIfCancellationRequested();
-
-            // Definitely do nothing for non-C#/VB projects.  We have nothing to freeze in that case.
-            var oldProjectState = this.SolutionState.GetRequiredProjectState(projectId);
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi (approved):

ToddGrun:
Test insertion PR: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/597687

ToddGrun:
> Test insertion PR: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/597687

Speedometer results look good:
image

image

src/Workspaces/Core/Portable/Workspace/Solution/SolutionCompilationState.cs
```diff
@@ -1482,6 +1492,9 @@ private SolutionCompilationState ComputeFrozenSnapshot(CancellationToken cancell
             newIdToTrackerMapBuilder[projectId] = newTracker;
         }
 
+        // Update _projectIdToTrackerMap to include all the newly created compilation trackers.
+        RoslynImmutableInterlocked.InterlockedExchange(ref _projectIdToTrackerMap, updatedIdToTrackerMapBuilder.ToImmutable());
```
CyrusNajmabadi, line 1496: i feel like there's some sort of race here.   where multiple threads come in, update the build, and the other maps, and then blast back into this field.  this actually feels like it could potentially be a problem.  convince me this is 100% safe.

CyrusNajmabadi: this may need to be a spin, with an interlocked compare-exchange.  and if the field changes out from underneath us, we have to go redo this work to pick up any comp trackers made by another thread.

ToddGrun: Actually, I'm now not convinced there wasn't a timing issue in this code before.

ToddGrun: Actually, this PR was the one that added the multiple usages of _projectIdToTrackerMap within the method. I've pushed out a commit that I *think* essentially does what you requested.

CyrusNajmabadi (approved):

ToddGrun:
Commit 2 VS Test insertion to get new speedometer numbers:

https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/598001

Issue #74423, opened 2024-07-17
False positive CS8618 null reference type warning when using required members and [SetsRequiredMembers] in derived class


SapiensAnatis:
**Version Used**: 

`Compiler version: '4.10.0-3.24314.14 (259e82e9)'. Language version: 12.0.`, but also reproduces on `main (11 Jun 2024)` on Sharplab.

**Steps to Reproduce**:

1. Compile the following code ([sharplab](https://sharplab.io/#v2:EYLgtghglgdgPgAQAwAIEEYB0ARKEDmMA9gM4AuUAxiZgMJEAmApgIIwQA2AniVCQNwBYAFAiEAZhQRg5AE4RKZNACYUAIQgkmIgN4iUBlAG0AykzIkASkwCOAVyiymDALJMwwJrJIBdfYYl1TSYACgxUBEj0JABKfwM9YUNklDIACz5MEzJZWHwUAF40KKQhJMMAX3iUasCne0dnKRkchSVwlGzcmHydFHxzfhRYKDIhquEJsUkEVWwvKAA3JpAgrV1q03MrWwcnV3dPbz9yg0D53OWGMOji8JiUVeBgsJK405RE5KmPwKJl2S5Zgoep7JodLp5T79QbDGCjcYiCZAA)):

```cs
using System.Diagnostics.CodeAnalysis;

public abstract class Base
{
    [SetsRequiredMembers]
    public Base(string @string)
    {
        this.String = @string;
    }
    
    public required abstract string String { get; init; }
}

public class Derived : Base
{
    [SetsRequiredMembers]
    public Derived(string @string) : base(@string)
    {
    }

    public override required string String { get; init; }
}
```

**Diagnostic Id**:

CS8618: Non-nullable property 'String' must contain a non-null value when exiting constructor. Consider declaring the property as nullable.

**Expected Behavior**:

No diagnostics, because the base class sets the value.

**Actual Behavior**:

A diagnotic on the constructor of `Derived`.

**Additional Information**

If I remove the `: Base` from `Derived` and make it an independent class, I get the same warning. This leads me to suspect this is a case of the nullable reference type analysis not being sophisticated enough to determine I have set `String` in the base constructor. I am aware that NRT analysis doesn't support every pattern under the sun and there are many ways you can 'trick' it. If this is one of those cases, I apologize - but I thought it was worth mentioning all the same.

jaredpar:
@333fred for comment as I'm unsure what the correct behavior is here.

[...]

PR #76363, opened 2024-12-10
Honor abstract required during state initialization in nullable

333fred:
Fixes https://github.com/dotnet/roslyn/issues/74423. When we're enforcing that all members of a type must have non-null values for `SetsRequiredMembers`, we don't want to include any members of the current type that are overrides of a base type. Instead, those members should be enforced when we check the chained constructor call, since the chained constructor may itself have `SetsRequiredMembers`, and therefore already enforced that overridden properties have a valid state.

### DIFF
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -910,13 +910,28 @@ namespace Microsoft.CodeAnalysis.CSharp
                                 => ImmutableArray<Symbol>.Empty,
 
                             (includeAllMembers: false, includeCurrentTypeRequiredMembers: true, includeBaseRequiredMembers: false)
-                                => containingType.GetMembersUnordered().SelectManyAsArray(predicate: SymbolExtensions.IsRequired, selector: getAllMembersToBeDefaulted),
+                                => containingType.GetMembersUnordered().SelectManyAsArray(
+                                    predicate: SymbolExtensions.IsRequired,
+                                    selector: symbol => getAllMembersToBeDefaulted(symbol, filterOverridingProperties: true)),
 
                             (includeAllMembers: false, includeCurrentTypeRequiredMembers: true, includeBaseRequiredMembers: true)
-                                => containingType.AllRequiredMembers.SelectManyAsArray(static kvp => getAllMembersToBeDefaulted(kvp.Value)),
+                                => containingType.AllRequiredMembers.SelectManyAsArray(static kvp => getAllMembersToBeDefaulted(kvp.Value, filterOverridingProperties: true)),
 
                             (includeAllMembers: true, includeCurrentTypeRequiredMembers: _, includeBaseRequiredMembers: false)
-                                => containingType.GetMembersUnordered().SelectAsArray(getFieldSymbolToBeInitialized),
+                                => containingType.GetMembersUnordered().SelectManyAsArray(
+                                    selector: symbol =>
+                                    {
+                                        var symbolToInitialize = getFieldSymbolToBeInitialized(symbol);
+                                        var prop = symbolToInitialize as PropertySymbol ?? (symbolToInitialize as FieldSymbol)?.AssociatedSymbol as PropertySymbol;
+                                        if (prop is not null && isFilterableOverrideOfAbstractProperty(prop))
+                                        {
+                                            return OneOrMany<Symbol>.Empty;
+                                        }
+                                        else
+                                        {
+                                            return OneOrMany.Create(symbolToInitialize);
+                                        }
+                                    }),
 
                             (includeAllMembers: true, includeCurrentTypeRequiredMembers: true, includeBaseRequiredMembers: true)
                                 => getAllTypeAndRequiredMembers(containingType),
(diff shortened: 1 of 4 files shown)


333fred:
@RikkiGibson @dotnet/roslyn-compiler for reviews.

RikkiGibson:
It looks like some diagnostics on non-required override properties and events are now missing.

https://helixr1107v0xdeko0k025g8.blob.core.windows.net/dotnet-roslyn-refs-pull-76363-merge-935b3591bda3452889/Microsoft.CodeAnalysis.CSharp.Semantic.UnitTests_9/1/console.7b0003b2.log?helixlogtype=result

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
```diff
@@ -5820,6 +5820,462 @@ public Derived() : base() { }
...
+            }
+
+            public class Derived : Base
+            {
+                [SetsRequiredMembers]
+                public Derived(string str) : base(str)
+                {
+                }
+
+                public override required string Str { get; set; }
```
RikkiGibson, line 6109: We should add a test which is similar to this except with `[AllowNull]` on this override. This may actually have the effect of removing both warnings.

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
```diff
@@ -5820,6 +5820,462 @@ public Derived() : base() { }
...
+            #nullable enable
+            using System.Diagnostics.CodeAnalysis;
+
+            public abstract class Base
+            {
+                public Base()
+                {
+                }
+    
+                public required abstract string Str { get; set; }
```
RikkiGibson, line 6055: Maybe we should also test some similar cases to this where required is missing on the base (which appears to be permitted)? The results might further validate the need for the logic added in this PR.

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
```diff
@@ -5820,6 +5820,462 @@ public Derived() : base() { }
...
+            using System.Diagnostics.CodeAnalysis;
+
+            public abstract class Base
+            {
+                [SetsRequiredMembers]
+                public Base(string str)
+                {
+                }
+
+                public required virtual string Str { get; set; } = "";
```
RikkiGibson, line 5870: So from this result, is the language effectively aware that a base constructor can satisfy its `required virtual` properties using property initializers, and skip calling the override accessors? But it can't do the same for a `required abstract`?

RikkiGibson (commented):
Done review pass, I will need a little more time to soak it all in and to see a few more tests.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -910,13 +910,28 @@ static ImmutableArray<Symbol> membersToBeInitialized(NamedTypeSymbol containingT
...
-                                => containingType.AllRequiredMembers.SelectManyAsArray(static kvp => getAllMembersToBeDefaulted(kvp.Value)),
+                                => containingType.AllRequiredMembers.SelectManyAsArray(static kvp => getAllMembersToBeDefaulted(kvp.Value, filterOverridingProperties: true)),
 
                             (includeAllMembers: true, includeCurrentTypeRequiredMembers: _, includeBaseRequiredMembers: false)
-                                => containingType.GetMembersUnordered().SelectAsArray(getFieldSymbolToBeInitialized),
+                                => containingType.GetMembersUnordered().SelectManyAsArray(
+                                    selector: symbol =>
+                                    {
+                                        var symbolToInitialize = getFieldSymbolToBeInitialized(symbol);
+                                        var prop = symbolToInitialize as PropertySymbol ?? (symbolToInitialize as FieldSymbol)?.AssociatedSymbol as PropertySymbol;
```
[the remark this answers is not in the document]
333fred: It cannot. `getFieldSymbolToBeInitialized` will return the backing field for autoproperties. The best normalization path I could fine was this one.

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
333fred: > which appears to be permitted

It is not permitted.

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
333fred: No, that's pretty much the opposite conclusion 😆. The intuition is to think about the fields: for an `abstract` property, we know that the abstract definition cannot introduce a property and `base.Str` is illegal. For virtual properties, however, an override introduces a new slot and `base.Str` could produce a `null` where `Str` wouldn't. Look at `SetsRequiredMembersHonoredForPropertyOverride_06`; the `DerivedDevived` constructor has 2 warnings on it. One is for the current type implementation of `Str`, and one is for the non-`abstract` overridden `base.Str` property.

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
RikkiGibson: I didn't see any error in the following scenario: [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoBBmABM3AYVwG9NcL88A3KAJwBcBXCAG3xQAZcAFOgewAOpXAHMApgwDcuAM6SZAX0zKMmHPiS4AIuLpRq4gCa4QRUuUoa64gI5N6x3P0N19R8R258hIidLkFXFVFIA==).

```cs
using System;
public class C {
    public virtual string Prop { get; set; }
}

public class Derived : C {
    public required override string Prop { get; set; }
}
```

Was there something I misunderstood here?

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
RikkiGibson: > the DerivedDevived constructor has 2 warnings on it.

I think I forgot to ask, how should user resolve both warnings from the context of that constructor?

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
333fred: Apparently I don't remember my own feature...

src/Compilers/CSharp/Test/Symbol/Symbols/RequiredMembersTests.cs
333fred: > I guess constructor needs to assign both Str and base.Str separately?

Correct. I don't _think_ there's a slot they can't assign outside of existing holes, such as explicit interface implementation.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -928,50 +943,94 @@ static ImmutableArray<Symbol> membersToBeInitialized(NamedTypeSymbol containingT
...
-                            var builder = ArrayBuilder<Symbol>.GetInstance(members.Length + requiredMembers.Count);
+                            var builder = ArrayBuilder<Symbol>.GetInstance(members.Length + baseRequiredMembers.Count);
                             builder.AddRange(members);
-                            foreach (var (_, requiredMember) in requiredMembers)
+                            foreach (var (_, requiredMember) in baseRequiredMembers)
                             {
                                 // We want to assume that all required members were _not_ set by the chained constructor
-                                builder.AddRange(getAllMembersToBeDefaulted(requiredMember));
+
+                                // We exclude any members that are abstract and have an implementation in the current type, when that implementation passes the same heuristics
```
RikkiGibson, line 959: It might be good to expand the comment to explain why it's good to do this filtering only when base is abstract.

RikkiGibson (approved):
I meant to hit approve before, must have misclicked.

333fred:

AlekseyTs:
@333fred Can you think of any VB scenario affected as well?

333fred:
> @333fred Can you think of any VB scenario affected as well?

VB [cannot subclass types with required members](https://github.com/dotnet/roslyn/blob/633f34af4fe019eaf0c2e830bdc07c891ecd49f7/src/Compilers/VisualBasic/Portable/Symbols/Source/SourceNamedTypeSymbol.vb#L1408-L1411), and also does not have NRT, so I do not think it can encounter this issue.

AlekseyTs (approved):
LGTM (commit 4)

PR #76370, opened 2024-12-11
Reduce allocations in WithUsingNamespacesAndTypesBinder.GetForwardedToAssemblyInUsingNamespaces

ToddGrun:
This method shows up as 1.1% of allocations in the typing scenario in the csharp editing speedometer test. Of that, 0.6% is in the string.Concat call merging the Namespace, '.', and type name.

Instead of creating and passing this merged string to GetForwardedToAssembly, we can instead construct a MetadataTypeName to pass to that method. That struct can either be constructed using a fullname or a namespace/typename combination.  In the caller where we are trying to get rid of the string concat, we can pass in the namespace/typename and avoid the concat operation completely.

The other caller to GetForwardedToAssembly also now passes in a MetadataTypeName, but it's constructed from the fullname as that method used to do itself before taking in the MetadataTypeName.

image


JoeRobich (approved):

ToddGrun:
@dotnet/roslyn-compiler ptal

333fred (approved):

cston (approved):

PR #76375, opened 2024-12-11
Remove an enumerator allocation in SymbolCompletionItem.CreateWorker

ToddGrun:
This is a small allocation, only showing up as about 0.1% of allocationsi n the typing scenario in the csharp editing speedometer test. However, the change is fairly simple, and I prefer the change as it passes around ImmutableArrays instead of IReadOnlyLists.

*** relevant allocations from the csharp editing speedometer test ***
image

### DIFF
--- a/src/Features/Core/Portable/Completion/CommonCompletionUtilities.cs
+++ b/src/Features/Core/Portable/Completion/CommonCompletionUtilities.cs
@@ -82,28 +82,6 @@ internal static class CommonCompletionUtilities
         return true;
     }
 
-    public static Func<CancellationToken, Task<CompletionDescription>> CreateDescriptionFactory(
-        SolutionServices workspaceServices,
-        SemanticModel semanticModel,
-        int position,
-        ISymbol symbol,
-        SymbolDescriptionOptions options)
-    {
-        return CreateDescriptionFactory(workspaceServices, semanticModel, position, options, [symbol]);
-    }
-
-    public static Func<CancellationToken, Task<CompletionDescription>> CreateDescriptionFactory(
-        SolutionServices workspaceServices, SemanticModel semanticModel, int position, SymbolDescriptionOptions options, IReadOnlyList<ISymbol> symbols)
-    {
-        return c => CreateDescriptionAsync(workspaceServices, semanticModel, position, symbols, options, supportedPlatforms: null, cancellationToken: c);
-    }
-
-    public static Func<CancellationToken, Task<CompletionDescription>> CreateDescriptionFactory(
-        SolutionServices workspaceServices, SemanticModel semanticModel, int position, IReadOnlyList<ISymbol> symbols, SymbolDescriptionOptions options, SupportedPlatformData supportedPlatforms)
-    {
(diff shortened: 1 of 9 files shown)


src/Features/Core/Portable/Completion/CommonCompletionUtilities.cs
```diff
@@ -82,28 +82,6 @@ public static bool IsStartingNewWord(SourceText text, int characterPosition, Fun
         return true;
     }
 
-    public static Func<CancellationToken, Task<CompletionDescription>> CreateDescriptionFactory(
```
ToddGrun, line 85: >public static Func<CancellationToken, Task<CompletionDescription>> CreateDescriptionFactory( [](http://example.com/codeflow?start=4&length=92)

These don't appear to be used anymore.

src/Features/Core/Portable/Completion/Providers/SymbolCompletionItem.cs
```diff
@@ -50,16 +50,14 @@ private static CompletionItem CreateWorker(
...
             builder.Add(KeyValuePairUtil.Create(InsertionTextProperty, insertionText));
-        }
 
         builder.Add(KeyValuePairUtil.Create("ContextPosition", contextPosition.ToString()));
         AddSupportedPlatforms(builder, supportedPlatforms);
         symbolEncoder(symbols, builder);
 
         tags = tags.NullToEmpty();
-        if (symbols.All(symbol => symbol.IsObsolete()) && !tags.Contains(WellKnownTags.Deprecated))
+        if (!tags.Contains(WellKnownTags.Deprecated) && symbols.All(static symbol => symbol.IsObsolete()))
```
ToddGrun, line 60: >!tags.Contains(WellKnownTags.Deprecated) [](http://example.com/codeflow?start=12&length=40)

order switched here as the IsObsolete checks allocate quite a bit. I doubt the caller has this tag set already in tags, but this seemed nicer.

src/Features/Core/Portable/ExternalAccess/Pythia/Api/PythiaCompletionProviderBase.cs
```diff
@@ -42,7 +43,7 @@ public static CompletionItem CreateSymbolCompletionItem(
         SupportedPlatformData? supportedPlatforms = null,
         ImmutableDictionary<string, string>? properties = null,
         ImmutableArray<string> tags = default)
-        => SymbolCompletionItem.CreateWithSymbolId(displayText, displayTextSuffix: null, symbols, rules, contextPosition, sortText, insertionText,
+        => SymbolCompletionItem.CreateWithSymbolId(displayText, displayTextSuffix: null, symbols.ToImmutableArray(), rules, contextPosition, sortText, insertionText,
```
ToddGrun, line 46: >ToImmutableArray [](http://example.com/codeflow?start=97&length=16)

This is potentially an allocation where it wasn't before, but only if they weren't using an ImmutableArray and it's limited to Pythia.

PR #76399, opened 2024-12-12
Optimize ProjectDependencyGraph.DoesProjectTransitivelyDependOnProject

ToddGrun:
I believe that this method unnecessarilly obtains the reverse dependencies when the forward references are already defined. If the forward references are present, we should just use the result of trying to find the potential dependencies in it.

DoesProjectTransitivelyDependOnProject shows as 2.9% of allocations and 1.6% of CPU in the typing scenario in the scharp editing speedometer test. Will kick off a test insertion and see what those numbers are with this change.

*** CPU before change ***
image

*** Allocations before change ***
image


CyrusNajmabadi (approved):

jasonmalinowski (approved):

ToddGrun:
Test Insertion: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/598266

PR #76435, opened 2024-12-14
Pass caller argument expression through RoslynDebug.Assert

jjonescz:
Makes RoslynDebug.Assert more consistent with Debug.Assert. Without this, on .NET 9, the message `"condition"` is effectively being passed into `Debug.Assert` which is not very useful.

### DIFF
--- a/src/Compilers/Core/Portable/InternalUtilities/Debug.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/Debug.cs
@@ -13,12 +13,21 @@ namespace Roslyn.Utilities
     internal static class RoslynDebug
     {
         /// <inheritdoc cref="Debug.Assert(bool)"/>
+#if NET9_0_OR_GREATER
+        [OverloadResolutionPriority(-1)]
+#endif
         [Conditional("DEBUG")]
         public static void Assert([DoesNotReturnIf(false)] bool condition) => Debug.Assert(condition);
 
         /// <inheritdoc cref="Debug.Assert(bool, string)"/>
         [Conditional("DEBUG")]
-        public static void Assert([DoesNotReturnIf(false)] bool condition, string message)
+        public static void Assert([DoesNotReturnIf(false)] bool condition,
+#if NET
+            [CallerArgumentExpression(nameof(condition))] string? message = null
+#else
+            string message
+#endif
+            )
             => Debug.Assert(condition, message);
 
         /// <inheritdoc cref="Debug.Assert(bool, string)"/>
--- a/src/Compilers/Core/Portable/InternalUtilities/Debug.cs
+++ b/src/Compilers/Core/Portable/InternalUtilities/Debug.cs
@@ -13,9 +13,6 @@ namespace Roslyn.Utilities
     internal static class RoslynDebug
     {
         /// <inheritdoc cref="Debug.Assert(bool)"/>
-#if NET9_0_OR_GREATER
-        [OverloadResolutionPriority(-1)]
-#endif
         [Conditional("DEBUG")]
         public static void Assert([DoesNotReturnIf(false)] bool condition) => Debug.Assert(condition);
 


src/Compilers/Core/Portable/InternalUtilities/Debug.cs
```diff
@@ -13,12 +13,21 @@ namespace Roslyn.Utilities
     internal static class RoslynDebug
     {
         /// <inheritdoc cref="Debug.Assert(bool)"/>
+#if NET9_0_OR_GREATER
+        [OverloadResolutionPriority(-1)]
```
Youssef1313, line 17: Is that really needed? Can't the other overload exist alone?
On .NET, they had to do this for backcompat. But as RoslynDebug is internal, I think it should be safe to just have the right overload?

jjonescz: Might not be needed as you say. But I don't want to risk some other DLL depending on this (via InternalsVisibleTo) and then getting MissingMethodExceptions.

jaredpar: The way our IVT is setup (no IVT outside our repo) it should be safe to remove this if the repo still compiles.

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
```diff
@@ -13,12 +13,21 @@ namespace Roslyn.Utilities
...
+#endif
         [Conditional("DEBUG")]
         public static void Assert([DoesNotReturnIf(false)] bool condition) => Debug.Assert(condition);
 
         /// <inheritdoc cref="Debug.Assert(bool, string)"/>
         [Conditional("DEBUG")]
-        public static void Assert([DoesNotReturnIf(false)] bool condition, string message)
+        public static void Assert([DoesNotReturnIf(false)] bool condition,
+#if NET
+            [CallerArgumentExpression(nameof(condition))] string? message = null
```
jaredpar, line 26: Why are we using the `#if NET` condition here? I get that only the core version of `Debug.Assert` has this overload but not sure if that should stop us from making `RoslynDebug.Assert` better.

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
jjonescz: Alright, I will remove the overload resolution priority attribute.

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
```diff
@@ -18,7 +18,13 @@ internal static class RoslynDebug
...
```
AlekseyTs, line 23: >[CallerArgumentExpression(nameof(condition))] [](http://example.com/codeflow?start=12&length=45)

It is not clear to me how this is going to affect any existing usages of ```RoslynDebug.Assert```, given that the parameter was not optional and, therefore, all call sites supply its value explicitly.

src/Compilers/Core/Portable/InternalUtilities/Debug.cs
```diff
@@ -18,7 +18,13 @@ internal static class RoslynDebug
 
         /// <inheritdoc cref="Debug.Assert(bool, string)"/>
         [Conditional("DEBUG")]
-        public static void Assert([DoesNotReturnIf(false)] bool condition, string message)
+        public static void Assert([DoesNotReturnIf(false)] bool condition,
```
AlekseyTs, line 21: >Assert [](http://example.com/codeflow?start=27&length=6)

Do we really need to maintain a private version of Assert? I was under impression that standard ```Debug.Assert``` has been appropriately decorated.

333fred: Agreed with this. I thought the main reason we still kept `RoslynDebug.Assert` around was that we haven't removed all of the callsites.

PR #76449, opened 2024-12-16
Map the path in InterceptableLocation.GetDisplayPath

### LINKED ISSUE #76126: InterceptableLocation.GetDisplayLocation breaks deterministic build
**Version Used**: .NET SDK 9.0.100 

The current implementation of `InterceptableLocation.GetDisplayName()` which is used in source generator to output human-readable location for the generated code, but it outputs full path which breaks deterministic build as the the generated source code could be embedded to the PDB file. Should the output location be normalized to restore the deterministic build?

https://github.com/dotnet/roslyn/blob/6cc106c0eaa9b0ae070dba3138a23aeab9b50c13/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs#L74-L78

https://github.com/dotnet/roslyn/blob/6cc106c0eaa9b0ae070dba3138a23aeab9b50c13/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs#L35-L37

The generated non-deterministic source code looks like below.

```csharp
    [GeneratedCode("Microsoft.Extensions.Configuration.Binder.SourceGeneration", "9.0.11.2809")]
    file static class BindingExtensions
    {
        #region IConfiguration extensions.
        /// <summary>Attempts to bind the given object instance to configuration values by matching property names against configuration keys recursively.</summary>
        [InterceptsLocation(1, "aSfEOMF93KOjHfGLzFdenPgFAABPbmVDb2xsZWN0b3JMb2dFeHBvcnRQcm9jZXNzb3JCdWlsZGVyLmNz")] // D:\a\opentelemetry-dotnet-contrib\opentelemetry-dotnet-contrib\src\OpenTelemetry.Exporter.OneCollector\Logs\OneCollectorLogExportProcessorBuilder.cs(47,74)
```

RikkiGibson:
Closes #76126

### DIFF
--- a/src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
+++ b/src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
@@ -5229,7 +5229,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             var lineNumberOneIndexed = lineSpan.Line + 1;
             var characterNumberOneIndexed = lineSpan.Character + 1;
 
-            return new InterceptableLocation1(checksum, path, nameSyntax.Position, lineNumberOneIndexed, characterNumberOneIndexed);
+            return new InterceptableLocation1(checksum, path, Compilation.Options.SourceReferenceResolver, nameSyntax.Position, lineNumberOneIndexed, characterNumberOneIndexed);
         }
 #nullable disable
 
--- a/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs
+++ b/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs
@@ -53,12 +53,13 @@ internal sealed class InterceptableLocation1 : InterceptableLocation
 
     private readonly ImmutableArray<byte> _checksum;
     private readonly string _path;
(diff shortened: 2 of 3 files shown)


src/Compilers/CSharp/Portable/Compilation/CSharpSemanticModel.cs
```diff
@@ -5229,7 +5229,7 @@ internal InterceptableLocation GetInterceptableLocationInternal(SyntaxNode nameS
             var lineNumberOneIndexed = lineSpan.Line + 1;
             var characterNumberOneIndexed = lineSpan.Character + 1;
 
-            return new InterceptableLocation1(checksum, path, nameSyntax.Position, lineNumberOneIndexed, characterNumberOneIndexed);
+            return new InterceptableLocation1(checksum, path, Compilation.Options.SourceReferenceResolver, nameSyntax.Position, lineNumberOneIndexed, characterNumberOneIndexed);
```
jaredpar, line 5232: A bit annoying that this is a nullable reference type when we could have used `SourceFileResolver.Default`. Not sure there is a ton of value in changing it now.

jaredpar (approved):

cston (approved):

PR #76452, opened 2024-12-16
Move Default VB Version to 17.13

AlekseyTs:


333fred (approved):
Is there some SDK component that will need to adjust to this @jcouv?

AlekseyTs:

src/Compilers/VisualBasic/Test/CommandLine/CommandLineTests.vb
```diff
@@ -2104,6 +2104,7 @@ End Module").Path
             Assert.Equal(LanguageVersion.VisualBasic16, LanguageVersion.VisualBasic16.MapSpecifiedToEffectiveVersion())
             Assert.Equal(LanguageVersion.VisualBasic16_9, LanguageVersion.VisualBasic16_9.MapSpecifiedToEffectiveVersion())
             Assert.Equal(LanguageVersion.VisualBasic17_13, LanguageVersion.VisualBasic17_13.MapSpecifiedToEffectiveVersion())
+            Assert.Equal(LanguageVersion.VisualBasic17_13, LanguageVersion.Default.MapSpecifiedToEffectiveVersion())
```
jcouv, line 2107: Looking at the comment below points to some instructions for adding a new VB language version. 
But the first bullet is not applicable (no UpgradeProject for VB) and the last bullet is no longer applicable (that document doesn't include specific version numbers since we added `/LangVersion:?`). Would you mind removing them?
 
```
  ' When a new version is added, this test will break. This list must be checked:
  ' - update the "UpgradeProject" codefixer (not yet supported in VB)
  ' - update all the tests that call this canary
  ' - update the command-line documentation (CommandLine.md)
```

jcouv:
@davkean @drewnoakes We've added a new language version for VB (VB 17.13) and made it the new default. Do we need any corresponding changes in SDK or project system?
Looking at previous PRs related to language version ([here](https://github.com/dotnet/roslyn/commits/main/src/Compilers/VisualBasic/Portable/LanguageVersion.vb) and [here](https://github.com/dotnet/project-system/pulls?q=is%3Apr+author%3Ajcouv+is%3Aclosed)) and from my recollection, I believe the roslyn change is enough, but it's been a while so would like to confirm.

jcouv (approved):

PR #76486, opened 2024-12-18
FIx build errors with live (SCI) runtime dependencies when targeting NetPrevious (net9.0) in source-build (VMR)

ViktorHofer:
Unblocks https://github.com/dotnet/sdk/pull/45427

Fixes the following two errors:
- src/Features/Core/Portable/ExternalAccess/Watch/Api/WatchHotReloadService.cs(217,13): error CS1503: Argument 4: cannot convert from 'System.Collections.Immutable.ImmutableHashSet<Microsoft.CodeAnalysis.Project>' to 'System.Collections.Generic.IReadOnlySet<Microsoft.CodeAnalysis.Project>' [src/Features/Core/Portable/Microsoft.CodeAnalysis.Features.csproj::TargetFramework=net9.0]
- src/roslyn/src/Features/Core/Portable/ExternalAccess/Watch/Api/WatchHotReloadService.cs(218,13): error CS1503: Argument 5: cannot convert from 'System.Collections.Immutable.ImmutableHashSet<Microsoft.CodeAnalysis.Project>' to 'System.Collections.Generic.IReadOnlySet<Microsoft.CodeAnalysis.Project>' [src/roslyn/src/Features/Core/Portable/Microsoft.CodeAnalysis.Features.csproj::TargetFramework=net9.0]

The underlying SDK being used is "10.0.100-alpha.1.24555.54"

### DIFF
--- a/src/Compilers/Core/Portable/Microsoft.CodeAnalysis.csproj
+++ b/src/Compilers/Core/Portable/Microsoft.CodeAnalysis.csproj
@@ -31,9 +31,9 @@
       will import everything but content files from Microsoft.CodeAnalysis.Analyzers, specifically, analyzers.
     -->
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Condition="'$(DotNetBuildSourceOnly)' != 'true'" PrivateAssets="ContentFiles" />
-    <PackageReference Include="System.Collections.Immutable" />
+    <PackageReference Include="System.Collections.Immutable" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
     <PackageReference Include="System.Memory" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
-    <PackageReference Include="System.Reflection.Metadata" />
+    <PackageReference Include="System.Reflection.Metadata" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="System.Text.Encoding.CodePages" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
--- a/src/Features/Core/Portable/Microsoft.CodeAnalysis.Features.csproj
+++ b/src/Features/Core/Portable/Microsoft.CodeAnalysis.Features.csproj
@@ -139,7 +139,7 @@
     <PackageReference Include="Microsoft.CodeAnalysis.AnalyzerUtilities" />
     <PackageReference Include="Microsoft.CodeAnalysis.Elfie" Condition="'$(DotNetBuildSourceOnly)' != 'true'" />
     <PackageReference Include="Microsoft.DiaSymReader" />
-    <PackageReference Include="System.Text.Json" />
+    <PackageReference Include="System.Text.Json" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
   </ItemGroup>
   <Import Project="..\..\..\Analyzers\Core\Analyzers\Analyzers.projitems" Label="Shared" />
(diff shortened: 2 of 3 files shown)


CyrusNajmabadi:
Why did this break?  Why can't an immutable hash set be passed to a read only set?

ViktorHofer:
Given that this only happened in the source-build leg in the VMR I double checked whether the SBRP Microsoft.NETCore.App.Ref/9.0.0 ref/System.Collections.Immutable.dll assembly has a wrong definition for `ImmutableHashSet<T>`. It correctly imports `IReadOnlySet<T>`:

image

Not sure what's going on here.

ViktorHofer:
Oh I see what's going on. This assembly reference is used:

/__w/1/vmr/src/roslyn/artifacts/sb/package-cache/system.collections.immutable/10.0.0-alpha.1.24617.1/lib/netstandard2.0/System.Collections.Immutable.dll

AlekseyTs (approved):

src/Compilers/Core/Portable/Microsoft.CodeAnalysis.csproj
```diff
@@ -31,9 +31,9 @@
       will import everything but content files from Microsoft.CodeAnalysis.Analyzers, specifically, analyzers.
     -->
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Condition="'$(DotNetBuildSourceOnly)' != 'true'" PrivateAssets="ContentFiles" />
-    <PackageReference Include="System.Collections.Immutable" />
+    <PackageReference Include="System.Collections.Immutable" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
     <PackageReference Include="System.Memory" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
-    <PackageReference Include="System.Reflection.Metadata" />
+    <PackageReference Include="System.Reflection.Metadata" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
```
jaredpar, line 36: Changes of this nature can't be merged unless we do a full VS validation insertion. This is too core of a change to go through without one.

ViktorHofer:
> Have that TFM be both NetCurrent and NetPrevious.

I thought that's what we did? Maybe I misunderstood the original change. @jaredpar who can help with that?

jaredpar:
> I thought that's what we did? 

No it was changed to be `$(NetPrevious)` only

https://github.com/dotnet/roslyn/pull/76354/files#diff-00f5693d71dbe7582970cb797b17a1d85497c08a3e2838c8ea7ce2026526b07aR58

> Btw how would that solve the above described issue?

Once everything has a `$(NetCurrent)` target I would assume that fixes the fact it's falling back to a `netstandard2.0` target.

ViktorHofer:
> Once everything has a $(NetCurrent) target I would assume that fixes the fact it's falling back to a netstandard2.0 target.

The issue here is that as long as you target `NetPrevious` in the source-build product, you will get the netstandard2.0 assemblies of runtime packages. Removing `NetPrevious` from the list should obviously fix the issues but I don't understand how consistently adding `NetPrevious` to all properties would fix this.

jaredpar:
> The issue here is that as long as you target NetPrevious in the source-build product, you will get the netstandard2.0 assemblies of runtime packages. 

Not sure what to do here then. Roslyn has built `NetPrevious` and `NetCurrent` in our SB TFMs for quite some time. It sounds like source build has changed such that this is not valid anymore? If so we can adjust but very unlikely to happen anytime this year. Too big of a change.

ViktorHofer:
> Not sure what to do here then. Roslyn has built NetPrevious and NetCurrent in our SB TFMs for quite some time. It sounds like source build has changed such that this is not valid anymore?

Nothing changed outside of roslyn. Runtime packages always only had `NetCurrent` and `netstandard2.0` assemblies in their source-built packages. The change that regressed this is https://github.com/dotnet/roslyn/commit/e75664fa1ebe4b064ea84814a006b2d6e7fe2e11. It wasn't caught in your repo source-build validation because you don't use the "live" runtime packages there.

jaredpar:
@tmat, @jasonmalinowski you all will need to decide how to handle this. At the moment the workspace layer is using `NetPrevious` in source build. That will end up using `netstandard2.0` versions of some libraries that have different API sets. The change https://github.com/dotnet/roslyn/commit/e75664fa1ebe4b064ea84814a006b2d6e7fe2e11 is hitting one of these gaps.

Overall we may want to just delete `NetPrevious` from full source build. That's not happening until Jan though.

ViktorHofer:
This can be closed now as product source-build only targets NetCurrent now.

PR #76492, opened 2024-12-18
Propagate marked attributes from method type parameters to type parameters of synthesized types.

AlekseyTs:
Related to https://github.com/dotnet/roslyn/issues/73920.

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceTypeParameterSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceTypeParameterSymbol.cs
@@ -20,7 +20,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
     /// <summary>
     /// Base class for type and method type parameters.
     /// </summary>
-    internal abstract class SourceTypeParameterSymbolBase : TypeParameterSymbol, IAttributeTargetSymbol
+    internal abstract class SourceTypeParameterSymbol : TypeParameterSymbol, IAttributeTargetSymbol
     {
         private readonly ImmutableArray<SyntaxReference> _syntaxRefs;
         private readonly ImmutableArray<Location> _locations;
@@ -31,7 +31,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
         private CustomAttributesBag<CSharpAttributeData> _lazyCustomAttributesBag;
         private TypeParameterBounds _lazyBounds = TypeParameterBounds.Unset;
 
-        protected SourceTypeParameterSymbolBase(string name, int ordinal, ImmutableArray<Location> locations, ImmutableArray<SyntaxReference> syntaxRefs)
+        protected SourceTypeParameterSymbol(string name, int ordinal, ImmutableArray<Location> locations, ImmutableArray<SyntaxReference> syntaxRefs)
         {
             Debug.Assert(!syntaxRefs.IsEmpty);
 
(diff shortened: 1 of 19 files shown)


AlekseyTs:

AlekseyTs:

src/Compilers/CSharp/Test/Semantic/Semantics/LambdaTests.cs
```diff
@@ -8553,5 +8553,40 @@ public void ParamsArray_ThisModifier_02()
                 // var lam = (params this int[] xs) => xs.Length;
                 Diagnostic(ErrorCode.ERR_ThisInBadContext, "this").WithLocation(1, 19));
         }
+
+        [Fact]
+        public void CompilerLoweringPreserveAttribute_01()
```
333fred, line 8558: Is it worth adding some tests for lambdas observed as expression trees, to make sure there's nothing different about how we do the machinery lowering for that scenario?

333fred (approved):

src/Compilers/CSharp/Test/Semantic/Semantics/LambdaTests.cs
AlekseyTs: > Is it worth adding some tests for lambdas observed as expression trees, to make sure there's nothing different about how we do the machinery lowering for that scenario?

I do not think so.

AlekseyTs:

src/Compilers/CSharp/Portable/Symbols/Source/SourceTypeParameterSymbol.cs
```diff
@@ -602,32 +602,71 @@ private TypeParameterConstraintKind GetConstraintKinds()
         }
     }
 
-    internal sealed class SourceMethodTypeParameterSymbol : SourceTypeParameterSymbolBase
+    internal abstract class SourceMethodTypeParameterSymbol : SourceTypeParameterSymbol
```
cston, line 605: >abstract class SourceMethodTypeParameterSymbol [](http://example.com/codeflow?start=13&length=46)

What does this refactoring enable?

src/Compilers/CSharp/Portable/Symbols/Synthesized/SynthesizedSubstitutedTypeParameterSymbol.cs
```diff
@@ -34,6 +34,19 @@ public override bool IsImplicitlyDeclared
 
         internal override void AddSynthesizedAttributes(PEModuleBuilder moduleBuilder, ref ArrayBuilder<CSharpAttributeData> attributes)
         {
+            if (ContainingSymbol.Kind == SymbolKind.NamedType &&
+                _underlyingTypeParameter.OriginalDefinition is SourceMethodTypeParameterSymbol &&
```
cston, line 38: >is SourceMethodTypeParameterSymbol [](http://example.com/codeflow?start=60&length=34)

Consider capturing in a local and reusing below: `is SourceMethodTypeParameterSymbol underlyingDefinition`

cston (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourceTypeParameterSymbol.cs
AlekseyTs: > What does this refactoring enable?

All implementations of method type parameters in source have a common base distinct from the base of type type parameters

src/Compilers/CSharp/Portable/Symbols/Synthesized/SynthesizedSubstitutedTypeParameterSymbol.cs
AlekseyTs: > > is SourceMethodTypeParameterSymbol
> 
> Consider capturing in a local and reusing below: `is SourceMethodTypeParameterSymbol underlyingDefinition`

I'll make this adjustment in the next PR

PR #76505, opened 2024-12-18
Only target NetCurrent for source build

mthalman:
The `NetCurrent` property is the only value that should be targeted in the context of source build. Otherwise, when `NetCurrent` is `net10.0` and `NetPrevious` is `net9.0`, you can run into this error when building with the latest .NET 10 SDK build:

```
/repos/dotnet/.dotnet/sdk/10.0.100-alpha.1.24611.6/Microsoft.Common.CurrentVersion.targets(1889,5): Project '../MSBuild.BuildHost/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.csproj' targets 'net10.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v9.0'. [/repos/dotnet/src/roslyn/src/Workspaces/Core/MSBuild/Microsoft.CodeAnalysis.Workspaces.MSBuild.csproj]
```

Found as part of the work on https://github.com/dotnet/sdk/pull/45435.

### DIFF
--- a/eng/targets/TargetFrameworks.props
+++ b/eng/targets/TargetFrameworks.props
@@ -40,7 +40,7 @@
         <!-- TODO until we figure out what is up with NetPrevious -->
         <NetPrevious>$(NetMinimum)</NetPrevious>
         <NetRoslyn>$(NetPrevious)</NetRoslyn>
-        <NetRoslynSourceBuild>$(NetCurrent);$(NetPrevious)</NetRoslynSourceBuild>
+        <NetRoslynSourceBuild>$(NetCurrent)</NetRoslynSourceBuild>
         <NetRoslynAll>$(NetCurrent);$(NetPrevious)</NetRoslynAll>
         <NetRoslynBuildHostNetCoreVersion>$(NetPrevious)</NetRoslynBuildHostNetCoreVersion>
         <NetRoslynNext>$(NetPrevious)</NetRoslynNext>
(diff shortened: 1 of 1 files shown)


ViktorHofer:
@mthalman this got fixed / worked around with https://github.com/dotnet/roslyn/commit/58d7d7b8b31f453e044da8e6aa2dcdce86e759a1 already.

mthalman:
> @mthalman this got fixed / worked around with [58d7d7b](https://github.com/dotnet/roslyn/commit/58d7d7b8b31f453e044da8e6aa2dcdce86e759a1) already.

That doesn't solve the problem. It still attempts to build Microsoft.CodeAnalysis with the net9.0 TFM in a source build configuration which leads to the following errors:

```
CSC error CS0006: Metadata file '/repos/dotnet/src/roslyn/artifacts/sb/package-cache/microsoft.netcore.app.ref/9.0.0/analyzers/dotnet/cs/Microsoft.Interop.ComInterfaceGenerator.dll' could not be found [/repos/dotnet/src/roslyn/src/Compilers/Core/Portable/Microsoft.CodeAnalysis.csproj]
CSC error CS0006: Metadata file '/repos/dotnet/src/roslyn/artifacts/sb/package-cache/microsoft.netcore.app.ref/9.0.0/analyzers/dotnet/cs/Microsoft.Interop.LibraryImportGenerator.dll' could not be found [/repos/dotnet/src/roslyn/src/Compilers/Core/Portable/Microsoft.CodeAnalysis.csproj]
CSC error CS0006: Metadata file '/repos/dotnet/src/roslyn/artifacts/sb/package-cache/microsoft.netcore.app.ref/9.0.0/analyzers/dotnet/cs/Microsoft.Interop.SourceGeneration.dll' could not be found [/repos/dotnet/src/roslyn/src/Compilers/Core/Portable/Microsoft.CodeAnalysis.csproj]
CSC error CS0006: Metadata file '/repos/dotnet/src/roslyn/artifacts/sb/package-cache/microsoft.netcore.app.ref/9.0.0/analyzers/dotnet/cs/System.Text.Json.SourceGeneration.dll' could not be found [/repos/dotnet/src/roslyn/src/Compilers/Core/Portable/Microsoft.CodeAnalysis.csproj]
CSC error CS0006: Metadata file '/repos/dotnet/src/roslyn/artifacts/sb/package-cache/microsoft.netcore.app.ref/9.0.0/analyzers/dotnet/cs/System.Text.RegularExpressions.Generator.dll' could not be found [/repos/dotnet/src/roslyn/src/Compilers/Core/Portable/Microsoft.CodeAnalysis.csproj]
```

This happens because it's loading the 9.0 targeting back from SBRP which has no analyzers. But the point is, it shouldn't be attempting to target net9.0 in full source build.

ViktorHofer:
Yes, I agree that when building roslyn from source `NetPrevious` should get built. We had a discussion about this in linked PR and it sounds like others agree on this as well and we should make the change eventually. From the changes in this PR, `NetPrevious` still gets built. I would expect L57 and L58 to also need to change.

PR #71546, opened 2024-01-09
Do not report conflicting types errors for source-generated APIs

333fred:
It is a common scenario for source generators to make `internal` attributes in an assembly via our `PostInit` callbacks. This is mostly fine, but can result in annoying CS0436 "this type conflicts" warnings. To better support this scenario, we suppress this warning for this specific case. In order to plumb this knowledge through, I did have to update a bit larger tail than I would have otherwise preferred, but we now have the building block for implementing https://github.com/dotnet/roslyn/issues/56810 as well.

PR #76523, opened 2024-12-19
Allow users to define EmbeddedAttribute

333fred:
Alternative approach to https://github.com/dotnet/roslyn/pull/71546. Today, our enforcement of whether to allow users to define `Microsoft.CodeAnalysis.EmbeddedAttribute` is inconsistent; if we need to generate it, we error if the user defines one. But if the we don't need to generate it (as we increasingly do not since .NET 6), we allow the user to define their own version. This PR standardizes our enforcement, and enables the compiler to use the user-declared attribute instead generating one. We also document the breaking change from the standardized enforcement.

### DIFF
--- a/docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
+++ b/docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
@@ -156,3 +156,24 @@ struct S
     public static void M([UnscopedRef] ref int x) { }
 }
 ```
+
+## `Microsoft.CodeAnalysis.EmbeddedAttribute` is validated on declaration
+
+***Introduced in Visual Studio 2022 version 17.13***
+
+The compiler now validates the shape of `Microsoft.CodeAnalysis.EmbeddedAttribute` when declared in source. Previously, the compiler
+would allow user-defined declarations of this attribute, but only when it didn't need to generate one itself. We now validate that:
+
+1. It must be internal
+2. It must be a class
+3. It must be sealed
+4. It must be non-static
+5. It must have an internal or public parameterless constructor
+6. It must be attributed with itself
+
+```cs
+namespace Microsoft.CodeAnalysis;
+
+// Previously, sometimes allowed. Now, CS9271
+public class EmbeddedAttribute : Attribute {}
+```
--- a/docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
+++ b/docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
@@ -169,7 +169,8 @@ would allow user-defined declarations of this attribute, but only when it didn't
 3. It must be sealed
 4. It must be non-static
 5. It must have an internal or public parameterless constructor
-6. It must be attributed with itself
+6. It must inherit from System.Attribute.
+7. It must be allowed on any type declaration (class, struct, interface, enum, or delegate)
 
 ```cs
 namespace Microsoft.CodeAnalysis;
--- a/docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
+++ b/docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
@@ -40,7 +40,7 @@ End Class
 ***Introduced in Visual Studio 2022 version 17.13***
 
 The compiler now validates the shape of `Microsoft.CodeAnalysis.EmbeddedAttribute` when declared in source. Previously, the compiler
-would allow user-defined declarations of this attribute, but only when it didn't need to generate one itself. We now validate that:
+would allow user-defined declarations of this attribute with any shape. We now validate that:
 
 1. It must be Friend
 2. It must be a Class
@@ -50,9 +50,12 @@ would allow user-defined declarations of this attribute, but only when it didn't
 6. It must inherit from System.Attribute.
 7. It must be allowed on any type declaration (class, struct, interface, enum, or delegate)
 
-```cs
-namespace Microsoft.CodeAnalysis;
+```vb
+Namespace Microsoft.CodeAnalysis
 
-// Previously, sometimes allowed. Now, CS9271
-public class EmbeddedAttribute : Attribute {}
-`
+    ' Previously allowed. Now, CS9271
+    Public Class EmbeddedAttribute
+        Inherits Attribute
+    End Class
+End Namespace
+```
--- a/docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
+++ b/docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
@@ -48,12 +48,12 @@ would allow user-defined declarations of this attribute with any shape. We now v
 4. It must not be a Module
 5. It must have a Public parameterless constructor
 6. It must inherit from System.Attribute.
-7. It must be allowed on any type declaration (class, struct, interface, enum, or delegate)
+7. It must be allowed on any type declaration (Class, Struct, Interface, Enum, or Delegate)
 
 ```vb
 Namespace Microsoft.CodeAnalysis
 
-    ' Previously allowed. Now, CS9271
+    ' Previously allowed. Now, BC37335
     Public Class EmbeddedAttribute
         Inherits Attribute
     End Class
--- a/docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
+++ b/docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
@@ -35,3 +35,24 @@ Class C
 End Class
 ```
 
+## `Microsoft.CodeAnalysis.EmbeddedAttribute` is validated on declaration
+
+***Introduced in Visual Studio 2022 version 17.13***
+
+The compiler now validates the shape of `Microsoft.CodeAnalysis.EmbeddedAttribute` when declared in source. Previously, the compiler
+would allow user-defined declarations of this attribute, but only when it didn't need to generate one itself. We now validate that:
+
+1. It must be Friend
+2. It must be a Class
+3. It must be NotInheritable
+4. It must not be a Module
+5. It must have a Public parameterless constructor
+6. It must inherit from System.Attribute.
+7. It must be allowed on any type declaration (class, struct, interface, enum, or delegate)
+
+```cs
+namespace Microsoft.CodeAnalysis;
+
+// Previously, sometimes allowed. Now, CS9271
+public class EmbeddedAttribute : Attribute {}
+`
--- a/docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
+++ b/docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
@@ -48,7 +48,7 @@ would allow user-defined declarations of this attribute with any shape. We now v
 4. It must not be a Module
 5. It must have a Public parameterless constructor
 6. It must inherit from System.Attribute.
-7. It must be allowed on any type declaration (Class, Struct, Interface, Enum, or Delegate)
+7. It must be allowed on any type declaration (Class, Structure, Interface, Enum, or Delegate)
 
 ```vb
 Namespace Microsoft.CodeAnalysis
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -5909,7 +5909,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>The type name '{0}' is reserved to be used by the compiler.</value>
   </data>
   <data name="ERR_EmbeddedAttributeMustFollowPattern" xml:space="preserve">
-    <value>The type 'Microosft.CodeAnalysis.EmbeddedAttribute' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.</value>
+    <value>The type 'Microsoft.CodeAnalysis.EmbeddedAttribute' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.</value>
   </data>
   <data name="ERR_InExtensionMustBeValueType" xml:space="preserve">
     <value>The first 'in' or 'ref readonly' parameter of the extension method '{0}' must be a concrete (non-generic) value type.</value>
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -5908,6 +5908,9 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="ERR_TypeReserved" xml:space="preserve">
     <value>The type name '{0}' is reserved to be used by the compiler.</value>
   </data>
+  <data name="ERR_ReservedTypeMustFollowPattern" xml:space="preserve">
+    <value>The type '{0}' must be non-generic, internal, sealed, non-static, be attributed with 'Microsoft.CodeAnalysis.EmbeddedAttribute', and have a parameterless constructor.</value>
+  </data>
   <data name="ERR_InExtensionMustBeValueType" xml:space="preserve">
     <value>The first 'in' or 'ref readonly' parameter of the extension method '{0}' must be a concrete (non-generic) value type.</value>
   </data>
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -5909,7 +5909,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>The type name '{0}' is reserved to be used by the compiler.</value>
   </data>
   <data name="ERR_ReservedTypeMustFollowPattern" xml:space="preserve">
-    <value>The type '{0}' must be non-generic, internal, sealed, non-static, be attributed with 'Microsoft.CodeAnalysis.EmbeddedAttribute', and have a parameterless constructor.</value>
+    <value>The type '{0}' must be non-generic, internal, sealed, non-static, have a parameterless constructor, and be able to be applied to any type.</value>
   </data>
   <data name="ERR_InExtensionMustBeValueType" xml:space="preserve">
     <value>The first 'in' or 'ref readonly' parameter of the extension method '{0}' must be a concrete (non-generic) value type.</value>
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -5909,7 +5909,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>The type name '{0}' is reserved to be used by the compiler.</value>
   </data>
   <data name="ERR_ReservedTypeMustFollowPattern" xml:space="preserve">
-    <value>The type '{0}' must be non-generic, internal, sealed, non-static, have a parameterless constructor, and be able to be applied to any type.</value>
+    <value>The type '{0}' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.</value>
   </data>
   <data name="ERR_InExtensionMustBeValueType" xml:space="preserve">
     <value>The first 'in' or 'ref readonly' parameter of the extension method '{0}' must be a concrete (non-generic) value type.</value>
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -5908,8 +5908,8 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="ERR_TypeReserved" xml:space="preserve">
     <value>The type name '{0}' is reserved to be used by the compiler.</value>
   </data>
-  <data name="ERR_ReservedTypeMustFollowPattern" xml:space="preserve">
-    <value>The type '{0}' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.</value>
+  <data name="ERR_EmbeddedAttributeMustFollowPattern" xml:space="preserve">
+    <value>The type 'Microosft.CodeAnalysis.EmbeddedAttribute' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.</value>
   </data>
   <data name="ERR_InExtensionMustBeValueType" xml:space="preserve">
     <value>The first 'in' or 'ref readonly' parameter of the extension method '{0}' must be a concrete (non-generic) value type.</value>
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
@@ -1414,7 +1414,6 @@ next:;
         {
             get
             {
-
                 var data = GetEarlyDecodedWellKnownAttributeData();
                 return (data != null && data.HasCodeAnalysisEmbeddedAttribute)
                     // If this is Microsoft.CodeAnalysis.EmbeddedAttribute, we'll synthesize EmbeddedAttribute even if it's not applied.
@@ -1947,8 +1946,8 @@ next:;
                     || !this.DeclaringCompilation.IsAttributeType(this)
                     || (GetAttributeUsageInfo().ValidTargets & expectedTargets) != expectedTargets)
                 {
-                    // The type '{0}' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.
-                    diagnostics.Add(ErrorCode.ERR_ReservedTypeMustFollowPattern, GetFirstLocation(), AttributeDescription.CodeAnalysisEmbeddedAttribute.FullName);
+                    // The type 'Microsoft.CodeAnalysis.EmbeddedAttribute' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.
+                    diagnostics.Add(ErrorCode.ERR_EmbeddedAttributeMustFollowPattern, GetFirstLocation());
                 }
 
             }
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
@@ -1944,11 +1944,10 @@ next:;
                     || !IsSealed
                     || IsStatic
                     || !InstanceConstructors.Any(c => c is { ParameterCount: 0, DeclaredAccessibility: Accessibility.Internal or Accessibility.Public })
-                    || !HasCodeAnalysisEmbeddedAttribute
                     || !this.DeclaringCompilation.IsAttributeType(this)
                     || (GetAttributeUsageInfo().ValidTargets & ExpectedTargets) != ExpectedTargets)
                 {
-                    // The type '{0}' must be non-generic, internal, sealed, non-static, and have a parameterless constructor.
+                    // The type '{0}' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.
                     diagnostics.Add(ErrorCode.ERR_ReservedTypeMustFollowPattern, GetFirstLocation(), AttributeDescription.CodeAnalysisEmbeddedAttribute.FullName);
                 }
 
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
@@ -1922,7 +1922,7 @@ next:;
                     }
                 })
             {
-                // This is a user-defined implement of the special attribute Microsoft.CodeAnalysis.EmbeddedAttribute. It needs to follow specific rules:
+                // This is a user-defined implementation of the special attribute Microsoft.CodeAnalysis.EmbeddedAttribute. It needs to follow specific rules:
                 // 1. It must be internal
                 // 2. It must be a class
                 // 3. It must be sealed
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
@@ -4,6 +4,7 @@
 
 #nullable disable
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
@@ -1413,8 +1414,11 @@ next:;
         {
             get
             {
+
                 var data = GetEarlyDecodedWellKnownAttributeData();
-                return data != null && data.HasCodeAnalysisEmbeddedAttribute;
+                return (data != null && data.HasCodeAnalysisEmbeddedAttribute)
+                    // If this is Microsoft.CodeAnalysis.EmbeddedAttribute, we'll synthesize EmbeddedAttribute even if it's not applied.
+                    || this.IsMicrosoftCodeAnalysisEmbeddedAttribute();
             }
         }
 
@@ -1748,6 +1752,21 @@ next:;
                         ImmutableArray.Create(new TypedConstant(compilation.GetWellKnownType(WellKnownType.System_Type), TypedConstantKind.Type, originalType)),
                         isOptionalUse: true));
             }
+
+            if (this.IsMicrosoftCodeAnalysisEmbeddedAttribute() && GetEarlyDecodedWellKnownAttributeData() is null or { HasCodeAnalysisEmbeddedAttribute: false })
+            {
+                // This is Microsoft.CodeAnalysis.EmbeddedAttribute, and the user didn't manually apply this attribute to itself. Grab the parameterless constructor
+                // and apply it; if there isn't a parameterless constructor, there would have been a declaration diagnostic
+
+                var parameterlessConstructor = InstanceConstructors.FirstOrDefault(c => c.ParameterCount == 0);
+
+                if (parameterlessConstructor is not null)
+                {
+                    AddSynthesizedAttribute(
+                        ref attributes,
+                        SynthesizedAttributeData.Create(DeclaringCompilation, parameterlessConstructor, arguments: [], namedArguments: []));
+                }
+            }
         }
 
         #endregion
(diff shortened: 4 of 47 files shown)


333fred:
@dotnet/roslyn-compiler for review. I'll propose a source-generation API to allow authors to create this API easily, but this change can go in separately. For VB, there are existing tests verifying that it respects the EmbeddedAttribute definition, and allows defining in the same assembly. I opted not to enforce any shape on the VB as it doesn't do any form of validation in any scenario for the attribute definition today.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -156,3 +156,24 @@ struct S
...
+
+## `Microsoft.CodeAnalysis.EmbeddedAttribute` is validated on declaration
+
+***Introduced in Visual Studio 2022 version 17.13***
+
+The compiler now validates the shape of `Microsoft.CodeAnalysis.EmbeddedAttribute` when declared in source. Previously, the compiler
+would allow user-defined declarations of this attribute, but only when it didn't need to generate one itself. We now validate that:
+
+1. It must be internal
+2. It must be a class
```
AlekseyTs, line 168: >It must be a class [](http://example.com/codeflow?start=3&length=18)

Do we check that it inherits from System.Attribute?

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
```diff
@@ -1905,6 +1905,44 @@ protected override void AfterMembersCompletedChecks(BindingDiagnosticBag diagnos
                     diagnostics.Add(ErrorCode.ERR_RuntimeDoesNotSupportInlineArrayTypes, GetFirstLocation());
                 }
             }
+
+            if (this is
+                {
+                    Name: "EmbeddedAttribute",
+                    Arity: 0,
+                    ContainingType: null,
+                    ContainingNamespace:
```
AlekseyTs, line 1914: >ContainingNamespace [](http://example.com/codeflow?start=20&length=19)

Consider using ```IsContainedInNamespace``` helper

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -193,51 +198,312 @@ public static void Main()
...
+                    file sealed class EmbeddedAttribute : System.Attribute
+                    {
+                        public EmbeddedAttribute() { }
+                    }
+                }
+                class Test
+                {
+                    public void M(in int p)
+                    {
+                        // This should trigger generating another EmbeddedAttribute
```
AlekseyTs, line 330: >// This should trigger generating another EmbeddedAttribute [](http://example.com/codeflow?start=24&length=59)

Consider verifying this assumption

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
AlekseyTs: It looks like the 6th condition implies this requirement, but it might be good to state this explicitly un this document. And probably in the error message.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -193,51 +198,312 @@ public static void Main()
             CompileAndVerify(code, parseOptions: TestOptions.Regular.WithNoRefSafetyRulesAttribute(), verify: Verification.Passes, expectedOutput: "3");
         }
 
+        [Theory]
+        [InlineData("internal")]
+        [InlineData("public")]
+        public void EmbeddedAttributeFromSourceValidation_Valid(string ctorAccessModifier)
```
AlekseyTs, line 204: It looks like two source generator could easily end up in a situation when they define conflicting type in the same compilation. Even if the following partial declaration is used, there will be an error about a duplicate ```[Embedded]``` attribute.
```
namespace Microsoft.CodeAnalysis
{
    [Embedded]
    internal sealed partial class EmbeddedAttribute : System.Attribute
    {
    }
}
```

If the following declaration is used, there will be an error about duplicate `AttributeUsage` attribute.
```
namespace Microsoft.CodeAnalysis
{
    [Embedded]
    [AttributeUsage(AttributeTargets.All, AllowMultiple = true)]
    internal sealed partial class EmbeddedAttribute : System.Attribute
    {
    }
}
```

I think we should find a way for source generators to avoid a conflict like this.
For example, instead of reporting an error when `[Embedded]` is not applied, compiler could synthesize its application on a user defined type.

AlekseyTs: If we are planning VB source generators to rely on the same attribute, I think we should also test how similar situation can be handled in VB.

AlekseyTs:

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
333fred: > It looks like two source generator could easily end up in a situation when they define conflicting type in the same compilation. 

Indeed. I expect the next PR to add a public API for a source generator to register that they would like the attribute to be generated; we can then take care of doing that generation, only adding the source once. I think that's separable though, and this internal functionality should go in ahead of an api review meeting to add that api.

AlekseyTs: I think in order to continue we need to agree on the direction first. In my personal opinion a solution that doesn't require any special coordination between the parties involved (use of special API etc.) would be preferred. Especially when an explicit user's code itself can participate without any knowledge in advance.

333fred: I'd prefer an explicit API over an implicit API; automatically synthesizing an attribute definition when a lookup fails in source is an implicit API the compiler would be providing. I also don't like the types of downstream impacts that could have:

1. We'd expose a synthesized type that is applied to source. I don't know of another location we do that today, and it will likely have a number of downstream impacts on our API consumers. What would the IDE show for Go-To-Definition, for example?
2. Where does this attribute show up in our general APIs? Will it be returned from `GetTypeByMetadataName`, even if it was not applied to any source tree, or was applied to a source tree that has not yet been bound? Will it appear in the global namespace if I try to iterate all symbols defined in this compilation? Will it appear in the merged global namespace across all assemblies?

Having an explicit definition in a file, whether that's the user defining it, or the generator driver providing an API to define it if it hasn't been defined yet, is, in my opinion, the more logically and consistent approach, with less potential impact and implementation challenges. However, regardless of that, I think this PR is not dependent on a firm decision on what public API should be provided, and should be taken regardless. Our existing behavior around `EmbeddedAttribute` is inconsistent and should be cleaned up, even if we do not make any further changes in this area.

333fred: Clarified offline with @AlekseyTs. I misunderstood his original idea, and I like the approach. I'll take a look at that this afternoon.

333fred:
@AlekseyTs I believe I've address your feedback, thanks!

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -156,3 +156,25 @@ struct S
     public static void M([UnscopedRef] ref int x) { }
 }
 ```
+
+## `Microsoft.CodeAnalysis.EmbeddedAttribute` is validated on declaration
+
+
+The compiler now validates the shape of `Microsoft.CodeAnalysis.EmbeddedAttribute` when declared in source. Previously, the compiler
```
jjonescz, line 164: Unrelated to the PR: it's not clear to me what the `Microsoft.CodeAnalysis.EmbeddedAttribute` does. Is that documented somewhere?

docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -35,3 +35,24 @@ Class C
 End Class
 ```
 
+## `Microsoft.CodeAnalysis.EmbeddedAttribute` is validated on declaration
+
+
+The compiler now validates the shape of `Microsoft.CodeAnalysis.EmbeddedAttribute` when declared in source. Previously, the compiler
+would allow user-defined declarations of this attribute, but only when it didn't need to generate one itself. We now validate that:
```
AlekseyTs, line 43: >, but only when it didn't need to generate one itself [](http://example.com/codeflow?start=55&length=53)

", but only when it didn't need to generate one itself" Is this part accurate for VB?

docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -35,3 +35,24 @@ Class C
...
+1. It must be Friend
+2. It must be a Class
+3. It must be NotInheritable
+4. It must not be a Module
+5. It must have a Public parameterless constructor
+6. It must inherit from System.Attribute.
+7. It must be allowed on any type declaration (class, struct, interface, enum, or delegate)
+
+```cs
+namespace Microsoft.CodeAnalysis;
```
AlekseyTs, line 54: Are we supposed to use VB in this document?

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
```diff
@@ -1748,6 +1752,21 @@ internal override void AddSynthesizedAttributes(PEModuleBuilder moduleBuilder, r
                         ImmutableArray.Create(new TypedConstant(compilation.GetWellKnownType(WellKnownType.System_Type), TypedConstantKind.Type, originalType)),
                         isOptionalUse: true));
             }
+
+            if (this.IsMicrosoftCodeAnalysisEmbeddedAttribute() && GetEarlyDecodedWellKnownAttributeData() is null or { HasCodeAnalysisEmbeddedAttribute: false })
+            {
+                // This is Microsoft.CodeAnalysis.EmbeddedAttribute, and the user didn't manually apply this attribute to itself. Grab the parameterless constructor
```
AlekseyTs, line 1758: >// This is Microsoft.CodeAnalysis.EmbeddedAttribute [](http://example.com/codeflow?start=16&length=51)

Consider adding an assert to verify the fact that "This is Microsoft.CodeAnalysis.EmbeddedAttribute"

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
```diff
@@ -1905,6 +1924,34 @@ protected override void AfterMembersCompletedChecks(BindingDiagnosticBag diagnos
...
+                // 1. It must be internal
+                // 2. It must be a class
+                // 3. It must be sealed
+                // 4. It must be non-static
+                // 5. It must have an internal or public parameterless constructor
+                // 6. It must inherit from System.Attribute
+                // 7. It must be allowed on any type declaration (class, struct, interface, enum, or delegate)
+                // 8. It must be non-generic (checked as part of IsMicrosoftCodeAnalysisEmbeddedAttribute, we don't error on this because both types can exist)
+
+                const AttributeTargets ExpectedTargets = AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface | AttributeTargets.Enum | AttributeTargets.Delegate;
```
AlekseyTs, line 1940: >ExpectedTargets [](http://example.com/codeflow?start=39&length=15)

It looks like the code is not following naming conventions for locals. Despite the const modifier, this is a local.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -193,61 +198,399 @@ public static void Main()
...
+                    internal sealed class EmbeddedAttribute : System.Attribute
+                    {
+                        internal EmbeddedAttribute() { }
+                    }
+                }
+                class Test
+                {
+                    public void M(in int p)
+                    {
+                        // This should trigger generating another EmbeddedAttribute
```
AlekseyTs, line 549: >// This should trigger generating another EmbeddedAttribute [](http://example.com/codeflow?start=24&length=59)

Is this accurate?

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
AlekseyTs: I think we should have a test confirming that the following code can be compiled successfully:
```
namespace Microsoft.CodeAnalysis
{
    internal sealed partial class EmbeddedAttribute : System.Attribute
    {
    }
}

namespace Microsoft.CodeAnalysis
{
    internal sealed partial class EmbeddedAttribute : System.Attribute
    {
    }
}
```

src/Compilers/VisualBasic/Portable/Symbols/Source/SourceMemberContainerTypeSymbol.vb
```diff
@@ -60,6 +60,7 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
             ReportedVarianceDiagnostics = &H2    ' Set if variance diagnostics have been reported.
             ReportedBaseClassConstraintsDiagnostics = &H4    ' Set if base class constraints diagnostics have been reported.
             ReportedInterfacesConstraintsDiagnostics = &H8    ' Set if constraints diagnostics for base/implemented interfaces have been reported.
+            ReportedCodeAnalysisEmbeddedAttributeDiagnostics = &H10 ' Set if
```
AlekseyTs, line 63: >' Set if  [](http://example.com/codeflow?start=68&length=9)

It looks like the comment is not finished

src/Compilers/VisualBasic/Portable/Symbols/Source/SourceNamedTypeSymbol.vb
```diff
@@ -1687,16 +1690,52 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
...
+                ' 1. It must be Friend
+                ' 2. It must be a class
+                ' 3. It must be NotInheritable
+                ' 4. It must not be a Module
+                ' 5. It must have a Public parameterless constructor. This is different from C#, because VB requires public parameterless attribute constructors in general
+                ' 6. It must inherit from System.Attribute
+                ' 7. It must be allowed on any type declaration (class, struct, interface, enum, Or delegate)
+                ' 8. It must be non-generic
+
+                Dim expectedTargets = AttributeTargets.Class Or AttributeTargets.Struct Or AttributeTargets.Interface Or AttributeTargets.Enum Or AttributeTargets.Delegate
```
AlekseyTs, line 1717: >Dim [](http://example.com/codeflow?start=16&length=3)

```Const```?

File: src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs [hunk 13]
```diff
@@ -20,7 +22,8 @@ public void ReferencingEmbeddedAttributesFromTheSameAssemblySucceeds()
             var code = @"
 namespace Microsoft.CodeAnalysis
 {
-    internal class EmbeddedAttribute : System.Attribute { }
+    [Embedded]
```
AlekseyTs, line 25: >[Embedded] [](http://example.com/codeflow?start=4&length=10)

It looks like this change (the attribute application) is no longer necessary here and in other similar places.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -193,61 +198,399 @@ public static void Main()
...
+                    internal sealed class EmbeddedAttribute : System.Attribute
+                    {
+                        public EmbeddedAttribute() { }
+                    }
+                }
+                class Test
+                {
+                    public void M(in int p)
+                    {
+                        // This should trigger generating another EmbeddedAttribute
```
AlekseyTs, line 408: >// This should trigger generating another EmbeddedAttribute [](http://example.com/codeflow?start=24&length=59)

Is this accurate?

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
```diff
@@ -1748,6 +1752,21 @@ internal override void AddSynthesizedAttributes(PEModuleBuilder moduleBuilder, r
                         ImmutableArray.Create(new TypedConstant(compilation.GetWellKnownType(WellKnownType.System_Type), TypedConstantKind.Type, originalType)),
                         isOptionalUse: true));
             }
+
+            if (this.IsMicrosoftCodeAnalysisEmbeddedAttribute() && GetEarlyDecodedWellKnownAttributeData() is null or { HasCodeAnalysisEmbeddedAttribute: false })
```
AlekseyTs, line 1756: >&& GetEarlyDecodedWellKnownAttributeData() is null or { HasCodeAnalysisEmbeddedAttribute: false } [](http://example.com/codeflow?start=64&length=97)

Are we testing positive and negative effect of this condition?

src/Compilers/VisualBasic/Portable/Symbols/Source/SourceNamedTypeSymbol.vb
```diff
@@ -2559,6 +2600,26 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
                         ImmutableArray.Create(New TypedConstant(compilation.GetWellKnownType(WellKnownType.System_Type), TypedConstantKind.Type, originalType)),
                         isOptionalUse:=True))
             End If
+
+            If Me.IsMicrosoftCodeAnalysisEmbeddedAttribute() Then
+                Dim earlyAttributeData = GetEarlyDecodedWellKnownAttributeData()
+
+                If earlyAttributeData Is Nothing OrElse Not earlyAttributeData.HasCodeAnalysisEmbeddedAttribute Then
```
AlekseyTs, line 2607: >earlyAttributeData Is Nothing OrElse Not earlyAttributeData.HasCodeAnalysisEmbeddedAttribute [](http://example.com/codeflow?start=19&length=92)

Are we testing positive and negative effect of this condition?

AlekseyTs:
Done with review pass (commit 10)

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
333fred: Not documented afaik, though we will when we add the SG API. It prevents the compiler from resolving the type outside the current compilation. It's how the compiler itself gets around the "type defined in multiple assemblies" issues when it needs to synthesize an attribute.

docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
333fred: Technically, yes, in that it never generated one, so it never forbid a user definition, but I'll update the wording.

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
333fred: `EmbeddedAttributeFromSourceValidation_MissingEmbeddedAttributeApplicationGeneratesAttributeApplication` is the positive condition. There are a number of tests that cover the negative condition, where `Embedded` is applied to the attribute itself. Most directly, `EmbeddedAttributeFromSourceValidation_Valid`. However, I'll add a test with an unrelated early well-known attribute, to test the other positive condition.

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
333fred: The `if` condition verifies that. I'm not sure what an additional assert that just does the same logic adds?

AlekseyTs: You are right.

src/Compilers/VisualBasic/Portable/Symbols/Source/SourceNamedTypeSymbol.vb
333fred: Yes, similar to C#. I will also add a similar additional test for another early well-known attribute to test the `OrElse` bit.

333fred:
@AlekseyTs addressed feedback, thanks!

docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -35,3 +35,27 @@ Class C
...
+3. It must be NotInheritable
+4. It must not be a Module
+5. It must have a Public parameterless constructor
+6. It must inherit from System.Attribute.
+7. It must be allowed on any type declaration (class, struct, interface, enum, or delegate)
+
+```vb
+Namespace Microsoft.CodeAnalysis
+
+    ' Previously allowed. Now, CS9271
```
cston, line 56: >CS9271 [](http://example.com/codeflow?start=31&length=6)

`BC37335`

docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -35,3 +35,27 @@ Class C
...
+The compiler now validates the shape of `Microsoft.CodeAnalysis.EmbeddedAttribute` when declared in source. Previously, the compiler
+would allow user-defined declarations of this attribute with any shape. We now validate that:
+
+1. It must be Friend
+2. It must be a Class
+3. It must be NotInheritable
+4. It must not be a Module
+5. It must have a Public parameterless constructor
+6. It must inherit from System.Attribute.
+7. It must be allowed on any type declaration (class, struct, interface, enum, or delegate)
```
cston, line 51: >class, struct, interface, enum, or delegate [](http://example.com/codeflow?start=47&length=43)

Perhaps `Class, Structure, Interface, Enum, or Delegate`

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -5908,6 +5908,9 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="ERR_TypeReserved" xml:space="preserve">
     <value>The type name '{0}' is reserved to be used by the compiler.</value>
   </data>
+  <data name="ERR_ReservedTypeMustFollowPattern" xml:space="preserve">
```
cston, line 5911: >ERR_ReservedTypeMustFollowPattern [](http://example.com/codeflow?start=14&length=33)

Do we expect to use this error for other types? (The message is quite specific.) If not, consider giving the error a more specific name, and perhaps hardcoding the type name in the message.

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
```diff
@@ -1413,8 +1414,11 @@ internal override bool HasCodeAnalysisEmbeddedAttribute
         {
             get
             {
+
                 var data = GetEarlyDecodedWellKnownAttributeData();
```
cston, line 1418: Consider removing extra blank line.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -193,61 +195,400 @@ public static void Main()
...
+            {
+                var embeddedAttribute = module.ContainingAssembly.GetTypeByMetadataName(AttributeDescription.CodeAnalysisEmbeddedAttribute.FullName);
+                Assert.NotNull(embeddedAttribute);
+                Assert.Equal("Microsoft.CodeAnalysis.EmbeddedAttribute", embeddedAttribute.ToTestDisplayString());
+
+                var embeddedAttributeFile = module.ContainingAssembly.GetTypeByMetadataName(AttributeDescription.CodeAnalysisEmbeddedAttribute.FullName + "@<unknown>");
+                Assert.NotNull(embeddedAttribute);
+                Assert.Equal("Microsoft.CodeAnalysis.EmbeddedAttribute", embeddedAttribute.ToTestDisplayString());
+                Assert.NotSame(embeddedAttribute, embeddedAttributeFile);
+            });
```
cston, line 343: >} [](http://example.com/codeflow?start=12&length=1)

Consider verifying the `[Embedded]` attribute applied to the `file` type is `<unknown>_EmbeddedAttribute`.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -193,61 +195,400 @@ public static void Main()
...
+                    internal sealed class EmbeddedAttribute<T> : System.Attribute
+                    {
+                        public EmbeddedAttribute() { }
+                    }
+                }
+                class Test
+                {
+                    public void M(in int p)
+                    {
+                        // This would trigger generating another EmbeddedAttribute, if we hadn't already defined one
```
cston, line 440: >// This would trigger generating another EmbeddedAttribute, if we hadn't already defined one [](http://example.com/codeflow?start=24&length=92)

Should this be "This should trigger generating another EmbeddedAttribute"?

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -193,61 +195,400 @@ public static void Main()
...
+                    Assert.Equal(["Microsoft.CodeAnalysis.EmbeddedAttribute"], embeddedAttribute.GetAttributes().Where(a => a.AttributeClass.Name != "ObsoleteAttribute").Select(a => a.AttributeClass.ToTestDisplayString()));
+                })
+                .VerifyDiagnostics();
+        }
 
-            CreateCompilation(code, references: new[] { moduleRef }).VerifyEmitDiagnostics(
-                // error CS8004: Type 'EmbeddedAttribute' exported from module 'testModule.netmodule' conflicts with type declared in primary module of this assembly.
-                Diagnostic(ErrorCode.ERR_ExportedTypeConflictsWithDeclaration).WithArguments("Microsoft.CodeAnalysis.EmbeddedAttribute", "testModule.netmodule").WithLocation(1, 1));
+        [Fact]
+        public void EmbeddedAttributeFromSourceValidation_Generic()
```
cston, line 425: >EmbeddedAttributeFromSourceValidation_Generic [](http://example.com/codeflow?start=20&length=45)

Should we also test when both generic and non-generic are defined in source?

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -193,61 +195,400 @@ public static void Main()
...
+            CreateCompilation(code, assemblyName: "testModule", targetFramework: TargetFramework.NetStandard20).VerifyEmitDiagnostics(
+                // (3,6): error CS0616: 'EmbeddedAttribute' is not an attribute class
+                //     [Embedded]
+                Diagnostic(ErrorCode.ERR_NotAnAttributeClass, "Embedded").WithArguments("Microsoft.CodeAnalysis.EmbeddedAttribute").WithLocation(3, 6),
+                // (4,27): error CS9271: The type 'Microsoft.CodeAnalysis.EmbeddedAttribute' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.
+                //     internal sealed class EmbeddedAttribute
+                Diagnostic(ErrorCode.ERR_ReservedTypeMustFollowPattern, "EmbeddedAttribute").WithArguments("Microsoft.CodeAnalysis.EmbeddedAttribute").WithLocation(4, 27));
+        }
+
+        [Theory]
```
cston, line 525: >Theory [](http://example.com/codeflow?start=9&length=6)

Consider testing:
```
[InlineData("AttributeTargets.All")]
[InlineData("AttributeTargets.Class | <all types> | AttributeTargets.GenericParameter")]
```

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
```diff
@@ -267,15 +608,35 @@ public void M(in int p)
...
+                    // Methods
+                    .method public hidebysig specialname rtspecialname instance void .ctor () cil managed 
+                    {
+                        ldarg.0
+                        call instance void [mscorlib]System.Object::.ctor()
+                        ret
+                    }
+                }
+
+                .class public auto ansi beforefieldinit Microsoft.CodeAnalysis.EmbeddedAttribute
```
cston, line 629: >EmbeddedAttribute [](http://example.com/codeflow?start=79&length=17)

Why wasn't this `EmbeddedAttribute` definition used for `Test.M()` below? (This definition doesn't appear to be marked `[Embedded]`.)

src/Compilers/VisualBasic/Portable/Symbols/Source/SourceMemberContainerTypeSymbol.vb
```diff
@@ -60,6 +60,7 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
             ReportedVarianceDiagnostics = &H2    ' Set if variance diagnostics have been reported.
             ReportedBaseClassConstraintsDiagnostics = &H4    ' Set if base class constraints diagnostics have been reported.
             ReportedInterfacesConstraintsDiagnostics = &H8    ' Set if constraints diagnostics for base/implemented interfaces have been reported.
+            ReportedCodeAnalysisEmbeddedAttributeDiagnostics = &H10 ' Set if whether the symbol has been checked for Microsoft.CodeAnalysis.EmbeddedAttribute definition diagnostics.
```
cston, line 63: >' Set if whether the symbol [](http://example.com/codeflow?start=68&length=27)

"Set if ~~whether~~ the symbol has been checked ..."?

src/Compilers/VisualBasic/Portable/Symbols/Source/SourceNamedTypeSymbol.vb
```diff
@@ -1687,16 +1690,52 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
...
+        End Sub
+
+        Private Sub CheckEmbeddedAttributeImplementation()
+            If (m_lazyState And StateFlags.ReportedCodeAnalysisEmbeddedAttributeDiagnostics) <> 0 Then
+                Return
+            End If
+
+            Dim diagnostics As BindingDiagnosticBag = Nothing
+            If Me.IsMicrosoftCodeAnalysisEmbeddedAttribute() Then
+                ' This Is a user-defined implementation of the special attribute Microsoft.CodeAnalysis.EmbeddedAttribute. It needs to follow specific rules
```
cston, line 1707: >This Is [](http://example.com/codeflow?start=18&length=7)

This is (lowercase) ...

src/Compilers/VisualBasic/Test/Emit/Attributes/AttributeTests.vb
```diff
@@ -5081,6 +5083,254 @@ End Class
...
+        Inherits System.Attribute
+
+        {ctorAccessModifier} Sub New()
+        End Sub
+    End Class
+End Namespace
+"
+
+            If ctorAccessModifier <> "Private" Then
+            End If
```
cston, line 5176: Remove?

src/Compilers/VisualBasic/Test/Emit/Attributes/AttributeTests.vb
```diff
@@ -5081,6 +5083,254 @@ End Class
...
+BC37335: The type 'EmbeddedAttribute' must be non-generic, Friend, NotInheritable, have a Public parameterless constructor, inherit from System.Attribute, and be able to be applied to any type
+    <Embedded>
+    ~~~~~~~~~~~
+BC31504: 'EmbeddedAttribute' cannot be used as an attribute because it does not inherit from 'System.Attribute'.
+    <Embedded>
+     ~~~~~~~~
+]]>)
+        End Sub
+
+        <Theory>
```
cston, line 5291: ><Theory> [](http://example.com/codeflow?start=8&length=8)

Consider testing:
```
<InlineData("AttributeTargets.All")>
<InlineData("AttributeTargets.Class Or <all types> Or AttributeTargets.GenericParameter")>
```

AlekseyTs:

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests_Embedded.cs
333fred: When we look for an `EmbeddedAttribute` to put on a type, we never check any dependencies. We only check dependencies for the type that we're considering embedding (such as `IsReadOnlyAttribute`).

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
333fred: I'll update the assert.

333fred:
@AlekseyTs @cston for another review pass

docs/compilers/Visual Basic/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -35,3 +35,27 @@ Class C
...
+The compiler now validates the shape of `Microsoft.CodeAnalysis.EmbeddedAttribute` when declared in source. Previously, the compiler
+would allow user-defined declarations of this attribute with any shape. We now validate that:
+
+1. It must be Friend
+2. It must be a Class
+3. It must be NotInheritable
+4. It must not be a Module
+5. It must have a Public parameterless constructor
+6. It must inherit from System.Attribute.
+7. It must be allowed on any type declaration (Class, Struct, Interface, Enum, or Delegate)
```
AlekseyTs, line 51: >Struct [](http://example.com/codeflow?start=54&length=6)

"Structure"? If you are after VB keywords,

src/Compilers/VisualBasic/Test/Emit/Attributes/AttributeTests.vb
```diff
@@ -5081,6 +5083,276 @@ End Class
...
+            comp.AssertNoDiagnostics()
+
+            Dim sourceDeclaration = comp.SourceAssembly.GetTypeByMetadataName("Microsoft.CodeAnalysis.EmbeddedAttribute")
+            Assert.Equal(2, sourceDeclaration.Locations.Length)
+
+            CompileAndVerify(comp, symbolValidator:=Sub([module])
+                                                        Dim embeddedAttr = [module].ContainingAssembly.GetTypeByMetadataName(AttributeDescription.CodeAnalysisEmbeddedAttribute.FullName)
+                                                        Assert.NotNull(embeddedAttr)
+                                                        Assert.Equal({"Microsoft.CodeAnalysis.EmbeddedAttribute"},
+                                                            embeddedAttr.GetAttributes().Where(Function(a) a.AttributeClass.Name <> "ObsoleteAttribute").Select(Function(a) a.AttributeClass.ToTestDisplayString()))
```
AlekseyTs, line 5230: >.Where(Function(a) a.AttributeClass.Name <> "ObsoleteAttribute") [](http://example.com/codeflow?start=88&length=64)

The `Where` call looks redundant

AlekseyTs (approved):
LGTM (commit 12)

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -5908,6 +5908,9 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="ERR_TypeReserved" xml:space="preserve">
     <value>The type name '{0}' is reserved to be used by the compiler.</value>
   </data>
+  <data name="ERR_EmbeddedAttributeMustFollowPattern" xml:space="preserve">
+    <value>The type 'Microosft.CodeAnalysis.EmbeddedAttribute' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.</value>
```
cston, line 5912: >Microosft [](http://example.com/codeflow?start=21&length=9)

Typo.

cston (approved):

cston (approved):

PR #76567, opened 2024-12-26
Propagate marked attributes to synthesized fields.

AlekseyTs:
- From a parameter to a state machine field
- From a parameter to a closure field
- From a VB property/event to a backing field

Related to #73920.

### DIFF
--- a/src/Compilers/CSharp/Portable/Lowering/ClosureConversion/LambdaCapturedVariable.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/ClosureConversion/LambdaCapturedVariable.cs
@@ -10,18 +10,20 @@ using Microsoft.CodeAnalysis.CSharp.Symbols;
 using Roslyn.Utilities;
 using System.Collections.Immutable;
 using Microsoft.CodeAnalysis.Symbols;
+using Microsoft.CodeAnalysis.CSharp.Emit;
+using Microsoft.CodeAnalysis.PooledObjects;
 
 namespace Microsoft.CodeAnalysis.CSharp
 {
     /// <summary>
     /// A field of a frame class that represents a variable that has been captured in a lambda.
     /// </summary>
-    internal sealed class LambdaCapturedVariable : SynthesizedFieldSymbolBase, ISynthesizedMethodBodyImplementationSymbol
+    internal class LambdaCapturedVariable : SynthesizedFieldSymbolBase, ISynthesizedMethodBodyImplementationSymbol
     {
         private readonly TypeWithAnnotations _type;
         private readonly bool _isThis;
 
-        private LambdaCapturedVariable(SynthesizedClosureEnvironment frame, TypeWithAnnotations type, string fieldName, bool isThisParameter)
+        protected LambdaCapturedVariable(SynthesizedClosureEnvironment frame, TypeWithAnnotations type, string fieldName, bool isThisParameter)
             : base(frame,
                    fieldName,
                    isPublic: true,
@@ -45,7 +47,9 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             string fieldName = GetCapturedVariableFieldName(captured, ref uniqueId);
             TypeSymbol type = GetCapturedVariableFieldType(frame, captured);
-            return new LambdaCapturedVariable(frame, TypeWithAnnotations.Create(type), fieldName, IsThis(captured));
+            return captured is ParameterSymbol { IsThis: false } parameter ?
+                new LambdaCapturedVariableForRegularParameter(frame, TypeWithAnnotations.Create(type), fieldName, parameter) :
+                new LambdaCapturedVariable(frame, TypeWithAnnotations.Create(type), fieldName, IsThis(captured));
         }
 
         private static bool IsThis(Symbol captured)
@@ -150,5 +154,35 @@ namespace Microsoft.CodeAnalysis.CSharp
         /// </summary>
         public bool HasMethodBodyDependency
             => false;
+    }
+
+    internal sealed class LambdaCapturedVariableForRegularParameter : LambdaCapturedVariable
+    {
+        private readonly ParameterSymbol _parameter;
+
+        public LambdaCapturedVariableForRegularParameter(SynthesizedClosureEnvironment frame, TypeWithAnnotations type, string fieldName, ParameterSymbol parameter)
+            : base(frame, type, fieldName, isThisParameter: false)
+        {
+            Debug.Assert(parameter is { IsThis: false });
+            _parameter = parameter;
+        }
+
+        internal override void AddSynthesizedAttributes(PEModuleBuilder moduleBuilder, ref ArrayBuilder<CSharpAttributeData> attributes)
+        {
+            if (_parameter.OriginalDefinition is SourceParameterSymbolBase definition &&
+                ContainingModule == definition.ContainingModule)
+            {
+                foreach (CSharpAttributeData attr in definition.GetAttributes())
+                {
+                    if (attr.AttributeClass is { HasCompilerLoweringPreserveAttribute: true } attributeType &&
+                        (attributeType.GetAttributeUsageInfo().ValidTargets & System.AttributeTargets.Field) != 0)
+                    {
+                        AddSynthesizedAttribute(ref attributes, attr);
+                    }
+                }
+            }
+
+            base.AddSynthesizedAttributes(moduleBuilder, ref attributes);
+        }
     }
 }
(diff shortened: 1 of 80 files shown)


333fred (commented):
One minor reuse comment, otherwise looks good (commit 2)

src/Compilers/CSharp/Portable/Lowering/ClosureConversion/LambdaCapturedVariable.cs
```diff
@@ -45,7 +47,9 @@ public static LambdaCapturedVariable Create(SynthesizedClosureEnvironment frame,
 
             string fieldName = GetCapturedVariableFieldName(captured, ref uniqueId);
             TypeSymbol type = GetCapturedVariableFieldType(frame, captured);
-            return new LambdaCapturedVariable(frame, TypeWithAnnotations.Create(type), fieldName, IsThis(captured));
+            return captured is ParameterSymbol { IsThis: false } parameter ?
```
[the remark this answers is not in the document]
AlekseyTs: Do you have any specific suggestion in mind?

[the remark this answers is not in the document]
333fred: Probably by introducing an `out` for the `ParameterSymbol` and doing this:

```cs
var isThis = IsThis(captured, out ParameterSymbol? parameter);
return parameter is not null && !isThis ?
    new LambdaCapturedVariableForRegularParameter(frame, TypeWithAnnotations.Create(type), fieldName, parameter) :
    new LambdaCapturedVariable(frame, TypeWithAnnotations.Create(type), fieldName, isThis)
```

And then discard the `out` at the other use site.

333fred (approved):

src/Compilers/CSharp/Portable/Lowering/ClosureConversion/LambdaCapturedVariable.cs
```diff
@@ -39,21 +41,38 @@ private LambdaCapturedVariable(SynthesizedClosureEnvironment frame, TypeWithAnno
...
             string fieldName = GetCapturedVariableFieldName(captured, ref uniqueId);
             TypeSymbol type = GetCapturedVariableFieldType(frame, captured);
-            return new LambdaCapturedVariable(frame, TypeWithAnnotations.Create(type), fieldName, IsThis(captured));
+            bool isThis = IsThis(captured, out ParameterSymbol? parameter);
+            return parameter is not null && !isThis ?
+                new LambdaCapturedVariableForRegularParameter(frame, TypeWithAnnotations.Create(type), fieldName, parameter) :
+                new LambdaCapturedVariable(frame, TypeWithAnnotations.Create(type), fieldName, isThis);
         }
 
         private static bool IsThis(Symbol captured)
```
cston, line 58: >IsThis [](http://example.com/codeflow?start=28&length=6)

It looks like this overload is only used in one location, in `GetCapturedVariableFieldName()` below. Consider inlining the call there and removing this overload.

src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/StateMachineFieldSymbol.cs
```diff
@@ -83,4 +85,34 @@ internal override bool IsCapturedFrame
...
+                ContainingModule == definition.ContainingModule)
+            {
+                foreach (CSharpAttributeData attr in definition.GetAttributes())
+                {
+                    if (attr.AttributeClass is { HasCompilerLoweringPreserveAttribute: true } attributeType &&
+                        (attributeType.GetAttributeUsageInfo().ValidTargets & System.AttributeTargets.Field) != 0)
+                    {
+                        AddSynthesizedAttribute(ref attributes, attr);
+                    }
+                }
```
cston, line 112: This block to copy attributes is repeated in several locations. Would it make sense to extract a helper method?

src/Compilers/CSharp/Test/Symbol/Symbols/Source/EventTests.cs
```diff
@@ -2487,5 +2487,283 @@ class Test
         }
 
         #endregion
+
+        [Fact]
+        public void CompilerLoweringPreserveAttribute_01()
```
cston, line 2492: >CompilerLoweringPreserveAttribute_01 [](http://example.com/codeflow?start=20&length=36)

Did the behavior of these two tests change with this PR?

src/Compilers/CSharp/Test/Symbol/Symbols/Source/PropertyTests.cs
```diff
@@ -3026,5 +3026,74 @@ class C
             var property = compilation.GetMember<PropertySymbol>("C.P");
             Assert.True(property.RequiresInstanceReceiver);
         }
+
+        [Fact]
+        public void CompilerLoweringPreserveAttribute_01()
```
cston, line 3031: >CompilerLoweringPreserveAttribute_01 [](http://example.com/codeflow?start=20&length=36)

Did the behavior of these two tests change with this PR?

src/Compilers/CSharp/Test/Symbol/Symbols/Source/PropertyTests.cs
```diff
@@ -3026,5 +3026,74 @@ class C
...
+        }
+
+        [Fact]
+        public void CompilerLoweringPreserveAttribute_02()
+        {
+            string source1 = @"
+using System;
+
+[AttributeUsage(AttributeTargets.Field)]
+public class Preserve1Attribute : Attribute { }
```
cston, line 3072: >Preserve1Attribute [](http://example.com/codeflow?start=13&length=18)

What are we testing with no `[CompilerLoweringPreserve]` attribute?

src/Compilers/VisualBasic/Portable/Symbols/Source/SynthesizedFieldSymbol.vb
```diff
@@ -55,7 +55,22 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Symbols
...
+
+                If definition IsNot Nothing AndAlso
+                   ContainingModule = definition.ContainingModule Then
+                    For Each attr In definition.GetAttributes()
+                        Dim attributeType As NamedTypeSymbol = attr.AttributeClass
+                        If attributeType IsNot Nothing AndAlso attributeType.HasCompilerLoweringPreserveAttribute AndAlso
+                           (attributeType.GetAttributeUsageInfo().ValidTargets And System.AttributeTargets.Field) <> 0 Then
+                            AddSynthesizedAttribute(attributes, attr)
+                        End If
+                    Next
```
cston, line 70: This block is used in a couple of locations. Would it make sense to extract a helper method?

src/Compilers/CSharp/Portable/Lowering/ClosureConversion/LambdaCapturedVariable.cs
AlekseyTs: I prefer to not adjust existing consumers of this helper.

src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/StateMachineFieldSymbol.cs
AlekseyTs: > This block to copy attributes is repeated in several locations. Would it make sense to extract a helper method?

I do not think this is worth doing.

src/Compilers/CSharp/Test/Symbol/Symbols/Source/PropertyTests.cs
AlekseyTs: > Did the behavior of these two tests change with this PR?

No, it didn't.

src/Compilers/CSharp/Test/Symbol/Symbols/Source/PropertyTests.cs
AlekseyTs: > What are we testing with no ```[CompilerLoweringPreserve]``` attribute?

We are testing that the ```field:``` target allows to apply any attribute to the field explicitly. Therefore, no special treatment around ```CompilerLoweringPreserve``` is necessary.

src/Compilers/VisualBasic/Portable/Symbols/Source/SynthesizedFieldSymbol.vb
AlekseyTs: > This block is used in a couple of locations. Would it make sense to extract a helper method?

I do not think this is worth doing.

cston (approved):

Issue #66312, opened 2023-01-08
Errors about prohibiting pointer fields on records not shown in the code, they're only visible on the build output


Rekkonnect:
**Version Used**: Visual Studio *Version 17.5.0 Preview 2.0*

**Steps to Reproduce**:

1. Create a record with pointer fields on its primary constructor
2. Build the project

**Expected Behavior**: The errors are reported in the build and shown with red squiggles in the editor.

**Actual Behavior**: The errors are only reported in the build.

**Notes**: The errors are properly displayed on the SharpLab editor, and in Rider.

[**Minimal repro code**](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4wmApgZwA4QDGWABAMoD22AcprgcQNwCwAUGwAIDMJArjDggAzUgCcshCqIQkcAF1G9Cc8gApYcgFQkAChQ1ZRcEgiwAbLAHMIcrJoA8HACwA+EgBFzVm1gCULVm4+AWExCSkZeUVlcgBGdRgtXX1Ew39OHn5BERJxSWlZBSUVMgAmVVMLa1sHZzdPKp901iA===)

**Screen capture**:

https://user-images.githubusercontent.com/8298332/211193510-9506befd-b411-43d2-811a-d4fd8b102eb1.mp4

mavasani:
Couple of findings here:

1. The given minimal repro code produces CS8908 warnings, which seems to be generated only at emit time. All unit tests for this error code verifies it is part of compiler emit diagnostics: https://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/Errors/ErrorCode.cs,11d41e079bd1b80c,references
2. `ErrorFacts.IsBuildOnlyDiagnostic` method seems to return `false` for this error code: https://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/Errors/ErrorFacts.cs,2106. This seems to be a bug as it is indeed never reported by `SemanticModel.GetDiagnostics()` method. This in turn causes the error CS8908 to get filtered out from the error list after a build is executed in VS for code with this error.

Tagging @dotnet/roslyn-compiler team to confirm that `CS8908` is indeed a build-only compiler error. Once confirmed, I can create a PR for 2. above to fix this issue.

333fred:
It does look like it's a build-only error. However, I personally think that the bug is that it's a build-only error, and would rather the fix be moving the error into initial binding. There's no reason this can't reported much earlier.

[...]

PR #76588, opened 2024-12-30
Forbid unsafe types in records

333fred:
[The specification](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/records.md#members-of-a-record-type) for records forbids the use of unsafe types as an instance field of the type. However, we didn't check that in all cases; we only checked when generating the `Equals` method for the record type. This meant that if the user provided their own definition of `Equals`, they could bypass this restriction. We also did not check for nested unsafe types, so `int*[]` was permitted in any scenario. We have 2 options for fixing this:

1. Treat it as a specification bug, and update the spec to follow the current behavior. We'd therefore fix https://github.com/dotnet/roslyn/issues/66312 by moving the diagnostic to SourceNamedTypeSymbol.AfterMembersChecks, and only error when the user didn't provide their own implementation of `Equals`.
2. Treat it as a compiler bug, and forbid use of unsafe types in all scenarios.

I've opted for path 2 here, but we can discuss whether this is too aggressive and if we should opt for path 1 instead. Fixes https://github.com/dotnet/roslyn/issues/66312.


333fred:
We also need to decide what to do with manually-implemented properties that have unsafe types. As pointed out by @hamarb123, even with this change we'll still generate invalid IL for `PrintMembers` in such a case, where we cast the pointer to `object`. By my reading of the spec, it says we should make it work; nothing about the definition of `PrintMembers` specifies that this cast needs to occur, and we could instead convert the pointer to `IntPtr` before doing the print. The main question is whether we _should_ do this, or just forbid all pointer types in `records`, period.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -156,3 +156,30 @@ struct S
...
 }
 ```
+
+## `record` and `record struct` types cannot define pointer type members, even when providing their own Equals implementations
+
+PROTOTYPE: Figure what version this is going into
+
+***Introduced in Visual Studio 2022 version 17.14***
+
+The specification for `record class` and `record struct` types indicated that any unsafe types are disallowed. However, this was not
```
AlekseyTs, line 166: >indicated that any unsafe types are disallowed [](http://example.com/codeflow?start=63&length=46)

This sounds too broad. The spec says:

> It is an error for an instance field of a record to have an unsafe type.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -156,3 +156,30 @@ struct S
...
+enforced correctly in 2 scenarios:
+
+1. When the `record class` or `record struct` type defined its own `Equals` implementation.
+2. When the field type only used the unsafe type in a nested context, such as `int*[]`.
```
AlekseyTs, line 170: >When the field type only used the unsafe type in a nested context, such as `int*[]`. [](http://example.com/codeflow?start=3&length=84)

I couldn't find clear definition of "unsafe type" in the language spec. Therefore, it is not clear whether such array type should be considered as unsafe.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberFieldSymbol.cs
```diff
@@ -57,15 +57,20 @@ protected void TypeChecks(TypeSymbol type, BindingDiagnosticBag diagnostics)
...
             {
-                diagnostics.Add(ErrorCode.ERR_FieldCantBeRefAny, TypeSyntax?.Location ?? this.GetFirstLocation(), type);
+                diagnostics.Add(ErrorCode.ERR_FieldCantBeRefAny, getTypeErrorLocation(), type);
             }
             else if (type.IsRefLikeOrAllowsRefLikeType() && (this.IsStatic || !containingType.IsRefLikeType))
             {
-                diagnostics.Add(ErrorCode.ERR_FieldAutoPropCantBeByRefLike, TypeSyntax?.Location ?? this.GetFirstLocation(), type);
+                diagnostics.Add(ErrorCode.ERR_FieldAutoPropCantBeByRefLike, getTypeErrorLocation(), type);
+            }
+            else if (!this.IsStatic && type.ContainsPointerOrFunctionPointer() && (ContainingType.IsRecord || ContainingType.IsRecordStruct))
```
AlekseyTs, line 70: >(ContainingType.IsRecord || ContainingType.IsRecordStruct) [](http://example.com/codeflow?start=82&length=58)

I assume this check is cheaper than the type check for pointers. Consider performing reordering the checks.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -1839,9 +1839,18 @@ protected virtual void ValidatePropertyType(BindingDiagnosticBag diagnostics)
             {
                 diagnostics.Add(ErrorCode.ERR_FieldCantBeRefAny, TypeLocation, type);
             }
-            else if (this.IsAutoPropertyOrUsesFieldKeyword && type.IsRefLikeOrAllowsRefLikeType() && (this.IsStatic || !this.ContainingType.IsRefLikeType))
+
+            if (this.IsAutoPropertyOrUsesFieldKeyword)
             {
-                diagnostics.Add(ErrorCode.ERR_FieldAutoPropCantBeByRefLike, TypeLocation, type);
+                if (!this.IsStatic && type.ContainsPointerOrFunctionPointer() && (ContainingType.IsRecord || ContainingType.IsRecordStruct))
```
AlekseyTs, line 1845: >(ContainingType.IsRecord || ContainingType.IsRecordStruct) [](http://example.com/codeflow?start=81&length=58)

Similar suggestion

src/Compilers/CSharp/Test/Semantic/Semantics/RecordStructTests.cs
```diff
@@ -1707,9 +1802,8 @@ public unsafe record C
 }
 ";
 
-            var comp = CreateCompilation(new[] { src, IsExternalInitTypeDefinition }, options: TestOptions.UnsafeDebugDll);
-            comp.VerifyEmitDiagnostics(
-                // (12,22): error CS0723: Cannot declare a variable of static type 'C2'
+            var comp = CreateCompilation(new[] { src, IsExternalInitTypeDefinition }, options: TestOptions.UnsafeDebugDll, targetFramework: TargetFramework.Mscorlib461);
+            DiagnosticDescription[] expected = [// (12,22): error CS0723: Cannot declare a variable of static type 'C2'
```
AlekseyTs, line 1806: >// (12,22): error CS0723: Cannot declare a variable of static type 'C2' [](http://example.com/codeflow?start=48&length=71)

Consider placing the comment on the next line

AlekseyTs:
It looks like more test base-lines need an adjustment

AlekseyTs:

333fred:
@AlekseyTs addressed your feedback. The spec part of this change is at https://github.com/dotnet/csharplang/pull/9050, clarifying what "unsafe types" is actually supposed to mean.

src/Compilers/CSharp/Portable/Symbols/Synthesized/Records/SynthesizedRecordEquals.cs
```diff
@@ -136,12 +136,7 @@ internal override void GenerateMethodBody(TypeCompilationState compilationState,
                         fields.Add(f);
 
                         var parameterType = f.Type;
-                        if (parameterType.IsPointerOrFunctionPointer())
-                        {
-                            diagnostics.Add(ErrorCode.ERR_BadFieldTypeInRecord, f.GetFirstLocationOrNone(), parameterType);
-                            foundBadField = true;
-                        }
-                        else if (parameterType.IsRestrictedType())
+                        if (parameterType.ContainsPointerOrFunctionPointer() || parameterType.IsRestrictedType())
```
AlekseyTs, line 139: >ContainsPointerOrFunctionPointer [](http://example.com/codeflow?start=42&length=32)

I thought we decided to allow arrays of pointers

AlekseyTs:

src/Compilers/CSharp/Portable/Symbols/Synthesized/Records/SynthesizedRecordEquals.cs
333fred: Thanks for pointing it out, I added a test that would have failed.

AlekseyTs (approved):
LGTM (commit 4)

333fred:

hamarb123:
Did you end up blocking manually implemented unsafe properties, or allowing them, out of curiosity. Based on the spec change I assume they're allowed?

333fred:
> Did you end up blocking manually implemented pointer type properties, or allowing them (and fixing the `PrintMembers`), out of curiosity. Based on the spec change I assume they're allowed?

No change was made to locations without fields.

333fred:

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -193,3 +193,23 @@ namespace Microsoft.CodeAnalysis;
 // Previously, sometimes allowed. Now, CS9271
 public class EmbeddedAttribute : Attribute {}
 ```
+
+## `record` and `record struct` types cannot define pointer type members, even when providing their own Equals implementations
+
+PROTOTYPE: Figure what version this is going into
```
cston, line 199: Remove?

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -193,3 +193,23 @@ namespace Microsoft.CodeAnalysis;
...
+
+***Introduced in Visual Studio 2022 version 17.14***
+
+The specification for `record class` and `record struct` types indicated that any pointer types are disallowed as instance fields.
+However, this was not enforced correctly when the `record class` or `record struct` type defined its own `Equals` implementation.
+
+The compiler now correctly forbids this.
+
+```cs
+unsafe record struct R1(
```
cston, line 209: >R1 [](http://example.com/codeflow?start=21&length=2)

`R`, to match the `Equals(R)` declaration.

src/Compilers/CSharp/Test/Emit3/Semantics/RecordTests.cs
```diff
@@ -1271,12 +1272,73 @@ unsafe record C(
...
+                Diagnostic(ErrorCode.ERR_FieldAutoPropCantBeByRefLike, "RefLike").WithArguments("RefLike").WithLocation(15, 5)];
+
+            comp.VerifyDiagnostics(expected);
+            comp.VerifyEmitDiagnostics(expected);
+        }
+
+        [Fact]
+        [WorkItem(48115, "https://github.com/dotnet/roslyn/issues/48115")]
+        [WorkItem("https://github.com/dotnet/roslyn/issues/66312")]
+        public void RestrictedTypesAndPointerTypes_WithEquals()
```
cston, line 1284: >RestrictedTypesAndPointerTypes_WithEquals [](http://example.com/codeflow?start=20&length=41)

Since the expected behavior is identical to the previous test, consider using `[Theory]` with `bool withEquals` parameter.

src/Compilers/CSharp/Test/Semantic/Semantics/RecordStructTests.cs
```diff
@@ -1526,15 +1527,71 @@ unsafe record struct C(
...
-                Diagnostic(ErrorCode.ERR_BadFieldTypeInRecord, "P10").WithArguments("delegate*<void>").WithLocation(16, 21)
-                );
+                Diagnostic(ErrorCode.ERR_FieldAutoPropCantBeByRefLike, "RefLike").WithArguments("RefLike").WithLocation(15, 5)];
+            comp.VerifyDiagnostics(expected);
+            comp.VerifyEmitDiagnostics(expected);
         }
+        [Fact]
+        [WorkItem(48115, "https://github.com/dotnet/roslyn/issues/48115")]
+        [WorkItem("https://github.com/dotnet/roslyn/issues/66312")]
+        public void RestrictedTypesAndPointerTypes_WithEquals()
```
cston, line 1537: >RestrictedTypesAndPointerTypes_WithEquals [](http://example.com/codeflow?start=20&length=41)

Consider combining with previous test with `[Theory]` and `bool withEquals` parameter, so it's clear we expect the behavior to be the same.

src/Compilers/CSharp/Test/Emit3/Semantics/RecordTests.cs
```diff
@@ -1321,12 +1383,14 @@ public unsafe record C
...
+                Diagnostic(ErrorCode.ERR_FieldAutoPropCantBeByRefLike, "RefLike").WithArguments("RefLike").WithLocation(16, 12)];
+            comp.VerifyDiagnostics(expected);
+            comp.VerifyEmitDiagnostics(expected);
         }
 
-        [ConditionalFact(typeof(DesktopOnly), Reason = ConditionalSkipReason.RestrictedTypesNeedDesktop)]
+        [Fact]
         [WorkItem(48115, "https://github.com/dotnet/roslyn/issues/48115")]
+        [WorkItem("https://github.com/dotnet/roslyn/issues/66312")]
         public void RestrictedTypesAndPointerTypes_NominalMembers_AutoProperties()
```
cston, line 1394: >RestrictedTypesAndPointerTypes_NominalMembers_AutoProperties [](http://example.com/codeflow?start=20&length=60)

Consider testing at least one case with `field`.
```csharp
unsafe record C
{
    int* P { get => field; }
}
```

cston (approved):

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -196,7 +196,7 @@ public class EmbeddedAttribute : Attribute {}
 
 ## `record` and `record struct` types cannot define pointer type members, even when providing their own Equals implementations
 
-PROTOTYPE: Figure what version this is going into
+***Introduced in Visual Studio 2022 version 17.14***
```
cston, line 279: Duplicated below.

cston (approved):

PR #76601, opened 2025-01-02
Add option to return suggestion diagnostics as LSP Information diagnostics in vscode

dibarbet:
Related to https://github.com/dotnet/vscode-csharp/issues/6723

On release we returned LSP.DiagnosticSeverity.Information (rendered as blue squiggles in VSCode) for our suggestion diagnostics - however we got a ton of feedback that the the noise was undesired, and so we switch to using LSP.DiagnosticSeverity.Information (ellipse underneath).  

However hints are not shown in the problems pane, so we have some users requesting information be brought back.  This adds an option to do so.

client side - https://github.com/dotnet/vscode-csharp/pull/7892

### DIFF
--- a/src/LanguageServer/Protocol/Extensions/ProtocolConversions.Diagnostics.cs
+++ b/src/LanguageServer/Protocol/Extensions/ProtocolConversions.Diagnostics.cs
@@ -33,7 +33,10 @@ internal static partial class ProtocolConversions
             return [];
         }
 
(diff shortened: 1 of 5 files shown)


dibarbet:
different approach here

mikes-gh:
@dibarbet Is there a PR for the different approach?

dibarbet:
@JoeRobich is working on it, not sure if there is a PR yet

PR #76604, opened 2025-01-02
Remove Edgio CDN usage to release/dev17.8

jordi1215:
The changes are copied over from https://github.com/dotnet/roslyn/pull/76431

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -215,7 +215,7 @@
     <MicrosoftWin32PrimitivesVersion>4.3.0</MicrosoftWin32PrimitivesVersion>
     <MicrosoftWin32RegistryVersion>5.0.0</MicrosoftWin32RegistryVersion>
     <MicrosoftBuildLocatorVersion>1.6.10</MicrosoftBuildLocatorVersion>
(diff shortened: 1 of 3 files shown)


eng/Versions.props
```diff
@@ -316,7 +316,7 @@
       When updating the S.C.I or S.R.M version please let the MSBuild team know in advance so they
       can update to the same version. Version changes require a VS test insertion for validation.
     -->
-    <SystemCollectionsImmutableVersion>7.0.0</SystemCollectionsImmutableVersion>
+    <SystemCollectionsImmutableVersion>8.0.0</SystemCollectionsImmutableVersion>
```
dibarbet, line 319: I'm not 100% certain we can do this - what triggered the upgrade here?  If VS 17.8 itself isn't on 8.0 I believe we'll have issues

jordi1215: MSBuildStructuredLoggerVersion version 2.2.386 depends on this apparently. Maybe we can try with a lower MSBuildStructuredLoggerVersion?

dibarbet: @jordi1215 yeah lets try a lower version.  Try 2.2.350, if that doesn't work I can help get other versions added so restore works

dibarbet (approved):

PR #76613, opened 2025-01-03
WIP: Reduce allocations by adding more SyntaxList specializations

ToddGrun:
Created as a draft PR to get speedometer allocation information to determine whether this is worth pursuing.

Currently, we specialize and cache SyntaxLists with up to 3 children. Not only do these specializations avoid an array allocation, but these specialized lists are also pooled. This PR increases that number to 5 to determine if this would have a positive allocation impact.

Speedometer test shows 1.4% of codeanalysis process allocations are ArrayElement<GreenNode> arrays allocated from SyntaxListBuilder.ToListNode. Additionally, WithManyChildren objects account for 0.7% of allocations. I'm going to run a test insertion with this change to determine whether these numbers are reduced by increasing the specialization size from 3 to 5.

Local testing opening the Roslyn sln showed 22.7 million calls to SyntaxListBuilder.ToListNode. Of that, 1.32 million were for lists with length > 3, but only 600 thousand were for lists with length > 5.

If the speedometer numbers look promising, I'm considering creating another draft PR to get numbers on what adding specializations up to 9 in size would look like (there were only 172K lists that had size >= 10)

### DIFF
--- a/src/Compilers/CSharp/Test/Syntax/Syntax/SyntaxNodeCacheTests.cs
+++ b/src/Compilers/CSharp/Test/Syntax/Syntax/SyntaxNodeCacheTests.cs
@@ -74,5 +74,61 @@ namespace Microsoft.CodeAnalysis.CSharp.UnitTests
                 Assert.True(listOf3 != listOf2, $"{i} iterations");
             }
         }
+
+        [Fact]
+        public void TryGetNode_With3Of4Children()
+        {
+            const int nIterations = 1_000_000;
+
+            for (int i = 0; i < nIterations; i++)
+            {
+                var child0 = new SyntaxTokenWithTrivia(SyntaxKind.InternalKeyword, null, null);
+                var child1 = new SyntaxTokenWithTrivia(SyntaxKind.StaticKeyword, null, null);
+                var child2 = new SyntaxTokenWithTrivia(SyntaxKind.ReadOnlyKeyword, null, null);
+                var child3 = new SyntaxTokenWithTrivia(SyntaxKind.AsyncKeyword, null, null);
+                SyntaxNodeCache.AddNode(child0, child0.GetCacheHash());
+                SyntaxNodeCache.AddNode(child1, child1.GetCacheHash());
+                SyntaxNodeCache.AddNode(child2, child2.GetCacheHash());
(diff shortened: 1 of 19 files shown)


ToddGrun:
Results aren't quite as promising as hoped, still going to elevate out of draft status to get compiler folks thoughts.

I did two speedometer runs from the [test insertion](https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/600753) and measured the amount of allocations of ArrayElement<GreenNode>[] (filtered under SytnaxListbuilder.ToListNode), WithManyChildren, WithFourChildren, and WithFiveChildren.

Previous: (203 MB)
    Type Microsoft.CodeAnalysis.ArrayElement`1[Microsoft.CodeAnalysis.GreenNode][] 	139,775,232
    Type WithManyChildren     	 											63,205,840

Run 1: (170 MB)
    Type Microsoft.CodeAnalysis.ArrayElement`1[Microsoft.CodeAnalysis.GreenNode][] 	84,275,928
    Type WithFiveChildren     	 												33,838,840
    Type WithManyChildren     	 											31,819,968
    Type WithFourChildren     	 												20,441,968

Run 2: (186 MB)
    Type Microsoft.CodeAnalysis.ArrayElement`1[Microsoft.CodeAnalysis.GreenNode][] 	92,727,776
    Type WithManyChildren     	                            								36,063,728
    Type WithFiveChildren     	                            									35,550,104
    Type WithFourChildren     	                            									21,387,928

So, it looks like this would save about 25 MB of allocations, which isn't quite the improvement I had hoped for. So, while this saves about 12% of allocations when just looking at SyntaxLists with > 3 elements, it ends up saving < 0.1% of all allocations in the codeanalysis process that occur in the speedometer test. I'm open to tabling this PR if the compiler devs think this isn't worth the effort, but the changes are pretty simple conceptually since it just extends the current pattern from 3 elements to 5.

ToddGrun:
Actually, after a bit more consideration, I'm just going to close this PR. There are additional costs to cache calculations for the nodes which get more expensive as the number of child nodes increases. The WithManyChildren type just avoids caching altogether to avoid those CPU costs. I could make the 4/5 child nodes avoid caching too, but then the memory savings would be even smaller.

TL/DR, was an interesting investigation, but the CPU performance impact probably doesn't warrant the slight memory improvement numbers.

PR #76617, opened 2025-01-03
Fix assembly load when file path contains URI reserved chars

dibarbet:
Resolves https://github.com/microsoft/vscode-dotnettools/issues/1686


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/ExportProviderBuilderTests.cs
```diff
@@ -49,6 +54,98 @@ public async Task MultipleMefCompositionsAreCached()
...
+                options: options);
+#pragma warning restore RS0030 // Do not use banned APIs
+
+            // Write a dll to a subdir with a reserved character in the name.
+            var tempSubDir = directory.CreateDirectory(reservedCharacter);
+            var tempFile = tempSubDir.CreateFile($"{assemblyName}.dll");
+            var dllData = compilation.EmitToStream();
+            tempFile.WriteAllBytes(dllData.ToArray());
+
+            // Mark the file as read only to prevent mutations.
```
jasonmalinowski, line 142: Does this break us cleaning up the temp files later?

jasonmalinowski:
@dibarbet Just generally curious here:

1. Should VS MEF give some alternative here to avoid us having to work around this behavior? Or a common implementation of IAssemblyLoader that smooths this over? Because I imagine DevKit and other extensions would have the same issue, right?
2. Are we able to just avoid using the codebase parameter entirely, in some way?

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/ExportProviderBuilderTests.cs
dibarbet: I didn't see issues, but let me check.  I stole this from a compiler test

dibarbet:
> Should VS MEF give some alternative here to avoid us having to work around this behavior? Or a common implementation of IAssemblyLoader that smooths this over? Because I imagine DevKit and other extensions would have the same issue, right?

Possibly.  However we do want our own IASsemblyLoader here so we can load assemblies into matching ALCs.  So they would have to provide us with some consistent way to get the file path.

> Are we able to just avoid using the codebase parameter entirely, in some way?

I do not know of one.  As best as I can tell it is the only way to get the file path, which we need to put it in the directory ALC.

PR #76635, opened 2025-01-06
Release a SeparatedSyntaxListBuilder back to the SyntaxListPool

ToddGrun:
I noticed this when looking at the SyntaxListPool and that it's _allocated debug info was not reducing back to zero length. 

The early return codepaths in LanguageParser.ParseLocalDeclarationStatement are commonly hit, preventing the current code from releasing "variables" back into the pool. Changed to account for the early returns by utilizing the existing finally block similar to what is done for the "mods" pooled item.


RikkiGibson (approved):

ToddGrun:

chsienki (approved):

PR #76636, opened 2025-01-06
Ensure VB parser is disposed

ToddGrun:
Noticed then when I was trying out adding a pooled object to the parser and that the dispose wasn't always being called to release the object back to the pool. I'm still investigating whether I want to add the pooled data to the parser, but minimially, the existing dispose method should be called.

### DIFF
--- a/src/Compilers/VisualBasic/Portable/Scanner/Directives.vb
+++ b/src/Compilers/VisualBasic/Portable/Scanner/Directives.vb
@@ -52,10 +52,10 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Syntax.InternalSyntax
                 Me.GetNextTokenInState(ScannerState.VB)
 
             Else
-                Dim parser As New Parser(Me)
-
-                directiveTrivia = parser.ParseConditionalCompilationStatement()
-                directiveTrivia = parser.ConsumeStatementTerminatorAfterDirective(directiveTrivia)
+                Using parser = New Parser(Me)
+                    directiveTrivia = parser.ParseConditionalCompilationStatement()
+                    directiveTrivia = parser.ConsumeStatementTerminatorAfterDirective(directiveTrivia)
+                End Using
             End If
 
             Debug.Assert(directiveTrivia.FullWidth > 0, "should at least get #")
--- a/src/Compilers/VisualBasic/Portable/Scanner/XmlDocComments.vb
+++ b/src/Compilers/VisualBasic/Portable/Scanner/XmlDocComments.vb
@@ -68,62 +68,62 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Syntax.InternalSyntax
                 Me.MoveToNextSyntaxNodeInTrivia()
 
             Else
-                Dim parser As New Parser(Me)
-
-                Me.IsScanningXmlDoc = True
-                Me._isStartingFirstXmlDocLine = True
-
-                ' NOTE: Documentation comment syntax trivia must have at least one child xml node, because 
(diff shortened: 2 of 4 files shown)


src/Compilers/VisualBasic/Portable/Scanner/Directives.vb
```diff
@@ -52,10 +52,10 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Syntax.InternalSyntax
                 Me.GetNextTokenInState(ScannerState.VB)
 
             Else
-                Dim parser As New Parser(Me)
-
-                directiveTrivia = parser.ParseConditionalCompilationStatement()
-                directiveTrivia = parser.ConsumeStatementTerminatorAfterDirective(directiveTrivia)
+                Using parser = New Parser(Me)
```
jaredpar, line 55: In this case the `Dispose` is not actually returning the `Scanner` to the pool.

src/Compilers/VisualBasic/Portable/Scanner/XmlDocComments.vb
```diff
@@ -68,62 +68,62 @@ Namespace Microsoft.CodeAnalysis.VisualBasic.Syntax.InternalSyntax
...
-
-                ' The second phase
-                Me._endOfXmlInsteadOfLastDocCommentLineBreak = False
-                If nodes.Count = 0 AndAlso parser.CurrentToken.Kind = SyntaxKind.EndOfXmlToken Then
-                    ' This must be an empty documentation comment, we need to reset scanner so 
-                    ' that the doc comment exterior trivia ([''']) lands on the final XmlNode
-
-                    ResetLineBufferOffset()
-                    restorePoint.RestoreTokens(includeLookAhead:=False)
+                Using parser = New Parser(Me)
```
jaredpar, line 71: Same feedback.

jaredpar (changes requested):
The `Scanner` instance is only disposed when it is created by the `Parser`. When it is explicitly passed as an argument to the `Parser` then it is not disposed. This change seems to have no functional impact that I can see. Am I missing something?

ToddGrun:
As mentioned in the PR description, this became visible to me as I was locally trying out adding an item to the VB Parser that needed to be cleaned up during dispose, regardless of whether the parser needed to dispose the scanner. In those cases, the missing Dispose calls were problematic. So yes, this PR shouldn't have any effect on current code, but it makes adding code that needs the Dispose easier and less error prone.

If you prefer, I can not make this a separate PR and only move forward if the change I'm investigating that needs the Dispose ends up being useful, but I considered this a good change even outside that context as it took me a bit to figure out why my objects were leaking in that the Dispose isn't called from all construction paths.

---
In reply to: [2532544758](https://github.com/dotnet/roslyn/pull/76636#issuecomment-2532544758) [](http://example.com/codeflow?ancestors=2532544758)

ToddGrun:
@jaredpar -- Did I answer your question? Would you prefer for me to abandon this PR and just include it if I create a PR that adds something to the VB parser that needs to be disposed?

jaredpar:
Want to make sure the current pattern wasn't a deliberate choice (@AlekseyTs)

ToddGrun:
I've tested locally the disposable changes I'd like to make to the parser, and I'd like to get some speedometer test measurements. 

@AlekseyTs -- was the old pattern intentional to avoid adding disposable objects into the parser except for the scanner?

AlekseyTs (approved):
LGTM (commit 1)

ToddGrun:

jaredpar (approved):

PR #76640, opened 2025-01-06
Make full source build pure NetCurrent

jaredpar:
For a full source build there should be no need for `$(NetPrevious)` to be included. Having it there creates complications in the build as we end up depending on projects that are not built during source build.


jaredpar:
@ViktorHofer FYI

ViktorHofer (approved):

docs/contributing/target-framework-strategy.md
```diff
@@ -0,0 +1,122 @@
+# Target Framework Strategy
+
+## Layers
+
+The roslyn repository produces components for a number of different products that push varying ship and TFM constraints on us. A summary of some of our dependencies are : 
+
+- Build Tools: requires us to ship compilers on `net472`
+- .NET SDK: requires us to ship compilers on current servicing target framework (presently `net8.0`)
+- Source build: requires us to ship `$(NetCurrent)` and `$(NetPrevious)` in workspaces and below (presently `net9.0` and `net8.0` respectively)
```
dibarbet, line 9: Seems this is no longer true?

dibarbet (approved):

333fred (approved):

docs/contributing/target-framework-strategy.md
jaredpar: Split this out into both repo and full source build discussion

PR #76657, opened 2025-01-07
Fix ref safety of implicit calls that might capture refs in the receiver

jjonescz:
Alternative to https://github.com/dotnet/roslyn/pull/76237 and https://github.com/dotnet/roslyn/pull/76263.
Fixes https://github.com/dotnet/roslyn/issues/75802.
Fixes https://github.com/dotnet/roslyn/issues/63306.

There can be implicit calls generated by the compiler like `collection.Add` or `interpolatedStringHandler.AppendFormatted` that previously weren't fully analyzed by the ref safety analysis.

For collection initializers, the ref safety analysis went over all the initializers (`Add` calls) and intersected their val escape scopes, but that essentially did nothing since those `Add` methods usually return `void` so GetValEscape doesn't look at them.
For interpolated strings, the ref safety analysis went over all the arguments and intersected their val escape scopes.
But neither of those approaches detected when a ref passed as an `in` argument could be captured into the receiver (collection / interpolated string handler). So this PR analyzes the implicit calls properly to detect that.

All the existing get / check val escape utilities work on expressions and so they analyze calls only if they can actually return a ref. (Arg mixing checks if a ref can escape into the receiver but since here we are essentially trying to determine what the escape scope of the receiver should be, we cannot easily use that - I've tried that in the alternative PRs linked above but it wasn't very nice.) So I added other utilities to determine/check the escape scope of the receiver of a call.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
@@ -327,6 +327,16 @@ namespace Microsoft.CodeAnalysis.CSharp
                     ReceiverIsSubjectToCloning = propertyAccess.InitialBindingReceiverIsSubjectToCloning,
                     HasAnyErrors = propertyAccess.HasAnyErrors,
                 };
+
+            public static MethodInvocationInfo FromCollectionElementInitializer(BoundCollectionElementInitializer colElement)
+                => new MethodInvocationInfo
+                {
+                    MethodInfo = MethodInfo.Create(colElement.AddMethod),
+                    Parameters = colElement.AddMethod.Parameters,
+                    Receiver = colElement.ImplicitReceiverOpt,
+                    ArgsOpt = colElement.Arguments,
+                    ArgsToParamsOpt = colElement.ArgsToParamsOpt,
+                };
         }
 
         /// <summary>
@@ -2320,26 +2330,14 @@ namespace Microsoft.CodeAnalysis.CSharp
         }
 
         private SafeContext GetInvocationEscapeToReceiver(
-            in MethodInfo methodInfo,
-            ImmutableArray<ParameterSymbol> parameters,
-            BoundExpression? receiver,
-            ImmutableArray<BoundExpression> argsOpt,
-            ImmutableArray<RefKind> argRefKindsOpt,
-            ImmutableArray<int> argsToParamsOpt,
+            in MethodInvocationInfo methodInvocationInfo,
             SafeContext localScopeDepth)
         {
             // By default it is safe to escape.
             SafeContext escapeScope = SafeContext.CallingMethod;
 
             var escapeValues = ArrayBuilder<EscapeValue>.GetInstance();
-            GetFilteredInvocationArgumentsForEscapeToReceiver(
-                methodInfo,
-                parameters,
-                receiver,
-                argsOpt,
-                argRefKindsOpt,
-                argsToParamsOpt,
-                escapeValues);
+            GetFilteredInvocationArgumentsForEscapeToReceiver(in methodInvocationInfo, escapeValues);
 
             foreach (var (_, argument, _, isArgumentRefEscape) in escapeValues)
             {
@@ -2570,12 +2568,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
         private bool CheckInvocationEscapeToReceiver(
             SyntaxNode syntax,
-            in MethodInfo methodInfo,
-            ImmutableArray<ParameterSymbol> parameters,
-            BoundExpression? receiver,
-            ImmutableArray<BoundExpression> argsOpt,
-            ImmutableArray<RefKind> argRefKindsOpt,
-            ImmutableArray<int> argsToParamsOpt,
+            in MethodInvocationInfo methodInvocationInfo,
             bool checkingReceiver,
             SafeContext escapeFrom,
             SafeContext escapeTo,
@@ -2584,14 +2577,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             bool result = true;
 
             var escapeValues = ArrayBuilder<EscapeValue>.GetInstance();
-            GetFilteredInvocationArgumentsForEscapeToReceiver(
-                methodInfo,
-                parameters,
-                receiver,
-                argsOpt,
-                argRefKindsOpt,
-                argsToParamsOpt,
-                escapeValues);
+            GetFilteredInvocationArgumentsForEscapeToReceiver(in methodInvocationInfo, escapeValues);
 
             foreach (var (parameter, argument, _, isArgumentRefEscape) in escapeValues)
             {
@@ -2601,7 +2587,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                 if (!valid)
                 {
-                    ReportInvocationEscapeError(syntax, methodInfo.Symbol, parameter, checkingReceiver, diagnostics);
+                    ReportInvocationEscapeError(syntax, methodInvocationInfo.MethodInfo.Symbol, parameter, checkingReceiver, diagnostics);
                     result = false;
                     break;
                 }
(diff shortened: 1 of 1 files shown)


src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -5667,5 +5862,106 @@ void getParts(BoundInterpolatedString interpolatedString)
...
+            SafeContext getParts(BoundInterpolatedString interpolatedString, SafeContext localScopeDepth)
+            {
+                SafeContext scope = SafeContext.CallingMethod;
+
+                foreach (var part in interpolatedString.Parts)
+                {
+                    if (part is not BoundCall { Method.Name: BoundInterpolatedString.AppendFormattedMethod } call)
+                    {
+                        // Dynamic calls cannot have ref struct parameters, and AppendLiteral calls will always have literal
+                        // string arguments and do not require us to be concerned with escape
```
AlekseyTs, line 5896: >and do not require us to be concerned with escape [](http://example.com/codeflow?start=44&length=49)

Can ```AppendLiteral``` have an `in` parameter? I assume the ref-ness of the parameter is important, not just the argument value.

jjonescz: Yes, AppendLiteral calls might have the same issue, thanks.

jjonescz:

RikkiGibson:
I will try to get to this next week

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -2172,6 +2172,50 @@ private SafeContext GetInvocationEscapeToReceiver(
...
+            Debug.Assert(extensionParameter is not null);
+            parameters = parameters.IsDefault ? [extensionParameter] : [extensionParameter, .. parameters];
+
+            Debug.Assert(receiver is not null);
+            argsOpt = argsOpt.IsDefault ? [receiver] : [receiver, .. argsOpt];
+            receiver = null;
+
+            if (!argRefKindsOpt.IsDefault)
+            {
+                argRefKindsOpt = [RefKind.None, .. argRefKindsOpt];
```
RikkiGibson: I still need to review the larger PR, but, I am confused by this part. Doesn't the receiver argument ref kind need to be compatible with the extension parameter? e.g. `extension(ref int i)` needs a compatible ref kind here.

jjonescz: You seem to be commenting on a code that I haven't changed in the PR (it was only merged in from `main`). This code snippet is in the `ReplaceWithExtensionImplementationIfNeeded` method, right? But since I've reviewed extensions, I remember being confused by the same thing :D See https://github.com/dotnet/roslyn/pull/77967#discussion_r2030804759.

RikkiGibson: 😓

jjonescz:
@RikkiGibson @jaredpar @dotnet/roslyn-compiler for reviews, thanks

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -2613,6 +2687,89 @@ static bool hasRefLikeReturn(Symbol symbol)
...
+                // We have an instance method receiver.
+                if (!hasRefToRefStructThis(methodInfo.Method) && !hasRefToRefStructThis(methodInfo.SetMethod))
+                {
+                    return;
+                }
+            }
+            else
+            {
+                // We have an extension method receiver.
+                Debug.Assert(methodInfo.Method?.IsExtensionMethod != false);
```
RikkiGibson, line 2859: We'll want to either test new extensions here or add a tracking issue link for adding such tests.

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -5788,73 +5968,113 @@ private bool CheckInterpolatedStringHandlerConversionEscape(BoundExpression expr
...
+                        call.ArgsToParamsOpt,
+                        localScopeDepth));
+                }
+
+                return scope;
+            }
         }
 
-        private void GetInterpolatedStringHandlerArgumentsForEscape(BoundExpression expression, ArrayBuilder<BoundExpression> arguments)
+        private bool CheckValEscapeOfInterpolatedStringHandlerCalls(BoundExpression expression, SafeContext escapeFrom, SafeContext escapeTo, BindingDiagnosticBag diagnostics)
```
RikkiGibson, line 5834: Does this need to take in a `BoundExpression`? Or would it work to take in a `BoundInterpolatedString` and make the `expression` used for traversal a local?

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
jjonescz: The expression can _start_ as `BoundBinaryOperator` or `BoundInterpolatedString`. So yes, this needs to take `BoundExpression`.

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
jjonescz: I've realized the "correct" way is to call `ReplaceWithExtensionImplementationIfNeeded` like the ref safety analysis does in other places. I will update the code to do that.

jjonescz:

jjonescz:

jjonescz:

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -5790,73 +5985,115 @@ private bool CheckInterpolatedStringHandlerConversionEscape(BoundExpression expr
...
+                        return scope.Intersect(getParts(interpolatedString, localScopeDepth));
+
+                    default:
+                        throw ExceptionUtilities.UnexpectedValue(expression.Kind);
                 }
             }
 
-            arguments.Free();
-            return result;
+            SafeContext getParts(BoundInterpolatedString interpolatedString, SafeContext localScopeDepth)
```
333fred: If I see something named `getParts`, I expect it literally returns all the parts. Probably want to rename for clarity.
```suggestion
            SafeContext getPartsScope(BoundInterpolatedString interpolatedString, SafeContext localScopeDepth)
```

src/Compilers/CSharp/Portable/Binder/Binder.ValueChecks.cs
```diff
@@ -5790,73 +5985,115 @@ private bool CheckInterpolatedStringHandlerConversionEscape(BoundExpression expr
...
-            GetInterpolatedStringHandlerArgumentsForEscape(expression, arguments);
+            // Narrow the scope for implicit calls which allow the receiver to capture refs from the arguments.
+            result = result && CheckValEscapeOfInterpolatedStringHandlerCalls(expression, escapeFrom, escapeTo, diagnostics);
 
-            bool result = true;
-            foreach (var argument in arguments)
+            return result;
+        }
+
+        private SafeContext GetValEscapeOfInterpolatedStringHandlerCalls(BoundExpression expression, SafeContext localScopeDepth)
```
333fred, line 5792: Why do we need both of these functions? Surely `Check` can be implemented as a call to `Get` and a comparison to the allowed escape scope?

jjonescz: Same reason we have `Get*`/`Check*` pairs elsewhere in ref safety analysis - better error reporting.

jjonescz:

333fred (approved):

PR #76660, opened 2025-01-07
CodeGenerator.EmitStackAlloc - Avoid capturing blob array

alrz:
I'm not sure if conditional code on dotnet version is acceptable as this API is only available in 8.0+.

### DIFF
--- a/src/Compilers/CSharp/Portable/CodeGen/EmitStackAllocInitializer.cs
+++ b/src/Compilers/CSharp/Portable/CodeGen/EmitStackAllocInitializer.cs
@@ -45,7 +45,7 @@ namespace Microsoft.CodeAnalysis.CSharp.CodeGen
                 var sizeInBytes = elementType.EnumUnderlyingTypeOrSelf().SpecialType.SizeInBytes();
 
                 ImmutableArray<byte> data = GetRawData(initExprs);
-                if (data.All((d, first) => d == first, data[0]))
+                if (data.All(static (d, first) => d == first, data[0]))
                 {
                     // All bytes are the same, no need for metadata blob, just initblk to fill it with the repeated value.
                     _builder.EmitOpCode(ILOpCode.Dup);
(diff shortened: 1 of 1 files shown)


src/Compilers/CSharp/Portable/CodeGen/EmitStackAllocInitializer.cs
```diff
@@ -45,7 +45,11 @@ private void EmitStackAlloc(TypeSymbol type, BoundArrayInitialization? inits, Bo
                 var sizeInBytes = elementType.EnumUnderlyingTypeOrSelf().SpecialType.SizeInBytes();
 
                 ImmutableArray<byte> data = GetRawData(initExprs);
+#if NET8_0_OR_GREATER
+                if (!data.AsSpan().ContainsAnyExcept(data[0]))
```
333fred, line 49: We don't generally have a problem with conditional code, but we usually want to have a good reason to have it; do you have benchmarks or other data indicating this capture is a problem?

AlekseyTs, line 49: >if (!data.AsSpan().ContainsAnyExcept(data[0])) [](http://example.com/codeflow?start=16&length=46)

We can add an overload of ```All``` which takes an extra argument and passes it to the predicate. I think that would be preferred over a conditional compilation in this case.

AlekseyTs: In fact we already have the overload in ImmutableArrayExtensions:
```
public static bool All<T, TArg>(this ImmutableArray<T> array, Func<T, TArg, bool> predicate, TArg arg)
```

AlekseyTs:

src/Compilers/CSharp/Portable/CodeGen/EmitStackAllocInitializer.cs
```diff
@@ -45,7 +45,7 @@ private void EmitStackAlloc(TypeSymbol type, BoundArrayInitialization? inits, Bo
                 var sizeInBytes = elementType.EnumUnderlyingTypeOrSelf().SpecialType.SizeInBytes();
 
                 ImmutableArray<byte> data = GetRawData(initExprs);
-                if (data.All(datum => datum == data[0]))
+                if (data.All((d, first) => d == first, data[0]))
```
jjonescz, line 48:
```suggestion
                if (data.All(static (d, first) => d == first, data[0]))
```

jjonescz (approved):

AlekseyTs (approved):
LGTM (commit 3)

AlekseyTs:
@alrz Thanks for the contribution

PR #76666, opened 2025-01-07
Switch ObjectPool to use Interlocked.Exchange instead of CompareExchange

ToddGrun:
This may be beneficial if the location being queried has had a new non-null value placed into it since the optimistic read, allowing use of that value, instead of skipping it.

When opening Roslyn.sln, I verified both modified Allocate methods did indeed hit the case where Interlocked.Exchange returned a non-null value not equal to the optimistic read's value.

I'm a bit hesitant about this change, as the code has not changed for many years and there may be some low-level optimization that such a change will negate. However, on the surface, it seems like using Interlocked.Exchange should be sufficient, so I'll go ahead and create the PR and see if someone else can find something I'm missing.

### DIFF
--- a/src/Dependencies/PooledObjects/ObjectPool`1.cs
+++ b/src/Dependencies/PooledObjects/ObjectPool`1.cs
@@ -147,7 +147,8 @@ namespace Microsoft.CodeAnalysis.PooledObjects
             // We will interlock only when we have a candidate. in a worst case we may miss some
             // recently returned objects. Not a big deal.
             var inst = _firstItem;
-            if (inst == null || inst != Interlocked.CompareExchange(ref _firstItem, null, inst))
+            if (inst == null ||
+                (inst = Interlocked.Exchange(ref _firstItem, null)) == null)
             {
                 inst = AllocateSlow();
             }
@@ -176,7 +177,8 @@ namespace Microsoft.CodeAnalysis.PooledObjects
                 var inst = items[i].Value;
                 if (inst != null)
                 {
-                    if (inst == Interlocked.CompareExchange(ref items[i].Value, null, inst))
+                    inst = Interlocked.Exchange(ref items[i].Value, null);
+                    if (inst != null)
                     {
                         return inst;
                     }


CyrusNajmabadi (approved):
*seems* fine.  let's run a speedometer run to ensure no unpleasant surprises :)

ToddGrun:
> _seems_ fine. let's run a speedometer run to ensure no unpleasant surprises :)

Test PR insertion: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/601391

ToddGrun:
Test insertion is being wonky during build, will work to get another one going too.

In the meantime, adding some observant eyes directly as the code is ancient and a bit scary to change. Thanks!

@AlekseyTs @jasonmalinowski @sharwell

AlekseyTs:
@ToddGrun Could you clarify the goal of this PR? Fixing an problem in an algorithm, optimization, etc.

ToddGrun:
> @ToddGrun Could you clarify the goal of this PR? Fixing an problem in an algorithm, optimization, etc.

The intent is to slightly improve the efficiency of the Roslyn ObjectPool class by allowing the ObjectPool.Allocate method to return cached items that would have otherwise not been returned. By changing the code to use Interlocked.Exchange instead of Interlocked.CompareExchange, the Allocate method can find hits even when the value from the interlocked read differs from the optimistic read value.

333fred (approved):

ToddGrun:
Sadly, I think I might have to abandon this PR. The speedometer tests led to some interesting results when looking at CPU/Allocations when filtering with IncPaths under PooledObjects.ObjectPool:

In the Roslyn code analysis process (.net core), things look good:
  Allocations went from 36MB -> 29.7 MB
  CPU went from 939 ms -> 689 ms

However, in the devenv process (netfx), things are mixed: 
  Allocations went from 75.6 MB -> 50 MB
  CPU went from 567 ms -> 1384 ms.

Two different speedometer runs had the same trends, things improved except for CPU in the VS process. @stephentoub -- sorry to pull in the big guns here, but is it to be expected that Interlocked.CompareExchange and Interlocked.Exchange would have such different performance characteristics in these two contexts?

ToddGrun:
Closing as this didn't degraded CPU perf in VS as outlined above. Seems a bit counter-intuitive, but it's what the tests demonstrate.

stephentoub:
On the surface I'm not sure why you'd see such a significant CPU difference between core and netfx here; there have been improvements around {Compare}Exchange, but I wouldn't expect them to be so significant. My guess is there's something else subtle going on, e.g. Exchange is always going to incur a write whereas CompareExchange won't, so maybe there's something about the memory layout that's contributing to a false sharing issue on netfx, or something like that. Hard to say without spending more time measuring and analyzing.

Issue #76031, opened 2024-11-22
Update `field` keyword breaking change diagnostics


cston:
With `-langversion:preview`, the token `field`, when used as a *primary expression* in a property accessor, binds to the synthesized backing field for the property. That is a breaking change when there is another symbol in scope named `field` that the compiler would have bound to with earlier language versions. The compiler reports a warning in `-langversion:preview` for those cases.

A couple of issues have been reported regarding the current warning:

1. The warning does not match the [Breaking change warnings proposal](https://github.com/dotnet/csharplang/blob/main/proposals/breaking-change-warnings.md). In that proposal, warnings are reported in *earlier* language versions that the behavior *will change* rather than warning in *later* language versions that behavior *has changed*.

2. The warning is reported for a *reference* to `field`, but not for a *declaration* of a local or parameter named `field` within the accessor, even though `field` will not bind to that variable. (To bind to the variable, use `@field`.)

We should consider updating the diagnostics to address these issues.

jaredpar:
> The warning is reported for a reference to field, but not for a declaration of a local or parameter named field within the accessor, even though field will not bind to that variable. (To bind to the variable, use @field.)

I think that we need to draw inspiration from `value` here. Within an accessor it's not possible to declare a local named `value` because it would effectively shadow the implicit parameter. Given that we try and mirror `field` and `value` as much as possible, I think that we should error on local declarations. If you want to keep the local declaration then name it `@field`.

cston:
The second issue was resolved in https://github.com/dotnet/roslyn/pull/76671.

We will *not* fix the first issue. Instead of warning in earlier language versions that the behavior will change when binding `field` to an existing symbol in scope, the compiler warns in later language versions that `field` binds to the synthesized backing field instead of the existing symbol.

PR #76671, opened 2025-01-08
Field-backed properties: report diagnostic for variable named field declared in accessor

cston:
Report an error for a *declaration* of a local or parameter named `field` within an accessor, since `field` will not bind to that variable. See item 2 in https://github.com/dotnet/roslyn/issues/76031.

See proposal: https://github.com/dotnet/csharplang/pull/9038.

Based on the variable declaration checks reverted in https://github.com/dotnet/roslyn/pull/74164.


src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -6886,8 +6886,11 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="WRN_FieldIsAmbiguous_Title" xml:space="preserve">
     <value>The 'field' keyword binds to a synthesized backing field for the property.</value>
   </data>
-  <data name="INF_IdentifierConflictWithContextualKeyword_Title" xml:space="preserve">
```
cston, line 6889: >INF_IdentifierConflictWithContextualKeyword_Title [](http://example.com/codeflow?start=14&length=49)

Unused.

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -561,6 +561,8 @@ private BoundStatement BindLocalFunctionStatement(LocalFunctionStatementSyntax n
             var hasErrors = localSymbol.ScopeBinder
                 .ValidateDeclarationNameConflictsInScope(localSymbol, diagnostics);
 
+            ReportFieldContextualKeywordConflictIfAny(localSymbol, node, node.Identifier, diagnostics);
```
RikkiGibson, line 564: I was curious if there was any more "central" location to do this check, like where local symbols are gathered up into a scope, or some such. It looks like probably not. Places like `LocalScopeBinder.MakeLocal` are not really designed for reporting diagnostics.

cston: I haven't found a central location to catch these cases.

RikkiGibson (approved):

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -1488,6 +1488,31 @@ bool hasOtherFieldSymbolInScope()
             }
         }
 
+        internal void ReportFieldContextualKeywordConflictIfAny(Symbol? symbol, SyntaxNode syntax, SyntaxToken identifier, BindingDiagnosticBag diagnostics)
+        {
+            // The symbol should be locally declared. That is, it should be a symbol
+            // that would hide a backing field in earlier language versions.
+            // If a symbol is not provided, the caller is responsible for
+            // ensuring the identifier refers to a locally declared variable.
```
333fred, line 1496: Perhaps make this a doc comment?

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
333fred: Could we perhaps catch all parameters in `ParameterHelpers.MakeParameters`?

cston: `ParameterHelpers.MakeParameters()` will not work as is because that method is not used for lambda parameters (`LambdaSymbol.MakeParameters()` is used instead), and the method does not include a `Binder` which is used for reporting. I'll leave this as is.

333fred (approved):

Issue #76444, opened 2024-12-16
Unexpected value 'UnconvertedInterpolatedString' in CSharpOperationFactory


Cosifne:
**Version Used**: Version 17.13.0 Preview 3.0 [35615.4.main]

Many golden bars shows up when I typing in this file:
![Image](https://github.com/user-attachments/assets/d1b8d5f3-36bc-4608-8eaa-498887303a9e)

```
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

using Microsoft.CodeAnalysis.Options;

namespace Roslyn.VisualStudio.Next.UnitTests.UnifiedSettings.TestModel
{
    internal class ExpectedSetting(string path, IOption2 option, UnifiedSettingBase setting)
    {
        public string ExpectedUnifiedSettingsPath => path;
        public IOption2 Option => option;
        public UnifiedSettingBase ExpectedUnifiedSetting => setting;

        public static ExpectedSetting Create<T>(
            string ExpectedUnifiedSettingsPath,
            IOption2 roslynOption,
            string title,
            int order,
            T defaultValue,
            (IOption2 featureFlagOption, object value) featureFlagAndExperimentValue = default,
            (IOption2 enableWhenOption, object whenValue) enableWhenOptionAndValue = default,
            string? languageName = null)
        {
            var migration = new Migration
            {
                Pass = new Pass()
                {
                    Input = Input(roslynOption, languageName)
                }
            };

            var alternativeDefault = featureFlagAndExperimentValue is not default
                ? new AlternativeDefault<T>(featureFlagAndExperimentValue.featureFlagOption, featureFlagAndExperimentValue.value)
                : null;

            var enableWhen = enableWhenOptionAndValue is not default
                ? $"config:{enableWhenOptionAndValue.}"='{en}'"

        }
    }
}

```

The stack trace is:
```
StreamJsonRpc.RemoteInvocationException: Unexpected value 'UnconvertedInterpolatedString' of type 'Microsoft.CodeAnalysis.CSharp.BoundKind'
   at StreamJsonRpc.JsonRpc.<InvokeCoreAsync>d__156`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #76675, opened 2025-01-08
Ensure unconverted elements are converted when on LHS

333fred:
Fixes https://github.com/dotnet/roslyn/issues/76444.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
@@ -1431,10 +1431,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                 op2 = BindToNaturalType(op2, diagnostics);
                 op1 = InferTypeForDiscardAssignment((BoundDiscardExpression)op1, op2, diagnostics);
             }
-            else
-            {
-                op1 = BindToTypeForErrorRecovery(op1);
-            }
 
             return BindAssignment(node, op1, op2, isRef, diagnostics);
         }
@@ -1488,6 +1484,17 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             if (op1.Type is { } lhsType && !lhsType.IsErrorType())
             {
+                Debug.Assert(!op1.NeedsToBeConverted()
+                    // These are only required to be converted in Debug to increase confidence that we've called BindToTypeForErrorRecovery or BindToNaturalType
+                    // on the operand. We're calling BindToTypeForErrorRecovery below, so we're good.
+                    || op1 is BoundParameter or BoundLocal
+                    || op1.HasAnyErrors);
+
+                if (op1.HasAnyErrors)
+                {
+                    op1 = BindToTypeForErrorRecovery(op1);
+                }
+
                 // Build bound conversion. The node might not be used if this is a dynamic conversion
                 // but diagnostics should be reported anyways.
                 var conversion = GenerateConversionForAssignment(lhsType, op2,
@@ -1510,6 +1517,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
             else
             {
+                op1 = BindToTypeForErrorRecovery(op1);
                 op2 = BindToTypeForErrorRecovery(op2);
             }
 
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
@@ -1431,6 +1431,10 @@ namespace Microsoft.CodeAnalysis.CSharp
                 op2 = BindToNaturalType(op2, diagnostics);
                 op1 = InferTypeForDiscardAssignment((BoundDiscardExpression)op1, op2, diagnostics);
             }
+            else
+            {
+                op1 = BindToTypeForErrorRecovery(op1);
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -1431,6 +1431,10 @@ private BoundExpression BindAssignment(AssignmentExpressionSyntax node, BindingD
                 op2 = BindToNaturalType(op2, diagnostics);
                 op1 = InferTypeForDiscardAssignment((BoundDiscardExpression)op1, op2, diagnostics);
             }
+            else
+            {
+                op1 = BindToTypeForErrorRecovery(op1);
```
AlekseyTs, line 1436: >op1 = BindToTypeForErrorRecovery(op1); [](http://example.com/codeflow?start=16&length=38)

This doesn't look like a code path that is taken only in an error situation, therefore, calling ```BindToTypeForErrorRecovery``` unconditionally makes me uncomfortable.

AlekseyTs:

333fred:
@AlekseyTs addressed your feedback.

AlekseyTs (approved):
LGTM (commit 2)

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -1484,6 +1484,17 @@ private BoundAssignmentOperator BindAssignment(
 
             if (op1.Type is { } lhsType && !lhsType.IsErrorType())
             {
+                Debug.Assert(!op1.NeedsToBeConverted()
+                    // These are only required to be converted in Debug to increase confidence that we've called BindToTypeForErrorRecovery or BindToNaturalType
+                    // on the operand. We're calling BindToTypeForErrorRecovery below, so we're good.
+                    || op1 is BoundParameter or BoundLocal
+                    || op1.HasAnyErrors);
+
+                if (op1.HasAnyErrors)
```
cston, line 1493: >HasAnyErrors [](http://example.com/codeflow?start=24&length=12)

This check seems too broad since the error could in a nested expression with no error on `op1` directly. In short, it seems surprising we're binding for error recovery in that situation.

If we're just trying to ensure `op1` is converted, would it make sense to do that explicitly?
```csharp
if (op1.NeedsToBeConverted())
{
    op1 = BindToNaturalType(op1, BindingDiagnosticBag.Discard, reportNoTargetType: false);
}
```

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
```diff
@@ -1506,6 +1517,7 @@ private BoundAssignmentOperator BindAssignment(
             }
             else
             {
+                op1 = BindToTypeForErrorRecovery(op1);
```
cston, line 1520: >op1 = BindToTypeForErrorRecovery(op1); [](http://example.com/codeflow?start=16&length=38)

Are we testing this change?

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
333fred: My original approach did effectively this. Aleksey was concerned, rightly so, that discarding diagnostics here is probably not safe, as there is no guarantee that errors have been reported for `op1`. I do not think that it is surprising that we bind for error recovery in this scenario when we know that there are errors; in such a scenario, we simply want to be able to give the most precise info that we can.

cston: > discarding diagnostics here is probably not safe, as there is no guarantee that errors have been reported for `op1`.

In that case, we could pass in the `diagnostics` instance.

> I do not think that it is surprising that we bind for error recovery in this scenario when we know that there are errors

`HasAnyErrors` indicates there is an error somewhere beneath `op1`, but the outermost node `op1` might be otherwise valid.

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
333fred: I cannot find a way to hit this line in a way that would fail without this binding; however, I believe that, since this is an error scenario anyways, more defensive coding ensuring that this goes through proper conversion checks is a good thing. If you want, I can add an assert to fail if an interesting case ever goes through here, ie `Debug.Assert(op1 is BoundParameter or BoundLocal || !op1.NeedsToBeConverted(), "It is currently believed that this isn't hittable. If you trigger this assert, please an explicit test!");`?

cston: Thanks for investigating. I don't think we need an assert.

src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs
333fred: If you would prefer, I can switch to just checking `HasErrors`, rather than `HasAnyErrors`, so it's only on the node itself.

333fred:
@cston please take another look

cston (approved):

cston (approved):

AlekseyTs (approved):
LGTM (commit 4)

PR #76713, opened 2025-01-10
Stash and restore original culture in CultureNormalizer

333fred:


src/Compilers/Test/Utilities/CSharp/CSharpTestBase.cs
```diff
@@ -2544,10 +2544,18 @@ internal static string GetInterpolatedStringCustomHandlerType(string name, strin
             var cultureInfoHandler = @"
 public class CultureInfoNormalizer
 {
+    private static CultureInfo originalCulture;
+
     public static void Normalize()
```
cston, line 2549: >static [](http://example.com/codeflow?start=11&length=6)

Consider having `Normalize()` return an `IDisposable` to allow callers to use `using`.

cston (approved):

src/Compilers/Test/Utilities/CSharp/CSharpTestBase.cs
333fred: > > static
> 
> Consider having `Normalize()` return an `IDisposable` to allow callers to use `using`.

I intentionally did not do this to minimize change to existing test IL verification.

jaredpar (approved):

PR #76715, opened 2025-01-10
Fix XML Docs for IParameterSymbol.ExplicitDefaultValue

thomhurst:


333fred:
Hi @thomhurst, what makes you think these docs are incorrect? For example, I see this print `true`:
```cs
var code = """
void M(S i = default) {}

struct S {}
""";

var tree = CSharpSyntaxTree.ParseText(code);

var comp = CSharpCompilation.Create("test")
    .AddSyntaxTrees(tree)
    .WithReferenceAssemblies(ReferenceAssemblyKind.Net80);

var parameter = tree.GetRoot().DescendantNodes().OfType<ParameterSyntax>().Single();
var parameterSymbol = model.GetDeclaredSymbol(parameter)!;
Console.WriteLine(parameterSymbol.ExplicitDefaultValue is null);
```

thomhurst:
Let me re-investigate then, maybe I messed up debugging!

PR #76716, opened 2025-01-10
Handle nullable postcondition for non-static local function inside field initializer

### LINKED ISSUE #76528: [MemberNotNull(When)] annotations don't work on local functions inside member initializers inside non-static closure chains
**Version Used**: 
```
Branch main (19 Dec 2024)
Latest commit [395d298](https://github.com/dotnet/roslyn/commit/395d298bd9093150e41580be29c53955d2781874) by Cyrus Najmabadi:
Remove sync over async blocking in Extract Method/Interface options code (#76510)
```

**Steps to Reproduce**:

Compile the following code ([sharplab.io](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQAYACdbAOgBEoIBzGAewGcAXKAY2YoGFGEApgEEYEADYBPZlGYBuAgGIYAVzFiIwMQJIDRmgQXQBmMphI8CAbwIkyJ8gDYy2IgH4SAMygCxCefhs7ZzMABQAnRgAHbBJLEjoBVlkSAF8SAF4SAAoASgyAPkDba3xbMpJYKFZc/3LbcgBOLK8fBAoAGV06VgALHOSAegHbHgBlAA4HIkwQEioBMIEPBd0ObUYPEggSSJZpTQkSFTUSReXFmDWKAiKygG0AWQEwYAWAOUZWN9UxLNEwAQbZreXw5HIAXVu9QALBUYFVcgVPCCECRXK5MgAiTG1MopfwpIA))
```cs
using System;
#nullable enable
public class C
{
  public static string? field;

  public Action Prop1 { get; } = () =>
    {
      init();
      Console.WriteLine(field.Length); //  CS8602: Dereference of a possibly null reference.

      [MemberNotNull(nameof(field))]
      void init() => field ??= "";
    };
}
```

**Expected Behavior**:
No warning fo `field.Length` access in `Prop1`. The local function ensures that the member is not nullable and declares it in its annotation

**Actual Behavior**:
Incorrect CS8602 warning for `Prop1`

**Notes**:
You can get rid of the false warning by adding `static` to any of the closures in this example (either to the lambda itself or to the local function) e.g.
```
  public Action Prop1 { get; } = static () =>
    {
      init();
      Console.WriteLine(field.Length); // no warning

      [MemberNotNull(nameof(field))]
      void init() => field ??= "";
    };
```

However, it seems that this shouldn't be required - `static` only disallows closures and shouldn't affect the behavior when there are none.
`[MemberNotNull(When)]` doesn't require any static modifiers in other members, e.g.:
```
  public Action M() => () => 
    {
      init();
      Console.WriteLine(field.Length); // no warning

      [MemberNotNull(nameof(field))]
      void init() => field ??= "";
    };
```

RikkiGibson:
Closes #76528

### DIFF
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -7188,6 +7188,13 @@ namespace Microsoft.CodeAnalysis.CSharp
                 }
             }
 
+            if (current.ContainingSymbol is FieldSymbol)
+            {
+                // Functions inside a field initializer are effectively static.
+                Debug.Assert(method.MethodKind is MethodKind.LambdaMethod or MethodKind.LocalFunction);
+                return 0;
+            }
+
             if (current.TryGetThisParameter(out var thisParameter) && thisParameter is not null)
             {
                 return GetOrCreateSlot(thisParameter);
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -7147,7 +7147,7 @@ namespace Microsoft.CodeAnalysis.CSharp
         }
 
         /// <summary>
-        /// Applies the member postconditions of <paramref name="method"/> to members of <see cref="receiverOpt"/> or to the appropriate static members.
+        /// Applies the member postconditions of <paramref name="method"/> to members of <paramref name="receiverOpt"/> or to the appropriate static members.
         /// Used for the "use-site" analysis of MemberNotNullAttributes.
         /// </summary>
         private void ApplyMemberPostConditions(BoundExpression? receiverOpt, MethodSymbol? method)
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -1172,6 +1172,10 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
         }
 
+        /// <summary>
+        /// Gets a slot for a static member, or a member of 'this', which is being referenced by a postcondition.
+        /// Used for "declaration-site" analysis of MemberNotNullAttributes.
+        /// </summary>
         private int GetSlotForMemberPostCondition(Symbol member)
         {
             if (member.Kind != SymbolKind.Field &&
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Test/Semantic/Semantics/LocalFunctionTests.cs
```diff
@@ -10588,5 +10588,63 @@ void M()
...
+                #nullable enable
+
+                using System;
+                using System.Diagnostics.CodeAnalysis;
+
+                public class C
+                {
+                    public static string? field;
+
+                    public Action Prop1 { get; } = () =>
```
jjonescz, line 10607: Consider testing also when the lambdas are marked `static`.

jjonescz (approved):

RikkiGibson:

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7188,6 +7188,13 @@ private int GetReceiverSlotForMemberPostConditions(MethodSymbol? method)
                 }
             }
 
+            if (current.ContainingSymbol is FieldSymbol)
+            {
+                // Functions inside a field initializer are effectively static.
```
AlekseyTs, line 7193: >// Functions inside a field initializer are effectively static. [](http://example.com/codeflow?start=16&length=63)

Is this accurate? For example, it looks like in the following scenario both the lambda and the local function are implemented by an instance method.
```
#nullable enable

using System;

public class C(string p)
{
    public static string? field;

    public Action field2 = () =>
    {
        init();
        Console.WriteLine(field.Length);

        [MemberNotNull(nameof(field))]
        void init() => field ??= p;
    };
    
    string M() => p;
}
```

AlekseyTs:

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
RikkiGibson: You're right, when marked `static` the functions cannot use `p`. I'll adjust the comment and add the test.

AlekseyTs: It looks like the comment got removed, but it feels like we still need some rational to explain why returning 0 is probably not 100% accurate, but OK thing to do.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -7144,7 +7146,11 @@
             }
         }
 
+        /// <summary>
+        /// Applies the member postconditions of <paramref name="method"/> to members of <see cref="receiverOpt"/> or to the appropriate static members.
```
AlekseyTs, line 7150: >see [](http://example.com/codeflow?start=90&length=3)

`paramref`? It looks like CI is failing here.

RikkiGibson: This is fixed now.

AlekseyTs (approved):
LGTM (commit 8)

jjonescz (approved):

PR #76749, opened 2025-01-14
Test layering fixes

tmat:
Removes editor dependencies from workspace/LSP tests.


src/Workspaces/CoreTest/UtilityTest/IntervalTreeTests.cs
```diff
@@ -309,6 +307,9 @@ public void TestSortedEnumerable2()
         Assert.Equal(tree, [0, 1]);
     }
 
+    public static bool SpansOverlapsWith((int start, int length) left, (int start, int length) right)
+        => Math.Max(left.start, right.start) < Math.Min(left.start + left.length, right.start + right.length);
```
jasonmalinowski, line 311: You should be able to just have this call TextSpan.OverlapsWith, which would be our actual implementation in practice. Otherwise this test might just end up testing this test...

jasonmalinowski (approved):
:shipit:, although one concern about the IntervalTreeTests. We should be able to call our existing TextSpan helper methods there.

src/Workspaces/CoreTest/UtilityTest/IntervalTreeTests.cs
jasonmalinowski: I didn't review this for correctness under the assumption it'll just go away.

PR #76750, opened 2025-01-14
Change ideal Helix execution time

jaredpar:
This is an experiment to change the amount of time we spend executing a work item in Helix. Want to see if we get better machine utilization if we spend more time in the work items


jaredpar:
@dibarbet PTAL

dibarbet (approved):
lets see how it goes!

PR #76752, opened 2025-01-15
Add an equality comparer for AssemblyName in RazorAnalyzerAssemblyResolver

chsienki:
`AssemblyName` doesn't implement `GetHashCode` so each instance was unique, leading to duplicates in the hash set. This PR adds an equality comparer that uses the `FullName` of the `AssemblyName` as the comparison item for the hash set.

### DIFF
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -59,6 +59,8 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
 
         private sealed class AssemblyNameEqualityComparer : IEqualityComparer<AssemblyName>
         {
+            // Note that we explicitly consider the FullName here so that two assemblies with
+            // the same name but different versions are considered to be different
             public bool Equals(AssemblyName? x, AssemblyName? y) => x!.FullName.Equals(y!.FullName);
 
             public int GetHashCode(AssemblyName obj) => Hash.GetFNVHashCode(obj.FullName);
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -57,7 +57,7 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
             return resolver?.Invoke(assemblyName);
         }
 
-        private class AssemblyNameEqualityComparer : IEqualityComparer<AssemblyName>
+        private sealed class AssemblyNameEqualityComparer : IEqualityComparer<AssemblyName>
         {
             public bool Equals(AssemblyName? x, AssemblyName? y) => x!.FullName.Equals(y!.FullName);
 
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -10,6 +10,7 @@ using System.Reflection;
 using Microsoft.CodeAnalysis.Host.Mef;
 using Microsoft.CodeAnalysis.PooledObjects;
 using Microsoft.VisualStudio.Composition;
+using Roslyn.Utilities;
 
 namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
 {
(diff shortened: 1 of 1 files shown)


chsienki:
@dotnet/roslyn-compiler for review please :)

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -54,5 +55,12 @@ internal static bool TrySetAssemblyResolver(Func<AssemblyName, Assembly?> resolv
 
             return resolver?.Invoke(assemblyName);
         }
+
+        private class AssemblyNameEqualityComparer : IEqualityComparer<AssemblyName>
+        {
+            public bool Equals(AssemblyName? x, AssemblyName? y) => x?.FullName.Equals(y?.FullName) == true;
+
+            public int GetHashCode(AssemblyName obj) => Hash.GetFNVHashCode(obj.FullName);
```
RikkiGibson, line 63: I guess this direct call to a hash function is preferable to `string.GetHashCode()`?

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -54,5 +55,12 @@ internal static bool TrySetAssemblyResolver(Func<AssemblyName, Assembly?> resolv
 
             return resolver?.Invoke(assemblyName);
         }
+
+        private class AssemblyNameEqualityComparer : IEqualityComparer<AssemblyName>
+        {
+            public bool Equals(AssemblyName? x, AssemblyName? y) => x?.FullName.Equals(y?.FullName) == true;
```
RikkiGibson, line 61: It's not obvious to me this does the right thing when one or both AssemblyNames are null. Maybe it would be simpler to branch like `x is null ? y is null : x.FullName.Equals(y.FullName)`.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
AlekseyTs, line 63: >obj [](http://example.com/codeflow?start=76&length=3)

Should we be prepared to deal with `null` as an input?

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -54,5 +55,12 @@ internal static bool TrySetAssemblyResolver(Func<AssemblyName, Assembly?> resolv
 
             return resolver?.Invoke(assemblyName);
         }
+
+        private class AssemblyNameEqualityComparer : IEqualityComparer<AssemblyName>
```
jaredpar, line 59:
```suggestion
        private sealed class AssemblyNameEqualityComparer : IEqualityComparer<AssemblyName>
```

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
jaredpar: On .NET core only the short name is considered for loading purposes in the majority case. This code is going to differentiate based on the full name though.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
chsienki: `IEqualityComparer<T>.GetHashCode` is marked `[DisallowNull]` which I beleive the intent is that it will never be called with a null value. We also own the parent `IAnalyzerAssemblyResolver` interface, and this comparer is only used here, so annotations say we can never see a null value for `AssemblyName` even if the comparer did support calling it with a null value. I'm happy to handle null though if you'd prefer.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
chsienki: Yeah, so I mentioned this on Aleksey's comment, but we control the parent interface which is annotated as never being null, and the comparer is only ever used here. I've updated it to `!` those values to indicate we know it will never be null as opposed to making the test stronger. Happy to switch back to a stronger null check if you prefer though.

chsienki: >On .NET core only the short name is considered for loading purposes in the majority case. This code is going to differentiate based on the full name though.

This was deliberate. Full name includes the version string i.e something like `Microsoft.CodeAnalysis.Razor.Compiler, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e`

Whereas `.Name` (which I assume is what you mean by shortname?) only includes the assembly name itself (`Microsoft.CodeAnalysis.Razor.Compiler`). Given that this is AssemblyLoadContext code its completely valid to load two different versions of the same assembly.

If Razor specifically asks for version 9.0.0.0 I don't want to consider an 8.0.0.0 version to be equivalent.

chsienki:
@dotnet/roslyn-compiler for second round of feedback. Thanks!

jaredpar (approved):

333fred (approved):

chsienki:
Closing in favor of https://github.com/dotnet/roslyn/pull/76808

PR #76753, opened 2025-01-15
Move impl of ILspWorkpace to EditorTestWorkspace/LspTestWorkspace

tmat:
Removes dependency on LSP from workspace tests.


tmat:
@dibarbet @jasonmalinowski ptal

src/LanguageServer/Protocol.TestUtilities/Workspaces/LspTestWorkspace.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.LanguageServer;
+using Microsoft.CodeAnalysis.Text;
+using Roslyn.Utilities;
+
+namespace Microsoft.CodeAnalysis.Test.Utilities;
+
+public partial class LspTestWorkspace : TestWorkspace, ILspWorkspace
```
dibarbet, line 15: does this need to be partial?

src/LanguageServer/ProtocolUnitTests/Options/SolutionAnalyzerConfigOptionsUpdaterTests.cs
```diff
@@ -23,7 +23,7 @@ public class SolutionAnalyzerConfigOptionsUpdaterTests
 {
     private static TestWorkspace CreateWorkspace()
     {
-        var workspace = new TestWorkspace(LspTestCompositions.LanguageServerProtocol
+        var workspace = new LspTestWorkspace(LspTestCompositions.LanguageServerProtocol
```
dibarbet, line 26: With LspTestWorkspace - is it now possible to remove the windows only requirement from the LSP tests?  IIRC it was only there because the Editor test workspace targeted net472 or something like that?

definitely not asking for it in this PR, but wondering if I can do that in followups

dibarbet (approved):

src/LanguageServer/ProtocolUnitTests/Options/SolutionAnalyzerConfigOptionsUpdaterTests.cs
tmat: idk, feel free to try

src/LanguageServer/Protocol.TestUtilities/Workspaces/LspTestWorkspace.cs
tmat: Probably not

Issue #71374, opened 2023-12-20
Remove ServiceHub PackageReferences


tmat:
We have to include PackageReferences in some of our project to resolve ambiguities between the version of Microsoft.ServiceHub.Framework and Microsoft.ServiceHub.Client we are currently using. 

We can't update to newer version of Microsoft.ServiceHub.Client since it dropped support for netstandard2.0. This is being fixed in the latest version. When we can update to the latest we should remove the workaround.

```xml
    <PackageReference Include="Microsoft.ServiceHub.Framework" Version="$(MicrosoftServiceHubFrameworkVersion)" />
    <PackageReference Include="Microsoft.ServiceHub.Client" Version="$(MicrosoftServiceHubClientVersion)" />
```

See TODO in project files.

PR #76767, opened 2025-01-15
Make IServiceBrokerProvider available in all layers

tmat:
Previously the interface was not exported in OOP. MEF components that import it failed to compose when moved OOP.

Fixes https://github.com/dotnet/roslyn/issues/71374


tmat:
@CyrusNajmabadi ptal

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/BrokeredServices/ServiceBrokerProvider.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#pragma warning disable RS0030 // Do not use banned APIs (MEFv1)
```
tmat, line 5: Is there a way to make the import work with MEF v2?

CyrusNajmabadi (approved):

Issue #76765, opened 2025-01-15
NodeStateTable.GetModifiedItemAndState does not guard against a throwing .Equals method


chsienki:
When the comparer used here: https://github.com/dotnet/roslyn/blob/9db0f39f590e1e7d85b3330a11e35b9d5acc6d29/src/Compilers/Core/Portable/SourceGeneration/Nodes/NodeStateTable.cs#L557-L564

is *not* a user supplied comparer, and the call to `T.Equals()` throws, we don't catch the exception and convert into a `UserFunctionException`. This can cause a generator to bring the hosting process down. We should always wrap default comparers in a user wrapped user comparer to ensure we still do the translation.

See here for an example: https://github.com/microsoft/CsWinRT/issues/1897

PR #76769, opened 2025-01-16
Fix throw in generator comparer

chsienki:
When checking for generator item modification it's possible for an object to throw when calling `.Equals()`. User supplied comparers are wrapped in a special comparer that catches the exception and wraps it in a `UserFunctionException` that is used by the generator driver to handle the error and attribute it to the offending generator. 

However, if no comparer is supplied the generator uses the default comparer which does not translate exceptions, meaning it can crash the driver. 

Whilst fixing this I noticed that we were passing a lot of default comparers around and passing the same comparer to modify calls as we did in construction of the state table, none of which is required. By centralizing the comparer creation and passing we only need to change the default comparer to a wrapped comparer in a single place.

Commit by Commit review makes this simpler to see what / why I did. 

Fixes https://github.com/dotnet/roslyn/issues/76765

### DIFF
--- a/src/Compilers/Core/Portable/SourceGeneration/Nodes/CombineNode.cs
+++ b/src/Compilers/Core/Portable/SourceGeneration/Nodes/CombineNode.cs
@@ -72,7 +72,7 @@ namespace Microsoft.CodeAnalysis
                 };
 
                 var entry = (entry1.Item, input2);
-                if (state != EntryState.Modified || _comparer is null || !tableBuilder.TryModifyEntry(entry, _comparer, stopwatch.Elapsed, stepInputs, state))
+                if (state != EntryState.Modified || _comparer is null || !tableBuilder.TryModifyEntry(entry, stopwatch.Elapsed, stepInputs, state))
                 {
                     tableBuilder.AddEntry(entry, state, stopwatch.Elapsed, stepInputs, state);
                 }
--- a/src/Compilers/Core/Portable/SourceGeneration/Nodes/InputNode.cs
+++ b/src/Compilers/Core/Portable/SourceGeneration/Nodes/InputNode.cs
@@ -24,7 +24,7 @@ namespace Microsoft.CodeAnalysis
         private readonly Func<DriverStateTable.Builder, ImmutableArray<T>> _getInput;
         private readonly Action<IIncrementalGeneratorOutputNode> _registerOutput;
         private readonly IEqualityComparer<T> _inputComparer;
-        private readonly IEqualityComparer<T> _comparer;
+        private readonly IEqualityComparer<T>? _comparer;
         private readonly string? _name;
 
         public InputNode(Func<DriverStateTable.Builder, ImmutableArray<T>> getInput, IEqualityComparer<T>? inputComparer = null)
@@ -35,7 +35,7 @@ namespace Microsoft.CodeAnalysis
         private InputNode(Func<DriverStateTable.Builder, ImmutableArray<T>> getInput, Action<IIncrementalGeneratorOutputNode>? registerOutput, IEqualityComparer<T>? inputComparer = null, IEqualityComparer<T>? comparer = null, string? name = null)
         {
             _getInput = getInput;
-            _comparer = comparer ?? EqualityComparer<T>.Default;
+            _comparer = comparer;
             _inputComparer = inputComparer ?? EqualityComparer<T>.Default;
             _registerOutput = registerOutput ?? (o => throw ExceptionUtilities.Unreachable());
             _name = name;
@@ -83,7 +83,7 @@ namespace Microsoft.CodeAnalysis
                         // This allows us to correctly 'replace' items even when they aren't actually the same. In the case that the
                         // item really isn't modified, but a new item, we still function correctly as we mostly treat them the same,
                         // but will perform an extra comparison that is omitted in the pure 'added' case.
-                        var modified = tableBuilder.TryModifyEntry(inputItems[itemIndex], _comparer, elapsedTime, noInputStepsStepInfo, EntryState.Modified);
+                        var modified = tableBuilder.TryModifyEntry(inputItems[itemIndex], elapsedTime, noInputStepsStepInfo, EntryState.Modified);
                         Debug.Assert(modified);
                         itemsSet.Remove(inputItems[itemIndex]);
                     }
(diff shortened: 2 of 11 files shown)


chsienki:
@dotnet/roslyn-compiler for review please.

src/Compilers/Core/Portable/SourceGeneration/UserFunction.cs
```diff
@@ -25,6 +25,8 @@ internal sealed class WrappedUserComparer<T> : IEqualityComparer<T>
     {
         private readonly IEqualityComparer<T> _inner;
 
+        public static WrappedUserComparer<T> Default { get; } = new WrappedUserComparer<T>(EqualityComparer<T>.Default);
+
         public WrappedUserComparer(IEqualityComparer<T> inner)
```
jaredpar, line 30: When you originally wrote this rather than a ctor did you consider the following? 

```csharp
public static WrappedUserComparer Create(IEqualityComparer<T> comparer) => 
  comparer is WrappedUserComparer<T> w ? w : new WrappedUserComparer(comparer);
```

One concern I had when looking through this code was if we were inadvertently creating nested wraps.

src/Compilers/Core/Portable/SourceGeneration/Nodes/CombineNode.cs
```diff
@@ -72,7 +72,7 @@ public CombineNode(IIncrementalGeneratorNode<TInput1> input1, IIncrementalGenera
                 };
 
                 var entry = (entry1.Item, input2);
-                if (state != EntryState.Modified || _comparer is null || !tableBuilder.TryModifyEntry(entry, _comparer, stopwatch.Elapsed, stepInputs, state))
+                if (state != EntryState.Modified || _comparer is null || !tableBuilder.TryModifyEntry(entry, stopwatch.Elapsed, stepInputs, state))
```
jaredpar, line 75: What does the check `_comparer is null` serve here? The `null` case means it's effectively the default comparer so why are we not trying to use that for `TryModifyEntry`?

src/Compilers/Core/Portable/SourceGeneration/Nodes/PredicateSyntaxStrategy.cs
```diff
@@ -48,7 +48,7 @@ public Builder(PredicateSyntaxStrategy<T> owner, object key, StateTableStore tab
                 _name = name;
                 _comparer = comparer;
                 _key = key;
-                _filterTable = table.GetStateTableOrEmpty<SyntaxNode>(_owner._filterKey).ToBuilder(stepName: null, trackIncrementalSteps);
+                _filterTable = table.GetStateTableOrEmpty<SyntaxNode>(_owner._filterKey).ToBuilder(stepName: null, trackIncrementalSteps, equalityComparer: Roslyn.Utilities.ReferenceEqualityComparer.Instance);
```
jaredpar, line 51: `T` is not constrainted to `class` so why do we use a reference equality comparer vs. simply the default comparer?

src/Compilers/Core/Portable/SourceGeneration/Nodes/SourceOutputNode.cs
```diff
@@ -46,12 +46,12 @@ public NodeStateTable<TOutput> UpdateStateTable(DriverStateTable.Builder graphSt
                 this.LogTables(stepName, s_tableType, previousTable, previousTable, sourceTable);
                 if (graphState.DriverState.TrackIncrementalSteps)
                 {
-                    return previousTable.CreateCachedTableWithUpdatedSteps(sourceTable, stepName, EqualityComparer<TOutput>.Default);
+                    return previousTable.CreateCachedTableWithUpdatedSteps(sourceTable, stepName, equalityComparer: null);
```
jaredpar, line 49: This creates a new `NodeStateTable` that drops the comparer from `previousTable`. Is that intended?

src/Compilers/Core/Portable/SourceGeneration/Nodes/InputNode.cs
```diff
@@ -35,7 +35,7 @@ public InputNode(Func<DriverStateTable.Builder, ImmutableArray<T>> getInput, IEq
         private InputNode(Func<DriverStateTable.Builder, ImmutableArray<T>> getInput, Action<IIncrementalGeneratorOutputNode>? registerOutput, IEqualityComparer<T>? inputComparer = null, IEqualityComparer<T>? comparer = null, string? name = null)
         {
             _getInput = getInput;
-            _comparer = comparer ?? EqualityComparer<T>.Default;
+            _comparer = comparer;
             _inputComparer = inputComparer ?? EqualityComparer<T>.Default;
```
jjonescz, line 39: Should `_inputComparer` be handled similarly?

src/Compilers/Core/Portable/SourceGeneration/UserFunction.cs
chsienki: It's an internal class and you can only create it via an extension method that doesn't return it, so it shouldn't be possible to actually get an instance of it that can be passed to the ctor as an external user.

src/Compilers/Core/Portable/SourceGeneration/Nodes/PredicateSyntaxStrategy.cs
chsienki: This was code written by Cyrus originally, I'm refactoring it to keep it semantically the same.

src/Compilers/Core/Portable/SourceGeneration/Nodes/SourceOutputNode.cs
chsienki: Yes, if it's all cached the comparer would never be used. Later on, if there's a modification the table will be re-created and the real comparer will be passed to it.

src/Compilers/Core/Portable/SourceGeneration/Nodes/InputNode.cs
chsienki: No, we control the input comparer and all the inputs it compares. If that's throwing it's a real product bug and we don't want to falsely attribute it to a generator.

src/Compilers/Core/Portable/SourceGeneration/Nodes/CombineNode.cs
chsienki: It's an optimization. If we've already determined that one of the two sides is modified then the default comparer will always find the same result. However if there's a custom comparer it's possible that will override the result and return cached.

chsienki:
@dotnet/roslyn-compiler for any further feedback, thanks!

chsienki:
Closing in favor of https://github.com/dotnet/roslyn/pull/76808

jjonescz (approved):

jaredpar (approved):

Issue #76707, opened 2025-01-10
PEVerify fails compiling entrypoint in metadata-only mode


jjonescz:
**Version Used**: 9881daa0056e5a12ad968a088975ab9afdfb4f9f

**Steps to Reproduce**:

```cs
[Fact]
public void EmitMetadataOnly_TopLevelStatements()
{
    CompileAndVerify("""
        System.Console.WriteLine("a");
        """,
        emitOptions: EmitOptions.Default.WithEmitMetadataOnly(true))
        .VerifyDiagnostics();
}
```

**Expected Behavior**: Succeeds.

**Actual Behavior**: PEVerify fails with "Bad token as entry point in CLR header."

Note that using Main and TestOptions.ReleaseExe fails in the same way.

Perhaps it's expected that ref assemblies are only supported by PEVerify for class libraries, not console apps.

AlekseyTs:
CC @jcouv

jjonescz:
Working on a fix by emitting entrypoint in metadata-only EXE compilation. But it's a breaking change - previously a metadata-only EXE compilation wouldn't fail if the entrypoint was missing, now it does. Alternatively, I could make it so that it doesn't fail when the entrypoint is missing and continue with the previous behavior in that case.

PR #76783, opened 2025-01-16
Emit entrypoint to ref assemblies

jjonescz:
Fixes https://github.com/dotnet/roslyn/issues/76707.


src/Compilers/CSharp/Portable/Compilation/CSharpCompilation.cs
```diff
@@ -3503,6 +3503,20 @@ internal override bool CompileMethods(
...
 
                 SynthesizedMetadataCompiler.ProcessSynthesizedMembers(this, moduleBeingBuilt, cancellationToken);
+
+                if (moduleBeingBuilt.OutputKind.IsApplication())
+                {
+                    var entryPoint = GetEntryPointAndDiagnostics(cancellationToken);
+                    diagnostics.AddRange(entryPoint.Diagnostics.Diagnostics);
+                    if (entryPoint.MethodSymbol != null)
+                    {
+                        moduleBeingBuilt.SetPEEntryPoint(entryPoint.MethodSymbol, diagnostics);
```
AlekseyTs, line 3513: >moduleBeingBuilt.SetPEEntryPoint(entryPoint.MethodSymbol, diagnostics); [](http://example.com/codeflow?start=24&length=71)

Is this going to do a right thing for an `async` Main? It looks like ```GetEntryPoint``` helper synthesizes a wrapper method as an entry point in that case.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -3507,5 +3507,93 @@ public class InvalidOperationException();
...
+                stream.Position = 0;
+                Assert.NotEqual(0, new PEHeaders(stream).CorHeader.EntryPointTokenOrRelativeVirtualAddress);
+
+                stream.Position = 0;
+                var reference = AssemblyMetadata.CreateFromStream(stream).GetReference();
+                var comp = CreateCompilation("", references: [reference],
+                    options: TestOptions.DebugDll.WithMetadataImportOptions(MetadataImportOptions.All));
+                Assert.NotNull(comp.GetMember<MethodSymbol>("Program.Main"));
+            }
+        }
```
AlekseyTs, line 3597: >} [](http://example.com/codeflow?start=8&length=1)

I think we should add tests for `async` Main scenarios.

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests.cs
```diff
@@ -4879,12 +4879,14 @@ class C { }
         public void AttributeArgumentAsEnumFromMetadata()
         {
             var metadataStream1 = CSharpCompilation.Create("bar.dll",
+                options: new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary),
```
AlekseyTs, line 4882: >new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary) [](http://example.com/codeflow?start=25&length=65)

We have a bunch of predefined options in ```TestOptions```. consider using one of them instead.

src/Compilers/VisualBasic/Test/Emit/Emit/EmitMetadata.vb
```diff
@@ -973,5 +975,104 @@ End Class
...
+
+        <Fact, WorkItem("https://github.com/dotnet/roslyn/issues/76707")>
+        Public Sub ExcludePrivateMembers_PrivateMain()
+            Using peStream As New MemoryStream()
+                Using metadataStream As New MemoryStream()
+                    Dim comp = CreateCompilation(
+                        <compilation>
+                            <file>
+Module Program
+    Friend Sub Main()
```
AlekseyTs, line 1046: >Friend [](http://example.com/codeflow?start=4&length=6)

Private?

AlekseyTs:

src/Compilers/CSharp/Test/Emit3/Attributes/AttributeTests.cs
jjonescz: I intentionally didn't use those to have a minimal change to the test. The test doesn't use test helpers to create the compilation either.

src/Compilers/VisualBasic/Test/Emit/Emit/EmitMetadata.vb
jjonescz: That doesn't work in VB. I guess I can rename the test though.

src/Compilers/CSharp/Portable/Compilation/CSharpCompilation.cs
jjonescz: It will set the async Main as the entrypoint, not the generated one. I think that's fine for ref assembly.

docs/compilers/CSharp/Compiler Breaking Changes - DotNet 10.md
```diff
@@ -291,3 +291,23 @@ unsafe record struct R(
...
+***Introduced in Visual Studio 2022 version 17.14***
+
+Previously, the entrypoint was [unintentionally unset](https://github.com/dotnet/roslyn/issues/76707)
+when emitting executables in metadata-only mode (also known as ref assemblies).
+That is now corrected but it also means that a missing entrypoint is a compilation error:
+
+```cs
+// previously successful, now fails:
+CSharpCompilation.Create("test").Emit(new MemoryStream(),
+    options: EmitOptions.Default.WithEmitMetadataOnly(true))
```
AlekseyTs, line 306: >EmitOptions.Default.WithEmitMetadataOnly(true) [](http://example.com/codeflow?start=13&length=46)

Consider also adding an example that is using command line options. I think only small number of users work with the APIs directly.

src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs
```diff
@@ -218,7 +218,7 @@ public static void CompileMethodBodies(
 
         // Returns the MethodSymbol for the assembly entrypoint.  If the user has a Task returning main,
         // this function returns the synthesized Main MethodSymbol.
-        private static MethodSymbol GetEntryPoint(CSharpCompilation compilation, PEModuleBuilder moduleBeingBuilt, bool hasDeclarationErrors, bool emitMethodBodies, BindingDiagnosticBag diagnostics, CancellationToken cancellationToken)
+        internal static MethodSymbol GetEntryPoint(CSharpCompilation compilation, PEModuleBuilder moduleBeingBuilt, bool hasDeclarationErrors, bool emitMethodBodies, BindingDiagnosticBag diagnostics, CancellationToken cancellationToken)
```
AlekseyTs, line 221: >GetEntryPoint [](http://example.com/codeflow?start=37&length=13)

It looks like this function still performs a lot of work around body (creating, lowering, etc.) even when ```emitMethodBodies``` is `false`. Therefore, its effect in our scenario feels insufficient. It feels like additional changes to the body of this method are necessary to handle scenario when ```moduleBeingBuilt.EmitOptions.EmitMetadataOnly``` is `true`.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -3507,5 +3507,183 @@ public class InvalidOperationException();
...
+                }
+                """,
+                options: TestOptions.ReleaseExe.WithMetadataImportOptions(MetadataImportOptions.All),
+                emitOptions: EmitOptions.Default
+                    .WithEmitMetadataOnly(true)
+                    .WithIncludePrivateMembers(false),
+                symbolValidator: static (ModuleSymbol module) =>
+                {
+                    Assert.NotEqual(0, module.GetMetadata().Module.PEReaderOpt.PEHeaders.CorHeader.EntryPointTokenOrRelativeVirtualAddress);
+                    Assert.NotNull(module.GlobalNamespace.GetMember<MethodSymbol>("Program.<Main>"));
```
AlekseyTs, line 3606: >Assert.NotNull(module.GlobalNamespace.GetMember<MethodSymbol>("Program.<Main>")); [](http://example.com/codeflow?start=20&length=81)

Consider asserting accessibility of this method.

src/Compilers/VisualBasic/Test/Emit/Emit/EmitMetadata.vb
```diff
@@ -973,5 +975,104 @@ End Class
...
+
+        Private Shared Sub Verify(stream As Stream)
+            stream.Position = 0
+            Assert.NotEqual(0, New PEHeaders(stream).CorHeader.EntryPointTokenOrRelativeVirtualAddress)
+
+            stream.Position = 0
+            Dim reference = AssemblyMetadata.CreateFromStream(stream).GetReference()
+            Dim comp = CreateCompilation("", references:={reference}, options:=TestOptions.DebugDll.WithMetadataImportOptions(MetadataImportOptions.All))
+            Assert.NotNull(comp.GetMember(Of MethodSymbol)("Program.Main") IsNot Nothing)
+        End Sub
```
AlekseyTs, line 1076: >End Sub [](http://example.com/codeflow?start=8&length=7)

It looks like VB doesn't support async Main. Still let's add a test for such scenario and observe an error.

AlekseyTs:

src/Compilers/VisualBasic/Test/Emit/Emit/EmitMetadata.vb
```diff
@@ -973,5 +975,130 @@ End Class
...
+                symbolValidator:=Sub(m)
+                                     Assert.NotEqual(0, m.GetMetadata().Module.PEReaderOpt.PEHeaders.CorHeader.EntryPointTokenOrRelativeVirtualAddress)
+                                     Dim main = m.GlobalNamespace.GetMember(Of MethodSymbol)("Program.Main")
+                                     Assert.Equal(Accessibility.Public, main.DeclaredAccessibility)
+                                 End Sub
+            ).VerifyDiagnostics()
+        End Sub
+
+        <Fact, WorkItem("https://github.com/dotnet/roslyn/issues/76707")>
+        Public Sub EmitMetadataOnly_Exe_AsyncMain()
```
AlekseyTs, line 1002: >EmitMetadataOnly_Exe_AsyncMain [](http://example.com/codeflow?start=19&length=30)

Lets keep this test, but I was looking for a scenario that would produce ```ERRID.ERR_AsyncSubMain``` error

jjonescz: Adding the test. Turns out the VB's GetEntrypoint can return non-null result plus some error diagnostics, so updating the code a bit to check for that as well.

AlekseyTs (approved):
LGTM (commit 7)

jaredpar:
Chatted with @jjonescz about the breaking change and I'm okay with it.

jcouv (approved):
LGTM Thanks (iteration 7)

src/Compilers/Core/Portable/PEWriter/Members.cs
```diff
@@ -1047,6 +1047,11 @@ public static bool ShouldInclude(this ITypeDefinitionMember member, EmitContext
                 }
             }
 
+            if (method != null && (context.Module.PEEntryPoint == method || context.Module.DebugEntryPoint == method))
```
jcouv, line 1050: >|| context.Module.DebugEntryPoint == method [](http://example.com/codeflow?start=73&length=43)

Do we have a test affected by this DebugEntryPoint logic?

jjonescz: Added tests for that, thanks.

jcouv (approved):
LGTM Thanks (iteration 8)

PR #76803, opened 2025-01-18
Always create return conversions

333fred:
In `BindReturn`, we would try to suppress additional diagnostics and avoid extra work by seeing if there were any errors in the returned expression; if there were, we didn't do any further work and just returned a `BoundReturnStatement` without doing conversion work. However, this is bad for speculative scenarios; speculation can introduce false-positive "errors", such as having duplicated local names in pattern declarations because the speculation is about replacing some existing expression with a declaration pattern with some new expression. You then observe unexpected semantics in that `return` statement, because the conversion to the actual return type of the method is missing. To fix this, we now always do the return conversion, and only suppress diagnostics in that case. Fixes https://github.com/dotnet/roslyn/issues/76799.


333fred:
I should have looked closer at @RikkiGibson's open PRs, because this is a duplicate of https://github.com/dotnet/roslyn/pull/76699.

PR #76808, opened 2025-01-18
Load razor assembly directly:

chsienki:
- Pass the root directory to assembly resolvers
- When looking for razor or its deps load them into the same ALC as the compiler

### DIFF
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -17,7 +17,7 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
     [Export(typeof(IAnalyzerAssemblyResolver)), Shared]
     [method: ImportingConstructor]
     [method: Obsolete(MefConstruction.ImportingConstructorMessage, error: true)]
-    internal class RazorAnalyzerAssemblyResolver() : IAnalyzerAssemblyResolver
+    internal sealed class RazorAnalyzerAssemblyResolver() : IAnalyzerAssemblyResolver
     {
         private const string RazorCompilerAssemblyName = "Microsoft.CodeAnalysis.Razor.Compiler";
         private const string RazorUtilsAssemblyName = "Microsoft.AspNetCore.Razor.Utilities.Shared";
@@ -25,11 +25,10 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
 
         private static readonly object s_loaderLock = new();
 
-        public static Assembly? ResolveRazorAssembly(AssemblyName assemblyName, string directoryName)
+        public static Assembly? ResolveRazorAssembly(AssemblyName assemblyName, string rootDirectory)
         {
             if (assemblyName.Name is RazorCompilerAssemblyName or RazorUtilsAssemblyName or ObjectPoolAssemblyName)
             {
-                var assembly = Path.Combine(directoryName, $"{assemblyName.Name}.dll");
                 lock (s_loaderLock)
                 {
                     var compilerContext = AssemblyLoadContext.GetLoadContext(typeof(Compilation).Assembly)!;
@@ -37,6 +36,8 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
                     {
                         return loadedAssembly;
                     }
+
+                    var assembly = Path.Combine(rootDirectory, $"{assemblyName.Name}.dll");
                     return compilerContext.LoadFromAssemblyPath(assembly);
                 }
             }
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -1,14 +1,15 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
+#if NET
 
 using System;
-using System.Collections.Generic;
 using System.Composition;
-using System.Diagnostics;
+using System.IO;
+using System.Linq;
 using System.Reflection;
+using System.Runtime.Loader;
 using Microsoft.CodeAnalysis.Host.Mef;
-using Microsoft.CodeAnalysis.PooledObjects;
 using Microsoft.VisualStudio.Composition;
 
 namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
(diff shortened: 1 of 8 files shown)


src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -18,41 +19,31 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
...
+                lock (s_loaderLock)
                 {
-                    s_assemblyResolver = resolver;
-                    return true;
+                    var compilerContext = AssemblyLoadContext.GetLoadContext(typeof(Compilation).Assembly)!;
+                    if (compilerContext.Assemblies.SingleOrDefault(a => a.GetName().Name == assemblyName.Name) is Assembly loadedAssembly)
+                    {
+                        return loadedAssembly;
+                    }
+                    return compilerContext.LoadFromAssemblyPath(assembly);
```
chsienki, line 40: Note that here we're loading razor into the 'main' compiler ALC. We could create our own ALC to load into, but it will need to be custom code that defers to the main ALC for things like MS.CA.dll, that seemed overkill so I just used the main one (which is already conceptually the 'shared' one anyway).

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -18,41 +19,31 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
     [method: Obsolete(MefConstruction.ImportingConstructorMessage, error: true)]
     internal class RazorAnalyzerAssemblyResolver() : IAnalyzerAssemblyResolver
```
jjonescz, line 20: Sealed?

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -18,41 +19,31 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
...
+            if (assemblyName.Name is RazorCompilerAssemblyName or RazorUtilsAssemblyName or ObjectPoolAssemblyName)
             {
-                if (s_assemblyResolver is null && !s_assembliesRequested.Contains(canaryAssembly))
+                var assembly = Path.Combine(directoryName, $"{assemblyName.Name}.dll");
+                lock (s_loaderLock)
                 {
-                    s_assemblyResolver = resolver;
-                    return true;
+                    var compilerContext = AssemblyLoadContext.GetLoadContext(typeof(Compilation).Assembly)!;
+                    if (compilerContext.Assemblies.SingleOrDefault(a => a.GetName().Name == assemblyName.Name) is Assembly loadedAssembly)
```
jjonescz, line 36: Consider marking the lambda static if possible (by passing assemblyName.Name as an additional argument).

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -18,41 +19,31 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
...
-            Func<AssemblyName, Assembly?>? resolver = null;
-            lock (s_resolverLock)
-            {
-                s_assembliesRequested.Add(assemblyName);
-                resolver = s_assemblyResolver;
-            }
-
-            return resolver?.Invoke(assemblyName);
-        }
+        public Assembly? ResolveAssembly(AssemblyName assemblyName, string directoryName) => ResolveRazorAssembly(assemblyName, directoryName);
```
jjonescz, line 46: Consider using `rootDirectory` for the parameter name to match the implemented interface.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -18,41 +19,31 @@ namespace Microsoft.CodeAnalysis.ExternalAccess.Razor
...
-        /// <param name="canaryAssembly">The name of an assembly that is checked to see if it has been requested to load. Setting the resolver will fail if this has already been requested.</param>
-        /// <returns><c>true</c> when the resolver was successfully set.</returns>
-        internal static bool TrySetAssemblyResolver(Func<AssemblyName, Assembly?> resolver, AssemblyName canaryAssembly)
+        public static Assembly? ResolveRazorAssembly(AssemblyName assemblyName, string directoryName)
         {
-            lock (s_resolverLock)
+            if (assemblyName.Name is RazorCompilerAssemblyName or RazorUtilsAssemblyName or ObjectPoolAssemblyName)
             {
-                if (s_assemblyResolver is null && !s_assembliesRequested.Contains(canaryAssembly))
+                var assembly = Path.Combine(directoryName, $"{assemblyName.Name}.dll");
```
jjonescz, line 32: Consider constructing the path only when we need it (after the `if` below).

src/Workspaces/Remote/ServiceHub/Host/RemoteWorkspaceManager.cs
```diff
@@ -26,7 +26,7 @@ internal class RemoteWorkspaceManager
         MefHostServices.DefaultAssemblies
             .Add(typeof(AspNetCoreEmbeddedLanguageClassifier).Assembly)
             .Add(typeof(BrokeredServiceBase).Assembly)
-            .Add(typeof(RazorAnalyzerAssemblyResolver).Assembly)
+            .Add(typeof(IRazorLanguageServerTarget).Assembly)
```
jjonescz, line 29: Why has this changed?

jjonescz (approved):

src/Workspaces/Remote/ServiceHub/Host/RemoteWorkspaceManager.cs
chsienki: Because it targets both .NET and netstandard. In the netstandard version `RazorAnalyzerAssemblyResolver` no longer exists. It just seemed neater to change the type we're referencing instead of #if-defing the assembly add.

333fred (approved):

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -1,58 +1,50 @@
...
-        /// <param name="canaryAssembly">The name of an assembly that is checked to see if it has been requested to load. Setting the resolver will fail if this has already been requested.</param>
-        /// <returns><c>true</c> when the resolver was successfully set.</returns>
-        internal static bool TrySetAssemblyResolver(Func<AssemblyName, Assembly?> resolver, AssemblyName canaryAssembly)
+        public static Assembly? ResolveRazorAssembly(AssemblyName assemblyName, string rootDirectory)
         {
-            lock (s_resolverLock)
+            if (assemblyName.Name is RazorCompilerAssemblyName or RazorUtilsAssemblyName or ObjectPoolAssemblyName)
             {
-                if (s_assemblyResolver is null && !s_assembliesRequested.Contains(canaryAssembly))
+                lock (s_loaderLock)
```
jaredpar, line 32: Why do we need a lock here?

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -1,58 +1,50 @@
...
                 {
-                    s_assemblyResolver = resolver;
-                    return true;
+                    var compilerContext = AssemblyLoadContext.GetLoadContext(typeof(Compilation).Assembly)!;
+                    if (compilerContext.Assemblies.SingleOrDefault(a => a.GetName().Name == assemblyName.Name) is Assembly loadedAssembly)
+                    {
+                        return loadedAssembly;
+                    }
+
+                    var assembly = Path.Combine(rootDirectory, $"{assemblyName.Name}.dll");
```
jaredpar, line 40: Think that this code only works when the IDE has normalized the path to the razor assemblies. Essentially this logic only works if _every_ path to the razor compiler passed via `/analyzer:` is the same path once we get to this layer. If there is a different path then we will be in a "first one wins" situation (the first winner will be cached in the compiler ALC). 

If that is correct think we should add some explanation here.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -1,58 +1,50 @@
...
     [Export(typeof(IAnalyzerAssemblyResolver)), Shared]
     [method: ImportingConstructor]
     [method: Obsolete(MefConstruction.ImportingConstructorMessage, error: true)]
-    internal class RazorAnalyzerAssemblyResolver() : IAnalyzerAssemblyResolver
+    internal sealed class RazorAnalyzerAssemblyResolver() : IAnalyzerAssemblyResolver
     {
-        private static Func<AssemblyName, Assembly?>? s_assemblyResolver;
+        private const string RazorCompilerAssemblyName = "Microsoft.CodeAnalysis.Razor.Compiler";
+        private const string RazorUtilsAssemblyName = "Microsoft.AspNetCore.Razor.Utilities.Shared";
+        private const string ObjectPoolAssemblyName = "Microsoft.Extensions.ObjectPool";
```
jaredpar, line 24: Where in the code do we keep the list of generators that the IDE remaps from SDK to the VS copy? Wanted to make sure there is agreement on all the assemblies in that map and what we redirect here.

src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
chsienki: @jaredpar I filed #76868 to track addressing this feedback.

PR #76828, opened 2025-01-20
Change helix logging

jaredpar:
This changes helix logging in a few ways:

1. Includes helix.proj in the artifacts directory for tool / developer inspection
2. Adds the expected execution time of a work item into helix.proj so it can be compared to actual execution time.

The other change is more philosophical. Trying to make it so that only the structured / tooling data gets put into the artifacts dir. For any other logging that isn't structural in nature just print directly to the console so we can quickly see it at a glance


jaredpar:

dibarbet (approved):

PR #76835, opened 2025-01-21
Reduce work in ConversionsBase.AddUserDefinedConversionsToExplicitCandidateSet

ToddGrun:
The ArrayBuilder usage in this method is showing up slightly in speedometer CPU profiles (0.1%) but more heavily in local profiles I've taken using a much larger version of the file that Kayle provided than what the speedometer tests use (2.6%)

This method is called quite often and the (isExplicit && isChecked) check seems to not hit very often (specifically, never in my testing of opening/simple editing of the file). The usage of the pool and populating the array builder aren't necessary unless that case is hit, so I've reshuffled the code around a bit to avoid that.

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/UserDefinedExplicitConversions.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/UserDefinedExplicitConversions.cs
@@ -231,8 +231,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 }
                 else
                 {
-                    var originalOperators = ArrayBuilder<MethodSymbol>.GetInstance(operators.Count);
-                    originalOperators.AddRange(operators);
+                    var originalOperatorCount = operators.Count;
 
                     var operators2 = ArrayBuilder<MethodSymbol>.GetInstance();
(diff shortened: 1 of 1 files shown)


src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/UserDefinedExplicitConversions.cs
```diff
@@ -222,27 +222,27 @@ private void AddUserDefinedConversionsToExplicitCandidateSet(
...
-
                 // Add regular operators as well.
                 if (operators.IsEmpty)
                 {
-                    candidates.AddRange(operators2);
+                    declaringType.AddOperators(WellKnownMemberNames.ExplicitConversionName, operators);
                 }
                 else
                 {
+                    var originalOperators = ArrayBuilder<MethodSymbol>.GetInstance(operators.Count);
```
333fred, line 234: Consider, instead of allocating and copying here, just saving `operators.Count` here, and switching to `for` loop below.

ToddGrun: Nice idea!

333fred (approved):

AlekseyTs (approved):
LGTM (commit 2)

Issue #76130, opened 2024-11-28
Test plan for "extensions" feature


jcouv:
Spec for methods/properties: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-14.0/extensions.md
Spec for operators: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-14.0/extension-operators.md
Championed issue: https://github.com/dotnet/csharplang/issues/8697

- [ ] Signature conflict in context of extension type (the outer static class)
  - [ ] Might add more coverage for indexers and properties for completeness sake
- [ ] ctrl-F for `https://github.com/dotnet/roslyn/issues/76130` in source before done as we converted PROTOTYPE comments to reference this when we merged to `main`. 

# Core feature

- [x] Language version
- [x] Parsing
  - [x] `extension` keyword
  - [x] Type parameters
  - [x] Parameter list
    - [x] Empty
    - [x] One parameter
    - [x] More than one parameter
    - [x] Parameter name can be omitted
    - [x] The receiver parameter is not allowed to have modifiers if it is unnamed
    - [x] Named receiver parameter is only allowed to have the refness modifiers and `scoped`.
  - [x] Constraints
  - [x] Body
    - [x] `;`
    - [x] Empty
    - [x] Extension member declarations are syntactically identical to corresponding instance and static members in class and struct declarations
      - [x] Constructor declarations
      - [x] Method declarations
      - [x] Property declarations
      - [x] Operator declarations
      - [x] Event declarations
      - [x] Type declarations
- [x] Static classes as extension containers
  - [x] Extension declaration must be in a top level non-generic static class
  - [x] Can co-exist with other type members
- [x] Scope of names
  - [x] Type parameters are in scope within the body of the extension declaration.
  - [x] Receiver parameter is in scope within the body of the extension declaration.
  - [x] It is an error to refer to the receiver parameter from within a static member, except within a `nameof` expression.
  - [x] It is an error for members to declare type parameters or parameters (as well as local variables and local functions directly within the member body) with the same name as a type parameter or receiver parameter of the extension declaration.
  - [x] It is not an error for the members themselves to have the same name as the type parameters or receiver parameter of the enclosing extension declaration
- [x] Lookup
  - [x] Member names are not directly found in a simple name lookup from within the extension declaration
  - [x] Members of the enclosing type and namespace can be found in a simple name lookup
  - [x] Extension type parameters and extension parameter are found first
- [x] Extension members
[...]

PR #76867, opened 2025-01-22
Extensions parsing

jcouv:


src/Compilers/CSharp/Portable/Generated/CSharp.Generated.g4
```diff
@@ -81,6 +81,7 @@ member_declaration
   | base_type_declaration
   | delegate_declaration
   | enum_member_declaration
+  | extension_container
```
CyrusNajmabadi, line 84: don't love that it's not called _declaration.

src/Compilers/CSharp/Portable/Generated/CSharp.Generated.g4
```diff
@@ -348,6 +349,14 @@ delegate_declaration
   : attribute_list* modifier* 'delegate' type identifier_token type_parameter_list? parameter_list type_parameter_constraint_clause* ';'
   ;
 
+extension_container
+  : attribute_list* modifier* syntax_token type_parameter_list? '(' receiver_parameter ')' type_parameter_constraint_clause* '{' member_declaration* '}'
```
CyrusNajmabadi, line 353: syntax_token?  not `'extension'`?

CyrusNajmabadi: nit: allow an identifier.  people will try to write it, and it will allow for better error recovery.

CyrusNajmabadi: nit: allow a parameter_list, as people will write more, and it will enable better error recovery.

src/Compilers/CSharp/Portable/Generated/CSharp.Generated.g4
```diff
@@ -348,6 +349,14 @@ delegate_declaration
...
+  ;
+
+receiver_parameter
+  : attribute_list* modifier* type identifier_token?
```
CyrusNajmabadi, line 357: we should allow a parameter_syntax.  it will make it much easier ot recover and represent throughout the system.  (similar to how even `a => a` uses a parameter_syntax).  

overfitting syntax is often problematic.  both for normal typing cases, and for expanding on this in the future.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1732,7 +1732,7 @@ private TypeDeclarationSyntax ParseClassOrStructOrInterfaceDeclaration(SyntaxLis
             _termState |= TerminatorState.IsEndOfRecordOrClassOrStructOrInterfaceSignature;
 
             var saveTerm = _termState;
-            _termState |= TerminatorState.IsPossibleAggregateClauseStartOrStop;
+            //_termState |= TerminatorState.IsPossibleAggregateClauseStartOrStop; // TODO2 looking for affected test
```
CyrusNajmabadi, line 1735: mark as prototype?  so this doesn't get merged in?

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
             }
         }
 
+        private MemberDeclarationSyntax ParseExtensionContainer(SyntaxList<AttributeListSyntax> attributes, SyntaxListBuilder modifiers)
```
CyrusNajmabadi, line 1966: make return type strongly typed.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
...
+
+            cancellationToken.ThrowIfCancellationRequested();
+
+            Debug.Assert(this.CurrentToken.ContextualKind == SyntaxKind.ExtensionKeyword);
+            SyntaxToken extensionKeyword = this.EatContextualToken(SyntaxKind.ExtensionKeyword);
+
+            var outerSaveTerm = _termState;
+            _termState |= TerminatorState.IsPossibleAggregateClauseStartOrStop; // TODO2 add test affected
+
+            TypeParameterListSyntax typeParameters = this.ParseTypeParameterList();
```
CyrusNajmabadi, line 1978: check for identifier and recover gracefully.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
...
+            ReceiverParameterSyntax receiverParameter = parseReceiverParameter(out var openParenToken, out var closeParenToken);
```
CyrusNajmabadi, line 1979: parse a parameter list, with a requirement that at least a single parameter is required.  This can be done by updating ParseParameterList to take a parameter stating if a single parameter is required (and passing that to hte hlper it calls).

you can optionally error at parse time or binding time if there is >1 parameter.  I recommend binding time.

CyrusNajmabadi: By requiring a single parameter, you can be sure in later phases that it is safe to access [0]

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
...
+            SyntaxListBuilder<TypeParameterConstraintClauseSyntax> constraints = default;
+            SyntaxListBuilder<MemberDeclarationSyntax> members = default;
+            try
+            {
+                if (this.CurrentToken.ContextualKind == SyntaxKind.WhereKeyword)
+                {
+                    _termState |= TerminatorState.IsEndOfRecordOrClassOrStructOrInterfaceSignature;
+                    constraints = _pool.Allocate<TypeParameterConstraintClauseSyntax>();
+                    this.ParseTypeParameterConstraintClauses(constraints);
+                }
```
CyrusNajmabadi, line 1992: extract this into helper.  otehrwise it can get out of sync with type parameter parsing elsewhere.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
...
+                SyntaxToken? openBrace = this.EatToken(SyntaxKind.OpenBraceToken);
+                SyntaxToken? closeBrace;
+
+                // ignore members if missing open curly
+                if (openBrace.IsMissing)
+                {
+                    parseMembers = false;
+                }
+
+                if (parseMembers)
```
CyrusNajmabadi, line 2005: having a variable to just expand on `if (!openBrace.IsMissing)` seems excessive.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
...
+                    closeBrace = WithAdditionalDiagnostics(closeBrace, this.GetExpectedTokenError(SyntaxKind.CloseBraceToken, this.CurrentToken.Kind));
+                }
+                else
+                {
+                    closeBrace = this.EatToken(SyntaxKind.CloseBraceToken);
+                }
+
+                var modifiersList = (SyntaxList<SyntaxToken>)modifiers.ToList();
+                var membersList = (SyntaxList<MemberDeclarationSyntax>)members;
+                var constraintsList = (SyntaxList<TypeParameterConstraintClauseSyntax>)constraints;
```
CyrusNajmabadi, line 2056: this is confusing.  we have the .ToListAndFree helpers.  Why not use those?  then you don't need a try/finally either.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
...
+                        }
+                        else
+                        {
+                            // Error -- try to sync up with intended reality
+                            this.SkipBadMemberListTokens(ref openBrace, members);
+                        }
+                    }
+                }
+
+                if (openBrace.IsMissing)
```
CyrusNajmabadi, line 2044: here you check the openBrace, instead of parseMembers

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
...
+                {
+                    parseMembers = false;
+                }
+
+                if (parseMembers)
+                {
+                    members = _pool.Allocate<MemberDeclarationSyntax>();
+
+                    while (true)
```
CyrusNajmabadi, line 2009: afaict, this is a copy of ParseClassOrStructOrInterfaceDeclaration.  Seems like we can just expand that to parse extensions, just with some extension-specific code.  Note: as all teh rest support the primary cosntructor parameter list, this all falls out.   The only difference that i can tell so far is simple that in the case of extensions we want at least one parameter (though this could be a binding-time check), and that we do *not* have an identifier.  All of that seems similar to add to ParseClassOrStructOrInterfaceDeclaration vs this copy :)

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1962,6 +1963,151 @@ static TypeDeclarationSyntax constructTypeDeclaration(ContextAwareSyntax syntaxF
...
+
+                SyntaxListBuilder modifiers = _pool.Allocate();
+                this.ParseParameterModifiers(modifiers, isFunctionPointerParameter: false, isLambdaParameter: false);
+
+                TypeSyntax type = this.ParseType(mode: ParseTypeMode.Parameter);
+
+                SyntaxToken? identifier;
+                if (this.CurrentToken.Kind == SyntaxKind.CloseParenToken)
+                {
+                    identifier = null;
```
CyrusNajmabadi, line 2098: ah, you want identifiers to be optional.  mixed feelings. but that feels very doable still :)

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -3193,6 +3346,22 @@ private MemberDeclarationSyntax ParseMemberDeclarationCore(SyntaxKind parentKind
...
 
+        private bool IsExtensionContainerStart()
+        {
+            switch (this.CurrentToken.Kind)
+            {
+                case SyntaxKind.IdentifierToken:
+                    if (CurrentToken.ContextualKind == SyntaxKind.ExtensionKeyword)
+                    {
+                        return IsFeatureEnabled(MessageID.IDS_FeatureExtensions);
+                    }
```
CyrusNajmabadi, line 3357: do you want to look for the more explicit `extension(` ?

src/Compilers/Test/Core/Traits/CompilerFeature.cs
```diff
@@ -43,5 +43,6 @@ public enum CompilerFeature
         RecordStructs,
         RequiredMembers,
         RefLifetime,
+        Extensions,
     }
 }
```
CyrusNajmabadi, line 48: incremental tests that changing `class C(int i)` to `extension(int i)` and vvice versa works would be good.

CyrusNajmabadi:
Overall, this looks quite good to me.  My pref is to share more int eh parser.  either higher level parsing functinos, or lower level shared routines :)

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
jcouv: This PR is draft. Not ready for review. TODO2 comments are flagged by our build and are useful to track things that need to be handled in the PR ;-)

CyrusNajmabadi: oh.  gmail didn't mark this as draft.  sorry!

gtk about TODO2.  i didn't realize that.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
jcouv: We're going to reserve the `extension` keyword (types with that name will be errors and need to be escaped), so I don't see much point to additional subtlety here. Simple seems better.

CyrusNajmabadi: wfm.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -84,7 +84,7 @@ internal enum TerminatorState
             IsEndOfFunctionPointerParameterList = 1 << 23,
             IsEndOfFunctionPointerParameterListErrored = 1 << 24,
             IsEndOfFunctionPointerCallingConvention = 1 << 25,
-            IsEndOfRecordOrClassOrStructOrInterfaceSignature = 1 << 26,
+            IsEndOfRecordOrClassOrStructOrInterfaceOrExtensionSignature = 1 << 26,
```
CyrusNajmabadi, line 87: at this point, we might just say IsEndOfTypeDeclarationSignature :D

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1728,17 +1728,19 @@ private TypeDeclarationSyntax ParseClassOrStructOrInterfaceDeclaration(SyntaxLis
...
+            bool isExtension = keyword.Kind == SyntaxKind.ExtensionKeyword;
             var outerSaveTerm = _termState;
-            _termState |= TerminatorState.IsEndOfRecordOrClassOrStructOrInterfaceSignature;
+            _termState |= TerminatorState.IsEndOfRecordOrClassOrStructOrInterfaceOrExtensionSignature;
 
             var saveTerm = _termState;
             _termState |= TerminatorState.IsPossibleAggregateClauseStartOrStop;
 
-            var name = this.ParseIdentifierToken();
+            SyntaxToken? name = (isExtension && this.CurrentToken.Kind != SyntaxKind.IdentifierToken) ? null : this.ParseIdentifierToken();
```
CyrusNajmabadi, line 1738: nice.  consider stating that we will parse an identifier if present for extensions, but will error later in binding/symbols.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -3193,6 +3225,22 @@ private MemberDeclarationSyntax ParseMemberDeclarationCore(SyntaxKind parentKind
             }
         }
 
+        private bool IsExtensionContainerStart()
+        {
+            switch (this.CurrentToken.Kind)
```
CyrusNajmabadi, line 3230: nit: 
```suggestion
            return this.CurrentToken.ContextualKind == SyntaxKind.ExtensionKeyword && IsFeatureEnabled(MessageID.IDS_FeatureExtensions);
```

:)

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4584,21 +4632,27 @@ private SeparatedSyntaxList<ParameterSyntax> ParseParameterList(
             out SyntaxToken open,
             out SyntaxToken close,
             SyntaxKind openKind,
-            SyntaxKind closeKind)
+            SyntaxKind closeKind,
+            bool allowOptionalIdentifier)
```
CyrusNajmabadi, line 4636: thank you for makingt his non-optional.  i think it's good for the callers to be explicit.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4584,21 +4632,27 @@ private SeparatedSyntaxList<ParameterSyntax> ParseParameterList(
...
             var parameters = ParseCommaSeparatedSyntaxList(
                 ref open,
                 closeKind,
                 static @this => @this.IsPossibleParameter(),
-                static @this => @this.ParseParameter(),
+                parseElement,
                 skipBadParameterListTokens,
                 allowTrailingSeparator: false,
-                requireOneElement: false,
+                requireOneElement: allowOptionalIdentifier,
```
CyrusNajmabadi, line 4655: probably good to doc this.  it's subtle that there's a relationship here.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4696,19 +4750,33 @@ private ParameterSyntax ParseParameter()
...
+                    }
+                    else
+                    {
+                        identifier = null;
+                    }
+                }
+                else
+                {
+                    identifier = this.ParseIdentifierToken();
+                }
```
CyrusNajmabadi, line 4775:
```suggestion
                identifier = this.CurrentToken.Kind == SyntaxKind.IdentifierToken || !allowOptionalIdentifier
                    ? this.ParseIdentifierToken()
                    : null;
```

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1728,17 +1728,19 @@ private TypeDeclarationSyntax ParseClassOrStructOrInterfaceDeclaration(SyntaxLis
...
 
-            var name = this.ParseIdentifierToken();
+            SyntaxToken? name = (isExtension && this.CurrentToken.Kind != SyntaxKind.IdentifierToken) ? null : this.ParseIdentifierToken();
+
             var typeParameters = this.ParseTypeParameterList();
 
-            var paramList = CurrentToken.Kind == SyntaxKind.OpenParenToken
-                ? ParseParenthesizedParameterList() : null;
+            var paramList = CurrentToken.Kind == SyntaxKind.OpenParenToken || isExtension
+                ? ParseParenthesizedParameterList(allowOptionalIdentifier: isExtension) : null;
```
CyrusNajmabadi, line 1743: nit: doc this.  probably something like: In the case of extensions, there must always be a parameter list.  Requiring this here also means that later parts of the compiler do not have to be resilient to getting back a null parameter list from teh declaration.

CyrusNajmabadi (approved):
Signing off on the approach where we integrate with the existing TypeDecl hierarchy.  This will make this much more pleasant for IDE consumpgion, so we're a fan of this.

For future symbolic questions, please keep us in teh loop as we have some thoughts on this as well.

CyrusNajmabadi:
Reminder to probably update reusing parameters in incremental contexts if it has an default identifier.  and a few incremental tests on converting between normal type decls and extensions.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4724,6 +4714,12 @@ private static bool CanReuseParameter(CSharp.Syntax.ParameterSyntax parameter)
                 }
             }
 
+            // cannot reuse parameters without identifiers (found in extension declarations) as they are parsed with different rules
```
CyrusNajmabadi, line 4717:
```suggestion
            // cannot reuse parameters without identifiers (found in extension declarations) as they are parsed with different rules.
            // The reverse is fine though.  Normal parameters (from non extensions) can be re-used into an extension declaration
            // as all normal parameters are legal extension parameters.
```

CyrusNajmabadi: up to you.

AlekseyTs:
@jcouv We probably should create a new feature branch for the new approach.

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -2600,7 +2600,7 @@ A catch() block after a catch (System.Exception e) block can catch non-CLS excep
     <value>Invalid preprocessor expression</value>
   </data>
   <data name="ERR_InvalidMemberDecl" xml:space="preserve">
-    <value>Invalid token '{0}' in class, record, struct, or interface member declaration</value>
+    <value>Invalid token '{0}' in class, record, struct, interface, or extension member declaration</value>
```
AlekseyTs, line 2603: >class, record, struct, interface, or extension member [](http://example.com/codeflow?start=34&length=53)

Would it make sense to simply say "a member" instead of "class, record, struct, interface, or extension member"?

src/Compilers/CSharp/Portable/Generated/CSharpSyntaxGenerator/CSharpSyntaxGenerator.SourceGenerator/Syntax.xml.Main.Generated.cs
```diff
@@ -5146,6 +5147,44 @@
     public static EnumMemberDeclarationSyntax EnumMemberDeclaration(string identifier)
         => SyntaxFactory.EnumMemberDeclaration(default, default(SyntaxTokenList), SyntaxFactory.Identifier(identifier), default);
 
+    /// <summary>Creates a new ExtensionDeclarationSyntax instance.</summary>
+    public static ExtensionDeclarationSyntax ExtensionDeclaration(SyntaxList<AttributeListSyntax> attributeLists, SyntaxTokenList modifiers, SyntaxToken keyword, SyntaxToken identifier, TypeParameterListSyntax? typeParameterList, ParameterListSyntax? parameterList, BaseListSyntax? baseList, SyntaxList<TypeParameterConstraintClauseSyntax> constraintClauses, SyntaxToken openBraceToken, SyntaxList<MemberDeclarationSyntax> members, SyntaxToken closeBraceToken, SyntaxToken semicolonToken)
```
AlekseyTs, line 5151: >public static ExtensionDeclarationSyntax ExtensionDeclaration [](http://example.com/codeflow?start=4&length=61)

Is this supposed to be added to the PublicAPI.Unshipped.txt?

AlekseyTs, line 5151: >SyntaxToken identifier [](http://example.com/codeflow?start=162&length=22)

This parameter is unexpected

AlekseyTs, line 5151: >BaseListSyntax? baseList [](http://example.com/codeflow?start=266&length=24)

This parameter is unexpected

src/Compilers/CSharp/Portable/Generated/CSharpSyntaxGenerator/CSharpSyntaxGenerator.SourceGenerator/Syntax.xml.Main.Generated.cs
```diff
@@ -5146,6 +5147,44 @@
...
+    }
+
+    /// <summary>Creates a new ExtensionDeclarationSyntax instance.</summary>
+    public static ExtensionDeclarationSyntax ExtensionDeclaration(SyntaxList<AttributeListSyntax> attributeLists, SyntaxTokenList modifiers, SyntaxToken keyword, SyntaxToken identifier, TypeParameterListSyntax? typeParameterList, ParameterListSyntax? parameterList, BaseListSyntax? baseList, SyntaxList<TypeParameterConstraintClauseSyntax> constraintClauses, SyntaxList<MemberDeclarationSyntax> members)
```
AlekseyTs, line 5181: >SyntaxToken identifier [](http://example.com/codeflow?start=162&length=22)

This parameter is unexpected

AlekseyTs, line 5181: >BaseListSyntax? baseList [](http://example.com/codeflow?start=266&length=24)

This parameter is unexpected

jcouv:
> @jcouv We probably should create a new feature branch for the new approach.

Yes. I've created `features/extensions`, but forgot to target this PR to that branch. Fixed

src/Compilers/CSharp/Portable/Generated/CSharpSyntaxGenerator/CSharpSyntaxGenerator.SourceGenerator/Syntax.xml.Syntax.Generated.cs
```diff
@@ -11312,6 +11312,178 @@
...
 }
 
+/// <summary>Extension container syntax.</summary>
+/// <remarks>
+/// <para>This node is associated with the following syntax kinds:</para>
+/// <list type="bullet">
+/// <item><description><see cref="SyntaxKind.ExtensionDeclaration"/></description></item>
+/// </list>
+/// </remarks>
+public sealed partial class ExtensionDeclarationSyntax : TypeDeclarationSyntax
```
AlekseyTs, line 11322: >public sealed partial class ExtensionDeclarationSyntax [](http://example.com/codeflow?start=0&length=54)

Is this supposed to be added to the PublicAPI.Unshipped.txt?

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4584,21 +4623,26 @@ private SeparatedSyntaxList<ParameterSyntax> ParseParameterList(
...
             var parameters = ParseCommaSeparatedSyntaxList(
                 ref open,
                 closeKind,
                 static @this => @this.IsPossibleParameter(),
-                static @this => @this.ParseParameter(),
+                parseElement,
                 skipBadParameterListTokens,
                 allowTrailingSeparator: false,
-                requireOneElement: false,
+                requireOneElement: forExtension, // For extension declarations, we require one receiver parameter
```
AlekseyTs, line 4645: >requireOneElement: forExtension, // For extension declarations, we require one receiver parameter [](http://example.com/codeflow?start=16&length=97)

It looks like the meaning of this parameter is to ensure that the syntax list is not empty, i.e. at least one item is expected. That doesn't mean that the helper won't accept more that one item. The comment, however, might be interpreted as though we instruct the helper to accept exactly one item (no more, no less).

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4696,19 +4748,22 @@ private ParameterSyntax ParseParameter()
             }
 
             var type = this.ParseType(mode: ParseTypeMode.Parameter);
-            SyntaxToken identifier;
 
+            SyntaxToken? identifier;
             if (this.CurrentToken.Kind == SyntaxKind.IdentifierToken && IsCurrentTokenWhereOfConstraintClause())
             {
                 identifier = this.AddError(CreateMissingIdentifierToken(), ErrorCode.ERR_IdentifierExpected);
```
AlekseyTs, line 4755: >identifier = this.AddError(CreateMissingIdentifierToken(), ErrorCode.ERR_IdentifierExpected); [](http://example.com/codeflow?start=16&length=93)

It feels like this error recovery might not be appropriate when identifier is optional. Specifically, we probably do not want to create a missing identifier.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
AlekseyTs, line 4645: >requireOneElement: forExtension [](http://example.com/codeflow?start=16&length=31)

It feels like it might be good to suppress any parsing diagnostics from syntax errors in the second and any following parameter in the list. Not sure how easy to do this, given that diagnostics is attached to nodes. Consider making a note to follow up later.

src/Compilers/CSharp/Portable/Syntax/SyntaxKindFacts.cs
```diff
@@ -244,6 +244,7 @@ public static bool IsAnyToken(SyntaxKind kind)
                 case SyntaxKind.UnderscoreToken:
                 case SyntaxKind.MultiLineRawStringLiteralToken:
                 case SyntaxKind.SingleLineRawStringLiteralToken:
+                case SyntaxKind.ExtensionKeyword:
```
AlekseyTs, line 247: >case SyntaxKind.ExtensionKeyword [](http://example.com/codeflow?start=16&length=32)

Is this entry necessary? It looks like there is no entry for ```AllowsKeyword```.

AlekseyTs:
>             TestOptions.RegularNext);

We also test language version behavior with RegularPreview. 

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:134 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>             TestOptions.RegularNext);

I think tests that are not focused on language version should target RegularPreview.

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:200 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         TestOptions.RegularNext);

Test with RegularPreview as well? 

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:704 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         TestOptions.RegularNext);

Test with RegularPreview as well?

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:744 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>     extension(Type, Type) { }

Consider covering more error scenarios:
- missing comma
- some other syntax error in the second parameter
- missing `)` (with and without the second parameter)
- etc.

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:931 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         const int i = 0;

Do we expect a semantic error?

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1713 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         fixed int field[10];

Do we expect a semantic error?

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1783 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         event EventHandler eventField;

Do we expect a semantic error?

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1857 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         class D { }

Do we expect a semantic error?

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1921 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         var comp = CreateCompilation(src, parseOptions: TestOptions.RegularNext);

Consider testing with RegularPreview as well.

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:3261 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>             TestOptions.RegularNext,

Consider testing with RegularPreview as well.

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:3268 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>             TestOptions.RegularNext);

Consider testing with RegularPreview as well.

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:3311 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         var comp = CreateCompilation(src, parseOptions: TestOptions.RegularNext);

Consider testing with RegularPreview as well.

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:3397 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         UsingTree(src, TestOptions.RegularNext);

Consider testing with RegularPreview as well.

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:3400 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>             TestOptions.RegularNext);

Consider testing with RegularPreview as well.

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:3441 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>                             M(SyntaxKind.IdentifierToken);

The missing identifier feels unexpected

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:3676 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
Done with review pass (commit 9)

jcouv:
>         const int i = 0;

I think this may end up allowed, since emitted as static field.

---
In reply to: [2611226090](https://github.com/dotnet/roslyn/pull/76867#issuecomment-2611226090) [](http://example.com/codeflow?ancestors=2611226090)

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1713 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         const int i = 0;

> I think this may end up allowed, since emitted as static field.

The spec doesn't allow fields and we didn't even think what would it mean for implementation. Also, is this going to be an extension field?

---
In reply to: [2611341439](https://github.com/dotnet/roslyn/pull/76867#issuecomment-2611341439) [](http://example.com/codeflow?ancestors=2611341439,2611226090)

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1713 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         const int i = 0;

I suggest sticking to the spec for now

---
In reply to: [2611344424](https://github.com/dotnet/roslyn/pull/76867#issuecomment-2611344424) [](http://example.com/codeflow?ancestors=2611344424,2611341439,2611226090)

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1713 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

jcouv:
>         class D { }

This may be allowed (TBD)

---
In reply to: [2611227756](https://github.com/dotnet/roslyn/pull/76867#issuecomment-2611227756) [](http://example.com/codeflow?ancestors=2611227756)

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1921 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

AlekseyTs:
>         class D { }

> This may be allowed (TBD)

I suggest sticking to the spec for now

---
In reply to: [2611345773](https://github.com/dotnet/roslyn/pull/76867#issuecomment-2611345773) [](http://example.com/codeflow?ancestors=2611345773,2611227756)

---
Refers to: src/Compilers/CSharp/Test/Syntax/Parsing/ExtensionsParsingTests.cs:1921 in 51686c8. [](commit_id = 51686c8549ac1b11b528b21efd9ce8f171128524, deletion_comment = False)

src/Compilers/CSharp/Portable/Syntax/Syntax.xml
```diff
@@ -3617,6 +3617,30 @@
...
   </Node>
+  <Node Name="ExtensionDeclarationSyntax" Base="TypeDeclarationSyntax">
+    <TypeComment>
+      <summary>Extension container syntax.</summary>
+    </TypeComment>
+    <Kind Name="ExtensionDeclaration" />
+    <Field Name="AttributeLists" Type="SyntaxList&lt;AttributeListSyntax&gt;" Override="true"/>
+    <Field Name="Modifiers" Type="SyntaxList&lt;SyntaxToken&gt;" Override="true"/>
+    <Field Name="Keyword" Type="SyntaxToken" Override="true">
+      <ContextualKind Name="ExtensionKeyword"/>
```
CyrusNajmabadi, line 3628: i'm actually slightly surprised by this.  i would have thought this would be `<kind`

src/Compilers/CSharp/Portable/Syntax/Syntax.xml
```diff
@@ -3617,6 +3617,30 @@
...
+    <Field Name="ParameterList" Type="ParameterListSyntax" Optional="true" Override="true" />
+    <Field Name="ConstraintClauses" Type="SyntaxList&lt;TypeParameterConstraintClauseSyntax&gt;" Override="true"/>
+    <Field Name="OpenBraceToken" Type="SyntaxToken" Override="true" Optional="true">
+      <Kind Name="OpenBraceToken"/>
+    </Field>
+    <Field Name="Members" Type="SyntaxList&lt;MemberDeclarationSyntax&gt;" Override="true"/>
+    <Field Name="CloseBraceToken" Type="SyntaxToken" Override="true" Optional="true">
+      <Kind Name="CloseBraceToken"/>
+    </Field>
+    <Field Name="SemicolonToken" Type="SyntaxToken" Optional="true" Override="true">
```
jjonescz, line 3640: Why is the semicolon token allowed here? It's not in the speclet.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -1707,18 +1707,18 @@ private MemberDeclarationSyntax ParseTypeDeclaration(SyntaxList<AttributeListSyn
...
+                    Debug.Assert(CurrentToken.ContextualKind is SyntaxKind.RecordKeyword or SyntaxKind.ExtensionKeyword);
+                    return ParseClassOrStructOrInterfaceOrExtensionDeclaration(attributes, modifiers);
 
                 default:
                     throw ExceptionUtilities.UnexpectedValue(this.CurrentToken.Kind);
             }
         }
 
-        private TypeDeclarationSyntax ParseClassOrStructOrInterfaceDeclaration(SyntaxList<AttributeListSyntax> attributes, SyntaxListBuilder modifiers)
+        private TypeDeclarationSyntax ParseClassOrStructOrInterfaceOrExtensionDeclaration(SyntaxList<AttributeListSyntax> attributes, SyntaxListBuilder modifiers)
```
jjonescz, line 1718: Looks like this could be renamed to ParseTypeDeclaration as well

src/Compilers/CSharp/Portable/Syntax/Syntax.xml
jcouv: I kept it for consistency and simplicity with the other declaration kinds. Will [update](https://github.com/dotnet/csharplang/pull/9090) the speclet.

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
jcouv: That name would conflict. Shortened as ParseMainTypeDeclaration

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4670,12 +4726,20 @@ private static bool CanReuseParameter(CSharp.Syntax.ParameterSyntax parameter)
                 }
             }
 
+            // cannot reuse parameters without identifiers (found in extension declarations) as they are parsed with different rules.
+            // The reverse is fine though.  Normal parameters (from non extensions) can be re-used into an extension declaration
+            // as all normal parameters are legal extension parameters.
+            if (parameter.Identifier.Kind() == SyntaxKind.None)
+            {
+                return false;
```
AlekseyTs, line 4734: >return false [](http://example.com/codeflow?start=16&length=12)

It looks like in every place where the enclosing method is called, we know whether missing identifier is allowed. We can pass this information and take advantage of that here.

AlekseyTs:
It looks like there are legitimate test failures in IOperation leg.

AlekseyTs:
Done with review pass (commit 13).

src/Compilers/CSharp/Portable/Syntax/ExtensionDeclarationSyntax.cs
```diff
@@ -0,0 +1,22 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.CodeAnalysis.CSharp.Syntax
+{
+    public partial class ExtensionDeclarationSyntax
+    {
+        public override SyntaxToken Identifier => default;
+
+        internal override BaseTypeDeclarationSyntax WithIdentifierCore(SyntaxToken identifier)
+            => throw new System.NotImplementedException();
```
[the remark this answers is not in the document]
AlekseyTs: I am not sure why this thread was marked as resolved.

src/Compilers/CSharp/Portable/Syntax/ExtensionDeclarationSyntax.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.CodeAnalysis.CSharp.Syntax
+{
+    public partial class ExtensionDeclarationSyntax
+    {
+        public override SyntaxToken Identifier
+            => throw new System.NotImplementedException();
```
AlekseyTs, line 10: >throw new System.NotImplementedException(); [](http://example.com/codeflow?start=15&length=43)

I think returning `default` was the right thing to do. A call through base class should be allowed.

AlekseyTs: It is, however, Ok to throw when one tries to set identifier.

AlekseyTs:
Done with review pass (commit 17)

src/Compilers/CSharp/Portable/Parser/LanguageParser.cs
```diff
@@ -4670,14 +4726,22 @@ private static bool CanReuseParameter(CSharp.Syntax.ParameterSyntax parameter)
                 }
             }
 
+            // We can only reuse parameters without identifiers (found in extension declarations) in context that allow optional identifiers.
+            // The reverse is fine though.  Normal parameters (from non extensions) can be re-used into an extension declaration
+            // as all normal parameters are legal extension parameters.
+            if (!allowOptionalIdentifier && parameter.Identifier.Kind() == SyntaxKind.None)
```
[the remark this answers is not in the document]
jcouv: Added a test (last two commits show the before/after to show the impact). Had to fix a helper API in incremental tests which reveals some bad test baselines.
I'm considering cherry-picking that portion back to main branch. Let me know what you think.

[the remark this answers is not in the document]
AlekseyTs: > Let me know what you think.

Sounds good

AlekseyTs (approved):
LGTM (commit 22)

jcouv:
@jjonescz for second review. Thanks

jjonescz (approved):

Issue #76870, opened 2025-01-22
A follow up is needed for a process of merging backing fields for a partial property


AlekseyTs:
It looks like initially PartialDefinitionPart and PartialImplementationPart are created with their own unique backing field symbols. After the parts are merged, ```ContainingType.GetMembers()``` contains ```PartialDefinitionPart.BackingField``` and  ```PartialImplementationPart.BackingField``` is adjusted to point to that field symbolas well. The field symbol originally created for PartialImplementationPart is not among ```ContainingType.GetMembers()```, but if the implementation part has an initializer, corresponding bound field initializer refers to that field, rather than to the ```PartialDefinitionPart.BackingField```. Also, it is very likely that ```PartialDefinitionPart.BackingField.HasInitializer``` is set to false. Emitted code for a simple scenario looks correct, likely due to the fact that the fields share the same name. However, the internal compiler data structures look inconsistent, which potentially might cause observable problems in some scenarios.

AlekseyTs:
A workaround for the inconsistency is added to https://github.com/dotnet/roslyn/pull/76871,  otherwise several tests in  ```PartialPropertiesTests``` and ```FieldKeywordTests``` were asserting in ```AssertMemberExposure```.

Issue #76842, opened 2025-01-21
Partial indexer with an IndexerName attribute causes a crash in compiler


AlekseyTs:
https://sharplab.io/#v2:EYLgtghglgdgNAFxFANnAJiA1AHwAIAMABHgIwB0ASgK4wJRgCm5AwgPZgAOqjATgMp8AblADGjAM4BuALAAoeZwi96EFEQkJe1UQiL9S8gN7yiZkgGYiSlVDVFYehAAsoEgNqOiADwC6RIwBzRgQAXgA+bykAX3lTc3cASRh0Rm8+ADkIJgAKACIAWQBPLKY8gEpfeLM8KxtVdS8XN086H38gkJj5WLkgA= :
```
using System.Runtime.CompilerServices;

partial struct S1
{
    public partial int this[int x] {get=>x;}

    [IndexerName("MyName")]
    public partial int this[int x] {get;}
}
```

Or https://sharplab.io/#v2:EYLgtghglgdgNAFxFANnAJiA1AHwAIAMABHgIwB0ASgK4wJRgCm5AwgPZgAOqjATgMp8AblADGjAM4BuALAAoeZwi96EFEQkJe1UQiL9S8gN7yiZogG0AkjHSMAHnwByEJgAoARAFkAni6YeAJQAuqbmeADMREoqUGpEsHoIABZQEhaJRPbBREYA5owIALwAfPZSAL7yYWaR0cqq6pkpaRl0WTn5hZXyVXJAA=== :
```
using System.Runtime.CompilerServices;

partial struct S1
{
    [IndexerName("MyName")]
    public partial int this[int x] {get=>x;}

    public partial int this[int x] {get;}
}
```

Observed:
```
System.ArgumentException: adding a duplicate (Parameter 'key')
   at Roslyn.Utilities.ConcurrentDictionaryExtensions.Add[K,V](ConcurrentDictionary`2 dict, K key, V value) in /_/src/Compilers/Core/Portable/InternalUtilities/ConcurrentDictionaryExtensions.cs:line 24
   at Microsoft.CodeAnalysis.CSharp.MethodCompiler.CompileMethod(MethodSymbol methodSymbol, Int32 methodOrdinal, ProcessedFieldInitializers& processedInitializers, SynthesizedSubmissionFields previousSubmissionFields, TypeCompilationState compilationState) in /_/src/Compilers/CSharp/Portable/Compiler/MethodCompiler.cs:line 1343
[...]

PR #76871, opened 2025-01-22
Move merging of partial members to an earlier phase of members construction.

### LINKED ISSUE #75002: Ambiguity errors reported from `SemanticModel.GetDiagnostics()` for calls to partial members on type with primary constructor
**Version Used**: 2ff1a4768213b0e6d8d5762fae6cebd2af6b04bd

**Steps to Reproduce**:

Run the following unit test.

```csharp
[Fact]
public void PartialMembers()
{
    var source1 = """
        C c = null;
        c.M();
        _ = c.P;
        """;
    var source2 = """
        partial class C(int p)
        {
            public partial void M() { }
            public partial void M();
            public partial object P { get; }
            public partial object P { get => null; }
        }
        """;
    var comp = CreateCompilation([source1, source2]);
    var tree = comp.SyntaxTrees[0];
    model.GetDiagnostics().Verify(
        // (2,3): error CS0121: The call is ambiguous between the following methods or properties: 'C.M()' and 'C.M()'
        // c.M();
        Diagnostic(ErrorCode.ERR_AmbigCall, "M").WithArguments("C.M()", "C.M()").WithLocation(2, 3),
        // (3,7): error CS0229: Ambiguity between 'C.P' and 'C.P'
        // _ = c.P;
        Diagnostic(ErrorCode.ERR_AmbigMember, "P").WithArguments("C.P", "C.P").WithLocation(3, 7));
}
```

**Expected Behavior**:
No errors from the call to `model.GetDiagnostics()`.

**Actual Behavior**:
Ambiguity errors included above.

### LINKED ISSUE #76651: Incorrect "CS0229 Ambiguity between" for partial prop with primary ctor, at design time only
**Version Used**: 17.13.0 Preview 2.1

Thanks to @RoccoZero who reported that this issue has been happening for a few months.

1. Building succeeds, though the error list and code editor show the errors.
1. The repro requires separate source files.
1. The repro disappears if the primary ctor is deleted.

###### .csproj

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net9.0</TargetFramework>
  </PropertyGroup>

</Project>
```

###### File 1

```cs
partial class PartialClass
{
    public partial string Prop => null;
}

partial class PartialClass(int p)
{
    public partial string Prop { get; }
}
```

###### File 2

```cs
class Repro
{
    void M(PartialClass c)
    {
        // ❌ CS0229 Ambiguity between 'PartialClass.Prop' and 'PartialClass.Prop'
        _ = c.Prop;
        _ = c.Prop; // Yep, error still here.
        _ = new PartialClass(0).Prop; // No error here!
        _ = c.Prop; // ERROR DISAPPEARS
        _ = c.Prop; // IT'S STILL GONE
    }
}
```

Image

Image

AlekseyTs:
Fixes #76651.
Fixes #75002.
Related to #76842.
Related to #76870.

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
@@ -2896,7 +2896,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
 
         private sealed class DeclaredMembersAndInitializersBuilder
         {
-            public ArrayBuilder<Symbol> NonTypeMembersWithPartialImplementations = ArrayBuilder<Symbol>.GetInstance();
+            public ArrayBuilder<Symbol> NonTypeMembers = ArrayBuilder<Symbol>.GetInstance();
             public readonly ArrayBuilder<ArrayBuilder<FieldOrPropertyInitializer>> StaticInitializers = ArrayBuilder<ArrayBuilder<FieldOrPropertyInitializer>>.GetInstance();
             public readonly ArrayBuilder<ArrayBuilder<FieldOrPropertyInitializer>> InstanceInitializers = ArrayBuilder<ArrayBuilder<FieldOrPropertyInitializer>>.GetInstance();
             public bool HaveIndexers;
@@ -2909,7 +2909,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             public DeclaredMembersAndInitializers ToReadOnlyAndFree(CSharpCompilation compilation)
             {
                 return new DeclaredMembersAndInitializers(
-                    NonTypeMembersWithPartialImplementations.ToImmutableAndFree(),
+                    NonTypeMembers.ToImmutableAndFree(),
                     MembersAndInitializersBuilder.ToReadOnlyAndFree(StaticInitializers),
                     MembersAndInitializersBuilder.ToReadOnlyAndFree(InstanceInitializers),
                     HaveIndexers,
@@ -2939,7 +2939,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
 
             public void Free()
             {
-                NonTypeMembersWithPartialImplementations.Free();
+                NonTypeMembers.Free();
 
                 foreach (var group in StaticInitializers)
                 {
@@ -3281,11 +3281,11 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                 {
                     case TypeKind.Struct:
                         CheckForStructBadInitializers(builder, diagnostics);
-                        CheckForStructDefaultConstructors(builder.NonTypeMembersWithPartialImplementations, isEnum: false, diagnostics: diagnostics);
+                        CheckForStructDefaultConstructors(builder.NonTypeMembers, isEnum: false, diagnostics: diagnostics);
                         break;
 
                     case TypeKind.Enum:
-                        CheckForStructDefaultConstructors(builder.NonTypeMembersWithPartialImplementations, isEnum: true, diagnostics: diagnostics);
+                        CheckForStructDefaultConstructors(builder.NonTypeMembers, isEnum: true, diagnostics: diagnostics);
                         break;
 
                     case TypeKind.Class:
@@ -3596,7 +3596,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                         break;
 
                     case SyntaxKind.DelegateDeclaration:
-                        SourceDelegateMethodSymbol.AddDelegateMembers(this, builder.NonTypeMembersWithPartialImplementations, (DelegateDeclarationSyntax)syntax, diagnostics);
+                        SourceDelegateMethodSymbol.AddDelegateMembers(this, builder.NonTypeMembers, (DelegateDeclarationSyntax)syntax, diagnostics);
                         break;
 
                     case SyntaxKind.NamespaceDeclaration:
@@ -4083,7 +4083,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                     symbol = SourceEnumConstantSymbol.CreateImplicitValuedConstant(this, member, otherSymbol, otherSymbolOffset, diagnostics);
                 }
 
-                result.NonTypeMembersWithPartialImplementations.Add(symbol);
+                result.NonTypeMembers.Add(symbol);
 
                 if (valueOpt != null || otherSymbol is null)
                 {
@@ -4230,7 +4230,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
                 }
             }
 
-            if (hasInitializers && !builder.NonTypeMembersWithPartialImplementations.Any(member => member is MethodSymbol { MethodKind: MethodKind.Constructor }))
+            if (hasInitializers && !builder.NonTypeMembers.Any(member => member is MethodSymbol { MethodKind: MethodKind.Constructor }))
             {
                 diagnostics.Add(ErrorCode.ERR_StructHasInitializersAndNoDeclaredConstructor, GetFirstLocation());
             }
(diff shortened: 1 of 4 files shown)


AlekseyTs:
@RikkiGibson, @cston, @dotnet/roslyn-compiler Please review

AlekseyTs:
@jjonescz, FYI

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3691,9 +3755,31 @@ private void MergePartialMembers(
...
+            static void fixupNotMergedPartialProperty(Symbol symbol)
+            {
+                Debug.Assert(symbol.IsPartialMember());
+                if (symbol is SourcePropertySymbol { OtherPartOfPartial: null } property)
+                {
+                    property.SetMergedBackingField(property.DeclaredBackingField);
+                }
+            }
+
+            void mergePartialMethods(ArrayBuilder<Symbol> nonTypeMembers, SourceOrdinaryMethodSymbol currentMethod, SourceOrdinaryMethodSymbol prevMethod, BindingDiagnosticBag diagnostics)
```
jcouv, line 3782: >mergePartialMethods [](http://example.com/codeflow?start=17&length=19)

static?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3709,12 +3795,11 @@ void mergePartialMethods(ref Dictionary<ReadOnlyMemory<char>, ImmutableArray<Sym
...
                 else
                 {
-                    DuplicateMembersByNameIfCached(ref membersByName);
-                    membersByName[name] = FixPartialMethod(membersByName[name], prevMethod, currentMethod);
+                    FixPartialMethod(nonTypeMembers, prevMethod, currentMethod);
                 }
             }
 
-            void mergePartialProperties(ref Dictionary<ReadOnlyMemory<char>, ImmutableArray<Symbol>> membersByName, ReadOnlyMemory<char> name, SourcePropertySymbol currentProperty, SourcePropertySymbol prevProperty, BindingDiagnosticBag diagnostics)
+            void mergePartialProperties(ArrayBuilder<Symbol> nonTypeMembers, SourcePropertySymbol currentProperty, SourcePropertySymbol prevProperty, BindingDiagnosticBag diagnostics)
```
jcouv, line 3802: >mergePartialProperties [](http://example.com/codeflow?start=17&length=22)

static?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
AlekseyTs: > static?

I do not think I changed anything that affects ability to place the modifier on the method. Therefore, I won't try adding it, unless there is other feedback to take care.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
AlekseyTs: > static?

Same response

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
AlekseyTs: Scratch that, it looks like I removed ```DuplicateMembersByNameIfCached``` call

jcouv: It looks like the containing method `MergePartialMembers` could be marked static too

jcouv (approved):
LGTM Thanks (iteration 2)

AlekseyTs:
@RikkiGibson, @cston, @dotnet/roslyn-compiler For the second review

AlekseyTs:
@RikkiGibson, @cston, @dotnet/roslyn-compiler For the second review

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3816,27 +3890,28 @@ private static void FixPartialProperty(ref Dictionary<ReadOnlyMemory<char>, Immu
...
 
             SourcePropertySymbol.InitializePartialPropertyParts(definition, implementation);
 
             // a partial property is represented in the member list by its definition part:
-            membersByName[name] = Remove(membersByName[name], implementation);
+            Remove(nonTypeMembers, implementation);
         }
 
-        private static ImmutableArray<Symbol> Remove(ImmutableArray<Symbol> symbols, Symbol symbol)
+        private static void Remove(ArrayBuilder<Symbol> symbols, Symbol symbol)
```
cston, line 3902: >ArrayBuilder<Symbol> symbols [](http://example.com/codeflow?start=35&length=28)

It looks like we call `Remove` once for each partial method or property that has both parts. Does the `symbols` array contain all non-type members? If that's the case, the number of `ReferenceEquals()` calls across all `Remove` calls for the containing type will be roughly N*M where N is the number of non-type members and M is the number of partial members. Are there cases where that might be an issue?

cston (approved):

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
AlekseyTs: > Are there cases where that might be an issue?

It is hard to tell for sure. I assume that ```ReferenceEquals``` call is pretty fast, since it is optimized away and probably roughly close to a pointer comparison. Also, M is likely not very big and no more than N/2. Also, on average, the match will be found before we check all N items, and each removal decreases N by one.

AlekseyTs: The `M` that I am using is not the number of partial members. It is the number of partial members that we are going to remove, which is no more than the number of partial members divided by two.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
```diff
@@ -74,7 +75,7 @@ private enum Flags : ushort
 
 #nullable enable
         private SynthesizedBackingFieldSymbol? _lazyDeclaredBackingField;
-        private SynthesizedBackingFieldSymbol? _lazyMergedBackingField;
+        private StrongBox<SynthesizedBackingFieldSymbol?>? _lazyMergedBackingField;
```
RikkiGibson, line 78: I didn't understand what is accomplished by using a StrongBox here, could you please explain?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -2948,7 +2957,7 @@ public void Free()
 
         protected sealed class DeclaredMembersAndInitializers
         {
-            public readonly ImmutableArray<Symbol> NonTypeMembers;
+            public readonly ImmutableArray<Symbol> NonTypeMembersWithPartialImplementations;
```
RikkiGibson, line 2960: It might be good to also rename the corresponding field of `DeclaredMembersAndInitializersBuilder`.

src/Compilers/CSharp/Portable/Symbols/Source/SourcePropertySymbolBase.cs
AlekseyTs: > I didn't understand what is accomplished by using a StrongBox here, could you please explain?

`null` is a valid value for the backing field. Therefore, `null` couldn't be used as a sentinel value (a value representing a not initialized state). But it was used this way in ```BackingField.get```. Potentially causing us to get into the `if` over and over again. Also there was no guarantee that we wouldn't get into the `if` for a partial case. StrongBox allows us to know for sure whether the value is initialized, `null` - not initialized, `not null` initialized.

RikkiGibson: Thank you

RikkiGibson (approved):

PR #76872, opened 2025-01-22
Update runtime async spec

333fred:
Updates from the runtime side: https://github.com/dotnet/runtime/pull/110420.

### DIFF
--- a/docs/compilers/CSharp/Runtime Async Design.md
+++ b/docs/compilers/CSharp/Runtime Async Design.md
@@ -54,6 +54,8 @@ public static class RuntimeHelpers
 We presume the following `MethodImplOptions` bit is present. This is used to indicate to the JIT that it should generate an async state machine for the method. This bit is not allowed to be used manually on any method; it is added by the compiler
 to an `async` method.
 
+TODO: We may want to block directly calling `MethodImplOptions.Async` methods with non-`Task`/`ValueTask` return types.
+
 ```cs
 namespace System.Runtime.CompilerServices;
 
--- a/docs/compilers/CSharp/Runtime Async Design.md
+++ b/docs/compilers/CSharp/Runtime Async Design.md
@@ -697,7 +697,7 @@ _ = {
 }
 ```
 
-#### Implementor of INotifyCompletion
+##### Implementor of INotifyCompletion
 
 ```cs
 var c = new C();
--- a/docs/compilers/CSharp/Runtime Async Design.md
+++ b/docs/compilers/CSharp/Runtime Async Design.md
@@ -267,9 +267,6 @@ System.Runtime.CompilerServices.RuntimeHelpers.Await(local);
 
 #### Await local of concrete type `Task<T>`
 
-This strategy will also be used for `Task`-like types that are not `Task`, `ValueTask`, `Task<T>`, or `ValueTask<T>`, both in value form, and in direct method call form. We will use either `AwaitAwaiterFromRuntimeAsync` or
-`UnsafeAwaitAwaiterFromRuntimeAsync`, depending on the interface implemented by the custom awaitable.
-
 ```cs
 var local = M();
 var i = await local;
--- a/docs/compilers/CSharp/Runtime Async Design.md
+++ b/docs/compilers/CSharp/Runtime Async Design.md
@@ -27,20 +27,32 @@ We use the following helper APIs to indicate suspension points to the runtime, i
 ```cs
 namespace System.Runtime.CompilerServices;
 
-// These methods are used to await things that cannot use runtime async signature form
 // TODO: Clarify which of these should be preferred? Should we always emit the `Unsafe` version when awaiting something that implements `ICriticalNotifyCompletion`?
 namespace System.Runtime.CompilerServices;
 
 public static class RuntimeHelpers
 {
-    [RuntimeAsyncMethod]
+    // These methods are used to await things that cannot use the Await helpers below
+    [MethodImpl(MethodImplOptions.Async)]
     public static Task AwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter) where TAwaiter : INotifyCompletion;
-    [RuntimeAsyncMethod]
+    [MethodImpl(MethodImplOptions.Async)]
     public static Task UnsafeAwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter) where TAwaiter : ICriticalNotifyCompletion;
+
+    // These methods are used to directly await method calls
+    // TODO: Do we only use these when directly awaiting a method call, or do we always use these when a `Task` is awaited, even if that task is a local?
+    [MethodImpl(MethodImplOptions.Async)]
+    public static void Await(Task task);
+    [MethodImpl(MethodImplOptions.Async)]
+    public static T Await<T>(Task<T> task);
+    [MethodImpl(MethodImplOptions.Async)]
+    public static void Await(ValueTask task);
+    [MethodImpl(MethodImplOptions.Async)]
+    public static T Await<T>(ValueTask<T> task);
 }
 ```
 
-We presume the following `MethodImplOptions` bit is present. This is used to indicate to the JIT that it should generate an async state machine for the method.
+We presume the following `MethodImplOptions` bit is present. This is used to indicate to the JIT that it should generate an async state machine for the method. This bit is not allowed to be used manually on any method; it is added by the compiler
+to an `async` method.
 
 ```cs
 namespace System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


333fred:
@jcouv @jakobbotsch @VSadov @agocke, this is the expected Roslyn code-gen side of the intrinsic method await version. Please give it a look!

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -27,20 +27,32 @@ We use the following helper APIs to indicate suspension points to the runtime, i
...
+    [MethodImpl(MethodImplOptions.Async)]
+    public static void Await(ValueTask task);
+    [MethodImpl(MethodImplOptions.Async)]
+    public static T Await<T>(ValueTask<T> task);
 }
 ```
 
-We presume the following `MethodImplOptions` bit is present. This is used to indicate to the JIT that it should generate an async state machine for the method.
+We presume the following `MethodImplOptions` bit is present. This is used to indicate to the JIT that it should generate an async state machine for the method. This bit is not allowed to be used manually on any method; it is added by the compiler
+to an `async` method.
```
jcouv, line 55: Do you want to add a note for the rule Andy proposed yesterday (method marked with MethodImpl.Async and not returning a Task may not be invoked in user code)?

jcouv (approved):

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -27,20 +27,32 @@ We use the following helper APIs to indicate suspension points to the runtime, i
...
 
 public static class RuntimeHelpers
 {
-    [RuntimeAsyncMethod]
+    // These methods are used to await things that cannot use the Await helpers below
+    [MethodImpl(MethodImplOptions.Async)]
     public static Task AwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter) where TAwaiter : INotifyCompletion;
-    [RuntimeAsyncMethod]
+    [MethodImpl(MethodImplOptions.Async)]
     public static Task UnsafeAwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter) where TAwaiter : ICriticalNotifyCompletion;
```
jakobbotsch, line 39: I think it is most natural for these to be void returning, similar to the intrinsic Await helpers below.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -27,20 +27,32 @@ We use the following helper APIs to indicate suspension points to the runtime, i
...
+
+    // These methods are used to directly await method calls
+    // TODO: Do we only use these when directly awaiting a method call, or do we always use these when a `Task` is awaited, even if that task is a local?
```
jakobbotsch, line 42: It's up to you. `Await` will result in smaller IL over expanding the awaiter based pattern and using `AwaitAwaiterFromRuntimeAsync`.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -27,20 +27,32 @@ We use the following helper APIs to indicate suspension points to the runtime, i
...
+    [MethodImpl(MethodImplOptions.Async)]
+    public static void Await(Task task);
+    [MethodImpl(MethodImplOptions.Async)]
+    public static T Await<T>(Task<T> task);
+    [MethodImpl(MethodImplOptions.Async)]
+    public static void Await(ValueTask task);
+    [MethodImpl(MethodImplOptions.Async)]
+    public static T Await<T>(ValueTask<T> task);
```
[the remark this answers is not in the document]
jakobbotsch: If we are worried about users manually adding calls to these functions then I wonder if we should just mark them `Obsolete` with `error = true` in metadata?

docs/compilers/CSharp/Runtime Async Design.md
333fred: I missed that Andy had updated these signatures as well in his PR, I'll update this side.

docs/compilers/CSharp/Runtime Async Design.md
333fred: It's mainly a question for y'all :laughing:. If you're ok with the pattern being directly calling `Await`, I'm good with it as well.

docs/compilers/CSharp/Runtime Async Design.md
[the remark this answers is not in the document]
333fred: We talked about potentially having the compiler automatically apply `CompilerFeatureRequiredAttribute` when it sees the implementation of these methods in corelib. As to the methodimpl, I'm just copying the signature from the runtime spec; the runtime spec should be treated as the canonical definition of these methods, and any discrepancies should mean that Roslyn probably needs to catch up to something.

docs/compilers/CSharp/Runtime Async Design.md
jakobbotsch: I don't think there will be a difference perf wise (well, if there is it should be on the JIT to optimize it), so I think going with the smaller IL expansion is preferable.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -301,23 +314,44 @@ class C
 }
 ```
 
+Translated C#:
+
+```cs
+System.Runtime.CompilerServices.RuntimeHelpers.Await(C.M<Task>());
```
333fred, line 320: @jakobbotsch @VSadov is it going to complicate the JIT's pattern matching significantly if we use this approach for generic methods like this? In the old specialized call version, we would have needed go to through `UnsafeAwaitAwaiterFromRuntimeAsync`, but since this now just appears to be `Task`-returning method, I'm hoping we can keep IL size down and just call `Await`.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -255,38 +284,22 @@ Translated C#:
...
-var i =
-{
-    var awaiter = local.GetAwaiter();
-    if (!awaiter.IsComplete)
-    {
-        /* Runtime-Async Call */ System.Runtime.CompilerServices.RuntimeHelpers.AwaitAwaiterFromRuntimeAsync<System.Runtime.CompilerServices.TaskAwaiter>(awaiter);
-    }
-    awaiter.GetResult()
-};
+var i = System.Runtime.CompilerServices.RuntimeHelpers.Await<int>(local);
```
jcouv, line 287: The paragraph above still mentions `AwaitAwaiterFromRuntimeAsync` and `UnsafeAwaitAwaiterFromRuntimeAsync`. Can you clarify when we use those vs. `Await`/`Await<T>`?

docs/compilers/CSharp/Runtime Async Design.md
jakobbotsch: I think you should be fine to use `Await` in these situations. It is up to the runtime/JIT to recognize when it will be unable to optimize.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -301,23 +311,44 @@ class C
 }
 ```
 
+Translated C#:
```
jcouv, line 314: Could we reduce the number of sections by removing the now unnecessary distinction between awaiting a call and awaiting a local? We could have a single case for awaiting an expression of type Task/Task<T>/ValueTask/ValueTask<T>.

docs/compilers/CSharp/Runtime Async Design.md
VSadov: Yes. The details of how some methods get special sibling/variant methods (and some do not) is now irrelevant from the public API point. It makes sense to allow to use `Await` with anything that returns `Task`/`ValueTask`. 

JIT will sort it out what can be optimized.

docs/compilers/CSharp/Runtime Async Design.md
```diff
@@ -515,36 +602,152 @@ Translated C#:
...
+    IL_0013: ldloc.0
+    IL_0014: call void [System.Runtime]System.Runtime.CompilerServices.RuntimeHelpers::UnsafeAwaitAwaiterFromRuntimeAsync<class C/Awaiter>(!!0)
+
+    IL_0019: ldloc.0
+    IL_001a: callvirt instance void C/Awaiter::GetResult()
+    IL_001f: ret
+}
+```
+
+#### Implementor of INotifyCompletion
```
jcouv, line 700: >#### [](http://example.com/codeflow?start=0&length=4)

nit: should be one more indent to match previous section?

jcouv (approved):

docs/compilers/CSharp/Runtime Async Design.md
333fred:
```suggestion
##### Implementor of INotifyCompletion
```

PR #76888, opened 2025-01-23
Stop escaped compiler args from being mangled by MSBuild

### LINKED ISSUE #72014: MSBuildWorkspace does not cleanly load VB project on Linux or MacOS
**Version Used**: 4.10 Preview 1

Reported in https://github.com/dotnet/roslyn/pull/69225

**Steps to Reproduce**:

1. dotnet new console --language VB
2. use MSBuildWorkspace.OpenProjectAsync to load the newly created VB project

**Expected Behavior**:
Project loads without any workspace diagnostics and compilation reports no compiler diagnostics.

**Actual Behavior**:
Compilation reports "BC30002": Type 'Global.Microsoft.VisualBasic.ApplicationServices.ApplicationBase' is not defined.

JoeRobich:
The TaskItem implementation from MSBuild treats the ItemSpec as file path and tries to normalize the path separators. When running on non-windows machines this means changing '\\' to '/'.  However this breaks how double quotes are escaped in the compiler args. By using our own implementation of TaskItem we can use the compiler args as the ItemSpec without any modification.

Resolves https://github.com/dotnet/roslyn/issues/72014

### DIFF
--- a/src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
+++ b/src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
@@ -55,8 +55,10 @@ namespace Microsoft.CodeAnalysis.BuildTasks
 
         public void CopyMetadataTo(ITaskItem destinationItem)
         {
+            // Implementation notes that we should not overwrite existing metadata on the destination.
+            var sourceMetadataNames = MetadataNames.OfType<string>();
             var destinationMetadataNames = destinationItem.MetadataNames.OfType<string>();
-            var metadataNamesToCopy = _metadata.Keys.Intersect(destinationMetadataNames, StringComparer.OrdinalIgnoreCase).ToArray();
+            var metadataNamesToCopy = sourceMetadataNames.Except(destinationMetadataNames, StringComparer.OrdinalIgnoreCase).ToArray();
 
             foreach (var metadataName in metadataNamesToCopy)
             {
--- a/src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
+++ b/src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
@@ -5,9 +5,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.Build.Framework;
-using System.Collections.Immutable;
 
 namespace Microsoft.CodeAnalysis.BuildTasks
 {
@@ -43,24 +41,40 @@ namespace Microsoft.CodeAnalysis.BuildTasks
         public string ItemSpec { get; set; }
 
         // Implementation notes that we should include the built-in metadata names as well as our custom ones.
-        public ICollection MetadataNames => _metadata.Keys.Concat(WellKnownItemSpecMetadataNames).ToImmutableArray();
+        public ICollection MetadataNames
+        {
+            get
+            {
+                var clone = new List<string>(_metadata.Keys);
+                clone.AddRange(WellKnownItemSpecMetadataNames);
+                return clone;
+            }
+        }
 
         // Implementation notes that we should include the built-in metadata names as well as our custom ones.
         public int MetadataCount => _metadata.Count + WellKnownItemSpecMetadataNames.Length;
 
         public IDictionary CloneCustomMetadata()
         {
-            return _metadata.ToImmutableDictionary();
+            return new Dictionary<string, string>(_metadata, StringComparer.OrdinalIgnoreCase);
         }
 
         public void CopyMetadataTo(ITaskItem destinationItem)
         {
             // Implementation notes that we should not overwrite existing metadata on the destination.
-            var destinationMetadataNames = destinationItem.MetadataNames.OfType<string>();
-            var metadataNamesToCopy = _metadata.Keys.Except(destinationMetadataNames, StringComparer.OrdinalIgnoreCase).ToArray();
-
-            foreach (var metadataName in metadataNamesToCopy)
+            var destinationMetadataNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var name in destinationItem.MetadataNames)
             {
+                destinationMetadataNames.Add((string)name);
+            }
+
+            foreach (var metadataName in _metadata.Keys)
+            {
+                if (destinationMetadataNames.Contains(metadataName))
+                {
+                    continue;
+                }
+
                 var metadataValue = _metadata[metadataName];
                 destinationItem.SetMetadata(metadataName, metadataValue);
             }
--- a/src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
+++ b/src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
@@ -1,0 +1,83 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Build.Framework;
+using System.Collections.Immutable;
+
+namespace Microsoft.CodeAnalysis.BuildTasks
+{
+    internal sealed class ArgsTaskItem : ITaskItem
+    {
+        // This list is taken from src/Shared/Modifiers.cs in the dotnet/msbuild repo.
+        private static readonly string[] WellKnownItemSpecMetadataNames =
+        {
+            "FullPath",
+            "RootDir",
+            "Filename",
+            "Extension",
+            "RelativeDir",
+            "Directory",
+            "RecursiveDir",
+            "Identity",
+            "ModifiedTime",
+            "CreatedTime",
+            "AccessedTime",
+            "DefiningProjectFullPath",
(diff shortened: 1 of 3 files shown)


src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Linq;
+using Microsoft.Build.Framework;
+using System.Collections.Immutable;
+
+namespace Microsoft.CodeAnalysis.BuildTasks
+{
+    internal sealed class ArgsTaskItem : ITaskItem
+    {
+        // This list is taken from src/Shared/Modifiers.cs in the dotnet/msbuild repo.
+        private static readonly string[] WellKnownItemSpecMetadataNames =
```
jaredpar, line 17: nit: consider using a collection expression

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Build.Framework;
+using System.Collections.Immutable;
+
+namespace Microsoft.CodeAnalysis.BuildTasks
+{
+    internal sealed class ArgsTaskItem : ITaskItem
+    {
+        // This list is taken from src/Shared/Modifiers.cs in the dotnet/msbuild repo.
```
jaredpar, line 16: Let's permalink to the version we took this from.

src/Compilers/Core/MSBuildTask/ManagedToolTask.cs
```diff
@@ -159,7 +159,7 @@ protected static ITaskItem[] GenerateCommandLineArgsTaskItems(List<string> comma
             var items = new ITaskItem[commandLineArgs.Count];
             for (var i = 0; i < commandLineArgs.Count; i++)
             {
-                items[i] = new TaskItem(commandLineArgs[i]);
+                items[i] = new ArgsTaskItem(commandLineArgs[i]);
```
jaredpar, line 162: This will change from us unconditionally escaping to unconditionally not escaping. Won't this regress scenarios where escaping is making builds work today? Consider for example that `/keyfile:` is passed via `commandLineArgs`. If that was using `\` today it would be `/` on unix but after this change it woudl still be `\` correct?

JoeRobich: oof. Luckily many args such as Analyzers, References, Source all come from TaskItems of their own which apply the path correction. So I need to look at args which come from properties and may contain file paths.

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        public string ItemSpec { get; set; }
+
+        // Implementation notes that we should include the built-in metadata names as well as our custom ones.
+        public ICollection MetadataNames
+        {
+            get
+            {
+                var clone = new List<string>(_metadata.Keys);
```
jaredpar, line 48:
```suggestion
                var clone = new List<string>(capacity: MetadataCount);
                clone.AddRange(_metadata.Keys);
```

Avoid unnecessary allocations.

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
```diff
@@ -0,0 +1,98 @@
...
+                clone.AddRange(WellKnownItemSpecMetadataNames);
+                return clone;
+            }
+        }
+
+        // Implementation notes that we should include the built-in metadata names as well as our custom ones.
```
jaredpar, line 54: Which implementation notes this?

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
```diff
@@ -0,0 +1,98 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using Microsoft.Build.Framework;
+
+namespace Microsoft.CodeAnalysis.BuildTasks
+{
+    internal sealed class ArgsTaskItem : ITaskItem
```
jaredpar, line 12: Can we add a comment summarizing that we're using this to avoid unnecessary path escaping?

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
```diff
@@ -0,0 +1,98 @@
...
+        public void CopyMetadataTo(ITaskItem destinationItem)
+        {
+            // Implementation notes that we should not overwrite existing metadata on the destination.
+            var destinationMetadataNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var name in destinationItem.MetadataNames)
+            {
+                destinationMetadataNames.Add((string)name);
+            }
+
+            foreach (var metadataName in _metadata.Keys)
```
jaredpar, line 71: Why are we only using `_metadata.Keys` here and not `MetadataNames`? Seems like it's only copying our custom metadata not the full set.

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
```diff
@@ -0,0 +1,98 @@
...
+                {
+                    continue;
+                }
+
+                var metadataValue = _metadata[metadataName];
+                destinationItem.SetMetadata(metadataName, metadataValue);
+            }
+        }
+
+        public string GetMetadata(string metadataName)
```
jaredpar, line 83: @rainersigwald, @baronfel can you help me understand the contract of `ITaskItem` a bit better? One part of this impl, and others I've seen, is that they have this pattern of exporting a set of names via `MetadataNames` that will throw when you pass them to `GetMetadata`. My mind sees that as a dictionary like type exporting a key but then throwing when you ask for the value. Why does `ITaskItem` behave this way?

baronfel: Do you have a specific instance we could look at? From what we can see TaskItem MetadataNames + GetMetadata _should_ be safe for all of the returned names, which include explicit Item Metadata as well as the historically-significant set of metadata that all Items have (because all items are files, don'tchanknow).

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
JoeRobich: In my reading of the [interface comments](https://github.com/dotnet/msbuild/blob/291a8108761ed347562228f2f8f25477996a5a93/src/Framework/ITaskItem.cs#L74). There seems to be a distinction made between "built-in" and "custom" metadata. The "built-in" metadata seems to be related to the ItemSpec which this method is not supposed to update.

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
JoeRobich: Maybe I misinterpreted the [comments on the interface](https://github.com/dotnet/msbuild/blob/291a8108761ed347562228f2f8f25477996a5a93/src/Framework/ITaskItem.cs#L35) and the [implementation here](https://github.com/dotnet/msbuild/blob/291a8108761ed347562228f2f8f25477996a5a93/src/Utilities/TaskItem.cs#L223). `MetadataNames` includes the "built-in" metadata which I wasn't sure would actually by in my custom metadata dictionary.

src/Compilers/Core/MSBuildTask/ArgsTaskItem.cs
JoeRobich: Will add links.

src/Compilers/Core/MSBuildTask/ManagedToolTask.cs
jasonmalinowski: So is the expectation that MSBuild project files might do things like <SomeProperty>Foo\Bar</SomeProperty> and this conversion is what makes things still work on non-Windows? That seems terrifying, but I guess project files using \ to make paths is _very_ common...

jasonmalinowski, line 162: I had thought we were going to try some trick about cloning a custom implementation. Does something like this not work?

- We create an ArgsTaskItem that implements ITaskItem2 just to return [EvaluatedIncludeEscaped](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.framework.itaskitem2.evaluatedincludeescaped?view=msbuild-17-netcore)
- We wrap then call new TaskItem() passing that task specifically to hit [this line](https://github.com/dotnet/msbuild/blob/37eb419ad2c986ac5530292e6ee08e962390249e/src/Utilities/TaskItem.cs#L132) that looks like it'll bypass further escaping.

That way we're returning an existing implementation rather than returning our own which risks bugs in case we don't get the semantics of the interface right.

JoeRobich: Yep, let's do that instead.

jcouv:
Since there's been no update to the PR for a year, I'm going to mark is Draft for now to make room in our review queue.

JoeRobich:
Letting this go.

PR #76912, opened 2025-01-24
Add back intenral EE APIs used by the debugger until the debugger switches to the new overloads

tmat:
Fixes insertion break.


ToddGrun (approved):
:shipit:

PR #76918, opened 2025-01-24
Target test projects in Workspaces and Features layers to NetVSShared

tmat:
Our primary scenarios are devenv, VS OOP and VS Code, hence the tests should target `net472` and `net8.0`.

Switching to the right TFM uncovers a bug in Semantic Search. We targeted Semantic Search references to `net9.0` and tried to load the code to `net8.0` OOP.


tmat:
@dotnet/roslyn-ide @jaredpar  ptal

CyrusNajmabadi (approved):

PR #76932, opened 2025-01-26
Add new workspace event that gives handlers the opportunity to be processed immediately

ToddGrun:
This is in response to allocations seen when the CA process attempts to operate on a sourcetext version that has not been synchronized over. In those cases, the OOP process requests a full deserialization of the source text causing full buffer allocations in both VS and the CA process.

There are several points of asynchronicity in the current system around sending over buffer changes, reducing any of which would reduce the likelihood of needing this full serialization.

This PR removes one of those points of asynchronicity, specifically in the workspace eventing layer. Previously, all notiications were done on a delayed basis, by wrapping each notification in the WorkspaceSpace.ScheduleTask call from the RaiseWorkspaceChangedEventAsync method. This method allows callers to indicate they need to be called immediately upon the workspace change. As noted by a doc comment in the PR, these handlers should be very fast.

The numbers from speedometer came back looking pretty good (below). Another point of asynchronicity could be investigated around the usage of the _textChangeQueue ABWQ, but it might not be worth it as it looks like this change removes the majority of the serialization during typing. Other alternatives within the text syncing system such as allowing OOP to request a text change instead of the full buffer contents are also possible, but previous investigations into that ended up complicated and incomplete.

*** before allocations during typing in VS process ***
image

*** commit 3 allocations during typing in VS process ***
image

*** commit 17 allocations during typing in VS process ***
image

*** before allocations during typing in CA process ***
image

*** commit 3 allocations during typing in CA process ***
image

*** commit 17 allocations during typing in CA process ***
image


ToddGrun:
PR validation insertion pipeline: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10897976&view=results

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Log/FunctionId.cs
```diff
@@ -484,6 +484,8 @@ internal enum FunctionId
     Renamer_FindRenameLocationsAsync = 387,
     Renamer_ResolveConflictsAsync = 388,
 
+    Workspace_EventsImmediate = 389,
```
CyrusNajmabadi, line 487: I would keep this range free for rename

ToddGrun: Didn't see a good spot for a one-off id to add for Workspace events. Any suggestion?

CyrusNajmabadi: I'd just start a new range :-)

ToddGrun:
Moving out of draft mode as the numbers look pretty good. Still some serialization happening during typing, but it's a lot less than before and this change was pretty trivial. If code reviewers think adding this to the workspace level is not desirable, can look into other approaches.

CyrusNajmabadi (approved):

ToddGrun:
commit 5 PR validation pipeline run: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/results?buildId=10900641&view=results

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,45 +203,62 @@ private void DispatchSynchronizeTextChanges(
...
-            // send out that text changes to remote side.
-            //
-            // the remote side, once got the text change, will again see whether it can use that text change information
-            // without any high cost and create new snapshot from it.
-            //
-            // otherwise, it will do the normal behavior of getting full text from VS side. this optimization saves
-            // times we need to do full text synchronization for typing scenario.
-            if (!oldDocument.TryGetText(out var oldText) ||
-                !newDocument.TryGetText(out var newText))
+            var ableToSync = false;
```
CyrusNajmabadi, line 206: consider just having a helper method (or local function) that just returns a bool.

src/Workspaces/Remote/ServiceHub/Services/AssetSynchronization/RemoteAssetSynchronizationService.cs
```diff
@@ -60,22 +61,39 @@ public ValueTask SynchronizeTextChangesAsync(
 
             using (RoslynLogger.LogBlock(FunctionId.RemoteHostService_SynchronizeTextAsync, cancellationToken))
             {
-                // Try to get the text associated with baseTextChecksum
-                var text = await TryGetSourceTextAsync(WorkspaceManager, workspace, documentId, baseTextChecksum, cancellationToken).ConfigureAwait(false);
-                if (text == null)
+                var ableToSync = false;
```
CyrusNajmabadi, line 64: same feedback.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,57 +197,68 @@ await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
...
-        //
-        // the remote side, once got the text change, will again see whether it can use that text change information
-        // without any high cost and create new snapshot from it.
-        //
-        // otherwise, it will do the normal behavior of getting full text from VS side. this optimization saves
-        // times we need to do full text synchronization for typing scenario.
-        using var _ = ArrayBuilder<(DocumentId id, Checksum textChecksum, ImmutableArray<TextChange> changes, Checksum newTextChecksum)>.GetInstance(out var builder);
-
-        foreach (var (oldDocument, newDocument) in values)
+        _ = _threadingContext.JoinableTaskFactory.RunAsync(async () =>
```
CyrusNajmabadi, line 204: we should def doc the design here.  make it clear why we've written it this way.

note: you can also use IAsyncToken here to track the work, allowing you to properly be able to unit test as well.

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
```diff
@@ -59,14 +77,25 @@ protected Task RaiseWorkspaceChangedEventAsync(WorkspaceChangeKind kind, Solutio
...
 
-        var ev = GetEventHandlers<WorkspaceChangeEventArgs>(WorkspaceChangeEventName);
+        var ev = GetEventHandlers<WorkspaceChangeEventArgs>(WorkspaceChangedImmediateEventName);
+        WorkspaceChangeEventArgs args = null;
+        if (ev.HasHandlers)
+        {
+            using (Logger.LogBlock(FunctionId.Workspace_EventsImmediate, (s, p, d, k) => $"{s.Id} - {p} - {d} {kind.ToString()}", newSolution, projectId, documentId, kind, CancellationToken.None))
+            {
+                args = new WorkspaceChangeEventArgs(kind, oldSolution, newSolution, projectId, documentId);
+                ev.RaiseEvent(static (handler, arg) => handler(arg.self, arg.args), (self: this, args));
```
CyrusNajmabadi, line 87: aside, are these exception safe if the attached handler throws an exception?  i ask because the old system might have been ok ift he Task went into a faulted state.  i'm not sure about the new system.

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
```diff
@@ -42,6 +43,23 @@ public event EventHandler<WorkspaceChangeEventArgs> WorkspaceChanged
         }
     }
 
+    /// <summary>
+    /// An event raised *immediately* whenever the current solution is changed. Handlers
+    /// should be written to be very fast.
```
jasonmalinowski, line 48:
```suggestion
    /// An event raised *immediately* whenever the current solution is changed. Handlers
    /// should be written to be very fast. Called on the same thread changing the workspace,
    /// which may vary depending on the workspace.
```

Just to make it clear which thread it's on (and that it's probably not great to depend on that).

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
```diff
@@ -59,14 +77,25 @@ protected Task RaiseWorkspaceChangedEventAsync(WorkspaceChangeKind kind, Solutio
...
+        var ev = GetEventHandlers<WorkspaceChangeEventArgs>(WorkspaceChangedImmediateEventName);
+        WorkspaceChangeEventArgs args = null;
+        if (ev.HasHandlers)
+        {
+            using (Logger.LogBlock(FunctionId.Workspace_EventsImmediate, (s, p, d, k) => $"{s.Id} - {p} - {d} {kind.ToString()}", newSolution, projectId, documentId, kind, CancellationToken.None))
+            {
+                args = new WorkspaceChangeEventArgs(kind, oldSolution, newSolution, projectId, documentId);
+                ev.RaiseEvent(static (handler, arg) => handler(arg.self, arg.args), (self: this, args));
+            }
+        }
```
jasonmalinowski, line 89: Can we extract this bit into a local function so there's not repetition?

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
jasonmalinowski: Oops asked this same question here: https://github.com/dotnet/roslyn/pull/76932#discussion_r1931416071 I think this might cause an issue now if this were to throw so probably best to handle that better.

ToddGrun: Any concern with me just changing RaiseEvent from using FatalError.ReportAndPropagate to instead use FatalError.ReportAndCatch (also, would it make sense to have the try/catch inside the foreach in that method?)

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: The intent is to fire-and-forget, but only yield the current thread if an async operation necessitates it.

Is GetChangeRanges expensive? It doesn't look to be to me, but maybe I'm missing an override? If it is expensive, the two conditions that use it could probably be removed, as they are very minor and specialized optimizations.

Will add some doc here

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Log/FunctionId.cs
```diff
@@ -639,4 +639,9 @@ internal enum FunctionId
     VSCode_LanguageServer_Started = 860,
     VSCode_Project_Load_Started = 861,
     VSCode_Projects_Load_Completed = 862,
+
+    // 900-999 for items that don't fit into other categories.
```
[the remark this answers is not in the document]
ToddGrun: I wanted a catch-all group as I was tired scanning through trying to find the next available slot with all the other groups. Is the request here to change the starting range / size of the catch all group, or just not to have it?

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -160,6 +144,17 @@ private void OnWorkspaceChanged(object? sender, WorkspaceChangeEventArgs e)
         }
     }
 
+    private void OnWorkspaceChangedImmediate(object? sender, WorkspaceChangeEventArgs e)
+    {
+        if (e.Kind == WorkspaceChangeKind.DocumentChanged)
+        {
+            var oldDocument = e.OldSolution.GetDocument(e.DocumentId);
+            var newDocument = e.NewSolution.GetDocument(e.DocumentId);
+            if (oldDocument != null && newDocument != null)
```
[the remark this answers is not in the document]
ToddGrun: Would an assert be sufficient here? As the old code was resilient to either not being found, I'd be slightly hesitant to throw on that.

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
ToddGrun: Extracted into local method, slightly complicated by wanting to keep the args allocation to be just once

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -53,9 +44,13 @@ internal sealed class SolutionChecksumUpdater
...
     private bool _isSynchronizeWorkspacePaused;
 
+    private readonly IThreadingContext _threadingContext;
+    private readonly CancellationToken _shutdownToken;
+
     public SolutionChecksumUpdater(
         Workspace workspace,
         IAsynchronousOperationListenerProvider listenerProvider,
+        IThreadingContext threadingContext,
         CancellationToken shutdownToken)
```
CyrusNajmabadi, line 54: I'm curious if the shutdown token is needed. Instead of just using the disposal token on the threading context

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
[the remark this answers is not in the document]
CyrusNajmabadi: I don't mind the paranoia. Especially since the old code had it.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,57 +197,76 @@ await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
...
-        // times we need to do full text synchronization for typing scenario.
-        using var _ = ArrayBuilder<(DocumentId id, Checksum textChecksum, ImmutableArray<TextChange> changes, Checksum newTextChecksum)>.GetInstance(out var builder);
-
-        foreach (var (oldDocument, newDocument) in values)
+        // Attempt to inform the remote asset synchronization service as quickly as possible
+        // about the text changes between oldDocument and newDocument. By doing this, we can
+        // reduce the likelihood of the remote side encountering an unknown checksum and
+        // requiring a synchronization of the full document.
+        // This method uses JTF.RunAsync to create a fire-and-forget task. JTF.RunAsync will
+        // attempt to execute DispatchSynchronizeTextChangesHelperAsync synchronously if possible.
```
CyrusNajmabadi, line 209: But can this ever be synchronous?  It's an rpc call across processes after all.

src/VisualStudio/Core/Def/Remote/VisualStudioWorkspaceServiceHubConnector.cs
```diff
@@ -38,7 +40,7 @@ public void StartListening(Workspace workspace, object serviceOpt)
         }
 
         // only push solution snapshot from primary (VS) workspace:
-        _checksumUpdater = new SolutionChecksumUpdater(workspace, _listenerProvider, _disposalCancellationSource.Token);
+        _checksumUpdater = new SolutionChecksumUpdater(workspace, _listenerProvider, _threadingContext, _disposalCancellationSource.Token);
```
CyrusNajmabadi, line 43: If we have the threading context, we might be able to get rid of the disposal logic here and passing that ct along

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
```diff
@@ -59,22 +78,40 @@ protected Task RaiseWorkspaceChangedEventAsync(WorkspaceChangeKind kind, Solutio
...
         {
             return Task.CompletedTask;
         }
+
+        static void RaiseEventForHandlers(
+            EventMap.EventHandlerSet<EventHandler<WorkspaceChangeEventArgs>> handlers,
+            WorkspaceChangeEventArgs args,
+            bool isImmediate)
+        {
+            var functionId = isImmediate ? FunctionId.Workspace_EventsImmediate : FunctionId.Workspace_Events;
```
CyrusNajmabadi, line 108: Consider just passing the funding id along instead of the bool

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
```diff
@@ -59,22 +78,40 @@ protected Task RaiseWorkspaceChangedEventAsync(WorkspaceChangeKind kind, Solutio
...
+
+            using (Logger.LogBlock(functionId, (s, p, d, k) => $"{s.Id} - {p} - {d} {args.Kind.ToString()}", args.NewSolution, args.ProjectId, args.DocumentId, args.Kind, CancellationToken.None))
+            {
+                handlers.RaiseEvent(static (handler, args) => handler(args.NewSolution.Workspace, args), args);
```
CyrusNajmabadi, line 112: I def have concerns about exceptions. Can we try/catch with a fatalerror reporter?

src/Workspaces/Remote/ServiceHub/Services/AssetSynchronization/RemoteAssetSynchronizationService.cs
```diff
@@ -49,24 +49,39 @@ public ValueTask SynchronizeActiveDocumentAsync(DocumentId? documentId, Cancella
...
         {
-            var workspace = GetWorkspace();
+            var wasSynchronized = await SynchronizeTextChangesHelperAsync().ConfigureAwait(false);
 
-            using (RoslynLogger.LogBlock(FunctionId.RemoteHostService_SynchronizeTextAsync, cancellationToken))
+            var metricName = wasSynchronized ? "SucceededCount" : "FailedCount";
+            TelemetryLogging.LogAggregatedCounter(FunctionId.RemoteHostService_SynchronizeTextAsyncStatus, KeyValueLogMessage.Create(m =>
             {
-                foreach (var (documentId, baseTextChecksum, textChanges, newTextChecksum) in changes)
+                m[TelemetryLogging.KeyName] = nameof(RemoteAssetSynchronizationService) + "." + metricName;
```
CyrusNajmabadi, line 65: Consider constants for the two values so we don't keep allocating dupe strings.

CyrusNajmabadi (approved):
Signing mostly off.  I do want Jason to approve as well. And there look to be a few places we can clean up. I would like exception resilience thought about. Thanks!

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: I think they are different. VisualStudioWorkspaceServiceHubConnector has StartListening/StopListening methods, which drive the shutdownToken. However, I think the threadingContext's DisposalToken is driven by MEF and I don't think is disposed until VS shutdown.

ToddGrun: Just saw your other comment about VisualStudioWorkspaceServiceHubConnector. Let me think about that.

ToddGrun: It looks like right now, VisualStudioWorkspaceServiceHubConnector.StopListening is called on Workspace dispose. If I just send over the threadingContext and use it's cancellation token, are we ok with essentially just using a cancellation token from MEF dispose and not workspace dispose?

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: That's a good point, I'll update the comment. It can be sync right up to that RPC call (which is all we care about)

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
ToddGrun: done!

src/Workspaces/Remote/ServiceHub/Services/AssetSynchronization/RemoteAssetSynchronizationService.cs
ToddGrun: Done!

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
ToddGrun: From the other thread:

Any concern with me just changing RaiseEvent from using FatalError.ReportAndPropagate to instead use FatalError.ReportAndCatch (also, would it make sense to have the try/catch inside the foreach in that method?)

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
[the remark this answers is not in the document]
ToddGrun: Went ahead and added just an assert

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
ToddGrun: @jasonmalinowski -- any thoughts on change RaiseEvent to instead use FatalError.ReportAndCatch (and to move the try/catch inside the loop)?

ToddGrun: I went ahead and just made the change to switch to ReportAndCatch

ToddGrun:
@jasonmalinowski -- would love to get your feedback on the changes and any remaining concerns you may have that haven't been addressed as desired

sharwell:
For this specific performance improvement, I'm curious about the viability of an alternative approach:

1. Attempt to produce the solution for an incoming call
2. If any document is not already available, wait for queued synchronization events to be processed
3. Produce the solution for the incoming call under the original approach

CyrusNajmabadi:
@sharwell def feel free to explore that idea. The approach here is very simple and effective and sits on a model we have good understanding of. All this is doing is allowing the current approach to operate without unnecessary delays that impact its efficacy.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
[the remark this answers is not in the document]
CyrusNajmabadi: wfm.  if we hit it we can investigate.  but the retail code is sstill safe.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,57 +200,76 @@ await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
...
-            var state = await oldDocument.State.GetStateChecksumsAsync(cancellationToken).ConfigureAwait(false);
-            var newState = await newDocument.State.GetStateChecksumsAsync(cancellationToken).ConfigureAwait(false);
-
-            var textChanges = newText.GetTextChanges(oldText).AsImmutable();
-            builder.Add((oldDocument.Id, state.Text, textChanges, newState.Text));
-        }
+            var metricName = wasSynchronized ? "SucceededCount" : "FailedCount";
+            TelemetryLogging.LogAggregatedCounter(FunctionId.ChecksumUpdater_SynchronizeTextChangesStatus, KeyValueLogMessage.Create(m =>
+            {
+                m[TelemetryLogging.KeyName] = nameof(SolutionChecksumUpdater) + "." + metricName;
```
CyrusNajmabadi, line 222: it feels like this will continually allocate.  can we instead make these into constants?

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,57 +200,76 @@ await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
...
+                m[TelemetryLogging.KeyValue] = 1L;
```
CyrusNajmabadi, line 223: is the idea that this will add 1 each time this happens?

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,57 +200,76 @@ await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
...
+                m[TelemetryLogging.KeyMetricName] = metricName;
```
CyrusNajmabadi, line 224: why do we need this here, if it is encoded in KeyName above? is one of these redundant?

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,57 +200,76 @@ await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
...
             return;
 
-        await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
-            (service, cancellationToken) => service.SynchronizeTextChangesAsync(builder.ToImmutableAndClear(), cancellationToken),
-            cancellationToken).ConfigureAwait(false);
+            async Task<bool> DispatchSynchronizeTextChangesHelperAsync()
+            {
+                var client = await RemoteHostClient.TryGetClientAsync(_workspace, _shutdownToken).ConfigureAwait(false);
+                if (client == null)
+                    return false;
```
CyrusNajmabadi, line 233: slightly odd that this would show up as 'failure'.  this means a customer who has OOP off will show has massively failing this op.  i think this should maybe count as 'success' or perhaps NA so we don't get the wrong idea about being in a bad statee.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: Oops, I changed the other one to constants, but not this one. Will change

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: Yes, the LogAggregatedCounter will add the specified amount to the counter

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
CyrusNajmabadi: yeah.  let's keep this as is.  i've looked at hte code and i'm not sure how to reconcile the different Disposes/CTs.  So let's not change that  now.

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
CyrusNajmabadi: yes.  please ReportAndCatch. We really do not our eventing to ever get into a failed state (or corrupt the WS).

src/Workspaces/Remote/ServiceHub/Services/AssetSynchronization/RemoteAssetSynchronizationService.cs
```diff
@@ -49,24 +54,40 @@ public ValueTask SynchronizeActiveDocumentAsync(DocumentId? documentId, Cancella
...
-            var workspace = GetWorkspace();
+            var wasSynchronized = await SynchronizeTextChangesHelperAsync().ConfigureAwait(false);
+
+            var metricName = wasSynchronized ? SynchronizeTextChangesAsyncSucceededMetricName : SynchronizeTextChangesAsyncFailedMetricName;
+            var keyName = wasSynchronized ? SynchronizeTextChangesAsyncSucceededKeyName : SynchronizeTextChangesAsyncFailedKeyName;
+            TelemetryLogging.LogAggregatedCounter(FunctionId.RemoteHostService_SynchronizeTextAsyncStatus, KeyValueLogMessage.Create(m =>
+            {
+                m[TelemetryLogging.KeyName] = keyName;
+                m[TelemetryLogging.KeyValue] = 1L;
+                m[TelemetryLogging.KeyMetricName] = metricName;
```
CyrusNajmabadi, line 73: same comment if we need the keyname and metric name esp if hte keyname contains the metric name.

CyrusNajmabadi:
with your latest changes, i feel fine with this going in (even without Jason signing off).  he will eventuallyget around to looking at this.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: This ends up down in AbstractAggregatingLog.Log. That code requires KeyName/KeyValue to be specified, and if KeyMetricName isn't specified, falls back to KeyName to be the metricName.

I can look into whether there are callers that depend on that behavior, and if not, remove that parameter. However, I'd prefer not to as even if that could be removed, it would add an allocation to strip out the metricname from the keyname.

ToddGrun:
@sharwell has an alternative approach which he's going to try out. I'll get perf numbers on his approach once it's ready and then we can evaluate how to proceed.

---
In reply to: [2628571096](https://github.com/dotnet/roslyn/pull/76932#issuecomment-2628571096) [](http://example.com/codeflow?ancestors=2628571096)

ToddGrun:
Latest test insertion: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/607891 (based on commit 14)

ToddGrun:
Alternative proposed implementation by Sam here: https://github.com/dotnet/roslyn/pull/77059

Running speedometer tests against that and will update once those are available

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Log/FunctionId.cs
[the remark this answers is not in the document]
jasonmalinowski: Just don't have it. And if we think groups don't make sense in the first place, just get rid of all the groups. :smile:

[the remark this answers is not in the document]
CyrusNajmabadi: i really like groups.  just give things around 10-20 enries.  it's def nicer when you're adding a bunch of new telemetry markers to a speciifc area.

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
```diff
@@ -59,22 +78,38 @@ protected Task RaiseWorkspaceChangedEventAsync(WorkspaceChangeKind kind, Solutio
             projectId = documentId.ProjectId;
         }
 
-        var ev = GetEventHandlers<WorkspaceChangeEventArgs>(WorkspaceChangeEventName);
+        var ev = GetEventHandlers<WorkspaceChangeEventArgs>(WorkspaceChangedImmediateEventName);
+        WorkspaceChangeEventArgs args = null;
+        if (ev.HasHandlers)
+        {
+            args = new WorkspaceChangeEventArgs(kind, oldSolution, newSolution, projectId, documentId);
```
jasonmalinowski, line 85: You could just pull this assignment out of the if() block here; practically any workspace where we would care about allocations will have at least one handler. :smile:

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Utilities/EventMap.cs
```diff
@@ -161,20 +161,21 @@ public readonly void RaiseEvent<TArg>(Action<TEventHandler, TArg> invoker, TArg
             // will raise that as a fatal exception, but OperationCancelledExceptions might be able to propagate through and fault the task we are using in the
             // chain. I'm choosing to use ReportWithoutCrashAndPropagate, because if our theory here is correct, it seems the first exception isn't actually
             // causing crashes, and so if it turns out this is a very common situation I don't want to make a often-benign situation fatal.
```
jasonmalinowski, line 163: This comment is stale now. This looks like it was written by me, but since the comment has a link to a query (rather than the actual bug), it's quite useless.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
jasonmalinowski: If that's the intent, then I don't think the RunAsync() is doing anything for us here, right?

jasonmalinowski: I had _thought_ GetChangeRanges could potentially end up in a text diff, but maybe not.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -160,6 +146,21 @@ private void OnWorkspaceChanged(object? sender, WorkspaceChangeEventArgs e)
...
     }
 
+    private void OnWorkspaceChangedImmediate(object? sender, WorkspaceChangeEventArgs e)
+    {
+        if (e.Kind == WorkspaceChangeKind.DocumentChanged)
+        {
+            var oldDocument = e.OldSolution.GetDocument(e.DocumentId);
+            var newDocument = e.NewSolution.GetDocument(e.DocumentId);
+
+            Debug.Assert(oldDocument != null && newDocument != null);
```
jasonmalinowski, line 156:
```suggestion
            var oldDocument = e.OldSolution.GetRequiredDocument(e.DocumentId);
            var newDocument = e.NewSolution.GetRequiredDocument(e.DocumentId);
```

If we're wrong, telemetry will tell us since you're adding that catch-and-report. :smile:

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,57 +203,81 @@ await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
...
 
             // no benefit here. pulling from remote host is more efficient
             if (changeRanges is [{ Span.Length: var singleChangeLength }] && singleChangeLength == oldText.Length)
-                continue;
+                return true;
 
-            var state = await oldDocument.State.GetStateChecksumsAsync(cancellationToken).ConfigureAwait(false);
-            var newState = await newDocument.State.GetStateChecksumsAsync(cancellationToken).ConfigureAwait(false);
+            var state = await oldDocument.State.GetStateChecksumsAsync(_shutdownToken).ConfigureAwait(false);
+            var newState = await newDocument.State.GetStateChecksumsAsync(_shutdownToken).ConfigureAwait(false);
```
jasonmalinowski, line 272: I'm confused a bit by the semantics we're trying to have here for OnWorkspaceChangedImmediate. The statement was "it must be very fast", but now we're computing checksums? I imagine the async-ness here is we use an AsyncLazy<> under the covers, and so if somebody else is asking we won't block our thread. But this newDocument is very new (indeed, we're still raising events for it!) so would this be the first place it'll get called? And would that then be consuming CPU right here when we wanted this to be cheap?

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
```diff
@@ -202,57 +203,81 @@ await client.TryInvokeAsync<IRemoteAssetSynchronizationService>(
...
+        Document oldDocument,
+        Document newDocument)
     {
-        var client = await RemoteHostClient.TryGetClientAsync(_workspace, cancellationToken).ConfigureAwait(false);
-        if (client == null)
+        // Attempt to inform the remote asset synchronization service as quickly as possible
+        // about the text changes between oldDocument and newDocument. By doing this, we can
+        // reduce the likelihood of the remote side encountering an unknown checksum and
+        // requiring a synchronization of the full document contents.
+        // This method uses JTF.RunAsync to create a fire-and-forget task. JTF.RunAsync will
```
jasonmalinowski, line 214: Comment now stale? RunAsync not being used anymore?

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: This is a very valid concern, debugged into the hash computation for the newDocument, and this is forcing it on the calling (UI) thread. Need to think on this.

CyrusNajmabadi:
could we just do the simple thing of moving off the UI thread (use a TPL.Run or Yield), and just seeing if we still get the good results of not needing to sync anything?

ToddGrun:
Yeah, maybe a Task.Yield would be good enough. Can try that out and see what the numbers say.

---
In reply to: [2638638591](https://github.com/dotnet/roslyn/pull/76932#issuecomment-2638638591) [](http://example.com/codeflow?ancestors=2638638591)

ToddGrun:
And here's the commit 17 test insertion: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/608770

Numbers looked pretty good, some serialization still happens, but it's a lot less. Will update the PR description

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: Added the task.Yield as Cyrus suggested, and verified that all the calculations have moved to a bg thread, but still saw pretty good allocation numbers as outlined in the PR updated description.

src/EditorFeatures/Core/Remote/SolutionChecksumUpdater.cs
ToddGrun: Removed the stale bit.

src/Workspaces/Core/Portable/Workspace/Workspace_Events.cs
ToddGrun: Done!

src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Utilities/EventMap.cs
ToddGrun: Killed the comment, things are different enough now that it's really not applicable.

ToddGrun:
@jasonmalinowski, @sharwell -- Will go ahead and merge on Monday unless either of you feels otherwise

PR #76948, opened 2025-01-28
System.CommandLine update

adamsitnik:
This PR consists of #76851 (the maestro PR) and actual changes required to update S.CL.


src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Program.cs
```diff
@@ -103,7 +102,7 @@ static async Task RunAsync(ServerConfiguration serverConfiguration, Cancellation
 
     // LSP server doesn't have the pieces yet to support 'balanced' mode for source-generators.  Hardcode us to
     // 'automatic' for now.
-    var globalOptionService = exportProvider.GetExportedValue<IGlobalOptionService>();
+    var globalOptionService = exportProvider.GetExportedValue<Microsoft.CodeAnalysis.Options.IGlobalOptionService>();
```
adamsitnik, line 105: After removing the `Cli` prefix (to make sure we don't break every single System.CommandLine user), `Microsoft.CodeAnalysis.Options.Option<T>` conflicts with `System.CommandLine.Option<T>`. I decided to remove the `using Microsoft.CodeAnalysis.Options` line and specify full name here as other solutions would require more changes.

PR #76952, opened 2025-01-28
Merges "release/dev17.13" to "main"

JoeRobich:
In particular this is to flow localization changes to main.


jjonescz (approved):

dibarbet (approved):
diff looks like only resources changes which is what I would expect

Issue #74538, opened 2024-07-24
String interpolation of only strings should light-up with string.Concat(params ReadOnlySpan<string>)


stephentoub:
Currently string interpolation when all of the parts are strings prefers to lower to a string.Concat call when there's a string.Concat overload with the exact same number of string parameters. Beyond that, it falls back to using normal interpolation mechanisms. Now that in .NET 9 there's a `string.Concat(params ReadOnlySpan<string>)` overload, it'd be nice if string interpolation preferred to use that overload when it's available, when all the parts are strings, and when there isn't a better overload of string.Concat available. This statement is invalidated by the new params span overload, which can support passing in all of the strings without an allocation:
https://github.com/dotnet/roslyn/blob/91773064bdc85ab3f64d8118d7b1f562f70dccc6/src/Compilers/CSharp/Portable/Binder/Binder_InterpolatedString.cs#L231-L233

cc: @333fred

333fred:
@stephentoub do you want us to continue calling the discrete parameter versions for 4 or less strings, or just always use the `params ROS` version when available?

stephentoub:
> [@stephentoub](https://github.com/stephentoub) do you want us to continue calling the discrete parameter versions for 4 or less strings, or just always use the `params ROS` version when available?

Ideally it should use the discrete parameter versions when they exist, and fall back to the span overload when they don't. So, based on the APIs we have today, 1/2/3/4 strings will use the dedicated overloads and 5+ will use the `ReadOnlySpan<string>` overload. The span overload is just a bit more expensive as it needs to enumerate through the strings, whereas the dedicated overloads are all hardcoded to that exact number.

stephentoub:
(Thanks!)

stephentoub:
Also, and let me know if I should open a separate issue for this, currently if I do this:
```C#
string s1 = "abc";
string s2 = "def";
string s3 = "ghi";

string result = s1 + " " + s2 + " " + s3;
```
that compiles down to the equivalent of this:
```C#
string result = string.Concat(new string[] { s1, " ", s2, " ", s3 });
```
whereas it'd be nice if it instead compiled down to the equivalent of:
```C#
string result = string.Concat((ReadOnlySpan<string>)[s1, " ", s2, " ", s3]);
```
saving the string[] allocation and binding to the same overload we're talking about here for interpolation.

333fred:
Hopefully I can address at the same time, let's just keep it in one issue.

[...]

PR #76955, opened 2025-01-28
Rewrite string.Concat lowering code

333fred:
Our existing lowering strategy for string addition is complex, to say the least. It works in a bottom-up fashion, needing to constantly undo and redo existing work as VisitExpression returns and discovered ever-larger sequences of string additions. This is not easily maintainable, and also makes addressing issues like https://github.com/dotnet/roslyn/issues/74538 extremely difficult. As a first step in that direction, I've rewritten how we lower these to instead approach it top-down; when a `+` that operates on strings is encountered, we gather all possible operands up front, fold as many constant components as we can, and then emit the final `string.Concat` call all in one go. This should hopefully be significantly easier to maintain in the future, as well as be a much more flexible base for building further optimizations on, such as using `string.Concat(ReadOnlySpan<string>)` or `DefaultInterpolatedStringHandler` in the future.

### DIFF
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
@@ -128,7 +128,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             if (IsBinaryStringConcatenation(node))
             {
                 Debug.Assert(applyParentUnaryOperator is null);
-                return RewriteStringConcatenation(node);
+                return VisitStringConcatenation(node);
             }
 
             // In machine-generated code we frequently end up with binary operator trees that are deep on the left,
@@ -262,7 +262,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                     case BinaryOperatorKind.ObjectAndStringConcatenation:
                     case BinaryOperatorKind.StringAndObjectConcatenation:
                     case BinaryOperatorKind.StringConcatenation:
-                        //return RewriteStringConcatenation(syntax, operatorKind, loweredLeft, loweredRight, type);
                         throw ExceptionUtilities.UnexpectedValue(operatorKind);
 
                     case BinaryOperatorKind.StringEqual:
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
@@ -125,6 +125,12 @@ namespace Microsoft.CodeAnalysis.CSharp
                 return VisitUtf8Addition(node);
             }
 
+            if (IsBinaryStringConcatenation(node))
+            {
+                Debug.Assert(applyParentUnaryOperator is null);
+                return RewriteStringConcatenation(node);
+            }
+
             // In machine-generated code we frequently end up with binary operator trees that are deep on the left,
             // such as a + b + c + d ...
             // To avoid blowing the call stack, we make an explicit stack of the binary operators to the left, 
@@ -135,7 +141,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             for (BoundBinaryOperator? current = node; current != null && current.ConstantValueOpt == null; current = current.Left as BoundBinaryOperator)
             {
                 // The regular visit mechanism will handle this.
-                if (current.InterpolatedStringHandlerData is not null || current.OperatorKind is BinaryOperatorKind.Utf8Addition)
+                if (current.InterpolatedStringHandlerData is not null || current.OperatorKind is BinaryOperatorKind.Utf8Addition || IsBinaryStringConcatenation(current))
                 {
                     Debug.Assert(stack.Count >= 1);
                     break;
@@ -208,7 +214,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                     case BinaryOperatorKind.ObjectAndStringConcatenation:
                     case BinaryOperatorKind.StringAndObjectConcatenation:
                     case BinaryOperatorKind.StringConcatenation:
-                        return RewriteStringConcatenation(syntax, operatorKind, loweredLeft, loweredRight, type);
+                        throw ExceptionUtilities.UnexpectedValue(operatorKind);
                     case BinaryOperatorKind.DelegateCombination:
                         return RewriteDelegateOperation(syntax, operatorKind, loweredLeft, loweredRight, type, SpecialMember.System_Delegate__Combine);
                     case BinaryOperatorKind.DelegateRemoval:
@@ -256,7 +262,8 @@ namespace Microsoft.CodeAnalysis.CSharp
                     case BinaryOperatorKind.ObjectAndStringConcatenation:
                     case BinaryOperatorKind.StringAndObjectConcatenation:
                     case BinaryOperatorKind.StringConcatenation:
-                        return RewriteStringConcatenation(syntax, operatorKind, loweredLeft, loweredRight, type);
+                        //return RewriteStringConcatenation(syntax, operatorKind, loweredLeft, loweredRight, type);
+                        throw ExceptionUtilities.UnexpectedValue(operatorKind);
 
                     case BinaryOperatorKind.StringEqual:
                         return RewriteStringEquality(oldNode, syntax, operatorKind, loweredLeft, loweredRight, type, SpecialMember.System_String__op_Equality);
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
@@ -72,6 +72,7 @@ namespace Microsoft.CodeAnalysis.CSharp
         private BoundExpression CreateStringConcat(SyntaxNode originalSyntax, ArrayBuilder<BoundExpression> visitedArguments)
         {
             Debug.Assert(!_inExpressionLambda);
+            Debug.Assert(visitedArguments.All(arg => arg.Type!.SpecialType is SpecialType.System_String or SpecialType.System_Char or SpecialType.System_Object));
             // There are a few different lowering patterns that we take:
             //
             // 1. If all the added expressions were folded into a single constant, we can just return that.
@@ -286,8 +287,7 @@ fallbackStrings:
             {
                 argument = operand;
             }
-
-            if (argument is BoundCall call)
+            else if (argument is BoundCall call)
             {
                 if (wellKnownConcatOptimizationMethods.IsWellKnownConcatMethod(call, out var concatArguments))
                 {
@@ -303,9 +303,8 @@ fallbackStrings:
                     argument = charExpression;
                 }
             }
-
             // This is `strValue ?? ""`, possibly from a nested binary addition of an interpolated string. We can just directly use the left operand
-            if (argument is BoundNullCoalescingOperator { LeftOperand: { Type.SpecialType: SpecialType.System_String } left, RightOperand: BoundLiteral { ConstantValueOpt: { IsString: true, RopeValue.IsEmpty: true } } })
+            else if (argument is BoundNullCoalescingOperator { LeftOperand: { Type.SpecialType: SpecialType.System_String } left, RightOperand: BoundLiteral { ConstantValueOpt: { IsString: true, RopeValue.IsEmpty: true } } })
             {
                 argument = left;
             }
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
@@ -587,18 +587,9 @@ fallbackStrings:
                 }
             }
 
-            // If expression is of form `constantChar.ToString()` then unwrap it from a `ToString()` call so we can lower it to a constant later
-            // NOTE: We get `object.ToString()` from a compilation because we just need to compare symbols and don't need all error recovery of `TryGetSpecialTypeMethod`
-            if (expr is BoundCall { Type.SpecialType: SpecialType.System_String, Method: { Name: "ToString" } method, ReceiverOpt: { Type: NamedTypeSymbol { SpecialType: SpecialType.System_Char } charType, ConstantValueOpt.IsChar: true } } call &&
-                method.GetLeastOverriddenMember(charType) == _compilation.GetSpecialTypeMember(SpecialMember.System_Object__ToString))
-            {
-                expr = call.ReceiverOpt;
-            }
-
             // Is the expression a constant char?  If so, we can
             // simply make it a literal string instead and avoid any 
             // allocations for converting the char to a string at run time.
-            // Similarly if it's a literal null, don't do anything special.
             if (expr is { ConstantValueOpt: { } cv })
             {
                 if (cv.SpecialType == SpecialType.System_Char)
@@ -607,7 +598,8 @@ fallbackStrings:
                 }
                 else if (cv.IsNull)
                 {
-                    return expr;
+                    // Should have been dropped by now.
+                    throw ExceptionUtilities.Unreachable();
                 }
             }
 
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
@@ -30,16 +30,10 @@ namespace Microsoft.CodeAnalysis.CSharp
             }
 
             ArrayBuilder<BoundExpression> arguments;
-            if (unvisitedRight is BoundBinaryOperator rightBinary && IsBinaryStringConcatenation(rightBinary))
+            if (unvisitedRight is BoundBinaryOperator { InterpolatedStringHandlerData: null } rightBinary && IsBinaryStringConcatenation(rightBinary))
             {
-                CollectConcatArguments(rightBinary, left, out arguments);
-
+                CollectAndVisitConcatArguments(rightBinary, left, out arguments);
                 Debug.Assert(ReferenceEquals(arguments[0], left));
-
-                for (var i = 1; i < arguments.Count; i++)
-                {
-                    arguments[i] = VisitExpression(arguments[i]);
-                }
             }
             else
             {
@@ -68,12 +62,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
             // We'll walk the children in a depth-first order, pull all the arguments out, and then visit them. We'll fold any constant arguments as
             // we go, pulling them all into a string literal.
-            CollectConcatArguments(originalOperator, compoundAssignmentLeftRead: null, out var arguments);
-
-            for (int i = 0; i < arguments.Count; i++)
-            {
-                arguments[i] = VisitExpression(arguments[i]);
-            }
+            CollectAndVisitConcatArguments(originalOperator, visitedCompoundAssignmentLeftRead: null, out var arguments);
 
             return RewriteStringConcat(originalOperator.Syntax, arguments);
         }
@@ -202,21 +191,21 @@ fallbackStrings:
             }
         }
 
-        private void CollectConcatArguments(BoundBinaryOperator originalOperator, BoundExpression? compoundAssignmentLeftRead, out ArrayBuilder<BoundExpression> destinationArguments)
+        private void CollectAndVisitConcatArguments(BoundBinaryOperator originalOperator, BoundExpression? visitedCompoundAssignmentLeftRead, out ArrayBuilder<BoundExpression> destinationArguments)
         {
             destinationArguments = ArrayBuilder<BoundExpression>.GetInstance();
             var concatMethods = new WellKnownConcatRelatedMethods(_compilation);
             pushArguments(this, originalOperator, destinationArguments, ref concatMethods);
-            if (compoundAssignmentLeftRead is not null)
+            if (visitedCompoundAssignmentLeftRead is not null)
             {
                 // We don't expect to be able to optimize anything about the compound assignment left read, so we just add it as-is. This assert should be kept in sync
                 // with the cases that can be optimized by the addArgument local function below; if we ever find a case that can be optimized, we may need to consider
                 // whether to do so. The visiting logic in the parent function here depends on only one argument being added for a compound assignment left read, so if
                 // we ever do introduce optimizations here that result in more than one argument being added to destinationArguments, we'll need to adjust that logic.
-                Debug.Assert(compoundAssignmentLeftRead is
+                Debug.Assert(visitedCompoundAssignmentLeftRead is
                     not (BoundCall or BoundConversion { ConversionKind: ConversionKind.Boxing, Type.SpecialType: SpecialType.System_Object, Operand.Type.SpecialType: SpecialType.System_Char })
                     and { ConstantValueOpt: null });
-                destinationArguments.Add(compoundAssignmentLeftRead);
+                destinationArguments.Add(visitedCompoundAssignmentLeftRead);
             }
             destinationArguments.ReverseContents();
 
@@ -249,7 +238,7 @@ fallbackStrings:
                 static bool shouldRecurse(BoundExpression expr, [NotNullWhen(true)] out BoundBinaryOperator? binaryOperator)
                 {
                     binaryOperator = expr as BoundBinaryOperator;
-                    if (IsBinaryStringConcatenation(binaryOperator))
+                    if (IsBinaryStringConcatenation(binaryOperator) && binaryOperator.InterpolatedStringHandlerData is null)
                     {
                         return true;
                     }
@@ -262,6 +251,8 @@ fallbackStrings:
 
                 static void addArgument(LocalRewriter self, BoundExpression argument, ArrayBuilder<BoundExpression> finalArguments, ref WellKnownConcatRelatedMethods wellKnownConcatOptimizationMethods)
                 {
+                    argument = self.VisitExpression(argument);
+
                     if (argument is BoundConversion { ConversionKind: ConversionKind.Boxing, Type.SpecialType: SpecialType.System_Object, Operand: { Type.SpecialType: SpecialType.System_Char } operand })
                     {
                         argument = operand;
@@ -282,6 +273,12 @@ fallbackStrings:
                         {
                             argument = charExpression;
                         }
+                    }
+
+                    // This is `strValue ?? ""`, possibly from a nested binary addition of an interpolated string. We can just directly use the left operand
+                    if (argument is BoundNullCoalescingOperator { LeftOperand: { Type.SpecialType: SpecialType.System_String } left, RightOperand: BoundLiteral { ConstantValueOpt: { IsString: true, RopeValue.IsEmpty: true } } })
+                    {
+                        argument = left;
                     }
 
                     switch (argument.ConstantValueOpt)
--- a/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
+++ b/src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
@@ -38,8 +38,10 @@ namespace Microsoft.CodeAnalysis.CSharp
             else
             {
                 arguments = ArrayBuilder<BoundExpression>.GetInstance();
-                arguments.Add(left);
-                arguments.Add(VisitExpression(unvisitedRight));
+                var concatMethods = new WellKnownConcatRelatedMethods(_compilation);
+                AddConcatArgument(unvisitedRight, argumentAlreadyVisited: false, arguments, ref concatMethods);
+                AddConcatArgument(left, argumentAlreadyVisited: true, arguments, ref concatMethods);
+                arguments.ReverseContents();
             }
 
             return RewriteStringConcat(syntax, arguments);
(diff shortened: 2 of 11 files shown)


src/Compilers/CSharp/Test/Emit2/CodeGen/CodeGenSpanBasedStringConcatTests.cs
```diff
@@ -1479,22 +1504,19 @@ public void M()
...
               .locals init (char V_0,
-                            char V_1,
-                            char V_2)
-              IL_0000:  ldc.i4.s   97
-              IL_0002:  stloc.0
-              IL_0003:  ldloca.s   V_0
-              IL_0005:  newobj     "System.ReadOnlySpan<char>..ctor(ref readonly char)"
+                            char V_1)
+              IL_0000:  ldstr      "a"
+              IL_0005:  call       "System.ReadOnlySpan<char> string.op_Implicit(string)"
```
333fred, line 1509: This flip is going from making a ROS from a single character to loading a constant single-character string and converting it to a ROS. Talking with Stephen Toub, we expect it to be a neutral change; I can make the new code use the character-based path, but that seemed like more code to maintain for no real benefit.

src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenDynamicTests.cs
```diff
@@ -1370,24 +1370,24 @@ .locals init (C.<>c__DisplayClass11_0 V_0, //CS$<>8__locals0
   IL_02b5:  ldloc.0
   IL_02b6:  ldfld      ""dynamic C.<>c__DisplayClass11_0.dyn""
   IL_02bb:  stloc.2
-  IL_02bc:  ldsfld     ""System.Runtime.CompilerServices.CallSite<System.Func<System.Runtime.CompilerServices.CallSite, object, bool>> C.<>o__11.<>p__12""
+  IL_02bc:  ldsfld     ""System.Runtime.CompilerServices.CallSite<System.Func<System.Runtime.CompilerServices.CallSite, object, bool>> C.<>o__11.<>p__9""
```
333fred, line 1373: The changes in this file are just related to the different visit order in the local rewriter changing the order of the fields in the display class. I've manually verified the type IL before and after.

src/Compilers/CSharp/Test/Emit2/CodeGen/CodeGenSpanBasedStringConcatTests.cs
```diff
@@ -586,15 +586,40 @@ static void Main()
...
-        // Moreover, we previously reported these errors anyway, so this is not a behavioral change
-        comp.VerifyEmitDiagnostics(
-            // (13,47): error CS0656: Missing compiler required member 'System.Object.ToString'
-            //     static string M1(string s, char c) => s + c;
-            Diagnostic(ErrorCode.ERR_MissingPredefinedMember, "c").WithArguments("System.Object", "ToString").WithLocation(13, 47),
-            // (14,43): error CS0656: Missing compiler required member 'System.Object.ToString'
-            //     static string M2(string s, char c) => c + s;
-            Diagnostic(ErrorCode.ERR_MissingPredefinedMember, "c").WithArguments("System.Object", "ToString").WithLocation(14, 43));
+        // We don't use object.ToString() or char.ToString() in the final codegen.
+        var verifier = CompileAndVerify(comp, expectedOutput: ExecutionConditionUtil.IsCoreClr ? "sccs" : null, verify: Verification.FailsPEVerify);
```
333fred, line 590: For these missing `ToString` tests, it didn't seem important to me to preserve the existing behavior. The existing behavior was simply a factor of the incremental build-up process of the local rewriter, and while I do think it is good to verify our behavior with non-compliant standard libraries, I don't think it matters that we ensure an error is produced here. If we break it again in some future refactoring, oh well.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
333fred, line 1: I would not try and review diffs here. Treat the old and new code as completely separate; the old code can be used as a reference, but the changes are too extensive to try and actually compare line-by-line.

333fred:
@dotnet/roslyn-compiler, this is ready for review.

333fred:
@dotnet/roslyn-compiler for reviews.

src/Compilers/Core/Portable/Collections/Rope.cs
```diff
@@ -19,6 +19,7 @@ internal abstract class Rope
         public abstract override string ToString();
         public abstract string ToString(int maxLength);
         public abstract int Length { get; }
+        public abstract bool IsEmpty { get; }
```
AlekseyTs, line 22: >public abstract bool IsEmpty { get; } [](http://example.com/codeflow?start=8&length=37)

```public bool IsEmpty => Length == 0;```

File: src/Compilers/Core/Portable/Collections/Rope.cs [hunk 5]
```diff
@@ -116,6 +118,7 @@ private sealed class ConcatRope : Rope
         {
             private readonly Rope _left, _right;
             public override int Length { get; }
+            public override bool IsEmpty => _left.IsEmpty && _right.IsEmpty;
```
AlekseyTs, line 121: >_left.IsEmpty && _right.IsEmpty [](http://example.com/codeflow?start=44&length=31)

It looks like this is going to "visit" every involved ```Rope```, but ```Length``` is already readily precomputed.

src/Compilers/CSharp/Test/Symbol/Symbols/MissingSpecialMember.cs
```diff
@@ -1679,6 +1679,7 @@ static void Main()
 }";
             var compilation = CreateCompilationWithMscorlib461(source);
             compilation.MakeMemberMissing(SpecialMember.System_String__ConcatStringStringStringString);
+            compilation.MakeMemberMissing(SpecialMember.System_String__Concat_4ReadOnlySpans);
```
AlekseyTs, line 1682: >compilation.MakeMemberMissing(SpecialMember.System_String__Concat_4ReadOnlySpans); [](http://example.com/codeflow?start=12&length=82)

Could you please elaborate why changes to this unit-test are necessary (including the change to the ```source```)?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
```diff
@@ -135,7 +141,7 @@ public BoundExpression VisitBinaryOperator(BoundBinaryOperator node, BoundUnaryO
             for (BoundBinaryOperator? current = node; current != null && current.ConstantValueOpt == null; current = current.Left as BoundBinaryOperator)
             {
                 // The regular visit mechanism will handle this.
-                if (current.InterpolatedStringHandlerData is not null || current.OperatorKind is BinaryOperatorKind.Utf8Addition)
+                if (current.InterpolatedStringHandlerData is not null || current.OperatorKind is BinaryOperatorKind.Utf8Addition || IsBinaryStringConcatenation(current))
```
AlekseyTs, line 144: > || IsBinaryStringConcatenation(current) [](http://example.com/codeflow?start=128&length=40)

Do we have a test observing significance of this condition?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
-            Debug.Assert(
-                operatorKind == BinaryOperatorKind.StringConcatenation ||
-                operatorKind == BinaryOperatorKind.StringAndObjectConcatenation ||
-                operatorKind == BinaryOperatorKind.ObjectAndStringConcatenation);
+            Debug.Assert(IsBinaryStringConcatenation(operatorKind));
 
             if (_inExpressionLambda)
             {
-                return RewriteStringConcatInExpressionLambda(syntax, operatorKind, loweredLeft, loweredRight, type);
+                return RewriteStringConcatInExpressionLambda(left, unvisitedRight, operatorKind, syntax, resultType);
```
AlekseyTs, line 29: >return RewriteStringConcatInExpressionLambda(left, unvisitedRight, operatorKind, syntax, resultType); [](http://example.com/codeflow?start=16&length=101)

Is this a "dead" code? I think assignments are not allowed in expression lambdas.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+                charExpression = null;
+                return false;
+            }
 
-        private BoundExpression RewriteStringConcatenationManyExprs(SyntaxNode syntax, ImmutableArray<BoundExpression> loweredArgs)
-        {
-            Debug.Assert(loweredArgs.Length > 4);
-            Debug.Assert(loweredArgs.All(a => a.HasErrors || a.Type is { } && a.Type.IsStringType()));
+            private readonly void InitializeField(ref MethodSymbol? member, SpecialMember specialMember)
+                => member ??= _compilation.GetSpecialTypeMember(specialMember) as MethodSymbol;
```
AlekseyTs, line 388: >member ??= _compilation.GetSpecialTypeMember(specialMember) as MethodSymbol; [](http://example.com/codeflow?start=19&length=76)

I think ```GetSpecialTypeMember``` can return null and we will be calling it over and over again. We probably should use something other than `null` as the sentinel value. For example, ```ErrorMethodSymbol.UnknownMethod```

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
-        /// folds two concat operands into one expression if possible
-        /// otherwise returns null
-        /// </summary>
-        private BoundExpression? TryFoldTwoConcatOperands(BoundExpression loweredLeft, BoundExpression loweredRight)
+        private void AddConcatArgument(BoundExpression argument, bool argumentAlreadyVisited, ArrayBuilder<BoundExpression> finalArguments, ref WellKnownConcatRelatedMethods wellKnownConcatOptimizationMethods)
         {
-            // both left and right are constants
-            var leftConst = loweredLeft.ConstantValueOpt;
-            var rightConst = loweredRight.ConstantValueOpt;
+            if (!argumentAlreadyVisited)
```
AlekseyTs, line 258: >if [](http://example.com/codeflow?start=12&length=2)

Can we assert ```!IsBinaryStringConcatenation(argument)```?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
-            arguments = default;
-            return false;
         }
 
-        /// <summary>
-        /// folds two concat operands into one expression if possible
-        /// otherwise returns null
-        /// </summary>
-        private BoundExpression? TryFoldTwoConcatOperands(BoundExpression loweredLeft, BoundExpression loweredRight)
+        private void AddConcatArgument(BoundExpression argument, bool argumentAlreadyVisited, ArrayBuilder<BoundExpression> finalArguments, ref WellKnownConcatRelatedMethods wellKnownConcatOptimizationMethods)
```
AlekseyTs, line 256: >AddConcatArgument [](http://example.com/codeflow?start=21&length=17)

The name of this method feels too "innocent". 
- It doesn't reflect the fact that it might decompose the passed in ```argument``` into multiple entries in ```finalArguments```
- It doesn't  reflect the fact that it might add nothing to ```finalArguments```
- It doesn't reflect the fact that existing entries in ```finalArguments``` are expected to be in reversed order
- It doesn't reflect the fact that we expect the method to be called for concatenation arguments in reversed order. I.e. the ```argument``` must immediately precede arguments that are already in the ```finalArguments```.

It would be good to come up with a more descriptive name and add a doc comment about the contract, etc.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
 
-            FlattenConcatArg(loweredLeft, leftFlattened);
-            FlattenConcatArg(loweredRight, rightFlattened);
+            return RewriteStringConcat(originalOperator.Syntax, arguments);
+        }
 
-            if (leftFlattened.Any() && rightFlattened.Any())
+        private BoundExpression RewriteStringConcat(SyntaxNode originalSyntax, ArrayBuilder<BoundExpression> visitedArguments)
+        {
+            // The strategy of this rewrite is to pull out all the individual arguments of the concat, visit them as appropriate,
```
AlekseyTs, line 74: >to pull out all the individual arguments of the concat, visit them as appropriate, [](http://example.com/codeflow?start=47&length=82)

It doesn't look like this method is responsible for doing so. Perhaps we should say that we expect the caller to prepare the input this way?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+            // and then construct the final concat all in one go. There are a few different lowering patterns that we take:
+            //
+            // 1. If we're in an expression tree, we have to follow strict binary operator rules. Nothing fancy to do, just do a recursive
```
AlekseyTs, line 77: >If we're in an expression tree, we have to follow strict binary operator rules. Nothing fancy to do, just do a recursive [](http://example.com/codeflow?start=18&length=120)

This doesn't look like something that this method is concerned about.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
-                }
+                // All the arguments were null or the empty string. We can just return a constant empty string.
+                visitedArguments.Free();
+                return _factory.StringLiteral(string.Empty);
+            }
+            else if (visitedArguments is [{ ConstantValueOpt: { IsString: true } or { IsChar: true } } arg])
+            {
+                // We were able to fold a constant, so we can just return that constant.
+                visitedArguments.Free();
+                return arg as BoundLiteral ?? VisitExpression(arg);
```
AlekseyTs, line 97: >VisitExpression(arg) [](http://example.com/codeflow?start=46&length=20)

I am not sure what is going on here. Are we double visiting the ```arg```?

AlekseyTs, line 97: >arg as BoundLiteral [](http://example.com/codeflow?start=23&length=19)

It looks like we might end up with a char literal here. Shouldn't we make a string out of it?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+                        {
+                            concatKind = StringConcatenationRewriteKind.AllStringsOrChars;
+                        }
+                        break;
 
-                case 2:
-                    var left = leftFlattened[0];
-                    var right = leftFlattened[1];
+                    default:
+                        concatKind = StringConcatenationRewriteKind.AllStringsOrObjects;
```
AlekseyTs, line 121: >concatKind = StringConcatenationRewriteKind.AllStringsOrObjects; [](http://example.com/codeflow?start=24&length=64)

It looks like we can get out of the loop after this.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
-            {
-                return loweredOperand;
-            }
-            else
-            {
-                return _factory.Coalesce(loweredOperand, _factory.Literal(""));
-            }
+            AllStrings,
+            AllStringsOrChars,
+            AllStringsOrObjects,
```
AlekseyTs, line 332: >AllStringsOrObjects [](http://example.com/codeflow?start=12&length=19)

"AllStringsCharsOrObjects"?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
 
-            if (leftFlattened.Any() && rightFlattened.Any())
+        private BoundExpression RewriteStringConcat(SyntaxNode originalSyntax, ArrayBuilder<BoundExpression> visitedArguments)
+        {
+            // The strategy of this rewrite is to pull out all the individual arguments of the concat, visit them as appropriate,
+            // and then construct the final concat all in one go. There are a few different lowering patterns that we take:
+            //
+            // 1. If we're in an expression tree, we have to follow strict binary operator rules. Nothing fancy to do, just do a recursive
+            //    rewrite into nested `string.Concat` calls.
+            // 2. If all the added expressions are constants, we can fold them all together and return a single constant.
```
AlekseyTs, line 79: >we can fold them all together [](http://example.com/codeflow?start=62&length=29)

This doesn't look like something that this method is concerned about.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+            // 3. If all the added expressions are strings, then we want to use one of the `string.Concat(string)`-based overloads: if 4 or less,
+            //    we'll use one of the hardcoded overloads. Otherwise, we'll use `string.Concat(string[])`.
+            // 4. If all the added expressions are strings or chars, we can use the `string.Concat(ReadOnlySpan<char>)`-based overloads. If there are
+            //    more than 4 arguments, we will instead fall back to `string.Concat(string[])`.
+            // 5. If all the added expressions are strings or objects, we'll use the `string.Concat(string)`-based overloads, and call ToString on the
```
AlekseyTs, line 84: >If all the added expressions are strings or objects [](http://example.com/codeflow?start=18&length=51)

Perhaps saying "If there are objects among added expressions" will be more accurate

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
-        private bool TryExtractStringConcatArgs(BoundExpression lowered, out ImmutableArray<BoundExpression> arguments)
-        {
-            switch (lowered)
+            // We push these in reverse order to take advantage of the left-recursive nature of the tree and avoid needing a second stack
+            static void pushArguments(LocalRewriter self, BoundBinaryOperator binaryOperator, ArrayBuilder<BoundExpression> arguments, ref WellKnownConcatRelatedMethods concatMethods)
             {
-                case BoundCall boundCall:
-                    var method = boundCall.Method;
-                    if (method.IsStatic && method.ContainingType.SpecialType == SpecialType.System_String)
+                do
```
AlekseyTs, line 217: >do [](http://example.com/codeflow?start=16&length=2)

```while (true)```?

File: src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs [hunk 20]
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+            return sequence;
         }
 
         /// <summary>
         /// Most of the above optimizations are not applicable in expression trees as the operator
         /// must stay a binary operator. We cannot do much beyond constant folding which is done in binder.
         /// </summary>
-        private BoundExpression RewriteStringConcatInExpressionLambda(SyntaxNode syntax, BinaryOperatorKind operatorKind, BoundExpression loweredLeft, BoundExpression loweredRight, TypeSymbol type)
+        private BoundExpression RewriteStringConcatInExpressionLambda(
+            BoundExpression original,
```
AlekseyTs, line 504: >BoundExpression [](http://example.com/codeflow?start=12&length=15)

BoundBinaryOperator?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,549 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+            }
+
+            while (stack.TryPop(out current))
+            {
+                var right = VisitExpression(current.Right);
+
+                currentResult = createBinaryOperator(this, current.OperatorKind, currentResult, right, current.Syntax, current.Type);
+            }
+
+            if (compoundAssignmentLeft is not null)
```
AlekseyTs, line 540: >if (compoundAssignmentLeft is not null) [](http://example.com/codeflow?start=12&length=39)

Assuming that ```original``` can be statically typed to ```BoundBinaryOperator```,
 I think the code above could be simpler/clearer and more similar to
code in other places where we are dealing with recursion on the left.
Something like:
```
var stack = ArrayBuilder<BoundBinaryOperator>.GetInstance();

while (true)
{
    stack.Push(original);

    if (original.Left is BoundBinaryOperator binary && IsBinaryStringConcatenation(binary))
    {
        original = binary;
    }
    else
    {
         break;
    }
}

BoundExpression currentResult = VisitExpression(stack.Peek().Left;);

while (stack.TryPop(out current))
{
    var right = VisitExpression(current.Right);

    currentResult = createBinaryOperator(this, current.OperatorKind, currentResult, right, current.Syntax, current.Type);
}
```

AlekseyTs:
>         private BoundExpression ConvertConcatExprToString(BoundExpression expr)

Are all code paths in this helper still relevant?

---
Refers to: src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs:566 in 849602a. [](commit_id = 849602a019ce18d71f530c6f45f377ef839b0c47, deletion_comment = False)

AlekseyTs:

src/Compilers/CSharp/Test/Symbol/Symbols/MissingSpecialMember.cs
333fred: Sure. We needed to change the source because it used 5 strings, not 4. Previously, since we built from the bottom up, we'd hit the 4-string level and break. Now, however, when we see 5 strings we go straight for `string.Concat(params string[])`, and never check for the 4-string version. I believe I can remove this MakeMemberMissing call though, it's an artifact of an earlier iteration when I changed the compilation target instead, and I forgot to revert this line.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
333fred: Yes. `StringConcatTests.ConcatOneArgWithNullToString` will go through this path and hit one of the `throw ExceptionUtilities.UnexpectedValue()`s below.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
333fred: I didn't realize that, thanks. That will simplify a bunch of code.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
333fred: Good catch, adding `ConstantCharPlusNull` to cover this.

333fred:
@AlekseyTs I believe I've addressed your feedback, thanks.

AlekseyTs (approved):
LGTM (commit 10)

333fred:

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
```diff
@@ -208,7 +214,7 @@ private BoundExpression MakeBinaryOperator(
                     case BinaryOperatorKind.ObjectAndStringConcatenation:
                     case BinaryOperatorKind.StringAndObjectConcatenation:
                     case BinaryOperatorKind.StringConcatenation:
-                        return RewriteStringConcatenation(syntax, operatorKind, loweredLeft, loweredRight, type);
+                        throw ExceptionUtilities.UnexpectedValue(operatorKind);
```
AlekseyTs, line 217: >throw ExceptionUtilities.UnexpectedValue(operatorKind); [](http://example.com/codeflow?start=24&length=55)

It looks like ```MakeBinaryOperator``` helper is used in many places. Would it make sense to adjust the code to handle this code path as a stand-alone operator?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
```diff
@@ -256,7 +262,7 @@ private BoundExpression MakeBinaryOperator(
...
```
AlekseyTs, line 265: >throw ExceptionUtilities.UnexpectedValue(operatorKind); [](http://example.com/codeflow?start=24&length=55)

Similar question here.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_BinaryOperator.cs
333fred: It could be, but I don't think we need to do so right now. The code that I've written presumes that the input has not yet been visited, and `MakeBinaryOperator` is only called on already-visited operators. While we could update the new code to accept both pre-visited and post-visited operations, I think we can save that exercise for when we encounter a case that needs them.

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,558 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+                    else
+                    {
+                        self.VisitAndAddConcatArgumentInReverseOrder(binaryOperator.Left, argumentAlreadyVisited: false, arguments, ref concatMethods);
+                        break;
+                    }
+                }
 
-                        locals.Free();
-                        arguments = unwrappedArgsBuilder.ToImmutableAndFree();
+                static bool shouldRecurse(BoundExpression expr, [NotNullWhen(true)] out BoundBinaryOperator? binaryOperator)
```
jcouv, line 252: nit: I think we try to avoid nested local functions. Consider popping this one out

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,558 @@ namespace Microsoft.CodeAnalysis.CSharp
...
-                if (concatenated != null)
-                {
-                    return _factory.StringLiteral(concatenated);
-                }
+                argument = operand;
             }
 
-            // one or another is null. 
-            if (IsNullOrEmptyStringConstant(loweredLeft))
+            if (argument is BoundCall call)
```
jcouv, line 290: Can this `if` and the one below (for null-coalescing) be `else if` instead?

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
```diff
@@ -14,592 +14,558 @@ namespace Microsoft.CodeAnalysis.CSharp
...
+        /// have been done:
+        /// <list type="number">
+        /// <item>Any consecutive constant strings or chars have been folded.</item>
+        /// <item>Any nested string.Concat calls have had their arguments deconstructed into <paramref name="visitedArguments"/>.</item>
+        /// </list>
+        /// It is not valid to call this method inside an expression tree; that should be handled by a standard recursive rewrite.
+        /// </summary>
+        private BoundExpression CreateStringConcat(SyntaxNode originalSyntax, ArrayBuilder<BoundExpression> visitedArguments)
+        {
+            Debug.Assert(!_inExpressionLambda);
```
jcouv, line 74: nit: Consider asserting that the arguments have either string, char or object type.

jcouv (approved):
LGTM Thanks (iteration 11)

src/Compilers/CSharp/Portable/Lowering/LocalRewriter/LocalRewriter_StringConcat.cs
333fred: I prefer to keep local functions as scoped as possible. The outer functions/methods do not need access to this, so I will keep it where it is.

PR #76957, opened 2025-01-29
Nullable analysis: avoid tracking object creation for reference type without initializer

### LINKED ISSUE #76568: Slow build and uses 16GB of RAM with a large file and nullable enabled
**Version Used**: '4.12.0-3.24570.6 (913fb4ec)'. Language version: 13.0.

**Steps to Reproduce**:

The repro has one large generated file (~160k lines) with `#nullable enable` at the top. Most lines add an entry in a `ConcurrentDictionary`.

1. `git clone https://github.com/meziantou/repro-roslyn-nullable-largefile`
1. `dotnet build`

The problematic file is <https://github.com/meziantou/repro-roslyn-nullable-largefile/blob/main/HstsDomainPolicyCollection.g.cs>

**Expected Behavior**:
The application builds without using more than 16GB of RAM

**Actual Behavior**:
- On my ARM64 machine, the build takes about 40s and uses about 16GB of RAM.
- On GitHub CI, ubuntu-24.04, the build fails. I suspect the machine doesn't allow memory swapping and crash.
- On GitHub CI, windows-2022, it completes but it is very slow (> 10 minutes)
    - https://github.com/meziantou/repro-roslyn-nullable-largefile/actions/runs/12507256671

**Notes**:
Disabling nullables in the generated files seems to _fix_ the issue. It builds on Linux and it's faster on Windows. Note that there is a significant difference between Windows (80s) and Linux (20s), but maybe it's related to GitHub runners.

- With nullable: https://github.com/meziantou/repro-roslyn-nullable-largefile/actions/runs/12507256671
- without nullable (`#nullable disable`): https://github.com/meziantou/repro-roslyn-nullable-largefile/actions/runs/12507199016

I wonder if there is something in Roslyn that could be improved or if I should update the code generator to use another pattern.

cston:
Nullable analysis was creating a `LocalState` slot for each `new()` expression. In a method with many `new()` expressions, the result is large `BitVector` for the `LocalState`, and since the `LocalState` is cloned before each statement, if there are many statements, a large block is copied many times.

The fix here is to avoid creating a slot for `new()` unnecessarily, specifically when the target type is a reference type and there is no object initializer, since there should be no references to the fields of the `new()` instance.

A more general alternative, to avoid generating a large `LocalState` block, might be to free-up and re-use slots when the corresponding expression or local is no longer in scope. That would be a much more significant change though.

Fixes #76568

### DIFF
--- a/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
+++ b/src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
@@ -3926,10 +3926,11 @@ namespace Microsoft.CodeAnalysis.CSharp
             Debug.Assert(isTargetTyped == argumentsCompletion is not null);
 
             var type = node.Type;
-            (int slot, NullableFlowState resultState, Func<TypeSymbol, MethodSymbol?, int>? initialStateInferenceCompletion) = inferInitialObjectState(node, type, constructor, arguments, argumentResults, isTargetTyped);
+            var initializerOpt = node.InitializerExpressionOpt;
+            (int slot, NullableFlowState resultState, Func<TypeSymbol, MethodSymbol?, int>? initialStateInferenceCompletion) =
+                inferInitialObjectState(node, type, constructor, arguments, argumentResults, isTargetTyped, hasObjectInitializer: initializerOpt is { });
 
             Action<int, TypeSymbol>? initializerCompletion = null;
-            var initializerOpt = node.InitializerExpressionOpt;
             if (initializerOpt != null)
             {
                 initializerCompletion = VisitObjectCreationInitializer(slot, type, initializerOpt, delayCompletionForType: isTargetTyped);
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -4012,20 +4013,20 @@ void setAnalyzedNullabilityAsContinuation(
...
 
-                var argumentTypes = argumentResults.SelectAsArray(ar => ar.RValueType);
-
                 int slot = -1;
                 var resultState = NullableFlowState.NotNull;
-                if (type is object)
+                if (type is object &&
+                    (hasObjectInitializer || type.IsStructType()))
                 {
                     slot = GetOrCreatePlaceholderSlot(node);
```
cston, line 4031: A slot was created for each `new()` expression which resulted in a large `BitVector` for the `LocalState`, and since the `LocalState` is cloned before each statement, there were many copies created.

The fix is to avoid creating a slot for a `new()` expression unnecessarily. Specifically, we don't create a slot if the target type is a reference type and there is no object initializer, since there are no references to fields of the instance.

An alternative, to fix this more generally, we could consider freeing up and re-using slots when the corresponding expression or local is no longer in scope. That would be a much more significant change though.

AlekseyTs (approved):
LGTM (commit 1)

cston:

jjonescz (approved):

PR #76963, opened 2025-01-29
Load all razor generator dependencies

jjonescz:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2078732.

### DIFF
--- a/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
+++ b/src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Composition;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


src/Tools/ExternalAccess/Razor/RazorAnalyzerAssemblyResolver.cs
```diff
@@ -37,8 +38,20 @@ internal sealed class RazorAnalyzerAssemblyResolver() : IAnalyzerAssemblyResolve
                         return loadedAssembly;
                     }
 
-                    var assembly = Path.Combine(rootDirectory, $"{assemblyName.Name}.dll");
-                    return compilerContext.LoadFromAssemblyPath(assembly);
+                    // Load all the assemblies, otherwise the compiler ALC could be missing dependencies of the generator.
```
jaredpar, line 41: Oh ... I see. Because there is no actual handler in the compiler ALC that will call back into this.

jjonescz:
The issue seems to be also resolved by https://github.com/dotnet/razor/pull/11394.

PR #76968, opened 2025-01-30
Extensions: source symbol

jcouv:
Corresponding spec update: https://github.com/dotnet/csharplang/pull/9106

### DIFF
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
@@ -196,6 +196,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                 case TypeKind.Array:
                 case TypeKind.Dynamic:
                 case TypeKind.Submission:
+                case TypeKind.Extension:
                     this.LookupMembersInClass(result, type, name, arity, basesBeingResolved, options, originalBinder, diagnose, ref useSiteInfo);
                     break;
 
--- a/src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
+++ b/src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
@@ -196,7 +196,6 @@ namespace Microsoft.CodeAnalysis.CSharp
                 case TypeKind.Array:
                 case TypeKind.Dynamic:
                 case TypeKind.Submission:
-                case TypeKind.Extension:
                     this.LookupMembersInClass(result, type, name, arity, basesBeingResolved, options, originalBinder, diagnose, ref useSiteInfo);
                     break;
 
@@ -206,6 +205,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                 case TypeKind.Pointer:
                 case TypeKind.FunctionPointer:
+                case TypeKind.Extension:
                     result.Clear();
                     break;
 
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -8053,4 +8053,16 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="ERR_ExtensionDisallowsName" xml:space="preserve">
     <value>Extension declarations may not have a name.</value>
   </data>
+  <data name="ERR_ExtensionDisallowsMember" xml:space="preserve">
+    <value>Extension declarations can only include methods or properties</value>
+  </data>
+  <data name="ERR_BadExtensionContainingType" xml:space="preserve">
+    <value>Extensions must be declared in a top-level, non-generic, static class</value>
+  </data>
+  <data name="ERR_ExtensionParameterDisallowsDefaultValue" xml:space="preserve">
+    <value>The receiver parameter of an extension cannot have a default value</value>
+  </data>
+  <data name="ERR_ReceiverParameterOnlyOne" xml:space="preserve">
+    <value>An extension container can only have one receiver parameter</value>
+  </data>
 </root>
--- a/src/Compilers/CSharp/Portable/CSharpResources.resx
+++ b/src/Compilers/CSharp/Portable/CSharpResources.resx
@@ -8054,7 +8054,7 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>Extension declarations may not have a name.</value>
   </data>
   <data name="ERR_ExtensionDisallowsMember" xml:space="preserve">
-    <value>Extension declarations can only include methods or properties</value>
+    <value>Extension declarations can include only methods or properties</value>
   </data>
   <data name="ERR_BadExtensionContainingType" xml:space="preserve">
     <value>Extensions must be declared in a top-level, non-generic, static class</value>
@@ -8063,6 +8063,6 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
     <value>The receiver parameter of an extension cannot have a default value</value>
   </data>
   <data name="ERR_ReceiverParameterOnlyOne" xml:space="preserve">
-    <value>An extension container can only have one receiver parameter</value>
+    <value>An extension container can have only one receiver parameter</value>
   </data>
 </root>
--- a/src/Compilers/CSharp/Portable/Compilation/SyntaxTreeSemanticModel.cs
+++ b/src/Compilers/CSharp/Portable/Compilation/SyntaxTreeSemanticModel.cs
@@ -1402,7 +1402,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             var declarationSpan = extensionDeclaration.Span;
             foreach (var symbol in collection)
             {
-                if (symbol.HasLocationContainedWithin(this.SyntaxTree, declarationSpan, out var wasZeroWidthMatch))
+                if (symbol is NamedTypeSymbol { IsExtension: true } && symbol.HasLocationContainedWithin(this.SyntaxTree, declarationSpan, out var wasZeroWidthMatch))
                 {
                     if (!wasZeroWidthMatch)
                         return (NamedTypeSymbol)symbol;
--- a/src/Compilers/CSharp/Portable/Compilation/SyntaxTreeSemanticModel.cs
+++ b/src/Compilers/CSharp/Portable/Compilation/SyntaxTreeSemanticModel.cs
@@ -1402,7 +1402,7 @@ namespace Microsoft.CodeAnalysis.CSharp
             var declarationSpan = extensionDeclaration.Span;
             foreach (var symbol in collection)
             {
-                if (symbol is NamedTypeSymbol { IsExtension: true } && symbol.HasLocationContainedWithin(this.SyntaxTree, declarationSpan, out var wasZeroWidthMatch))
+                if (symbol is TypeSymbol { IsExtension: true } && symbol.HasLocationContainedWithin(this.SyntaxTree, declarationSpan, out var wasZeroWidthMatch))
                 {
                     if (!wasZeroWidthMatch)
                         return (NamedTypeSymbol)symbol;
--- a/src/Compilers/CSharp/Portable/Compilation/SyntaxTreeSemanticModel.cs
+++ b/src/Compilers/CSharp/Portable/Compilation/SyntaxTreeSemanticModel.cs
@@ -801,6 +801,7 @@ namespace Microsoft.CodeAnalysis.CSharp
                         break;
                     case SyntaxKind.ClassDeclaration:
                     case SyntaxKind.RecordDeclaration:
+                        // PROTOTYPE likely needs work for semantic model
                         {
                             var typeDecl = (TypeDeclarationSyntax)memberDecl;
 
@@ -871,6 +872,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                     case SyntaxKind.ClassDeclaration:
                     case SyntaxKind.RecordDeclaration:
+                        // PROTOTYPE likely needs work for semantic model
                         {
                             var typeDecl = (TypeDeclarationSyntax)memberDecl;
                             return typeDecl.ParameterList is object &&
@@ -1089,6 +1091,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
                 case SyntaxKind.ClassDeclaration:
                 case SyntaxKind.RecordDeclaration:
+                    // PROTOTYPE likely needs work for semantic model
                     {
                         SynthesizedPrimaryConstructor symbol = TryGetSynthesizedPrimaryConstructor((TypeDeclarationSyntax)node);
 
@@ -1378,8 +1381,35 @@ namespace Microsoft.CodeAnalysis.CSharp
         {
             Debug.Assert(declarationSyntax != null);
 
+            if (declarationSyntax is ExtensionDeclarationSyntax extensionDeclaration)
+            {
+                return GetDeclaredExtension(extensionDeclaration);
+            }
+
             var name = declarationSyntax.Identifier.ValueText;
             return GetDeclaredNamedType(declarationSyntax, name);
+        }
+
+        private NamedTypeSymbol GetDeclaredExtension(ExtensionDeclarationSyntax extensionDeclaration)
+        {
+            Debug.Assert(extensionDeclaration != null);
+
+            var container = GetDeclaredTypeMemberContainer(extensionDeclaration);
+            Debug.Assert(container is not null);
+
+            // look for any extension declaration with same declaration location
+            var collection = container.GetMembersUnordered();
+            var declarationSpan = extensionDeclaration.Span;
+            foreach (var symbol in collection)
+            {
+                if (symbol.HasLocationContainedWithin(this.SyntaxTree, declarationSpan, out var wasZeroWidthMatch))
+                {
+                    if (!wasZeroWidthMatch)
+                        return (NamedTypeSymbol)symbol;
+                }
+            }
+
+            return null;
         }
 
         private NamedTypeSymbol GetDeclaredType(DelegateDeclarationSyntax declarationSyntax)
@@ -1621,6 +1651,7 @@ namespace Microsoft.CodeAnalysis.CSharp
 
         private string GetDeclarationName(CSharpSyntaxNode declaration)
         {
+            // PROTOTYPE likely needs work for semantic model
             switch (declaration.Kind())
             {
                 case SyntaxKind.MethodDeclaration:
--- a/src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
+++ b/src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
@@ -287,7 +287,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             Debug.Assert(((Cci.ITypeReference)this).AsTypeDefinition(context) != null);
             NamedTypeSymbol baseType = AdaptedNamedTypeSymbol.BaseTypeNoUseSiteDiagnostics;
 
-            if (AdaptedNamedTypeSymbol.IsScriptClass)
+            if (AdaptedNamedTypeSymbol.IsScriptClass || AdaptedNamedTypeSymbol.IsExtension)
             {
                 // although submission and scripts semantically doesn't have a base we need to emit one into metadata:
                 Debug.Assert((object)baseType == null);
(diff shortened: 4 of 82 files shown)


src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -8053,4 +8053,16 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
   <data name="ERR_ExtensionDisallowsName" xml:space="preserve">
     <value>Extension declarations may not have a name.</value>
   </data>
+  <data name="ERR_ExtensionDisallowsMember" xml:space="preserve">
+    <value>Extension declarations can only include methods or properties</value>
```
AlekseyTs, line 8057: >can only include [](http://example.com/codeflow?start=34&length=16)

"can include only"?

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -8053,4 +8053,16 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
...
+  </data>
+  <data name="ERR_BadExtensionContainingType" xml:space="preserve">
+    <value>Extensions must be declared in a top-level, non-generic, static class</value>
+  </data>
+  <data name="ERR_ExtensionParameterDisallowsDefaultValue" xml:space="preserve">
+    <value>The receiver parameter of an extension cannot have a default value</value>
+  </data>
+  <data name="ERR_ReceiverParameterOnlyOne" xml:space="preserve">
+    <value>An extension container can only have one receiver parameter</value>
```
AlekseyTs, line 8066: >can only have one [](http://example.com/codeflow?start=34&length=17)

"can have only one"?

src/Compilers/CSharp/Portable/Compilation/SyntaxTreeSemanticModel.cs
```diff
@@ -1378,10 +1381,37 @@ private NamedTypeSymbol GetDeclaredType(BaseTypeDeclarationSyntax declarationSyn
...
+
+            var container = GetDeclaredTypeMemberContainer(extensionDeclaration);
+            Debug.Assert(container is not null);
+
+            // look for any extension declaration with same declaration location
+            var collection = container.GetMembersUnordered();
+            var declarationSpan = extensionDeclaration.Span;
+            foreach (var symbol in collection)
+            {
+                if (symbol.HasLocationContainedWithin(this.SyntaxTree, declarationSpan, out var wasZeroWidthMatch))
```
AlekseyTs, line 1405: >symbol [](http://example.com/codeflow?start=20&length=6)

Consider checking that it is an extension first.

src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
```diff
@@ -287,7 +287,7 @@ Cci.ITypeReference Cci.ITypeDefinition.GetBaseClass(EmitContext context)
             Debug.Assert(((Cci.ITypeReference)this).AsTypeDefinition(context) != null);
             NamedTypeSymbol baseType = AdaptedNamedTypeSymbol.BaseTypeNoUseSiteDiagnostics;
 
-            if (AdaptedNamedTypeSymbol.IsScriptClass)
+            if (AdaptedNamedTypeSymbol.IsScriptClass || AdaptedNamedTypeSymbol.IsExtension)
```
AlekseyTs, line 290: >|| AdaptedNamedTypeSymbol.IsExtension [](http://example.com/codeflow?start=53&length=37)

Are we checking presence of `object` somewhere? Otherwise, let's add a PROTOTYPE comment

src/Compilers/CSharp/Portable/Symbols/TypeSymbol.cs
```diff
@@ -555,6 +555,8 @@ public virtual bool IsAnonymousType
         /// </summary>
         internal virtual bool IsNativeIntegerWrapperType => false;
 
+        public virtual bool IsExtension => false;
```
AlekseyTs, line 558: >public virtual bool IsExtension => false; [](http://example.com/codeflow?start=8&length=41)

I think we prefer APIs like that to be abstract

AlekseyTs: Given that its implementation is just a type kind check. Why not make this method non-virtual.

src/Compilers/CSharp/Portable/Symbols/Source/ParameterHelpers.cs
```diff
@@ -135,11 +135,19 @@ private static ImmutableArray<TParameterSymbol> MakeParameters<TParameterSyntax,
...
             foreach (var parameterSyntax in parametersList)
             {
                 if (parameterIndex > lastIndex) break;
 
-                CheckParameterModifiers(parameterSyntax, diagnostics, parsingFunctionPointer, parsingLambdaParams: false, parsingAnonymousMethodParams: false);
+                if (inExtension && parameterIndex > 0)
+                {
+                    // We don't report detailed diagnostics on extension parameters beyond the first
+                    diagnostics.Add(ErrorCode.ERR_ReceiverParameterOnlyOne, parameterSyntax.GetLocation());
+                    diagnostics = BindingDiagnosticBag.Discarded;
```
AlekseyTs, line 147: >diagnostics = BindingDiagnosticBag.Discarded; [](http://example.com/codeflow?start=20&length=45)

```break;```?  It feels like we shouldn't be creating these parameter symbols.

File: src/Compilers/CSharp/Portable/Symbols/Source/ParameterHelpers.cs [hunk 7]
```diff
@@ -675,19 +693,23 @@ public static void ReportParameterErrors(
...
             {
-                Debug.Assert(containingSymbol is null || (containingSymbol is FunctionPointerMethodSymbol or { ContainingType: not null }));
-                // error CS0721: '{0}': static types cannot be used as parameters
-                diagnostics.Add(
-                    ErrorFacts.GetStaticClassParameterCode(containingSymbol?.ContainingType?.IsInterfaceType() ?? false),
-                    syntax.Type?.Location ?? syntax.GetLocation(),
-                    typeWithAnnotations.Type);
+                Debug.Assert(containingSymbol is null || (containingSymbol is FunctionPointerMethodSymbol or { ContainingType: not null }) || inExtension);
+
+                if (!inExtension || isNamed)
```
AlekseyTs, line 705: >isNamed [](http://example.com/codeflow?start=36&length=7)

It is not obvious why presence of the name should make a difference.

src/Compilers/CSharp/Portable/Symbols/Source/SourceComplexParameterSymbol.cs
```diff
@@ -1537,13 +1537,16 @@ private void DoMiscValidation()
...
-                if (IsParams && ParameterSyntax?.Modifiers.Any(SyntaxKind.ParamsKeyword) == true)
+                bool inExtension = this.ContainingSymbol is NamedTypeSymbol { IsExtension: true };
+                // An error is already reported for `params` in extensions
+                if (IsParams && ParameterSyntax?.Modifiers.Any(SyntaxKind.ParamsKeyword) == true && !inExtension)
                 {
                     validateParamsType(diagnostics);
                 }
 
-                if (DeclaredScope == ScopedKind.ScopedValue && !Type.IsErrorOrRefLikeOrAllowsRefLikeType())
+                if (DeclaredScope == ScopedKind.ScopedValue && !Type.IsErrorOrRefLikeOrAllowsRefLikeType() && (!inExtension || this.Ordinal == 0))
```
AlekseyTs, line 1547: >this.Ordinal == 0 [](http://example.com/codeflow?start=127&length=17)

I think we shouldn't be creating extension parameters that fail the ordinal check.

src/Compilers/CSharp/Portable/Symbols/Source/SourceParameterSymbol.cs
```diff
@@ -46,9 +46,10 @@ public static SourceParameterSymbol Create(
             var name = identifier.ValueText;
             var location = new SourceLocation(identifier);
 
-            if (hasParamsModifier && parameterType.IsSZArray())
+            if (hasParamsModifier && parameterType.IsSZArray() && !identifier.IsKind(SyntaxKind.None))
```
AlekseyTs, line 49: > && !identifier.IsKind(SyntaxKind.None) [](http://example.com/codeflow?start=62&length=39)

Can we simply ensure that ```hasParamsModifier``` is never true for an extension parameter? For example, `params` can be completely disallowed as a modifier on it. This way we won't need to special case various checks for extension parameters.

AlekseyTs:
>                         paramsKeyword = modifier;

I think we should ignore presence of `params` on an extension parameter, except for error reporting.

---
Refers to: src/Compilers/CSharp/Portable/Symbols/Source/ParameterHelpers.cs:992 in 79ae7ed. [](commit_id = 79ae7edcbcf8e24171a66040626075069185bc0e, deletion_comment = False)

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
```diff
@@ -196,6 +196,7 @@ protected void LookupMembersInType(LookupResult result, TypeSymbol type, string
                 case TypeKind.Array:
                 case TypeKind.Dynamic:
                 case TypeKind.Submission:
+                case TypeKind.Extension:
```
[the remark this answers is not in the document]
jcouv: Multiple tests crash without some handling here: Attributes_* (during lookup for attribute name),, *__ExplicitInterfaceImplementation (during lookup for `I`), Skeleton (during lookup for `System`).
But none rely on actually finding something yet. I'll change this to find nothing for now (next PR will be focused on lookup scenarios).

src/Compilers/CSharp/Portable/Symbols/Source/SourceMethodSymbolWithAttributes.cs
```diff
@@ -63,6 +63,8 @@ protected SourceMethodSymbol(SyntaxReference syntaxReferenceOpt)
                     return recordDecl;
                 case ClassDeclarationSyntax classDecl:
                     return classDecl;
+                case ExtensionDeclarationSyntax extensionDecl:
+                    return extensionDecl;
```
AlekseyTs, line 67: >return extensionDecl; [](http://example.com/codeflow?start=20&length=21)

What is the scenario when doing this is necessary?

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol_Extension.cs
```diff
@@ -0,0 +1,82 @@
...
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Roslyn.Utilities;
+
+namespace Microsoft.CodeAnalysis.CSharp.Symbols
+{
+    internal partial class SourceNamedTypeSymbol
+    {
+        private ImmutableArray<ParameterSymbol> _lazyExtensionParameters;
```
AlekseyTs, line 13: >_lazyExtensionParameters [](http://example.com/codeflow?start=48&length=24)

I think we should have at most one parameter

src/Compilers/CSharp/Portable/Emitter/Model/NamedTypeSymbolAdapter.cs
```diff
@@ -1043,6 +1048,9 @@ internal virtual IEnumerable<PropertySymbol> GetPropertiesToEmit()
                 }
             }
         }
+
+        internal virtual string ExtensionName
```
AlekseyTs, line 1052: >internal virtual string ExtensionName [](http://example.com/codeflow?start=8&length=37)

Since usage of this method goes beyond this file, I think it should be declared abstract.

AlekseyTs, line 1052: >ExtensionName [](http://example.com/codeflow?start=32&length=13)

It would be good to document what this method is supposed to return.

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
```diff
@@ -1812,6 +1815,16 @@ internal bool IsSimpleProgram
             }
         }
 
+        public override string MetadataName
+        {
+            get
+            {
+                return IsExtension
+                    ? MetadataHelpers.ComposeAritySuffixedMetadataName(Name, Arity, this.ExtensionName)
```
AlekseyTs, line 1823: >Name [](http://example.com/codeflow?start=71&length=4)

What is the ```Name``` value here?

src/Compilers/CSharp/Portable/Symbols/Source/ParameterHelpers.cs
jcouv: If you write `extension(Type x)`, it means we'll have an instance `x`, so the type may not be static.
If you write `extension(Type)`, you may be extending a static or an instance type (there will not be a `this` parameter holding an instance of this type). `extension(StaticType)` should not produce an error.

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
jcouv: It's blank, but I should be using `ExtensionName` here instead of `Name`. Thanks

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol_Extension.cs
jcouv: FWIW, I took a stab at not parsing __arglist, so we could have a non-nullable parameter here, but it snowballed a bit, so I ended up keeping the parser as-is.

src/Compilers/CSharp/Test/Emit3/Semantics/ExtensionTests.cs
```diff
@@ -0,0 +1,2616 @@
...
+using Microsoft.CodeAnalysis.CSharp.Symbols;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Test.Utilities;
+using Roslyn.Test.Utilities;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.CSharp.UnitTests.Semantics;
+
+[CompilerTrait(CompilerFeature.Extensions)]
+public class ExtensionsTests : CompilingTestBase
```
jjonescz, line 17: The file name does not match the class name (singular vs plural Extension/s).

src/Compilers/CSharp/Test/Emit3/Semantics/ExtensionTests.cs
```diff
@@ -0,0 +1,2616 @@
...
+{
+    private static void VerifyTypeIL(CompilationVerifier compilation, string typeName, string expected)
+    {
+        // .Net Core has different assemblies for the same standard library types as .Net Framework, meaning that that the emitted output will be different to the expected if we run them .Net Core
+        // Since we do not expect there to be any meaningful differences between output for .Net Core and .Net Framework, we will skip these tests on .Net Framework
```
jjonescz, line 22: Alternatively, I think it would work if you just replace `[netstandard]` with `[mscorlib]` in the expected string.

src/Compilers/CSharp/Portable/Symbols/Source/ParameterHelpers.cs
```diff
@@ -421,7 +503,8 @@ internal static void CheckParameterModifiers(
             BindingDiagnosticBag diagnostics,
             bool parsingFunctionPointerParams,
             bool parsingLambdaParams,
-            bool parsingAnonymousMethodParams)
+            bool parsingAnonymousMethodParams,
+            bool extensionReceiverParameter)
```
AlekseyTs, line 507: >extensionReceiverParameter [](http://example.com/codeflow?start=17&length=26)

It looks like the last four booleans are all about parameter location and at most one of them can be true at a given time. Consider merging them into a single parameter, probably an enum.

src/Compilers/CSharp/Portable/Symbols/Source/ParameterHelpers.cs
```diff
@@ -662,6 +750,7 @@ public static void ReportParameterErrors(
             SyntaxToken thisKeyword,
             SyntaxToken paramsKeyword,
             int firstDefault,
+            bool isNamed,
```
AlekseyTs, line 753: >isNamed [](http://example.com/codeflow?start=17&length=7)

This parameter looks redundant, we have ```BaseParameterSyntax syntax```.

AlekseyTs:

src/Compilers/CSharp/Portable/Symbols/PublicModel/TypeSymbol.cs
```diff
@@ -26,6 +26,11 @@ protected TypeSymbol(CodeAnalysis.NullableAnnotation nullableAnnotation)
 
         ITypeSymbol ITypeSymbol.WithNullableAnnotation(CodeAnalysis.NullableAnnotation nullableAnnotation)
         {
+            if (UnderlyingTypeSymbol.IsExtension)
+            {
+                throw new System.NotSupportedException();
```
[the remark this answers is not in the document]
jcouv: I don't think it's necessary. Looking at the codebase, only a few instance of NotSupportedException have a message.

src/Compilers/CSharp/Portable/Symbols/Source/SourceDestructorSymbol.cs
```diff
@@ -28,7 +28,7 @@ internal SourceDestructorSymbol(
             bool hasBlockBody = syntax.Body != null;
             bool isExpressionBodied = IsExpressionBodied;
 
-            if (syntax.Identifier.ValueText != containingType.Name)
+            if (syntax.Identifier.ValueText != containingType.Name && !containingType.IsExtension)
```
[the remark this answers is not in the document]
jcouv: See `Member_Finalizer`. It would have a cascading diagnostic without this

src/Compilers/CSharp/Portable/SymbolDisplay/SymbolDisplayVisitor.Types.cs
```diff
@@ -332,54 +332,70 @@ private void AddNameAndTypeArgumentsOrParameters(INamedTypeSymbol symbol)
...
-            var illegalGenericInstantiationSymbol = underlyingTypeSymbol as NoPiaIllegalGenericInstantiationSymbol;
-
-            if (illegalGenericInstantiationSymbol is not null)
+            if (symbol.IsExtension)
             {
-                symbol = illegalGenericInstantiationSymbol.UnderlyingSymbol.GetPublicSymbol();
+                if (Format.CompilerInternalOptions.HasFlag(SymbolDisplayCompilerInternalOptions.UseMetadataMemberNames))
+                {
+                    var extensionIdentifier = underlyingTypeSymbol!.ExtensionName;
+                    Builder.Add(CreatePart(SymbolDisplayPartKind.ModuleName, symbol, extensionIdentifier));
```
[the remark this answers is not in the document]
jcouv: Changed to ClassName, since it's literally the class name. But left a PROTOTYPE marker to review with IDE (since they use this categorization to affect colorization). Thanks

src/Compilers/CSharp/Test/Emit3/Semantics/ExtensionTests.cs
jcouv: That would work too. I followed the same model as some existing tests. I think it's good enough

jjonescz (approved):

AlekseyTs (approved):
LGTM (commit 7)

src/Compilers/CSharp/Test/Emit3/Semantics/ExtensionTests.cs
```diff
@@ -0,0 +1,2616 @@
...
+""";
+        var comp = CreateCompilation(src);
+        comp.VerifyEmitDiagnostics();
+
+        var tree = comp.SyntaxTrees[0];
+        var model = comp.GetSemanticModel(tree);
+        var parameter = tree.GetRoot().DescendantNodes().OfType<ParameterSyntax>().Single();
+        var parameterSymbol = model.GetDeclaredSymbol(parameter);
+        Assert.Null(parameterSymbol);
+        //AssertEx.SetEqual(["MyAttribute"], parameterSymbol.GetAttributes().Select(a => a.ToString()));
```
jcouv, line 2108: Missed this commented out line

Issue #29495, opened 2018-08-24
Override completion should place cursor near arrow


jcouv:
Currently, the cursor is placed at the end of the inserted code (as shown below).

At least in the case of a completion with an arrow body, that isn't most useful, since you won't type more statements in the body and you likely won't leave the inserted code as-is.

I'd propose that the cursor be left between the fat arrow and the inserted code.

This way, you could easily:
- convert the body to a block body,
- delete the body to type what you want.

![override-completion](https://user-images.githubusercontent.com/12466233/44597518-c5272a00-a784-11e8-8f87-15f545e4b586.gif)

Also, I think the override completion should not add a body when there is already one. This happens when you copy&paste one override method, then rename the method name.

sharwell:
> I'd propose that the cursor be left between the fat arrow and the inserted code.

I like this because it enables both wrapping and Use Block Body scenarios. Though we normally wrap these before the arrow not after.

CyrusNajmabadi:
> Though we normally wrap these before the arrow not after.

Note: we really should just have a formatting option for this.  :)       We were hesitant originally because it was a new language feature and we didn't know how it would be used.  At this point the major patterns are known and can be accounted for.

Neme12:
Like @CyrusNajmabadi suggested, I would prefer if this just selected the expression in the body, because it's very likely that user will want to replace that.

> This way, you could easily:
> 
> * convert the body to a block body,
> * delete the body to type what you want.

If we put the caret before the arrow, I'm not sure deleting the body is so easy (it requires moving a few characters right, and then selecting the whole expression). Converting the body to a block body should always be easy using the refactoring: 
![image](https://user-images.githubusercontent.com/11444821/48034018-eaa9ba00-e15d-11e8-9661-45952346a49e.png)

Neme12:
Actually, this issue mentions putting the caret to the right of the arrow:
> I'd propose that the cursor be left between the fat arrow and the inserted code.

I was looking at the PR when writing above comment, which implements care to the left of the arrow. I'm a little confused. @jcouv ?

jcouv:
You're correct, the PR puts the caret to the left of the arrow, so that the second scenario you listed becomes easier (deleting the body). I forgot to update the issue to reflect this. Sorry for the confusion.

There are three scenarios considered so far:
1) insert breakline before arrow,
2) convert the body to a block body,
3) delete the body to type what you want.

And some options:
1) caret on left of arrow (PR)
[...]

PR #76969, opened 2025-01-30
Allow expression body refactorings on non empty selections

ToddGrun:
Currently, all expression/body refactorings only work with an empty selection. I've started looking into https://github.com/dotnet/roslyn/issues/29495 which requests overload completion to set the selection upon commit. Upon commit (and setting the selection), it's desirable to be able to do the expression/body refactoring. This change should allow that.

1st commit == entirety of product code change.
2nd commit == new tests for each of the refactorings with selections
3rd commit == raw string changes in the modified test files

### DIFF
--- a/src/Features/CSharp/Portable/UseExpressionBody/UseExpressionBodyCodeRefactoringProvider.cs
+++ b/src/Features/CSharp/Portable/UseExpressionBody/UseExpressionBodyCodeRefactoringProvider.cs
@@ -58,12 +58,10 @@ internal sealed class UseExpressionBodyCodeRefactoringProvider() : SyntaxEditorB
     public override async Task ComputeRefactoringsAsync(CodeRefactoringContext context)
     {
         var (document, textSpan, cancellationToken) = context;
-        if (textSpan.Length > 0)
-            return;
 
         var position = textSpan.Start;
         var root = await document.GetRequiredSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
-        var node = root.FindToken(position).Parent!;
+        var node = root.FindNode(textSpan);
 
         var containingLambda = node.FirstAncestorOrSelf<LambdaExpressionSyntax>();
         if (containingLambda != null &&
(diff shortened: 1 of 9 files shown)


src/Features/CSharp/Portable/UseExpressionBody/UseExpressionBodyCodeRefactoringProvider.cs
```diff
@@ -58,12 +58,10 @@ internal sealed class UseExpressionBodyCodeRefactoringProvider() : SyntaxEditorB
     public override async Task ComputeRefactoringsAsync(CodeRefactoringContext context)
     {
         var (document, textSpan, cancellationToken) = context;
-        if (textSpan.Length > 0)
-            return;
 
         var position = textSpan.Start;
         var root = await document.GetRequiredSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
-        var node = root.FindToken(position).Parent!;
+        var node = root.FindNode(textSpan);
```
sharwell, line 64: Can we now use the refactoring helper in this code (the code that locates a node of a particular type in th ecurrent context)?

ToddGrun: I've verified that this code path is utilized for both body->expression and expression->body conversions (required setting csharp_style_expression_bodied_methods to true/false to see both ways come through). I think I'd prefer not to use TryGetRelevantNodeAsync is it isn't completely clear to me which type to be using in both cases, and throwing in the lambda expression check right below it.

ToddGrun: @sharwell -- the current change makes sense to me and fits in well with how this method worked before. I'm willing to give the helper a try if you think it fits well here, just would like some clarification on which type(s) would be used on the call(s).

ToddGrun: @sharwell -- really want to move forward on this conversation.

CyrusNajmabadi (approved):

PR #76970, opened 2025-01-30
Partial events and constructors: merge and check symbols

jjonescz:


src/Compilers/CSharp/Portable/Binder/BinderFactory.BinderFactoryVisitor.cs
```diff
@@ -613,6 +613,7 @@ bool checkSymbol(Symbol sym, TextSpan memberSpan, SymbolKind kind, out Symbol re
                             {
                                 MethodSymbol method => (Symbol)method.PartialImplementationPart,
                                 SourcePropertySymbol property => property.PartialImplementationPart,
+                                SourceEventSymbol ev => ev.PartialImplementationPart,
```
RikkiGibson, line 616: I wonder if this should be a helper. Feel free to extract if you find there are several consistent enough usages like this one.

Also, the comment that `member.Locations` includes both parts is not true, feel free to delete it.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3899,6 +3971,50 @@ private static void FixPartialProperty(ArrayBuilder<Symbol> nonTypeMembers, Sour
             Remove(nonTypeMembers, implementation);
         }
 
+        /// <summary>Links together the definition and implementation parts of a partial constructor. Returns a member list which has the implementation part removed.</summary>
```
RikkiGibson, line 3974: Doc comment for this and 'FixPartialEvent' looks out of date.

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -1171,7 +1171,7 @@
     <value>Cannot implicitly convert type '{0}' to '{1}'. An explicit conversion exists (are you missing a cast?)</value>
   </data>
   <data name="ERR_PartialMisplaced" xml:space="preserve">
-    <value>The 'partial' modifier can only appear immediately before 'class', 'record', 'struct', 'interface', or a method or property return type.</value>
+    <value>The 'partial' modifier can only appear immediately before 'class', 'record', 'struct', 'interface', 'event', an instance constructor identifier, or a method or property return type.</value>
```
RikkiGibson, line 1174: Our [learning docs](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/constructors) say: "A constructor is a method with the same name as its type." If you want, we could say "a constructor name" here.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
```diff
@@ -55,6 +57,11 @@ internal SourceEventSymbol(
             bool modifierErrors;
             _modifiers = MakeModifiers(modifiers, isExplicitInterfaceImplementation, isFieldLike, _location, diagnostics, out modifierErrors);
             this.CheckAccessibility(_location, diagnostics, isExplicitInterfaceImplementation);
+
+            if (IsPartial)
+            {
+                ModifierUtils.CheckFeatureAvailabilityForPartialEventsAndConstructors(_location, diagnostics);
```
RikkiGibson, line 63: nit: it seems a little inconsistent that feature availability is checked in constructor for event symbols but during modifier checks for constructor symbols.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
```diff
@@ -780,5 +803,42 @@ private void CheckExplicitImplementationAccessor(MethodSymbol? thisAccessor, Met
...
             }
         }
+
+        internal bool IsPartial => (this.Modifiers & DeclarationModifiers.Partial) != 0;
+
+        /// <summary>
+        /// <see langword="true"/> if this symbol corresponds to a semi-colon body declaration.
+        /// <see langword="false"/> if this symbol corresponds to a declaration with custom <see langword="add"/> and <see langword="remove"/> accessors.
+        /// </summary>
+        protected abstract bool IsFieldLike { get; }
```
RikkiGibson, line 813: Was there a reason that existing member `EventSymbol.HasAssociatedField` wasn't suitable?

src/Compilers/CSharp/Portable/Symbols/Source/SourceFieldLikeEventSymbol.cs
```diff
@@ -144,14 +153,14 @@ public override TypeWithAnnotations TypeWithAnnotations
             get { return _type; }
         }
 
-        public override MethodSymbol AddMethod
+        public override MethodSymbol? AddMethod
         {
-            get { return _addMethod; }
+            get { return PartialImplementationPart?.AddMethod ?? _addMethod; }
```
RikkiGibson, line 158: Something feels off about this to me. Partial property accessors are themselves represented as partial methods in the symbol model. It feels most straightforward for partial event accessors to work in a similar way, even though the accessors on the definition part are not "addressable", e.g. user cannot put attributes on them, and so on.

Mainly I am concerned about consistency of the symbol model, e.g., someone might expect to say `eventSymbol.AddMethod.ContainingSymbol` and get `eventSymbol` as the answer.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
jjonescz: Yes, `SourceFieldLikeEventSymbol` doesn't have `AssociatedField` if it's partial (or abstract or extern). But every partial event is composed of `SourceFieldLikeEventSymbol` and `SourceCustomEventSymbol` and this property is used to distinguish these two parts.

src/Compilers/CSharp/Portable/Symbols/Source/SourceFieldLikeEventSymbol.cs
jjonescz: > Mainly I am concerned about consistency of the symbol model, e.g., someone might expect to say `eventSymbol.AddMethod.ContainingSymbol` and get `eventSymbol` as the answer.

That makes sense. But I don't know how to fix this - do you think we should create a "duplicate" symbol for the same event accessor?

RikkiGibson: Yeah I think you would make a distinct accessor symbol on the definition part, which is a partial method, and let it plunk through the existing partial method machinery, much like partial property accessors do.

But, before moving forward with that, it might be good to raise it with the broader compiler team. @AlekseyTs in case you have any thoughts here.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
RikkiGibson: Ah I think I get it. There's a need for essentially checking which syntactic form was used, which is unrelated to whether event actually has a field. And this is preferred over `is SourceFieldLikeEventSymbol` or other such tests. Thanks!

src/Compilers/CSharp/Portable/Symbols/MemberSymbolExtensions.cs
```diff
@@ -569,9 +573,24 @@ internal static bool IsPartialDefinition(this Symbol member)
...
                 or SourcePropertySymbol { IsPartialDefinition: true }
-                or SourcePropertyAccessorSymbol { IsPartialDefinition: true };
+                or SourcePropertyAccessorSymbol { IsPartialDefinition: true }
+                or SourceConstructorSymbol { IsPartialDefinition: true }
+                or SourceEventSymbol { IsPartialDefinition: true };
         }
 
+#nullable enable
+        internal static Symbol? GetPartialImplementationPart(this Symbol member)
+        {
```
cston, line 583: >{ [](http://example.com/codeflow?start=8&length=1)

`Debug.Assert(member.IsDefinition);`?

src/Compilers/CSharp/Portable/Symbols/Source/SourceConstructorSymbol.cs
```diff
@@ -176,6 +183,15 @@ private void CheckModifiers(MethodKind methodKind, bool hasBody, Location locati
...
                 diagnostics.Add(ErrorCode.ERR_ConstructorInStaticClass, location);
             }
+            else if (IsPartial && !ContainingType.IsPartial())
+            {
+                diagnostics.Add(ErrorCode.ERR_PartialMemberOnlyInPartialClass, location);
+            }
+
+            if (IsPartial)
+            {
+                ModifierUtils.CheckFeatureAvailabilityForPartialEventsAndConstructors(location, diagnostics);
```
cston, line 193: >CheckFeatureAvailabilityForPartialEventsAndConstructors [](http://example.com/codeflow?start=30&length=55)

Does this check result in any diagnostics? When compiling with earlier language versions, I thought we treated `partial` as a type name for a regular method rather than as a constructor modifier.

src/Compilers/CSharp/Portable/Symbols/Source/SourceConstructorSymbol.cs
```diff
@@ -213,5 +229,35 @@ protected override bool IsWithinExpressionOrBlockBody(int position, out int offs
...
+        public override MethodSymbol? PartialDefinitionPart => IsPartialImplementation ? OtherPartOfPartial : null;
+
+        public override MethodSymbol? PartialImplementationPart => IsPartialDefinition ? OtherPartOfPartial : null;
+
+        internal static void InitializePartialConstructorParts(SourceConstructorSymbol definition, SourceConstructorSymbol implementation)
+        {
+            Debug.Assert(definition.IsPartialDefinition);
+            Debug.Assert(implementation.IsPartialImplementation);
+
+            Debug.Assert(definition._otherPartOfPartial is not { } alreadySetImplPart || alreadySetImplPart == implementation);
```
cston, line 253: >alreadySetImplPart == implementation [](http://example.com/codeflow?start=89&length=36)

Consider using `ReferenceEquals()` for this check and the one below.

src/Compilers/CSharp/Portable/Symbols/Source/SourceConstructorSymbol.cs
```diff
@@ -213,5 +229,35 @@ protected override bool IsWithinExpressionOrBlockBody(int position, out int offs
...
+            Debug.Assert(implementation._otherPartOfPartial is not { } alreadySetDefPart || alreadySetDefPart == definition);
+
+            definition._otherPartOfPartial = implementation;
+            implementation._otherPartOfPartial = definition;
+
+            Debug.Assert(definition._otherPartOfPartial == implementation);
```
cston, line 259: >definition._otherPartOfPartial == implementation [](http://example.com/codeflow?start=25&length=48)

Consider using `ReferenceEquals()` for this check and the one below.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
```diff
@@ -780,5 +803,42 @@ private void CheckExplicitImplementationAccessor(MethodSymbol? thisAccessor, Met
...
+        internal override EventSymbol? PartialDefinitionPart => SourcePartialDefinitionPart;
+
+        internal override EventSymbol? PartialImplementationPart => SourcePartialImplementationPart;
+
+        internal static void InitializePartialEventParts(SourceEventSymbol definition, SourceEventSymbol implementation)
+        {
+            Debug.Assert(definition.IsPartialDefinition);
+            Debug.Assert(implementation.IsPartialImplementation);
+
+            Debug.Assert(definition._otherPartOfPartial is not { } alreadySetImplPart || alreadySetImplPart == implementation);
```
cston, line 834: >alreadySetImplPart == implementation [](http://example.com/codeflow?start=89&length=36)

Consider using `ReferenceEquals` for this check and the one below.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
```diff
@@ -780,5 +803,42 @@ private void CheckExplicitImplementationAccessor(MethodSymbol? thisAccessor, Met
...
+            Debug.Assert(implementation._otherPartOfPartial is not { } alreadySetDefPart || alreadySetDefPart == definition);
+
+            definition._otherPartOfPartial = implementation;
+            implementation._otherPartOfPartial = definition;
+
+            Debug.Assert(definition._otherPartOfPartial == implementation);
```
cston, line 840: >_otherPartOfPartial == implementation [](http://example.com/codeflow?start=36&length=37)

Consider using `ReferenceEquals` for this check and the one below.

src/Compilers/CSharp/Portable/Symbols/Source/SourceFieldLikeEventSymbol.cs
```diff
@@ -15,15 +15,15 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
     /// <summary>
     /// This class represents an event declared in source without explicit accessors.
     /// It implicitly has thread safe accessors and an associated field (of the same
-    /// name), unless it does not have an initializer and is either extern or inside
+    /// name), unless it does not have an initializer and is extern, partial, or inside
```
cston, line 18: >partial [](http://example.com/codeflow?start=69&length=7)

Are we creating a `SourceFieldLikeEventSymbol` for the declaration part of a `partial event` such as below? If so, that seems surprising to me, since there is no field implied.
```csharp
partial class C
{
    partial event System.Action E { add { } remove { } }
    partial event System.Action E;
}
```

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3847,6 +3879,46 @@ static bool hasInitializer(SourcePropertySymbol property)
...
+                    (prevConstructor.IsPartialImplementation || (prevConstructor.OtherPartOfPartial is { } otherImplementation && (object)otherImplementation != currentConstructor)))
+                {
+                    // A partial constructor may not have multiple implementing declarations
+                    diagnostics.Add(ErrorCode.ERR_PartialMemberDuplicateImplementation, currentConstructor.GetFirstLocation(), currentConstructor);
+                }
+                else if (currentConstructor.IsPartialDefinition &&
+                    (prevConstructor.IsPartialDefinition || (prevConstructor.OtherPartOfPartial is { } otherDefinition && (object)otherDefinition != currentConstructor)))
+                {
+                    // A partial constructor may not have multiple defining declarations
+                    diagnostics.Add(ErrorCode.ERR_PartialMemberDuplicateImplementation, currentConstructor.GetFirstLocation(), currentConstructor);
```
cston, line 3895: >ERR_PartialMemberDuplicateImplementation [](http://example.com/codeflow?start=46&length=40)

Should this be `ERR_PartialMemberDuplicateDefinition`?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3847,6 +3879,46 @@ static bool hasInitializer(SourcePropertySymbol property)
...
+                    (prevEvent.IsPartialImplementation || (prevEvent.OtherPartOfPartial is { } otherImplementation && (object)otherImplementation != currentEvent)))
+                {
+                    // A partial event may not have multiple implementing declarations
+                    diagnostics.Add(ErrorCode.ERR_PartialMemberDuplicateImplementation, currentEvent.GetFirstLocation(), currentEvent);
+                }
+                else if (currentEvent.IsPartialDefinition &&
+                    (prevEvent.IsPartialDefinition || (prevEvent.OtherPartOfPartial is { } otherDefinition && (object)otherDefinition != currentEvent)))
+                {
+                    // A partial event may not have multiple defining declarations
+                    diagnostics.Add(ErrorCode.ERR_PartialMemberDuplicateImplementation, currentEvent.GetFirstLocation(), currentEvent);
```
cston, line 3915: >ERR_PartialMemberDuplicateImplementation [](http://example.com/codeflow?start=46&length=40)

Should this be `ERR_PartialMemberDuplicateDefinition`?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3847,6 +3879,46 @@ static bool hasInitializer(SourcePropertySymbol property)
                     return property.DeclaredBackingField?.HasInitializer == true;
                 }
             }
+
+            static void mergePartialConstructors(ArrayBuilder<Symbol> nonTypeMembers, SourceConstructorSymbol currentConstructor, SourceConstructorSymbol prevConstructor, BindingDiagnosticBag diagnostics)
+            {
+                if (currentConstructor.IsPartialImplementation &&
+                    (prevConstructor.IsPartialImplementation || (prevConstructor.OtherPartOfPartial is { } otherImplementation && (object)otherImplementation != currentConstructor)))
```
cston, line 3886: >(prevConstructor.OtherPartOfPartial is { } otherImplementation && (object)otherImplementation != currentConstructor) [](http://example.com/codeflow?start=64&length=116)

Are we testing this case?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3847,6 +3879,46 @@ static bool hasInitializer(SourcePropertySymbol property)
...
+                {
+                    // A partial constructor may not have multiple implementing declarations
+                    diagnostics.Add(ErrorCode.ERR_PartialMemberDuplicateImplementation, currentConstructor.GetFirstLocation(), currentConstructor);
+                }
+                else if (currentConstructor.IsPartialDefinition &&
+                    (prevConstructor.IsPartialDefinition || (prevConstructor.OtherPartOfPartial is { } otherDefinition && (object)otherDefinition != currentConstructor)))
```
cston, line 3892: >(prevConstructor.OtherPartOfPartial is { } otherDefinition && (object)otherDefinition != currentConstructor) [](http://example.com/codeflow?start=60&length=108)

Are we testing this case?

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3847,6 +3879,46 @@ static bool hasInitializer(SourcePropertySymbol property)
...
+                else
+                {
+                    FixPartialConstructor(nonTypeMembers, prevConstructor, currentConstructor);
+                }
+            }
+
+            static void mergePartialEvents(ArrayBuilder<Symbol> nonTypeMembers, SourceEventSymbol currentEvent, SourceEventSymbol prevEvent, BindingDiagnosticBag diagnostics)
+            {
+                if (currentEvent.IsPartialImplementation &&
+                    (prevEvent.IsPartialImplementation || (prevEvent.OtherPartOfPartial is { } otherImplementation && (object)otherImplementation != currentEvent)))
```
cston, line 3906: >(prevEvent.OtherPartOfPartial is { } otherImplementation && (object)otherImplementation != currentEvent) [](http://example.com/codeflow?start=58&length=104)

Are we testing this case?

src/Compilers/CSharp/Portable/Symbols/Source/SourceOrdinaryMethodSymbol.cs
```diff
@@ -679,6 +679,17 @@ protected override void MethodChecks(BindingDiagnosticBag diagnostics)
...
             CheckModifiers(MethodKind == MethodKind.ExplicitInterfaceImplementation, _location, diagnostics);
         }
+
+        internal override void AfterAddingTypeMembersChecks(ConversionsBase conversions, BindingDiagnosticBag diagnostics)
+        {
+            base.AfterAddingTypeMembersChecks(conversions, diagnostics);
+
+            if (this.ReturnType?.IsErrorType() == true && this.ReturnType.Name == "partial")
+            {
+                var available = MessageID.IDS_FeaturePartialEventsAndConstructors.CheckFeatureAvailability(diagnostics, DeclaringCompilation, ReturnTypeLocation);
```
cston, line 689: >MessageID.IDS_FeaturePartialEventsAndConstructors.CheckFeatureAvailability(diagnostics, DeclaringCompilation, ReturnTypeLocation) [](http://example.com/codeflow?start=32&length=129)

Does this check result in a language version error for the following when compiling with earlier language versions? If so, that seems incorrect.
```csharp
class C
{
    @partial F() => null;
}
```

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -112,4 +125,1093 @@ class @partial;
...
+            """;
+        // PROTOTYPE: Mismatch between accessibility modifiers of parts should be reported.
+        CreateCompilation(source).VerifyDiagnostics(
+            // (4,1): error CS0122: 'C.E.add' is inaccessible due to its protection level
+            // c.E += () => { };
+            Diagnostic(ErrorCode.ERR_BadAccess, "c.E += () => { }").WithArguments("C.E.add").WithLocation(4, 1),
+            // (5,1): error CS0122: 'C.E.remove' is inaccessible due to its protection level
+            // c.E -= () => { };
+            Diagnostic(ErrorCode.ERR_BadAccess, "c.E -= () => { }").WithArguments("C.E.remove").WithLocation(5, 1));
+    }
```
cston, line 1216: >} [](http://example.com/codeflow?start=4&length=1)

Consider testing mismatched static and instance parts.

src/Compilers/CSharp/Portable/Symbols/Source/SourceFieldLikeEventSymbol.cs
jjonescz: Yes, we do, but that's the same we do for `extern` events and they have no fields either.

src/Compilers/CSharp/Portable/Symbols/Source/SourceConstructorSymbol.cs
jjonescz: Good catch, that's outdated code, I will remove it, thanks.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
jjonescz: Yes, thanks!

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
jjonescz: This particular condition is hit by test `DuplicateDeclarations_01`, but some of the others are not. I will try adding more tests for them.

src/Compilers/CSharp/Portable/Symbols/Source/SourceOrdinaryMethodSymbol.cs
jjonescz: Thanks, fixed.

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
jjonescz: Checks for differences between parts are coming in the next PR, I have a test for static mismatch prepared there.

src/Compilers/CSharp/Portable/Symbols/Source/ModifierUtils.cs
```diff
@@ -223,6 +223,26 @@ internal static void CheckFeatureAvailabilityForStaticAbstractMembersInInterface
             }
         }
 
+#nullable enable
+        internal static void CheckFeatureAvailabilityForPartialEventsAndConstructors(Location location, BindingDiagnosticBag diagnostics)
+        {
+            Debug.Assert(location.SourceTree is not null);
+
+            LanguageVersion availableVersion = ((CSharpParseOptions)location.SourceTree.Options).LanguageVersion;
+            LanguageVersion requiredVersion = MessageID.IDS_FeaturePartialEventsAndConstructors.RequiredVersion();
```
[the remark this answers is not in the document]
jjonescz: They are part of the same speclet, and I assume they will be communicated as one feature of C# 14. But yeah, I don't know, can definitely split the feature in two in the implementation if you think that would be better.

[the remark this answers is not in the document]
cston: No need to change anything at this time.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
cston, line 813: >IsFieldLike [](http://example.com/codeflow?start=32&length=11)

This name feels confusing to me since "field-like" has a specific meaning in the spec as an event that can be used like a field within the containing type, and which is not the case for partial events.

If this property just intended for use in `IsPartialDefinition` and `IsPartialImplementation`, would it make sense to rename this to something more specific? For instance, match `SourcePropertySymbol` and name this `AccessorsHaveImplementation`.

src/Compilers/CSharp/Portable/Symbols/Source/SourceFieldLikeEventSymbol.cs
cston: Ok, thanks. It feels like this type name is not quite correct then, since this doesn't necessarily represent "field-like" events as defined in the spec. But that is perhaps an existing issue.

src/Compilers/CSharp/Portable/Symbols/Source/SourceFieldLikeEventSymbol.cs
```diff
@@ -15,15 +15,15 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
...
     /// an interface, in which case it only has accessors.
     /// </summary>
     internal sealed class SourceFieldLikeEventSymbol : SourceEventSymbol
     {
         private readonly string _name;
         private readonly TypeWithAnnotations _type;
-        private readonly SynthesizedEventAccessorSymbol _addMethod;
-        private readonly SynthesizedEventAccessorSymbol _removeMethod;
+        private readonly SourceEventAccessorSymbol? _addMethod;
```
cston, line 25: >? [](http://example.com/codeflow?start=50&length=1)

Should these fields be not-null?

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventDefinitionAccessorSymbol.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.CodeAnalysis.CSharp.Symbols;
+
+/// <summary>
+/// Accessor of a <see cref="SourceFieldLikeEventSymbol"/> which is a partial definition.
+/// </summary>
+internal sealed class SourceEventDefinitionAccessorSymbol : SourceEventAccessorSymbol
```
cston, line 10: >SourceEventDefinitionAccessorSymbol [](http://example.com/codeflow?start=22&length=35)

Would it make sense to make this a private nested type within `SourceFieldLikeEventSymbol`?

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventDefinitionAccessorSymbol.cs
```diff
@@ -0,0 +1,40 @@
...
+            syntaxReference: ev.SyntaxReference,
+            location: ev.Location,
+            explicitlyImplementedEventOpt: null,
+            aliasQualifierOpt: null,
+            isAdder: isAdder,
+            isIterator: false,
+            isNullableAnalysisEnabled: ev.DeclaringCompilation.IsNullableAnalysisEnabledIn(ev.CSharpSyntaxNode),
+            isExpressionBodied: false)
+    {
+        CheckFeatureAvailabilityAndRuntimeSupport(ev.CSharpSyntaxNode, ev.Location, hasBody: false, diagnostics: diagnostics);
```
cston, line 27: >CheckFeatureAvailabilityAndRuntimeSupport [](http://example.com/codeflow?start=8&length=41)

Are we testing cases where this check reports diagnostics?

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventDefinitionAccessorSymbol.cs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public override Accessibility DeclaredAccessibility => AssociatedEvent.DeclaredAccessibility;
+
+    public override bool IsImplicitlyDeclared => true;
+
+    internal override bool GenerateDebugInfo => true;
```
cston, line 34: >GenerateDebugInfo => true [](http://example.com/codeflow?start=27&length=25)

Is this correct? If we are generating code based on this accessor symbol, what if the event is marked `extern`?

jjonescz: Looks like this is never accessed on partial definitions, it is only accessed on partial implementations by the emit layer. So in theory this could be changed to throw unreachable exception if you think that would be better.

Generally `true` seems correct here - this accessor symbol represents a user-defined method and GenerateDebugInfo controls emit of sequence points and Debug-style IL which we want I think for all user-defined methods. 

Note that property accessors also return constant `true` for this.

Regarding `extern`, those have empty bodies, so GenerateDebugInfo should have no effect there.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventDefinitionAccessorSymbol.cs
jjonescz: Adding test `InInterface`.

src/Compilers/CSharp/Portable/Emitter/Model/MethodSymbolAdapter.cs
```diff
@@ -164,7 +164,11 @@ private Cci.IMethodDefinition ResolvedMethodImpl(EmitContext context)
             if (AdaptedMethodSymbol.IsDefinition && // can't be generic instantiation
                 AdaptedMethodSymbol.ContainingModule == moduleBeingBuilt.SourceModule) // must be declared in the module we are building
             {
-                Debug.Assert((object)AdaptedMethodSymbol.PartialDefinitionPart == null); // must be definition
+                if (AdaptedMethodSymbol.PartialDefinitionPart is { } definition)
+                {
+                    return definition.GetCciAdapter().ResolvedMethodImpl(context);
```
jjonescz, line 169: This change was needed to avoid crash in the sequence point test (added in the same commit). Similar test would also crash for partial methods/properties. Basically the VerifyMethodBody calls GetResolvedMethod on the implementation Cci symbol - that probably doesn't happen normally during emit, that's why the assert was fine for normal scenarios.

cston (approved):

jjonescz:
@RikkiGibson for another look, thanks

jjonescz:

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -1208,13 +1403,12 @@ void M()
                 }
             }
             """;
-        // PROTOTYPE: Mismatch between accessibility modifiers of parts should be reported.
```
RikkiGibson: I assume this is still tracked to be completed in a future PR.

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -678,10 +786,36 @@ partial class C
...
+                .class private auto ansi beforefieldinit C
+                    extends [mscorlib]System.Object
+                {
+                    // Methods
+                    .method private hidebysig specialname 
+                        instance void add_E (
+                            class [mscorlib]System.Action 'value'
+                        ) cil managed 
+                    {
+                    } // end of method C::add_E
```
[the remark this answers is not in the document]
jjonescz: I think they have only signatures in the metadata, not bodies. The IL decompiler shows it like this though.

src/Compilers/CSharp/Portable/Emitter/Model/MethodSymbolAdapter.cs
AlekseyTs: This change doesn't look right. I think the API is not supposed to return a different instance, it should either return an original instance, or null. Basically, for a pair of partial declarations, we should never "create" two different adapters, translation layer should make sure of that. Symbols should never be casted to CCI interfaces, they should be translated.

AlekseyTs: If I remember correctly, PEModuleBuilder performs the translation. There are APIs called "Translate..."

jjonescz: Okay, I think I can skip the failing test for now and look at this separately.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol.cs
```diff
@@ -3903,21 +3905,35 @@ static void mergePartialConstructors(ArrayBuilder<Symbol> nonTypeMembers, Source
...
                 else
                 {
+                    mergeAccessors(nonTypeMembers, (SourceEventAccessorSymbol?)currentEvent.AddMethod, (SourceEventAccessorSymbol?)prevEvent.AddMethod);
+                    mergeAccessors(nonTypeMembers, (SourceEventAccessorSymbol?)currentEvent.RemoveMethod, (SourceEventAccessorSymbol?)prevEvent.RemoveMethod);
                     FixPartialEvent(nonTypeMembers, prevEvent, currentEvent);
                 }
+
+                static void mergeAccessors(ArrayBuilder<Symbol> nonTypeMembers, SourceEventAccessorSymbol? currentAccessor, SourceEventAccessorSymbol? prevAccessor)
+                {
+                    if (currentAccessor?.IsPartialImplementation == true)
```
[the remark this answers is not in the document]
jjonescz: Yes, that's correct.

src/Compilers/CSharp/Portable/Symbols/Source/SourceFieldLikeEventSymbol.cs
```diff
@@ -200,5 +203,40 @@ internal override void ForceComplete(SourceLocation? locationOpt, Predicate<Symb
...
+                    @event: ev,
+                    syntaxReference: ev.SyntaxReference,
+                    location: ev.Location,
+                    explicitlyImplementedEventOpt: null,
+                    aliasQualifierOpt: null,
+                    isAdder: isAdder,
+                    isIterator: false,
+                    isNullableAnalysisEnabled: ev.DeclaringCompilation.IsNullableAnalysisEnabledIn(ev.CSharpSyntaxNode),
+                    isExpressionBodied: false)
+            {
```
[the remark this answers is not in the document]
jjonescz: Thanks, turns out that wasn't true for `extern` events (because the `extern` partial part is the implementation part), so I fixed that as well.

cston (approved):

RikkiGibson (approved):

PR #76983, opened 2025-01-31
Change override completion to select text after updating the buffer.

ToddGrun:
This (along with https://github.com/dotnet/roslyn/pull/76969) will allow a user to commit an override completion and easily replace the default text (and still use the expression body fixers).

This required the CompletionChange class to now allow the data to represent a selection instead of just a cursor position.

Addresses https://github.com/dotnet/roslyn/issues/29495


ToddGrun:
@genlu -- can you take a look at your convenience? Thanks!

src/Features/CSharp/Portable/Completion/CompletionProviders/CompletionUtilities.cs
```diff
@@ -188,30 +188,53 @@ public static SyntaxNode GetTargetCaretPositionForMethod(BaseMethodDeclarationSy
...
+            return methodDeclaration.ExpressionBody.Expression.Span;
+        }
+        else if (methodDeclaration.Body is not null)
+        {
+            // select the last statement in the method
+            return methodDeclaration.Body.Statements.Last().Span;
+        }
+        else
+        {
+            return methodDeclaration.Span;
```
genlu, line 205: Is this path reachable?

genlu (approved):
:shipit:

src/Features/CSharp/Portable/Completion/CompletionProviders/CompletionUtilities.cs
ToddGrun: I don't think so, but I wasn't sure. I saw some places that throw under that assumption and others that handle that case. I decided to be safe and just return the method decl's span if this does happen.

ToddGrun:
@dibarbet -- FYI: had some merge conflicts with some recent changes of yours, think all is merged fine now

dibarbet (approved):
merge looks good

PR #76984, opened 2025-01-31
Update PR val pipeline to restore internal tools in build (match main)

dibarbet:
restore internal tools was moved out of yaml in official build, also needs to happen with PR val build
See https://github.com/dotnet/roslyn/pull/76911

pr val build (just needs to get past build step) - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10928060&view=logs&j=d8f66498-7756-5707-3b43-f2312930a35f&t=bf681905-b801-52f4-803f-be09a0d10104


akhera99 (approved):
build passes 👍

PR #76991, opened 2025-01-31
Support declaration of instance increment operators

AlekseyTs:


src/Compilers/CSharp/Portable/Symbols/Source/SourceUserDefinedOperatorSymbolBase.cs
```diff
@@ -52,6 +52,29 @@ protected SourceUserDefinedOperatorSymbolBase(
 
             this.CheckUnsafeModifier(declarationModifiers, diagnostics);
 
+            bool isIncrementDecrement = syntax is OperatorDeclarationSyntax { OperatorToken.RawKind: (int)SyntaxKind.PlusPlusToken or (int)SyntaxKind.MinusMinusToken };
```
333fred, line 55: Nit: this pattern is done 3 times in this file, consider extracting a helper.

src/Compilers/CSharp/Portable/CSharpResources.resx
```diff
@@ -8047,4 +8047,16 @@ To remove the warning, you can use /reference instead (set the Embed Interop Typ
...
     <value>'InterceptsLocationAttribute(string, int, int)' is not supported. Move to 'InterceptableLocation'-based generation of these attributes instead. (https://github.com/dotnet/roslyn/issues/72133)</value>
   </data>
+  <data name="IDS_FeatureUserDefinedCompoundAssignmentOperators" xml:space="preserve">
+    <value>user-defined compound assignment operators</value>
+  </data>
+  <data name="ERR_OperatorsMustBePublic" xml:space="preserve">
+    <value>User-defined operator '{0}' must be declared public</value>
+  </data>
+  <data name="ERR_BadIncrementOpArgs" xml:space="preserve">
+    <value>Overloaded instance increment operator '{0}' takes no parameters</value>
```
333fred, line 8057: Consider using the imperative form here: _must_ take no parameters, for example. Right now, it's unclear from the wording how many parameters are supposed to be present.

src/Compilers/CSharp/Portable/Symbols/Source/SourceUserDefinedOperatorSymbolBase.cs
```diff
@@ -52,6 +52,29 @@ protected SourceUserDefinedOperatorSymbolBase(
...
+                    if (parameterCount is not (1 or 2))
+                    {
+                        diagnostics.Add(ErrorCode.ERR_BadUnOpArgs, this.GetFirstLocation(), SyntaxFacts.GetText(((OperatorDeclarationSyntax)syntax).OperatorToken.Kind()));
+                    }
+                }
+                else
+                {
+                    Binder.CheckFeatureAvailability(syntax, MessageID.IDS_FeatureUserDefinedCompoundAssignmentOperators, diagnostics, ((OperatorDeclarationSyntax)syntax).OperatorToken.GetLocation());
+
+                    if (parameterCount is not (0 or 2))
```
333fred, line 71: What's the reason for allowing 2 here? I would think that we should prefer giving errors about the specific operator the user typed out, rather than presuming the user meant to type a binary operator instead.

src/Compilers/CSharp/Test/Emit3/Symbols/UserDefinedCompoundAssignmentOperatorsTests.cs
```diff
@@ -0,0 +1,2802 @@
...
+                Assert.False(m.IsOverride);
+                Assert.True(m.HasSpecialName);
+                Assert.False(m.HasRuntimeSpecialName);
+                Assert.Equal(Accessibility.Public, m.DeclaredAccessibility);
+            }
+        }
+
+        [Theory]
+        [CombinatorialData]
+        public void Increment_022_VirtualIsOptionalInInterface([CombinatorialValues("++", "--")] string op)
```
333fred, line 827: Got to here today, will pick up again tomorrow.

src/Compilers/CSharp/Portable/Symbols/Source/SourceUserDefinedOperatorSymbolBase.cs
AlekseyTs: > What's the reason for allowing 2 here? I would think that we should prefer giving errors about the specific operator the user typed out, rather than presuming the user meant to type a binary operator instead.

Parser reports an error for the 2 parameters case. I don't want to move all that parameter checking out of the parser, so I just made the smallest change to let "might be a success" cases through in the parser.

333fred (approved):

AlekseyTs:

cston (approved):

PR #77044, opened 2025-02-05
Consume mibc data from VSCode profiling runs in official builds

dibarbet:
Should help resolve https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2283633/
 
We now produce mibc profiling data for the C# extension in a dnceng/internal pipeline that runs every night.  This updates Roslyn to consume pass the mibc to crossgen when building the language server.  This is adapted from similar support in devkit.

Test build - https://dnceng.visualstudio.com/internal/_build/results?buildId=2635982&view=results

Crossgen successfully runs in CI:
image

image

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -16,6 +16,11 @@ resources:
     type: git
     name: 1ESPipelineTemplates/1ESPipelineTemplates
     ref: refs/tags/release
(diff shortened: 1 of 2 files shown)


dibarbet:
going to hold off merging this just a bit until the .net9 changes go through.

PR #77061, opened 2025-02-05
Detect data section string literal hash collisions

jjonescz:


src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -350,7 +353,18 @@ internal MappedField GetOrAddDataField(ImmutableArray<byte> data, ushort alignme
...
+            (@this, ImmutableCollectionsMarshal.AsImmutableArray(data), diagnostics));
+
+            var previousText = @this._dataSectionStringLiteralNames.GetOrAdd(holder.Name, text);
+            if (previousText != text)
+            {
+                // If there is a hash collision, we cannot fallback to normal string literal emit strategy
+                // because the selection of which literal would get which emit strategy would not be deterministic.
+                var messageProvider = @this.ModuleBuilder.CommonCompilation.MessageProvider;
+                diagnostics.Add(messageProvider.CreateDiagnostic(messageProvider.ERR_DataSectionStringLiteralHashCollision, syntaxNode.GetLocation(), previousText));
+            }
```
ToddGrun, line 365: Curious whether this block could go inside the _dataSectionStringLiteralTypes.GetOrAdd call.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
```diff
@@ -3175,6 +3175,27 @@ .maxstack  1
...
+                System.Console.Write("a");
+                System.Console.Write("b");
+                System.Console.Write("aa");
+                """;
+            CreateCompilation(source,
+                parseOptions: TestOptions.Regular.WithFeature("experimental-data-section-string-literals", "0"))
+                .VerifyEmitDiagnostics(emitOptions,
+                    // (3,22): error CS9274: Cannot emit this string literal into the data section because it has XXHash128 collision with another string literal: a
+                    // System.Console.Write("aa");
+                    Diagnostic(ErrorCode.ERR_DataSectionStringLiteralHashCollision, @"""aa""").WithArguments("a").WithLocation(3, 22));
```
cston, line 3196: >ERR_DataSectionStringLiteralHashCollision [](http://example.com/codeflow?start=41&length=41)

How should I user work around this issue if they hit it? By disabling the feature?

src/Compilers/VisualBasic/Portable/Errors/MessageProvider.vb
```diff
@@ -594,6 +594,12 @@ Namespace Microsoft.CodeAnalysis.VisualBasic
             End Get
         End Property
 
+        Public Overrides ReadOnly Property ERR_DataSectionStringLiteralHashCollision As Integer
```
cston, line 597: >ERR_DataSectionStringLiteralHashCollision [](http://example.com/codeflow?start=43&length=41)

Are we testing this from VB?

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -333,24 +336,33 @@ internal MappedField GetOrAddDataField(ImmutableArray<byte> data, ushort alignme
...
                 Cci.IMethodDefinition bytesToStringHelper = @this.GetOrSynthesizeBytesToStringHelper(diagnostics);
 
+                var previousText = @this._dataSectionStringLiteralNames.GetOrAdd(name, text);
+                if (previousText != text)
+                {
+                    // If there is a hash collision, we cannot fallback to normal string literal emit strategy
+                    // because the selection of which literal would get which emit strategy would not be deterministic.
+                    var messageProvider = @this.ModuleBuilder.CommonCompilation.MessageProvider;
+                    diagnostics.Add(messageProvider.CreateDiagnostic(messageProvider.ERR_DataSectionStringLiteralHashCollision, syntaxNode.GetLocation(), previousText));
+                }
```
cston, line 356: >} [](http://example.com/codeflow?start=16&length=1)

`return null;`?

src/Compilers/Core/Portable/Emit/EmitOptions.cs
```diff
@@ -297,7 +299,8 @@ public bool Equals(EmitOptions? other)
                 IncludePrivateMembers == other.IncludePrivateMembers &&
                 InstrumentationKinds.NullToEmpty().SequenceEqual(other.InstrumentationKinds.NullToEmpty(), (a, b) => a == b) &&
                 DefaultSourceFileEncoding == other.DefaultSourceFileEncoding &&
-                FallbackSourceFileEncoding == other.FallbackSourceFileEncoding;
+                FallbackSourceFileEncoding == other.FallbackSourceFileEncoding &&
+                TestOnly_DataToHexViaXxHash128 == other.TestOnly_DataToHexViaXxHash128;
```
cston, line 303: >TestOnly_DataToHexViaXxHash128 [](http://example.com/codeflow?start=16&length=30)

Do we need to include this in `Equals()` and `GetHashCode()`? We're not doing that for `_testOnly_AllowLocalStateTracing`.

src/Compilers/CSharp/Test/Emit/Emit/EmitMetadataTests.cs
jjonescz: Yes, or changing the threshold to exclude one of the conflicting strings.

Or they should go buy a lottery ticket since they seem to be very lucky to hit a xxh128 collision :-D

src/Compilers/VisualBasic/Portable/Errors/MessageProvider.vb
jjonescz: This cannot be hit from VB, the feature is not implemented there (yet?). I guess I can change this to throw Unreachable exception then.

src/Compilers/Core/Portable/Emit/EmitOptions.cs
jjonescz: We don't, I thought it might be a safer option, but I can remove it.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
jjonescz: Then the dictionary would contain `null` values, which would need to be handled elsewhere. I'd prefer to leave this as it is (note the emit would work even without the error, just the emitted binary would contain two types with the same name).

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -333,24 +336,33 @@ internal MappedField GetOrAddDataField(ImmutableArray<byte> data, ushort alignme
...
 
                 Cci.IMethodDefinition bytesToStringHelper = @this.GetOrSynthesizeBytesToStringHelper(diagnostics);
 
+                var previousText = @this._dataSectionStringLiteralNames.GetOrAdd(name, text);
+                if (previousText != text)
+                {
+                    // If there is a hash collision, we cannot fallback to normal string literal emit strategy
+                    // because the selection of which literal would get which emit strategy would not be deterministic.
+                    var messageProvider = @this.ModuleBuilder.CommonCompilation.MessageProvider;
+                    diagnostics.Add(messageProvider.CreateDiagnostic(messageProvider.ERR_DataSectionStringLiteralHashCollision, syntaxNode.GetLocation(), previousText));
```
AlekseyTs, line 355: >previousText [](http://example.com/codeflow?start=154&length=12)

The string could be quite long. Would it make sense to truncate it?

cston (approved):

AlekseyTs (approved):
LGTM (commit 4)

jjonescz:
Thinking more about this, I conclude it's better to not emit an error.

1. There are actually two potential collisions - one for the generated `<S>` type names and one for the generated data field. Currently this PR only handles the former one.
2. It seems that those collisions don't matter much - the code can be emitted and run without problems since the IL references types and fields by tokens, not names.
3. Emitting an error for a collision reintroduces the problem that the feature aims to solve - if a user hits the error, they cannot compile anymore.
4. Perhaps we could emit a warning? I guess some metadata inspection tools could have problems with the binary if there are duplicate names. But that doesn't seem to be such a big issue and hence could be left as a future work.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
```diff
@@ -333,24 +336,33 @@ internal MappedField GetOrAddDataField(ImmutableArray<byte> data, ushort alignme
...
 
                 MappedField dataField = @this.GetOrAddDataField(data, alignment: 1);
 
                 Cci.IMethodDefinition bytesToStringHelper = @this.GetOrSynthesizeBytesToStringHelper(diagnostics);
 
+                var previousText = @this._dataSectionStringLiteralNames.GetOrAdd(name, text);
+                if (previousText != text)
+                {
+                    // If there is a hash collision, we cannot fallback to normal string literal emit strategy
+                    // because the selection of which literal would get which emit strategy would not be deterministic.
```
jkotas, line 353: Would it be possible to delay the assignment of the type name that gets emitted into the binary until after the typedef token is assigned? Assuming that typedef tokens are deterministic, the name generated from the typedef token would be deterministic as well. Also, the unique names generated from the typedef tokens can be a lot shorter than the hash.

jkotas:
It is an error to have duplicate type names in an assembly. From ECMA-335: "There shall be no duplicate rows in the TypeDef table, based on TypeNamespace+TypeName (unless this is a nested type - see below) [ERROR]". The runtime behavior for malformed binaries is undefined. I understand why it happens to work fine in the current runtime. 

I do not think it is a good idea for the compiler to generate malformed binaries silently even if it happens to work at the moment. It is better to produce an error. We have number of similar corner-case situations where the user needs to alter their code to workaround the internal compiler limitations. For example, very complex expression may fail to compile and users need to alter their code to make it work.

I have checked the behavior of a few tools on duplicate type names: ildasm/ilasm roundtrip fails, native aot compilation happens to handle it gracefully. I would not be surprised if we find a tool with silent bad codegen for malformed input with duplicate type names.

src/Compilers/Core/Portable/CodeGen/PrivateImplementationDetails.cs
jjonescz: It might be possible, but I'm not sure the compiler's internal object model is ready for that, the Name of the symbol is likely used in other places before the typedef token is assigned.

Chuck has suggested another alternative where we would collect the string literals during binding, sort them by length and content, and then assign indices to them (and names based on that) just before emit.

Also, as I mentioned above, currently we share the machinery for synthesizing data fields with array initializers and u8 literals, and these fields are named using sha256. So ideally we would change those too so they also get names based on indices instead of hashes.

I can add that to the spec as future work and go with an error for now.

PR #77072, opened 2025-02-05
 Pass the ITestOutputHelper to the MSBuildWorkspace logger

jasonmalinowski:
This allows us to get logs to help diagnose failing tests. The bulk of this PR is refactoring out the existing logger we had for this to reuse it better.


dibarbet (approved):

src/Workspaces/CoreTestUtilities/Logging/TestOutputLogger.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Logging;
+using Xunit.Abstractions;
+
+namespace Microsoft.CodeAnalysis.LanguageServer.UnitTests;
+
```
JoeRobich, line 9: Delete this file

JoeRobich (approved):

src/Workspaces/CoreTestUtilities/Logging/TestOutputLogger.cs
jasonmalinowski: Fixed, good catch.

src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer.UnitTests/Microsoft.CodeAnalysis.LanguageServer.UnitTests.csproj
```diff
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Compilers\Test\Core\Microsoft.CodeAnalysis.Test.Utilities.csproj" />
+    <ProjectReference Include="..\..\Workspaces\CoreTestUtilities\Microsoft.CodeAnalysis.Workspaces.Test.Utilities.csproj" />
```
JoeRobich, line 20: I think we should just link the TestOutputLoggerProvider class in instead of bringing the whole project reference. The Utilities project references Remote.ServiceHub, which references Razor EA, which references EditorFeatures and everything that brings in.

jasonmalinowski: > The Utilities project references Remote.ServiceHub, which references Razor EA, which references EditorFeatures and everything that brings in.

That seems like the real problem here, since this is our "most simple" test helpers library for all workspace layer stuff. Our file linking as it is for helpers is already quite a mess and I'd like to not make it worse. I'll take a quick look to see if that can be trimmed out.

JoeRobich (approved):

PR #77078, opened 2025-02-06
Revert "System.CommandLine update"

adamsitnik:
Reverts dotnet/roslyn#76948

**Edit:** What's the reason for the revert?

Short answer: chicken-and-egg problem with the update in SDK and NuGet.Client repos that will be easier to solve in April.

Long answer: The breaking change introduced in System.CommandLine requires an update to SDK, which uses both System.CommandLine and NuGet.CommandLine.XPlat (a NuGet package from NuGet.Client repo that also relies on System.CommandLine). NuGet.Client repo uses a hardcoded version of SDK (9.0.2) to run their integration tests, which is using an older version of System.CommandLine and fails at runtime (because the types have been renamed). The NuGet.Client repo release schedule is not tied to .NET, but to VS, so their dev branch updates flow to both SDK/main and SDK/release/9.0.x.  Unblocking the flow would require them to create dedicated branches (and cause a lot of work to sync all the changes), or use a workaround with multi-targeting (https://github.com/NuGet/NuGet.Client/pull/6252). It was decided that the simplest solution will be to just wait until NuGet.Client creates a new branch for next VS release in April.

cc @jaredpar @ViktorHofer

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -13,14 +13,14 @@
       <Sha>3f926184cacc477e3ddccd1eb4f3ebcb9a2b81a5</Sha>
       <SourceBuild RepoName="source-build-reference-packages" ManagedOnly="true" />
     </Dependency>
-    <Dependency Name="System.CommandLine" Version="2.0.0-beta4.25072.1">
+    <Dependency Name="System.CommandLine" Version="2.0.0-beta4.24528.1">
       <Uri>https://github.com/dotnet/command-line-api</Uri>
-      <Sha>060374e56c1b2e741b6525ca8417006efb54fbd7</Sha>
+      <Sha>feb61c7f328a2401d74f4317b39d02126cfdfe24</Sha>
     </Dependency>
     <!-- Intermediate is necessary for source build. -->
-    <Dependency Name="Microsoft.SourceBuild.Intermediate.command-line-api" Version="0.1.607201">
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.command-line-api" Version="0.1.552801">
       <Uri>https://github.com/dotnet/command-line-api</Uri>
-      <Sha>060374e56c1b2e741b6525ca8417006efb54fbd7</Sha>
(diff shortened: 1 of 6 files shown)


eng/Version.Details.xml
```diff
@@ -13,14 +13,14 @@
       <Sha>3f926184cacc477e3ddccd1eb4f3ebcb9a2b81a5</Sha>
       <SourceBuild RepoName="source-build-reference-packages" ManagedOnly="true" />
     </Dependency>
-    <Dependency Name="System.CommandLine" Version="2.0.0-beta4.25072.1">
+    <Dependency Name="System.CommandLine" Version="2.0.0-beta4.24528.1">
```
jjonescz, line 16: What's the reason for the revert?

jjonescz (approved):

eng/Version.Details.xml
ViktorHofer: @adamsitnik you might want to mention the reason for the revert in an issue somewhere (maybe in command-line-api) and then link to that in all the revert PRs.

adamsitnik: Short answer: chicken-and-egg problem with the update in SDK and NuGet.Client repos.

Long answer: The breaking change introduced in System.CommandLine requires an update to SDK, which uses both System.CommandLine and a NuGet package from NuGet.Client repo that also relies on System.CommandLine. NuGet.Client repo uses a hardcoded version of SDK (9.0.2) to run their integration tests, which is using an older version of System.CommandLine and fails at runtime (because the types have been renamed). The NuGet.Client repo release schedule is not tied to .NET, but to VS, so their dev branch updates flow to both SDK/main and SDK/release/9.0.x.  Unblocking the flow would require them to create dedicated branches (and cause a lot of work to sync all the changes), or use a workaround with multi-targeting (https://github.com/NuGet/NuGet.Client/pull/6252). It was decided that the simplest solution will be to just wait until NuGet.Client creates a new branch for next VS release in April.

PR #77088, opened 2025-02-06
Publish symbols manually to internal feeds for PR validation

dibarbet:
https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/635258

This switches our PR val builds to publish symbols to dnceng internal feeds only.  They do not need to be public. 

Previously, symbols were published by the automatic dnceng publishing step, but that published symbols using the channel configured for main, which is public.  Instead, we now publish manually using darc to add to an internal only channel.

Previously we had considered manually using microbuild for this, but due to the way arcade creates symbol packages, it turned out to be more effort than manually invoking darc to publish to a specific channel.

### DIFF
--- a/azure-pipelines-pr-validation.yml
+++ b/azure-pipelines-pr-validation.yml
@@ -290,10 +290,39 @@ extends:
           pool:
             name: VSEngSS-MicroBuild2022-1ES
 
+    - stage: publish
+      displayName: Publish
+      dependsOn:
+      - build
+      jobs:
+      - job: publishDarc
+        displayName: Publish using darc
+        pool:
+          name: VSEngSS-MicroBuild2022-1ES
+        steps:
+        - checkout: none
+        # Make BarBuildId available.
+        - template: /eng/common/templates/post-build/setup-maestro-vars.yml@self
+
+        - powershell: |
+            $versionEndpoint = 'https://maestro.dot.net/api/assets/darc-version?api-version=2019-01-16'
+            $darcVersion = $(Invoke-WebRequest -Uri $versionEndpoint -UseBasicParsing).Content
+            $arcadeServicesSource = 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json'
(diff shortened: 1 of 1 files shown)


jaredpar (approved):

azure-pipelines-pr-validation.yml
```diff
@@ -264,6 +264,31 @@ extends:
...
+        - task: MicroBuildArchiveSymbols@5
+          displayName: 'Upload Symbols to Symweb/MSDL'
+          inputs:
+            SymbolsFeatureName: 'RoslynPRValidation'
+            SymbolsProject: 'DDE'
+            SymbolsAgentPath: '$(Build.SourcesDirectory)\artifacts\tmp\$(BuildConfiguration)\SymbolPackages'
+            SubmitToInternet: false
+          condition: succeeded()
+
+        - powershell: Write-Host "##vso[build.updatebuildnumber]$(BuildNumber)"
```
JoeRobich, line 288:
```suggestion
        - powershell: Write-Host "##vso[build.updatebuildnumber]$(FancyBuildNumber)"
```

azure-pipelines-pr-validation.yml
```diff
@@ -264,6 +264,31 @@ extends:
             arguments: '-configuration $(BuildConfiguration) -branchName "$(SourceBranchName)" -prValidation'
           condition: succeeded()
 
+        # Microbuild uses the build number in the drop name, and it doesn't like the fancy build number.
+        # Rest it here before we publish, then set it back when we're done.
```
JoeRobich, line 268:
```suggestion
        # Reset it here before we publish, then set it back when we're done.
```

azure-pipelines-pr-validation.yml
dibarbet: FancyBuildNumber doesnt actually exist here.  FancyBuild step seems to set BuildNumber instead

JoeRobich: Sorry, saw it in the remove code below. Double checked and you are absolutely right.

JoeRobich (approved):

dibarbet:
passing symbol check - https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/635258

JoeRobich (approved):

Issue #77039, opened 2025-02-04
SemanticSearch.ReferenceAssemblies is rebuilding the references every time, even if not needed


jasonmalinowski:
This is more visible after https://github.com/dotnet/roslyn/pull/77038 is merged.

Problem: when Roslyn.VisualStudio.Setup is built, it's calling into targets in SemanticSearch.ReferenceAssemblies.csproj to rebuild the reference assemblies. This seems to rebuild every single time, even if the inputs or outputs are up to date. This means repeatedly pressing F6 in Visual Studio will never each a steady state, since each time Roslyn.VisualStudio.Setup builds, it'll rebuild the reference assemblies.

Running a few local builds I see this for the GenerateFilteredReferenceAssemblies task:

Input file "F:\Projects\Roslyn\artifacts\bin\Microsoft.CodeAnalysis.CSharp\Debug\net8.0\Microsoft.CodeAnalysis.CSharp.dll" is newer than output file "F:\Projects\Roslyn\src\Tools\SemanticSearch\ReferenceAssemblies\Apis\Microsoft.CodeAnalysis.CSharp.txt".

It seems odd to me that we have an output file pointing to our repo, but this will never get up to date properly.

jasonmalinowski:
@tmat can you chime in to what is the intent here?

jasonmalinowski:
It appears to me that the task is intentionally doing its own version checking and avoiding the file writes if they're not changed, but this trips up MSBuild doing the same thing.

tmat:
> Running a few local builds I see this for the GenerateFilteredReferenceAssemblies task:
> 
> Input file "F:\Projects\Roslyn\artifacts\bin\Microsoft.CodeAnalysis.CSharp\Debug\net8.0\Microsoft.CodeAnalysis.CSharp.dll" is newer than output file "F:\Projects\Roslyn\src\Tools\SemanticSearch\ReferenceAssemblies\Apis\Microsoft.CodeAnalysis.CSharp.txt".
> 
> It seems odd to me that we have an output file pointing to our repo, but this will never get up to date properly.

The output file is a list of filtered APIs. This is checked in, so that we can review any changes in APIs exposed to semantic search.

tmat:
> It appears to me that the task is intentionally doing its own version checking and avoiding the file writes if they're not changed, but this trips up MSBuild doing the same thing.

Should we remove the Api files from the output set? Ie. the only output would be the rewritten assembly?

[...]

PR #77097, opened 2025-02-07
Always write the Api output files, even if they don't change

jasonmalinowski:
The calling MSBuild targets already specify inputs and outputs; skipping writing the output files if they're not changed prevents the output timestamp from being updated, so the builds always run (and then force a rebuild of our setup projects).

Fixes https://github.com/dotnet/roslyn/issues/77039

### DIFF
--- a/src/Tools/SemanticSearch/BuildTask/GenerateFilteredReferenceAssembliesTask.cs
+++ b/src/Tools/SemanticSearch/BuildTask/GenerateFilteredReferenceAssembliesTask.cs
@@ -149,31 +149,14 @@ public sealed class GenerateFilteredReferenceAssembliesTask : Task
         var newContent = "# Generated, do not update manually\r\n" +
             string.Join("\r\n", apis);
 
-        string currentContent;
         try
         {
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi (approved):

src/Tools/SemanticSearch/BuildTask/GenerateFilteredReferenceAssembliesTask.cs
```diff
@@ -149,31 +149,14 @@ internal void WriteApis(string outputFilePath, byte[] peImage)
         var newContent = "# Generated, do not update manually\r\n" +
             string.Join("\r\n", apis);
 
-        string currentContent;
         try
         {
-            currentContent = File.ReadAllText(outputFilePath, Encoding.UTF8);
+            File.WriteAllText(outputFilePath, newContent);
+            Log.LogMessage($"Baseline updated: '{outputFilePath}'");
```
tmat, line 155: Nit: might be useful to output different message when the content differs.

tmat (approved):

PR #77118, opened 2025-02-08
Partial events and constructors: check differences

jjonescz:

### DIFF
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
@@ -55,8 +55,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             _syntaxRef = syntax.GetReference();
 
             var isExplicitInterfaceImplementation = interfaceSpecifierSyntaxOpt != null;
-            _modifiers = MakeModifiers(modifiers, isExplicitInterfaceImplementation, isFieldLike, _location, diagnostics, out _, out bool hasExplicitAccessModifier);
-            _hasExplicitAccessModifier = hasExplicitAccessModifier;
+            _modifiers = MakeModifiers(modifiers, isExplicitInterfaceImplementation, isFieldLike, _location, diagnostics, out _, out _hasExplicitAccessModifier);
             this.CheckAccessibility(_location, diagnostics, isExplicitInterfaceImplementation);
         }
 
@@ -407,7 +406,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             get { return (_modifiers & DeclarationModifiers.ReadOnly) != 0; }
         }
 
-        internal bool IsUnsafe
+        private bool IsUnsafe
         {
             get { return (_modifiers & DeclarationModifiers.Unsafe) != 0; }
         }
@@ -416,8 +415,6 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
         {
             get { return ModifierUtils.EffectiveAccessibility(_modifiers); }
         }
-
-        internal bool HasExplicitAccessModifier => _hasExplicitAccessModifier;
 
         internal sealed override bool MustCallMethodsDirectly
         {
@@ -851,7 +848,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             }
 
             if (DeclaredAccessibility != implementation.DeclaredAccessibility
-                || HasExplicitAccessModifier != implementation.HasExplicitAccessModifier)
+                || _hasExplicitAccessModifier != implementation._hasExplicitAccessModifier)
             {
                 diagnostics.Add(ErrorCode.ERR_PartialMemberAccessibilityDifference, implementation.GetFirstLocation());
             }
--- a/src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
+++ b/src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
@@ -27,6 +27,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
         private readonly Location _location;
         private readonly SyntaxReference _syntaxRef;
         private readonly DeclarationModifiers _modifiers;
+        private readonly bool _hasExplicitAccessModifier;
         internal readonly SourceMemberContainerTypeSymbol containingType;
 
         private SymbolCompletionState _state;
@@ -54,8 +55,8 @@ namespace Microsoft.CodeAnalysis.CSharp.Symbols
             _syntaxRef = syntax.GetReference();
 
             var isExplicitInterfaceImplementation = interfaceSpecifierSyntaxOpt != null;
-            bool modifierErrors;
-            _modifiers = MakeModifiers(modifiers, isExplicitInterfaceImplementation, isFieldLike, _location, diagnostics, out modifierErrors);
+            _modifiers = MakeModifiers(modifiers, isExplicitInterfaceImplementation, isFieldLike, _location, diagnostics, out _, out bool hasExplicitAccessModifier);
+            _hasExplicitAccessModifier = hasExplicitAccessModifier;
             this.CheckAccessibility(_location, diagnostics, isExplicitInterfaceImplementation);
         }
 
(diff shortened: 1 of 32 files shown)


src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
```diff
@@ -54,8 +55,8 @@ internal SourceEventSymbol(
             _syntaxRef = syntax.GetReference();
 
             var isExplicitInterfaceImplementation = interfaceSpecifierSyntaxOpt != null;
-            bool modifierErrors;
-            _modifiers = MakeModifiers(modifiers, isExplicitInterfaceImplementation, isFieldLike, _location, diagnostics, out modifierErrors);
+            _modifiers = MakeModifiers(modifiers, isExplicitInterfaceImplementation, isFieldLike, _location, diagnostics, out _, out bool hasExplicitAccessModifier);
```
cston, line 58: >out bool hasExplicitAccessModifier [](http://example.com/codeflow?start=129&length=34)

Minor: Could remove the local and use `out _hasExplicitAccessModifier` instead.

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
```diff
@@ -406,11 +407,18 @@ internal bool IsReadOnly
             get { return (_modifiers & DeclarationModifiers.ReadOnly) != 0; }
         }
 
+        internal bool IsUnsafe
```
cston, line 410: >internal [](http://example.com/codeflow?start=8&length=8)

`private`?

src/Compilers/CSharp/Portable/Symbols/Source/SourceEventSymbol.cs
```diff
@@ -406,11 +407,18 @@ internal bool IsReadOnly
...
+        {
+            get { return (_modifiers & DeclarationModifiers.Unsafe) != 0; }
+        }
+
         public sealed override Accessibility DeclaredAccessibility
         {
             get { return ModifierUtils.EffectiveAccessibility(_modifiers); }
         }
 
+        internal bool HasExplicitAccessModifier => _hasExplicitAccessModifier;
```
cston, line 420: >internal [](http://example.com/codeflow?start=8&length=8)

`private`, or remove the property and use the field directly?

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -1411,4 +1430,625 @@ void M()
...
+                partial event System.Action J { add { } remove { } }
+            }
+            """;
+        CreateCompilation(source).VerifyDiagnostics(
+            // (12,36): error CS9255: Both partial member declarations must have the same type.
+            //     partial event System.Func<int> E { add { } remove { } }
+            Diagnostic(ErrorCode.ERR_PartialMemberTypeDifference, "E").WithLocation(12, 36),
+            // (14,49): error CS8142: Both partial member declarations, 'C.G' and 'C.G', must use the same tuple element names.
+            //     partial event System.Action<(int A, int B)> G { add { } remove { } }
+            Diagnostic(ErrorCode.ERR_PartialMemberInconsistentTupleNames, "G").WithArguments("C.G", "C.G").WithLocation(14, 49),
```
cston, line 1567: >ERR_PartialMemberInconsistentTupleNames [](http://example.com/codeflow?start=33&length=39)

Are we reporting this error for partial constructor parameters?

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -1411,4 +1430,625 @@ void M()
...
+            partial class C1
+            {
+                protected virtual partial event System.Action E, F;
+                protected partial event System.Action E { add { } remove { } }
+                protected virtual partial event System.Action F { add { } remove { } }
+            }
+            partial class C2 : C1
+            {
+                protected override partial event System.Action E;
+                protected sealed partial event System.Action E { add { } remove { } }
```
cston, line 1876: Consider testing `override` and `sealed` independently.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberMethodSymbol.cs
```diff
@@ -39,6 +39,7 @@ protected struct Flags
             // b = HasAnyBody. 1 bit.
             // a = IsVararg. 1 bit.
             // t = HasThisInitializer. 1 bit.
+            // a = HasExplicitAccessModifier. 1 bit.
```
cston, line 42: >a [](http://example.com/codeflow?start=15&length=1)

Consider using a new character, perhaps `m`, and adding to the layout above.

src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberMethodSymbol.cs
```diff
@@ -204,6 +213,7 @@ public Flags(
                 int isMetadataVirtualIgnoringInterfaceImplementationChangesInt = isMetadataVirtual ? IsMetadataVirtualIgnoringInterfaceChangesBit : 0;
                 int isMetadataVirtualInt = isMetadataVirtual ? IsMetadataVirtualBit : 0;
                 int hasThisInitializerInt = hasThisInitializer ? HasThisInitializerBit : 0;
+                var hasExplicitAccessModifierInt = hasExplicitAccessModifier ? HasExplicitAccessModifierBit : 0;
```
cston, line 216: >var [](http://example.com/codeflow?start=16&length=3)

Consider using `int` for consistency.

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -1411,4 +1430,625 @@ void M()
...
+                {
+                    add { E += null; }
+                    remove { E += null; }
+                }
+                public static partial event System.Action F
+                {
+                    add { E += null; } // 3
+                    remove { E += null; } // 4
+                }
+            }
```
cston, line 1461: Was the intent to include use of `F` here as well? Such as:
```csharp
public partial event System.Action E
{
    add { this.E += null; }
    remove { this.F += null; } // 3
}
public static partial event System.Action F
{
    add { E += null; } // 4
    remove { F += null; }
}
```

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
cston: Never mind, it looks like that case results in `WRN_PartialMemberSignatureDifference` instead, in test below.

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -1411,4 +1430,625 @@ void M()
...
+                partial C1(int x) { }
+            }
+            partial class C2
+            {
+                partial C2(int x);
+                partial C2(int x = 1) { }
+            }
+            """;
+        CreateCompilation(source).VerifyDiagnostics();
+    }
```
cston, line 2005: >} [](http://example.com/codeflow?start=4&length=1)

Consider testing use, so it's clear which signature is used at the call site.
```csharp
var c1 = new C1();
var c2 = new C2();
```

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -1411,4 +1430,625 @@ void M()
...
+    {
+        var source = """
+            partial class C
+            {
+                partial C(int x = 1);
+                partial C(int x = 2) { }
+            }
+            """;
+        CreateCompilation(source).VerifyDiagnostics();
+    }
```
cston, line 2018: >} [](http://example.com/codeflow?start=4&length=1)

Consider testing use, `new C()`, and reporting the value of the argument passed.

cston (approved):

jjonescz:

RikkiGibson (approved):

src/Compilers/CSharp/Test/Emit3/PartialEventsAndConstructorsTests.cs
```diff
@@ -1411,4 +1430,693 @@ void M()
...
+                partial C1(int x) { }
+            }
+            partial class C2
+            {
+                partial C2(dynamic x);
+                partial C2(object x) { }
+            }
+            partial class C3
+            {
+                partial C3((int X, int Y) x);
```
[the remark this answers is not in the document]
jjonescz: Good catch, thanks. That should be an error in both cases if we want to be consistent with other partial member kinds.

PR #77127, opened 2025-02-10
Implement field null resilience analysis

### LINKED ISSUE #76662: Field-backed properties: support lazily initialized backing field in nullable analysis
Implement nullable analysis for `field` from [proposals/field-keyword.md](https://github.com/dotnet/csharplang/blob/main/proposals/field-keyword.md#nullability).

RikkiGibson:
Closes #76662
See also dotnet/csharplang#9184

Mainly what I would like feedback on here is: does the way the API works here make sense? We basically have a side API `SynthesizedBackingFieldSymbol.GetInferredNullableAnnotation()`, which at this point we take care to call only from the NullableWalker.

Since calling this API results in binding and flow analysis, I am concerned about re-entrancy. What if doing this flow analysis, causes us to ask about the field's nullable annotation, which causes us to flow analyze, etc... Hence the current limited implementation.

But I am also concerned about coherence of the public APIs if we do not expose this information in a broader/more holistic way. For example, if we hover on `field` in `public string Prop => field ??= "a";`, and it says `string Prop.field`, it is confusing, because we in fact inferred `string?` for it. So my question is essentially how can we surface this inferred nullability in a way which reduces the risk of stack cycles.


RikkiGibson:
@cston @333fred for review

docs/features/field-nullability.md
RikkiGibson, line 1: This file may not necessarily make it into main but it helped me organize my thoughts about how to do this.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -10737,5 +10722,653 @@ struct S
...
+        public void Nullable_NotResilient_AccessorWarningsAreDisabled_02()
+        {
+            var source = """
+                #nullable enable
+
+                class C
+                {
+                    public string Prop
+                    {
+                #pragma warning disable 8603 // Possible null reference return.
```
333fred, line 11301: What's causing the difference in warnings in the getter then?

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
333fred, line 1: Would like to see some more variations on tests. For example:
* passing `field` by ref with conflicting nullable annotations
* warnings caused by use of `field` inside nested local functions (both called and uncalled)
    * For bonus points, a local function that unconditionally deferences `field`, called before `??=` (should produce a warning), called after the `??=` (should not produce a warning), not called (should produce a warning).
* warnings caused by use of `field` inside nested lambdas
* Similar scenarios to `??=`, except with just `??` instead.
* Use in target-typed constructs like collection expressions

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
RikkiGibson: What's happening is: the nullable analysis is performed and diagnostics collected, but, the step which filters out diagnostics based on pragmas is not performed. So `ComputeInferredNullableAnnotation()` sees the diagnostic anyway.

In the `#nullable disable warnings` case, the `AnalyzeIfNeeded` call is able to see that nullable analysis should be completely skipped. So `ComputeInferredNullableAnnotation()` sees no diagnostics.

We could consider adjusting behavior here by either:
1) doing the pragma diagnostic filtering, to make it so `Annotated` is inferred in both `#nullable disable` and `#pragma warning disable` cases
2) forcing the nullable analysis to occur even with `#nullable disable warnings`, and using the resulting diagnostics without filtering, so that `NotAnnotated` is inferred in both cases.

I don't have much leaning here. It is expected that using `!` to suppress a "hypothetical" nullable warning, e.g. `get => field!;`, can result in the field being inferred as nullable. That seems good and fine to me. But deciding the field's nullability effectively based on pragmas or project-level `<NoWarn>` settings, etc., I really don't know. If I had to pick I guess I would say I lean toward (2).

333fred: I think I'd prefer option 1. Seems like the most consistent approach.

RikkiGibson: I can run with it.

One other scenario that comes to mind is: what if a `DiagnosticSuppressor` is suppressing a nullable warning? I think we probably shouldn't run any diagnostic suppressors in this context.

src/Compilers/CSharp/Portable/Errors/SameDiagnosticComparer.cs
```diff
@@ -0,0 +1,16 @@
...
+using Roslyn.Utilities;
+
+namespace Microsoft.CodeAnalysis.CSharp;
+
+internal sealed class SameDiagnosticComparer : EqualityComparer<Diagnostic>
+{
+    public static readonly SameDiagnosticComparer Instance = new SameDiagnosticComparer();
+    public override bool Equals(Diagnostic? x, Diagnostic? y) => x is null ? y is null : x.Equals(y);
+    public override int GetHashCode(Diagnostic obj) =>
+        Hash.Combine(Hash.CombineValues(obj.Arguments), Hash.Combine(obj.Location.GetHashCode(), obj.Code));
```
cston, line 15: >Hash.Combine(Hash.CombineValues(obj.Arguments), Hash.Combine(obj.Location.GetHashCode(), obj.Code)) [](http://example.com/codeflow?start=8&length=99)

`obj.GetHashCode()`

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -2793,13 +2812,47 @@ private void InheritDefaultState(TypeSymbol targetType, int targetSlot)
...
+
+        /// <summary>Gets the TypeWithAnnotations of a symbol, possibly using the inferred nullable annotation for backing fields.</summary>
+        private TypeWithAnnotations GetTypeOrReturnTypeWithAnnotations(Symbol symbol)
+        {
+            var typeWithAnnotations = symbol.GetTypeOrReturnType();
+            if (symbol is SynthesizedBackingFieldSymbol { InfersNullableAnnotation: true } backingField)
+            {
+                NullableAnnotation nullableAnnotation;
+                if (_getterNullResilienceData is var (analyzedField, assumedNullableAnnotation)
+                    && (object)analyzedField == backingField)
```
cston, line 2830: >(object)analyzedField == backingField [](http://example.com/codeflow?start=23&length=37)

Does this ever fail? If not, should we assert this somewhere?

RikkiGibson: I wasn't able to think of any existing code where this fails. I believe this could only happen if a getter is using a backing field of another property. Then, when we are inferring annotation of the current backing field, we would start inferring an annotation of a different backing field.

The 'else' block effectively asserts this, though, I am open to some more straightforward way of writing this.

I do think it is good to keep doing this check in release mode, though. I actually wouldn't mind just throwing here in release mode if this is not met, and dropping the assertion.

cston:
>                 CheckDisallowedNullAssignment(rightState, leftAnnotations, right.Syntax);

Are we testing the call to `CheckDisallowedNullAssignment(rightState, leftAnnotations, right.Syntax)` when `leftAnnotations` are explicit annotations on the backing field?

---
Refers to: src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs:10108 in 963f0b3. [](commit_id = 963f0b3e0a2d51b7e853b7fa65a7199c644b02ad, deletion_comment = False)

RikkiGibson:
> > ```
> >             CheckDisallowedNullAssignment(rightState, leftAnnotations, right.Syntax);
> > ```
> 
> Are we testing the call to `CheckDisallowedNullAssignment(rightState, leftAnnotations, right.Syntax)` when `leftAnnotations` are explicit annotations on the backing field?
> 
> Refers to: src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs:10108 in [963f0b3](https://github.com/dotnet/roslyn/commit/963f0b3e0a2d51b7e853b7fa65a7199c644b02ad). [](commit_id = [963f0b3](https://github.com/dotnet/roslyn/commit/963f0b3e0a2d51b7e853b7fa65a7199c644b02ad), deletion_comment = False)

Sounds like you are asking about a scenario like `[field: DisallowNull] int? Prop { get => field; }`?

src/Compilers/CSharp/Portable/Symbols/Synthesized/SynthesizedBackingFieldSymbol.cs
```diff
@@ -122,6 +125,110 @@ public override ImmutableArray<Location> Locations
...
+            {
+                var diagnostics = DiagnosticBag.GetInstance();
+                NullableWalker.AnalyzeIfNeeded(binder, boundGetAccessor, boundGetAccessor.Syntax, diagnostics, getterNullResilienceData: (getAccessor, _property.BackingField, assumedNullableAnnotation));
+                if (diagnostics.IsEmptyWithoutResolution)
+                {
+                    return diagnostics;
+                }
+
+                var filteredDiagnostics = DiagnosticBag.GetInstance();
+                _ = DeclaringCompilation.FilterAndAppendAndFreeDiagnostics(filteredDiagnostics, ref diagnostics, cancellationToken: default);
```
cston, line 226: >FilterAndAppendAndFreeDiagnostics [](http://example.com/codeflow?start=41&length=33)

Does the inferred nullability of the field depend on which warnings are enabled or disabled? Shouldn't the pragmas and compiler options only control which diagnostics are reported rather than the nullability of the field?

RikkiGibson: https://github.com/dotnet/roslyn/pull/77127#discussion_r1974484381

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -10737,5 +10723,842 @@ struct S
...
+            var prop = comp.GetMember<SourcePropertySymbol>("C.Prop");
+            Assert.Equal(NullableAnnotation.NotAnnotated, prop.BackingField.GetInferredNullableAnnotation());
+            Assert.Equal(NullableAnnotation.NotAnnotated, prop.BackingField.TypeWithAnnotations.NullableAnnotation);
+        }
+
+        [Fact]
+        public void Nullable_NotResilient_UseInSetter_01()
+        {
+            // Setter analysis is not yet implemented.
+            // https://github.com/dotnet/roslyn/issues/77215
```
cston, line 11210: Why is setter analysis needed here? It looks like we're already warning that the field is not initialized to a non-nullable value.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -10737,5 +10723,842 @@ struct S
...
+                {
+                    public string Prop
+                    {
+                        get => field;
+                        set => field = field.ToString();
+                    }
+                }
+                """;
+
+            var comp = CreateCompilation([source, AllowNullAttributeDefinition]);
```
cston, line 11224: >AllowNullAttributeDefinition [](http://example.com/codeflow?start=50&length=28)

Minor: It looks like this test and the next several do not use this attribute.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -10737,5 +10723,842 @@ struct S
...
+                Diagnostic(ErrorCode.WRN_UninitializedNonNullableBackingField, "Prop").WithArguments("property", "Prop").WithLocation(5, 19),
+                // (8,24): warning CS8625: Cannot convert null literal to non-nullable reference type.
+                //         set => field = null;
+                Diagnostic(ErrorCode.WRN_NullAsNonNullable, "null").WithLocation(8, 24));
+            var prop = comp.GetMember<SourcePropertySymbol>("C.Prop");
+            Assert.Equal(NullableAnnotation.NotAnnotated, prop.BackingField.GetInferredNullableAnnotation());
+            Assert.Equal(NullableAnnotation.NotAnnotated, prop.BackingField.TypeWithAnnotations.NullableAnnotation);
+
+            // The 8603 is a "hypothetical" diagnostic which occurs only internally in the compiler.
+            // We don't run DiagnosticSuppressors in that context.
```
cston, line 11483: >// We don't run DiagnosticSuppressors in that context. [](http://example.com/codeflow?start=12&length=54)

If there's no difference between the results with or without the diagnostic suppressor, consider using `[Theory]` and a `bool` parameter to include the suppressor or not, so it's clear there's no difference.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
```diff
@@ -10737,5 +10723,842 @@ struct S
...
+            comp0.VerifyEmitDiagnostics(
+                // (8,16): warning CS8607: A possible null value may not be used for a type marked with [NotNull] or [DisallowNull]
+                //         Prop = null; // 1
+                Diagnostic(ErrorCode.WRN_DisallowNullAttributeForbidsMaybeNullAssignment, "null").WithLocation(8, 16));
+
+            var sourceField = comp0.GetMember<SynthesizedBackingFieldSymbol>("C.<Prop>k__BackingField");
+            Assert.False(sourceField.InfersNullableAnnotation);
+            Assert.Equal(NullableAnnotation.Annotated, sourceField.TypeWithAnnotations.NullableAnnotation);
+            Assert.Equal("System.Int32?", sourceField.TypeWithAnnotations.ToTestDisplayString());
+        }
```
cston, line 11562: >} [](http://example.com/codeflow?start=8&length=1)

Consider testing a `static` property and associated constructor.

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
RikkiGibson: Done

docs/features/field-nullability.md
```diff
@@ -0,0 +1,21 @@
...
+...
+
+It's tempting to avoid exposing the inferred nullable annotation in the public symbol model. This might be problematic for automated tooling which is trying to reason about nullable initialization. For example, if a diagnostic suppressor wants to suppress `CS8618` nullable initialization warnings under certain conditions. How should it decide whether/why a property like `string Prop => field ??= GetValue();` requires initialization in constructors? It's unclear to me whether this matters.
+
+In the interests of *simplicity* (first make it correct, then make it fast), I'd like to move forward by *not* exposing the inferred annotation in the symbol model. Rather, we introduce a new internal API `SynthesizedBackingFieldSymbol.GetInferredNullableAnnotation()`, which `NullableWalker` will use to decide initial nullable state, report warnings, and so on. The implementation simply does an on-demand binding of the get accessor, then nullable analyzes and decides the nullable annotation. Care needs to be taken to avoid using this API in a re-entrant manner--so, the `NullableWalker` passes which are used to infer the nullability annotation must avoid calling it.
```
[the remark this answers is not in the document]
RikkiGibson: Done

src/Compilers/CSharp/Portable/FlowAnalysis/NullableWalker.cs
```diff
@@ -2793,13 +2812,47 @@ private void InheritDefaultState(TypeSymbol targetType, int targetSlot)
...
+                {
+                    // Currently in the process of inferring the nullable annotation for 'backingField'.
+                    // Therefore don't try to access the inferred nullable annotation, use a temporary assumedNullableAnnotation instead.
+                    nullableAnnotation = assumedNullableAnnotation;
+                }
+                else
+                {
+                    // If we have a '_getterNullResilienceData' in this code path,
```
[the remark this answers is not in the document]
RikkiGibson: Fixed

src/Compilers/CSharp/Portable/Symbols/Synthesized/SynthesizedBackingFieldSymbol.cs
```diff
@@ -122,6 +125,98 @@ public override ImmutableArray<Location> Locations
...
+
+                return true;
+            }
+        }
+
+        internal NullableAnnotation GetInferredNullableAnnotation()
+        {
+            if (_inferredNullableAnnotation == (int)NullableAnnotation.Ignored)
+            {
+                Interlocked.CompareExchange(ref _inferredNullableAnnotation, (int)ComputeInferredNullableAnnotation(), (int)NullableAnnotation.Ignored);
```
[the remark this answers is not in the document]
RikkiGibson: Done

src/Compilers/CSharp/Portable/Symbols/Synthesized/SynthesizedBackingFieldSymbol.cs
RikkiGibson: Resolved in offline discussion. I think we are comfortable moving forward with the suppressions behavior currently implemented in the PR.

src/Compilers/CSharp/Test/Emit3/FieldKeywordTests.cs
RikkiGibson: Done

cston (approved):

333fred (approved):

RikkiGibson:
I have updated the speclet links, to refer to the exact version of the spec which this PR is implementing.

There has been an additional suggestion in dotnet/csharplang#9184, to instead make it so the backing field is always nullable a la `var`, and instead define things in terms of what is the initial state of `field` in the getter, and when is a constructor exit warning reported on the field.

I started implementing that approach, and I think it is promising, but, it is proving to be a *lot* of churn, especially since backing fields are all over the place in our tests. The number of increased attributes etc. is something that may need more scrutiny. So, I would like to **start by merging this PR as-is**, and sending a follow up depending on how the next step of the work goes.

Please let me know if there are any objections/concerns. Thanks.

PR #77128, opened 2025-02-10
Prototype field nullability adopt

RikkiGibson:
#77127 + #76906


RikkiGibson:
@CyrusNajmabadi FWIW, the diagnostics I am seeing in https://github.com/dotnet/roslyn/pull/77128/checks?check_run_id=36932129488 all look correct/expected. So I think we are on the right track here.

https://github.com/dotnet/roslyn/blob/220b267164b845737dba9fa72b1aedf4a79bb7e7/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/NamingStyles/Serialization/NamingStylePreferences.cs#L88

https://github.com/dotnet/roslyn/blob/220b267164b845737dba9fa72b1aedf4a79bb7e7/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Utilities/AbstractSpeculationAnalyzer.cs#L88

PR #77133, opened 2025-02-10
Use ActivationConstraint.FeatureFlag API to control when Semantic Search is enabled

tmat:
Removes UIContext based workaround for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1985204

Blocked on https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/609808


ToddGrun:
If this change is good, it might be best to have this target the post 17.14 branch we have as that is where the package load changes are going right now.

ToddGrun (approved):
:shipit:

PR #77137, opened 2025-02-10
Make `InterceptableLocation` implement `IEquatable<InterceptableLocation>`

### LINKED ISSUE #77135: Make `InterceptableLocation` implement `IEquatable<InterceptableLocation>`
It has the necessary [semantics](https://github.com/dotnet/roslyn/blob/dfa7fc6bdea31a858a402168384192b633c811fa/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs#L157-L170), and the original proposal included `: IEquatable<InterceptableLocation>` — see [here](https://github.com/dotnet/roslyn/issues/72133#:~:text=%3A%20IEquatable%3CInterceptableLocation%3E).

aradalvand:
Closes #77135

@RikkiGibson Let me know if this is not the ideal setup.

### DIFF
--- a/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs
+++ b/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs
@@ -181,7 +181,7 @@ internal sealed class InterceptableLocation1 : InterceptableLocation
             _position);
     }
 
-    public override bool Equals(InterceptableLocation obj)
+    public override bool Equals(InterceptableLocation? obj)
     {
         return obj is InterceptableLocation1 other
             && _checksum.SequenceEqual(other._checksum)
--- a/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs
+++ b/src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs
@@ -18,7 +18,7 @@ namespace Microsoft.CodeAnalysis.CSharp;
 /// <summary>Denotes an interceptable call. Used by source generators to generate '[InterceptsLocation]' attributes.</summary>
 /// <seealso href="https://github.com/dotnet/roslyn/issues/72133" />
 /// <seealso href="https://github.com/dotnet/csharplang/issues/7009" />
-public abstract class InterceptableLocation
+public abstract class InterceptableLocation : IEquatable<InterceptableLocation>
 {
     private protected InterceptableLocation() { }
 
@@ -40,6 +40,8 @@ public abstract class InterceptableLocation
 
     public abstract override bool Equals(object? obj);
     public abstract override int GetHashCode();
+
+    public abstract bool Equals(InterceptableLocation other);
 }
 
 #pragma warning disable RSEXPERIMENTAL002 // internal usage of experimental API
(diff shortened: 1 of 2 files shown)


src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs
```diff
@@ -40,6 +40,8 @@ private protected InterceptableLocation() { }
 
     public abstract override bool Equals(object? obj);
     public abstract override int GetHashCode();
+
+    public abstract bool Equals(InterceptableLocation other);
```
RikkiGibson, line 44: Just occurred to me that this is a public API change. I don't see any reason this would be *problematic*--any more than adding the interface, which was included in the originally approved proposal, would be problematic here. But let's check with @dotnet/roslyn-api-owners to make sure.

RikkiGibson: Alternatively, Equals(InterceptableLocation) could be an explicit interface implementation.

aradalvand: > Alternatively, Equals(InterceptableLocation) could be an explicit interface implementation.

Yeah, I could do that as well.

> I don't see any reason this would be problematic

Neither do I. But I'll be waiting for a response from @dotnet/roslyn-api-owners

jjonescz, line 44:
```suggestion
    public abstract bool Equals(InterceptableLocation? other);
```

src/Compilers/CSharp/Portable/Utilities/InterceptableLocation.cs
```diff
@@ -183,4 +180,14 @@ public override int GetHashCode()
             BinaryPrimitives.ReadInt32LittleEndian(_checksum.AsSpan()),
             _position);
     }
+
+    public override bool Equals(InterceptableLocation obj)
```
jjonescz, line 184:
```suggestion
    public override bool Equals(InterceptableLocation? obj)
```

RikkiGibson:
Thanks for addressing the feedback on this PR. There is still a public API check failure:

`src\Compilers\CSharp\Portable\Utilities\InterceptableLocation.cs(44,26): error RS0016: (NETCORE_ENGINEERING_TELEMETRY=Build) Symbol 'abstract Microsoft.CodeAnalysis.CSharp.InterceptableLocation.Equals(Microsoft.CodeAnalysis.CSharp.InterceptableLocation? other) -> bool' is not part of the declared public API (https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md)`

This can be resolved by adding `abstract Microsoft.CodeAnalysis.CSharp.InterceptableLocation.Equals(Microsoft.CodeAnalysis.CSharp.InterceptableLocation? other) -> bool` to a new line of `src\Compilers\CSharp\Portable\PublicAPI.Unshipped.txt`. Maybe grouping it with the other lines for InterceptableLocation would be good.

Public API fix should be verifiable locally with `dotnet build src\Compilers\CSharp\Portable -p:RunAnalyzersDuringBuild=true`. Thanks!

RikkiGibson (approved):

jjonescz (approved):

jjonescz:
Thanks for the contribution @aradalvand

PR #77138, opened 2025-02-10
Update Versions.props for Dev18

akhera99:


dibarbet (approved):
If we're creating a dev18 branch, IMHO we should do some kind of merge from main to it.  Otherwise it will get out of date quickly.  And it is different from servicing branches in that it should be up to date with main (even infra).

Probably the easiest way would be to utilize arcade's branch merge github action.  We would configure it to merge main to the dev18 branch maybe once a day or once a week.

Similar to https://github.com/dotnet/vscode-csharp/blob/main/.github/workflows/branch-merge.yml , except with a scheduled trigger instead of on push

eng/config/PublishData.json
```diff
@@ -253,6 +253,16 @@
       "insertionTitlePrefix": "[d17.14 P1]",
       "insertionCreateDraftPR": false
     },
+    "release/dev18.0": {
+      "nugetKind": [
+        "Shipping",
+        "NonShipping"
+      ],
+      "vsBranch": "dev/monicaro/versioning",
```
dibarbet, line 261: I assume this is temporary until we get an actual branch

eng/config/PublishData.json
```diff
@@ -253,6 +253,16 @@
       "insertionTitlePrefix": "[d17.14 P1]",
       "insertionCreateDraftPR": false
     },
+    "release/dev18.0": {
```
dibarbet, line 256: (spoke offline) - this change needs to be made in main

eng/config/PublishData.json
akhera99: correct, that's in progress

dibarbet (approved):

PR #77143, opened 2025-02-10
Semantic Snippets - switch snippet

akhera99:
Adds a switch snippet provider.
Creates a switch snippet which produces a default statement.

### DIFF
--- a/src/Features/CSharp/Portable/Snippets/CSharpSnippetIdentifiers.cs
+++ b/src/Features/CSharp/Portable/Snippets/CSharpSnippetIdentifiers.cs
@@ -15,15 +15,16 @@ internal static class CSharpSnippetIdentifiers
     public const string Else = "else";
     public const string Enum = "enum";
     public const string For = "for";
-    public const string ReversedFor = "forr";
     public const string ForEach = "foreach";
-    public const string InitOnlyProperty = "propi";
-    public const string RequiredProperty = "propr";
+    public const string ReversedFor = "forr";
     public const string If = "if";
     public const string Interface = "interface";
     public const string Lock = "lock";
+    public const string InitOnlyProperty = "propi";
+    public const string RequiredProperty = "propr";
     public const string StaticIntMain = "sim";
     public const string Struct = "struct";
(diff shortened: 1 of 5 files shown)


src/Features/Core/Portable/Snippets/SnippetProviders/AbstractSwitchSnippetProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.CodeAnalysis.Snippets.SnippetProviders
+{
+    internal abstract class AbstractSwitchSnippetProvider<TSwitchStatementSyntax> : AbstractStatementSnippetProvider<TSwitchStatementSyntax>
+        where TSwitchStatementSyntax : SyntaxNode
+    {
+        protected sealed override Task<TextChange> GenerateSnippetTextChangeAsync(Document document, int position, CancellationToken cancellationToken)
+        {
+            var generator = SyntaxGenerator.GetGenerator(document);
+            var statement = generator.SwitchStatement(generator.IdentifierName("switch_on"),
```
akhera99, line 22: Can I create the identifier name like so or does it need to a resources file?

src/Features/CSharp/Portable/Snippets/CSharpSnippetIdentifiers.cs
```diff
@@ -15,15 +15,16 @@ internal static class CSharpSnippetIdentifiers
     public const string Else = "else";
     public const string Enum = "enum";
     public const string For = "for";
-    public const string ReversedFor = "forr";
```
akhera99, line 18: just reordering alphabetically

JoeRobich (changes requested):
The current switch snippet supports generating all cases statements when an enum variable is chosen for `switch_on`. Semantic snippets do not support this and it would regress the experience.

We need to determine whether:
1. the Editor team can provide us a callback similar to the current snippets `<Function>` capability.
2. we could build it on our side reusing parts of inline rename
3. we can't fully replace the current snippet model

Issue #77040, opened 2025-02-04
Pipe is broken exception in MSBuildWorkspace tests


jasonmalinowski:
We're seeing this randomly in CI, often in our Spanish queue:

```
System.AggregateException : An error occurred while writing to logger(s). (Workspace failure Failure:[BuildHost PID 996] Exception while shutting down the BuildHost process.
System.IO.IOException: Pipe is broken.
   at System.IO.Pipes.PipeStream.CheckWriteOperations()
   at System.IO.Pipes.PipeStream.FlushAsync(CancellationToken cancellationToken)
--- End of stack trace from previous location ---
   at System.IO.StreamWriter.<FlushAsyncInternal>g__Core|79_0(Boolean flushStream, Boolean flushEncoder, CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.MSBuild.RpcClient.InvokeCoreAsync(Int32 targetObject, String methodName, List`1 parameters, Type expectedReturnType, CancellationToken cancellationToken) in /_/src/Workspaces/MSBuild/Core/Rpc/RpcClient.cs:line 166
   at Microsoft.CodeAnalysis.MSBuild.BuildHostProcessManager.BuildHostProcess.DisposeAsync() in /_/src/Workspaces/MSBuild/Core/MSBuild/BuildHostProcessManager.cs:line 434)
---- System.Exception : Workspace failure Failure:[BuildHost PID 996] Exception while shutting down the BuildHost process.
System.IO.IOException: Pipe is broken.
   at System.IO.Pipes.PipeStream.CheckWriteOperations()
   at System.IO.Pipes.PipeStream.FlushAsync(CancellationToken cancellationToken)
--- End of stack trace from previous location ---
   at System.IO.StreamWriter.<FlushAsyncInternal>g__Core|79_0(Boolean flushStream, Boolean flushEncoder, CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.MSBuild.RpcClient.InvokeCoreAsync(Int32 targetObject, String methodName, List`1 parameters, Type expectedReturnType, CancellationToken cancellationToken) in /_/src/Workspaces/MSBuild/Core/Rpc/RpcClient.cs:line 166
   at Microsoft.CodeAnalysis.MSBuild.BuildHostProcessManager.BuildHostProcess.DisposeAsync() in /_/src/Workspaces/MSBuild/Core/MSBuild/BuildHostProcessManager.cs:line 434
  Seguimiento de la pila:
[...]

PR #77151, opened 2025-02-11
Fix race during shutdown of the BuildHost server

jasonmalinowski:
Our RPC mechanism to the build host is a simple one: we send JSON payloads that are formatted as a single textual line, similar to other methods. We did the writing to the build host via a TextWriter, and called Flush() at the end to ensure that everything is written to the other end, otherwise things might deadlock if we wait for the server to reply to something that's not coming.

We recently switched the communication over to named pipes, which have a bit of a subtle problem here: with PipeStream, Flush() doesn't do anything, but it will check the pipe is still connected and throw if it's not. This had a problem during shutdown: we might write to the pipe, and the other side will see the shutdown request, respond, and close, before that Flush() call could ever happen. And once it did, we'd then throw an exception becase the other end did exactly what wanted it to!

I could special-case the handling of pipes in the RPC code, but there's no reason to allow the use of arbitrary streams here. Instead, it's best to just update the tests to also use named pipes (which means we're better testing what we're doing in real code), and it also means we can dispense of flushing entirely. Unfortunately TextWriter gives no way to disable flushing that I can see, so I also dispose of TextWriter and just write to the PipeStream directly.

Fix https://github.com/dotnet/roslyn/issues/77040

### DIFF
--- a/src/Workspaces/MSBuild/BuildHost/Rpc/RpcServer.cs
+++ b/src/Workspaces/MSBuild/BuildHost/Rpc/RpcServer.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Pipes;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
@@ -25,19 +26,19 @@ namespace Microsoft.CodeAnalysis.MSBuild;
 /// </remarks>
 internal sealed class RpcServer
 {
-    private readonly TextWriter _sendingStream;
+    private readonly TextWriter _streamWriter;
     private readonly SemaphoreSlim _sendingStreamSemaphore = new SemaphoreSlim(initialCount: 1);
-    private readonly TextReader _receivingStream;
+    private readonly TextReader _streamReader;
 
     private readonly ConcurrentDictionary<int, object> _rpcTargets = [];
     private volatile int _nextRpcTargetIndex = -1; // We'll start at -1 so the first value becomes zero
 
     private readonly CancellationTokenSource _shutdownTokenSource = new CancellationTokenSource();
 
-    public RpcServer(Stream sendingStream, Stream receivingStream)
+    public RpcServer(PipeStream stream)
     {
-        _sendingStream = new StreamWriter(sendingStream, JsonSettings.StreamEncoding);
-        _receivingStream = new StreamReader(receivingStream, JsonSettings.StreamEncoding);
+        _streamWriter = new StreamWriter(stream, JsonSettings.StreamEncoding);
+        _streamReader = new StreamReader(stream, JsonSettings.StreamEncoding);
     }
 
     public int AddTarget(object rpcTarget)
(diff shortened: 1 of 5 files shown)


src/Workspaces/MSBuild/BuildHost/Rpc/RpcServer.cs
```diff
@@ -25,19 +26,19 @@ namespace Microsoft.CodeAnalysis.MSBuild;
...
-    private readonly TextReader _receivingStream;
+    private readonly TextReader _streamReader;
 
     private readonly ConcurrentDictionary<int, object> _rpcTargets = [];
     private volatile int _nextRpcTargetIndex = -1; // We'll start at -1 so the first value becomes zero
 
     private readonly CancellationTokenSource _shutdownTokenSource = new CancellationTokenSource();
 
-    public RpcServer(Stream sendingStream, Stream receivingStream)
+    public RpcServer(PipeStream stream)
```
jasonmalinowski, line 38: This change wasn't necessary, but keeps it consistent with RpcClient only accepting a PipeStream.

src/Workspaces/MSBuild/Core/Rpc/RpcClient.cs
```diff
@@ -153,20 +158,22 @@ public async Task<T> InvokeAsync<T>(int targetObject, string methodName, List<ob
...
-#else
-                await _sendingStream.FlushAsync().ConfigureAwait(false);
-#endif
+                // Write out the request to the stream. This previously used a TextWriter and called Flush(), but it was discovered that wasn't safe
+                // around PipeStream: the TextWriter might call Flush() on the underlying stream, and PipeStream's Flush() implementation doesn't do
+                // anything other than check if the pipe was disconnected. This created a race condition during us trying to shutdown the build host:
+                // we'd send a shutdown message, but if TextWriter calls PipeStream.Flush() at any point (either because we ask it to flush, or it were
+                // to decide it wants to flush), it might do so once the pipe has disconnected and we'll get an IOException -- even though the other
+                // process did exactly what we wanted it to! By just writing to the pipe directly we avoid any surprises here.
+                await _stream.WriteAsync(requestJsonBytes, 0, requestJsonBytes.Length, cancellationToken).ConfigureAwait(false);
```
jasonmalinowski, line 176: This is the core to the actual fix; everything else is renames or refactorings.

src/Workspaces/MSBuild/Test/RpcTests.cs
```diff
@@ -203,6 +215,18 @@ public async Task CancelledTaskDoesNotLeakRequest()
...
 
+        [Fact]
+        [WorkItem("https://github.com/dotnet/roslyn/issues/77040")]
+        public async Task RequestThatClosesServerDoesNotThrow()
+        {
+            await using var rpcPair = new RpcPair();
+            rpcPair.Server.AddTarget(new ObjectWithMethodThatShutsDownServer(rpcPair.Server));
+
+            // Invoke the method, and ensure we don't get exceptions back due to the shutdown if the pipe closed too early.
+            // This is not guaranteed to catch any bug, since any bug is fundamentally a race, but it at least ensures we aren't always broken.
```
jasonmalinowski, line 226: I wasn't able to get it to fail locally, but the test should ensure we don't break this too badly.

JoeRobich (approved):

PR #77158, opened 2025-02-11
Use the new test framework for missing IDE0060 diagnostics

sharwell:
Also fixes a regression where writes to `a.b` or `a[b]` were not counted as a read of `a`.

### DIFF
--- a/src/Analyzers/CSharp/Tests/RemoveUnusedParametersAndValues/RemoveUnusedParametersTests.cs
+++ b/src/Analyzers/CSharp/Tests/RemoveUnusedParametersAndValues/RemoveUnusedParametersTests.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


src/Analyzers/CSharp/Tests/RemoveUnusedParametersAndValues/RemoveUnusedParametersTests.cs
```diff
@@ -1241,14 +1265,50 @@ void M3(int p5)
...
+                {
+                    var x = p[0];
+                }
+            }
+            """);
+    }
+
+    [Theory]
+    [InlineData("int[]")]
+    [InlineData("Span<int>", Skip = "https://github.com/dotnet/sdk/pull/46546")]
```
sharwell, line 1288: @CyrusNajmabadi This is the regression we just assigned to you

sharwell: ➡️ I fixed this regression in 0b8cd8b

arunchndr (approved):

CyrusNajmabadi (approved):

PR #77168, opened 2025-02-12
Reduce allocations in SymbolCompletionItem.GetSupportedPlatforms

ToddGrun:
This method shows up as 0.8% of VS allocations during typing in the CSharpEditingTests.ScrollingAndTyping speedometer test. These changes should cut this in under half.

This method was previously using string.Split, which allocates an int[] for processing and allocates a return array. Instead, we can roll our own here and avoid both of those allocations

Highlighted allocations from below should be removed
image


ToddGrun:
@dotnet/roslyn-ide -- ptal -- small, simple, allocation improvement in completion

JoeRobich (approved):

PR #77171, opened 2025-02-12
Collection expressions: add `params` cases to tests for invalid builder methods

cston:
Add `params` cases to existing tests for invalid *collection builder* methods, since error reporting for collection builder methods may be affected by changes for *collection arguments* (see https://github.com/dotnet/roslyn/pull/76998).

This is a test-only change.


cston:
Merged into `features/dictionary-expressions` in https://github.com/dotnet/roslyn/pull/76998.

PR #73239, opened 2024-04-25
Resolve extension members in all non-invocation contexts

jcouv:
When we bind a member access expression there are two cases:
1. the expression is considered "invoked" (syntactically part of an `invocation_expression`), in which case the extension member lookup will filter out non-invocable members,
2. the expression is not considered "invoked", so if member lookup failed, we can go ahead and use normal extension member lookup (ie. not filtering out non-invocable members) to resolve the expression to a non-method extension member.

This PR implements the second case.

It also removes previous logic we had for resolving to a non-method extension member during a conversion. That logic relied on an incorrect understanding of the spec (that conversions to a delegate type would cause the expression to be treated as "invoked", so we had to wait until processing the conversion to perform the extension member lookup). 

So the following scenario becomes an error:
```
var x = (System.Action)D.f; // binds to field, so error: string cannot convert to Action
System.Action a = D.f; // ditto

class D { }
 
implicit extension E1 for D
{
    public static string f = null;
}
 
implicit extension E2 for object
{
    public static void f() { }
}
```

----

Here are the relevant sections of the spec that justify this understanding:

The [Member Lookup](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#125-member-lookup) has concept of a member being "invoked":
> "If the simple_name or member_access occurs as the primary_expression of an invocation_expression, the member is said to be invoked."
> "Next, if the member is invoked, all non-invocable members are removed from the set."

But in [Method group conversions](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/conversions.md#108-method-group-conversions), although treat the conversion as an invocation:
> A single method M is selected corresponding to a method invocation of the form `E(A)` ... [using arguments constructed from the parameter types of delegate type]

The method group conversion only kicks in when we have already determined that we have a method group:
> "An implicit conversion exists from a method group ([§12.2](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#122-expression-classifications)) to a compatible delegate type ([§20.4](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/delegates.md#204-delegate-compatibility)). If D is a delegate type, and E is an expression that is classified as a method group, ..."

In short, a conversion to a delegate type does not cause the expression to be treated as "invoked", although we process it somewhat like an invocation.

Relates to test plan https://github.com/dotnet/roslyn/issues/66722

PR #68629, opened 2023-06-14
Implement "extension invocation" section for methods

jcouv:
This PR includes:
- Add handling of methods and other members from extension types in "extension invocations"
- In the type handling case in "member access", we remove "Otherwise, `E.I` is an invalid member reference, and a compile-time error occurs.", so that we continue to extension invocation in the invocation scenario and to extension member lookup in the non-invocation scenario.
- ~~Add extension methods to "extension member lookup".~~ (this will be in a later PR, the rules need to be refined first)

The updated design goes as follows:
- an empty result from a lookup is represented by an empty method group (as before, but now you can also get that from a lookup on a type receiver)
- such empty result can trigger an extension resolution (including both extension methods and extension type members) in multiple places: conversion, `BindToNaturalType`, `CheckValue`, ...
- an extension resolution may yield a method group (like before) or a single extension member. In the latter case, we replace the empty method group with a bound node accessing that member.

Known issues:
- some diagnostics could be improved
- some places use `BindValue` but don't immediately convert or `BindToNaturalType`. Those places do not properly account for the resolved extension members. For example: `new [] { object.Member }` (expected to work)
- function type inference needs to be updated to account for extension members
- `GetMemberGroup` API from semantic model need to be updated to account for extension members
- Filed a [spec issue](https://github.com/dotnet/csharpstandard/issues/820) for current method group conversion rules.
- Filed https://github.com/dotnet/roslyn/issues/68796

Spec: https://github.com/jcouv/csharplang/blob/roles-spec/proposals/extensions.md#extension-invocations
Test plan https://github.com/dotnet/roslyn/issues/66722

```mermaid
graph LR;
    BindMethodGroupInvocation-->ResolveMethodGroup;
    BindInvocationExrepssion-->BindMethodGroupInvocation;
    MakeDeconstructionInvocationExpression-->BindMethodGroupInvocation;
    TryBindIndexOrRangeImplicitIndexerParts-->BindMethodGroupInvocation;
    PerformPatternMethodLookup["PerformPatternMethodLookup (Dispose and Fixed)"]-->BindMethodGroupInvocation;
    CheckValue-->ResolveMethodGroup;
    MakeQueryInvocation-->ResolveMethodGroup;
    MethodGroupReturnType-->ResolveMethodGroup;
    EnsureNameofExpressionSymbols-->ResolveMethodGroup;
    BindMemberAccessWithBoundLeft-->ResolveMethodGroup;
[...]

PR #77172, opened 2025-02-12
Extensions: Bind invocations and member accesses

jcouv:
This is mostly re-using logic from the roles branch. 
The core change/method is `ResolveExtension` (in `Binder_Expression.cs`) and the various callers. 
It's used when resolving method groups (binding invocations, MakeMemberAccessValue, ...) and binding member accesses (BindMemberAccessWithBoundLeft and BindInstanceMemberAccess)
The `MethodGroupResolution` which it returns can now also represent a non-method result (ie. we found a property instead of methods).

Function type scenarios are out-of-scope for this PR.

----

Note that we don't need to resolve to a non-method extension member during conversions.
Here are the relevant sections of the spec that justify this understanding:

The [Member Lookup](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#125-member-lookup) has concept of a member being "invoked":
> "If the simple_name or member_access occurs as the primary_expression of an invocation_expression, the member is said to be invoked."
> "Next, if the member is invoked, all non-invocable members are removed from the set."

But in [Method group conversions](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/conversions.md#108-method-group-conversions), although treat the conversion as an invocation:
> A single method M is selected corresponding to a method invocation of the form `E(A)` ... [using arguments constructed from the parameter types of delegate type]

The method group conversion only kicks in when we have already determined that we have a method group:
> "An implicit conversion exists from a method group ([§12.2](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#122-expression-classifications)) to a compatible delegate type ([§20.4](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/delegates.md#204-delegate-compatibility)). If D is a delegate type, and E is an expression that is classified as a method group, ..."

In short, a conversion to a delegate type does not cause the expression to be treated as "invoked", although we process it somewhat like an invocation.


src/Compilers/CSharp/Portable/Binder/Binder_Query.cs
```diff
@@ -928,6 +928,8 @@ protected BoundCall MakeQueryInvocation(CSharpSyntaxNode node, BoundExpression r
             Debug.Assert(receiver.Type is object || ultimateReceiver.Type is null);
             if ((object?)ultimateReceiver.Type == null)
             {
+                Debug.Assert(ultimateReceiver.Kind != BoundKind.MethodGroup || ultimateReceiver.HasAnyErrors);
```
jcouv, line 931: :memo: See previous discussion on this change [https://github.com/dotnet/roslyn/pull/73239#discussion_r1586625386](https://github.com/dotnet/roslyn/pull/73239#discussion_r1586625386)

src/Compilers/CSharp/Portable/Binder/Binder_Conversions.cs
```diff
@@ -1367,7 +1386,13 @@ static bool bindMethodGroupInvocation(
 
                 if (!methodGroup.HasAnyErrors) diagnostics.AddRange(resolution.Diagnostics); // Suppress cascading.
 
-                if (resolution.HasAnyErrors)
+                if (resolution.IsNonMethodExtensionMember(out Symbol? extensionMember))
+                {
+                    ReportMakeInvocationExpressionBadMemberKind(syntax, WellKnownMemberNames.CollectionInitializerAddMethodName, methodGroup, diagnostics);
```
AlekseyTs, line 1391: >ReportMakeInvocationExpressionBadMemberKind(syntax, WellKnownMemberNames.CollectionInitializerAddMethodName, methodGroup, diagnostics); [](http://example.com/codeflow?start=20&length=135)

It is not obvious why this is necessary an error. For example, if a property returns a delegate, the property access is a valid primary expression for an invocation.

src/Compilers/CSharp/Portable/Symbols/Source/SourceNamedTypeSymbol.cs
```diff
@@ -1975,7 +1975,6 @@ protected override void AfterMembersCompletedChecks(BindingDiagnosticBag diagnos
                     // The type 'Microsoft.CodeAnalysis.EmbeddedAttribute' must be non-generic, internal, sealed, non-static, have a parameterless constructor, inherit from System.Attribute, and be able to be applied to any type.
                     diagnostics.Add(ErrorCode.ERR_EmbeddedAttributeMustFollowPattern, GetFirstLocation());
                 }
-
```
AlekseyTs, line 1978: It doesn't look like there are meaningful changes in this file. Consider reverting it to its previous state.

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
```diff
@@ -176,6 +176,107 @@ protected void LookupMembersInternal(LookupResult result, NamespaceOrTypeSymbol
...
+#if !DEBUG
+            // In DEBUG mode, we prefer to exercise the code below even in the absence of extensions
+            if (extensions.Count == 0)
+            {
+                return;
+            }
+#endif
+
+            PooledHashSet<NamedTypeSymbol>? visited = null;
+            var baseType = type;
```
AlekseyTs, line 230: > var baseType = type; [](http://example.com/codeflow?start=11&length=21)

A couple of comments:
1. It looks like we are not handling scenario when ```type``` is a generic type parameter.
2. It is not clear to me why we are not using the same approach that ```ReducedExtensionMethodSymbol.Create``` use to check compatibility and a general type argument inference algorithm like in ```ReducedExtensionMethodSymbol.InferExtensionMethodTypeArguments```. We have a parameter and we have a type for the corresponding argument. What is the reason to use different approach?

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -7887,17 +7888,32 @@ BoundExpression tryBindMemberAccessWithBoundTypeLeft(
...
                 else
                 {
                     CompoundUseSiteInfo<AssemblySymbol> useSiteInfo = GetNewCompoundUseSiteInfo(diagnostics);
                     this.LookupMembersWithFallback(lookupResult, leftType, rightName, rightArity, ref useSiteInfo, basesBeingResolved: null, options: options);
                     diagnostics.Add(right, useSiteInfo);
+
                     if (lookupResult.IsMultiViable)
                     {
-                        return BindMemberOfType(node, right, rightName, rightArity, indexed, boundLeft, typeArgumentsSyntax, typeArguments, lookupResult, BoundMethodGroupFlags.None, diagnostics: diagnostics);
+                        return BindMemberOfType(node, right, rightName, rightArity, indexed, boundLeft, typeArgumentsSyntax, typeArguments, lookupResult, BoundMethodGroupFlags.SearchExtensions, diagnostics: diagnostics);
```
AlekseyTs, line 7901: >BoundMethodGroupFlags.SearchExtensions [](http://example.com/codeflow?start=154&length=38)

Should this be conditional? Perhaps, based on ```options```?

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -7887,17 +7888,32 @@ BoundExpression tryBindMemberAccessWithBoundTypeLeft(
...
+                            typeArgumentsSyntax, typeArguments, diagnostics);
+
+                        if (nonMethodExtensionMember is not null)
+                        {
+                            return nonMethodExtensionMember;
+                        }
                     }
+
+                    return MakeBoundMethodGroupAndCheckOmittedTypeArguments(boundLeft, rightName, typeArguments, lookupResult,
+                        flags: BoundMethodGroupFlags.SearchExtensions, node, typeArgumentsSyntax, diagnostics);
```
AlekseyTs, line 7916: >BoundMethodGroupFlags.SearchExtensions [](http://example.com/codeflow?start=31&length=38)

Similar question.

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
jcouv: The PR re-uses the algorithm from the roles branch as we'd discussed. But in email I suggested that we use method type inference here instead of type unification, which is what `ReducedExtensionMethodSymbol` does. I was planning to do that in a subsequent PR, but am happy to include that change here.

Yes, this PR doesn't handle type parameter receivers (it's explicitly blocked in places, marked with PROTOTYPE). I probably can revisit that after change to type inference.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -7952,7 +7969,46 @@ private BoundExpression MakeMemberAccessValue(BoundExpression expr, BindingDiagn
...
+        // - if the extension member lookup finds nothing, then we return nothing and let the caller represent the failed member lookup with a BoundMethodGroup.
+        internal BoundExpression? ResolveExtensionMemberAccessIfResultIsNonMethod(SyntaxNode syntax, BoundExpression receiver, string name,
+            SeparatedSyntaxList<TypeSyntax> typeArgumentsSyntax, ImmutableArray<TypeWithAnnotations> typeArgumentsOpt,
+            BindingDiagnosticBag diagnostics)
+        {
+            CompoundUseSiteInfo<AssemblySymbol> useSiteInfo = GetNewCompoundUseSiteInfo(diagnostics);
+            // Note: we're resolving without arguments, which means we're not treating the member access as invoked
+            var resolution = this.ResolveExtension(
+                syntax, name, analyzedArguments: null, receiver, typeArgumentsOpt, options: OverloadResolution.Options.None,
+                returnRefKind: default, returnType: null, withDependencies: useSiteInfo.AccumulatesDependencies);
```
AlekseyTs, line 7993: >useSiteInfo.AccumulatesDependencies [](http://example.com/codeflow?start=76&length=35)

It feels like an overkill to create ```useSiteInfo``` just to use this property. We should be able to get this information directly from ```diagnostics```.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -7952,7 +7969,46 @@ private BoundExpression MakeMemberAccessValue(BoundExpression expr, BindingDiagn
...
+
+                return GetExtensionMemberAccess(syntax, receiver, extensionMember, diagnostics);
+            }
+
+            resolution.Free();
+            return null;
+        }
+#nullable disable
+
+        private BoundExpression BindMemberAccessWithBoundLeftCore(
```
AlekseyTs, line 8011: >BindMemberAccessWithBoundLeftCore [](http://example.com/codeflow?start=32&length=33)

It is not clear what motivated the rename. Especially that "Core" suffix isn't rely informative. Is there "non-Core" version? What does that do vs. this one doesn't?

AlekseyTs: Was the "Instance" part of the original name inaccurate? I think we need to come up with a clearer, more descriptive name.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -8389,40 +8472,182 @@ protected MethodGroupResolution BindExtensionMethod(
...
+                    // (regardless of whether a best applicable method could be determined) then our search is complete.
+                    // Otherwise, store aside the first non-applicable result and continue searching for an applicable result.
+                    if (extensionResult.HasAnyApplicableMethod)
+                    {
+                        firstResult.Free();
+                        return extensionResult;
+                    }
+                    else if (firstResult.IsEmpty)
+                    {
+                        firstResult = extensionResult;
```
AlekseyTs, line 8503: >firstResult = extensionResult; [](http://example.com/codeflow?start=24&length=30)

We don't need to free ```firstResult``` here?

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -8389,40 +8472,182 @@ protected MethodGroupResolution BindExtensionMethod(
...
+                    options, returnRefKind, returnType, in callingConvention, withDependencies, scope,
+                    ref actualArguments, out MethodGroupResolution extensionMethodResult))
+                {
+                    // If the search in the current scope resulted in any applicable method (regardless of whether a best
+                    // applicable method could be determined) then our search is complete. Otherwise, store aside the
+                    // first non-applicable result and continue searching for an applicable result.
+                    if (extensionMethodResult.HasAnyApplicableMethod)
+                    {
+                        firstResult.Free();
+                        return extensionMethodResult;
```
AlekseyTs, line 8522: >return extensionMethodResult; [](http://example.com/codeflow?start=24&length=29)

Are we supposed to free ```actualArguments``` here?

src/Compilers/CSharp/Portable/Binder/Binder_Invocation.cs
```diff
@@ -715,6 +715,20 @@ private BoundExpression BindMethodGroupInvocation(
...
+
+            if (resolution.IsNonMethodExtensionMember(out Symbol extensionMember))
+            {
+                diagnostics.AddRange(resolution.Diagnostics);
+                BoundExpression extensionMemberAccess = GetExtensionMemberAccess(expression, methodGroup.ReceiverOpt, extensionMember, diagnostics);
+
+                Debug.Assert(extensionMemberAccess.Kind != BoundKind.MethodGroup);
+
+                extensionMemberAccess = CheckValue(extensionMemberAccess, BindValueKind.RValue, diagnostics);
+                BoundExpression extensionMemberInvocation = BindInvocationExpression(syntax, expression, methodName, extensionMemberAccess, analyzedArguments, diagnostics);
```
AlekseyTs, line 727: >methodName [](http://example.com/codeflow?start=105&length=10)

This looks suspicious, it looks like we already resolved the name.

src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversions.cs
```diff
@@ -262,8 +265,11 @@ private static MethodGroupResolution ResolveDelegateOrFunctionPointerMethodGroup
             }
             else
             {
-                return binder.ResolveMethodGroup(source, analyzedArguments: null, ref useSiteInfo, options: OverloadResolution.Options.IsMethodGroupConversion);
+                resolution = binder.ResolveMethodGroup(source, analyzedArguments: null, ref useSiteInfo, options: OverloadResolution.Options.IsMethodGroupConversion);
             }
+
+            Debug.Assert(!resolution.IsNonMethodExtensionMember(out _));
+            return resolution;
```
AlekseyTs, line 272: Did you mean to remove ```return resolution;``` above? Otherwise this code can be moved to the `else` block.

src/Compilers/CSharp/Test/IOperation/IOperation/IOperationTests_InvalidExpression.cs
```diff
@@ -33,9 +33,7 @@ static void Main(string[] args)
             string expectedOperationTree = @"
 IInvalidOperation (OperationKind.Invalid, Type: ?, IsInvalid) (Syntax: 'Console.WriteLine2()')
   Children(1):
-      IInvalidOperation (OperationKind.Invalid, Type: ?, IsInvalid) (Syntax: 'Console.WriteLine2')
```
AlekseyTs, line 36: >IInvalidOperation (OperationKind.Invalid, Type: ?, IsInvalid) (Syntax: 'Console.WriteLine2') [](http://example.com/codeflow?start=6&length=92)

Do we understand why this node disappeared?

AlekseyTs:

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
jcouv: From offline discussion, yes the issue was with "Instance". The bound left may not be an instance. I'll take a stab at improving the name

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
jcouv: `firstResult.IsEmpty` indicates that don't have anything to free (in particular, no analyzed arguments). I'll add an assertion to clarify.

src/Compilers/CSharp/Test/IOperation/IOperation/IOperationTests_InvalidExpression.cs
jcouv: Yes, it is due to the change in `tryBindMemberAccessWithBoundTypeLeft` (Binder_Expression.cs line 7916)
See explanation: [https://github.com/dotnet/roslyn/pull/68629#discussion_r1253432174](https://github.com/dotnet/roslyn/pull/68629#discussion_r1253432174)

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
jcouv: I think the flag should be set unconditionally, unless we can identify a type member access scenario where we shouldn't resolve extensions.

In short:

Previously, we'd only set the flag in instance member access scenarios. It would generally indicate that the member group came from an instance member access, with a couple of limited exceptions (see below).
Now, we also want to set this flag for method groups resulting for a type member access. The previous exceptions are not applicable and we haven't found new exceptions that would require using this flag conditionally here.
Setting this flag on type member access enables invocation scenarios (`ResolveMethodGroup` should call `ResolveExtensions`, see `methodGroup.SearchExtensions` check) and natural function type scenarios (`GetUniqueSignatureFromMethodGroup` should inspect extensions in a future PR).

Note: although `LookupOptions` does have an `IncludeExtensionMethods` flag, I don't think it's relevant here (it's only used for lookup APIs in the semantic model). The `options` chosen in `BindMemberAccessWithBoundLeft` don't need to use it.

My thought process:

This flag gets stored into the `BoundMethodGroup` in cases where one is returned:
- in `BindMemberOfType`, we either return a method group with this flag (when a method group is found or no result), or a non-method result such as a type, property access, event access, field access (which don't use this flag)
- in `MakeBoundMethodGroupAndCheckOmittedTypeArguments` always returns a method group with this flag

The flag indicates that subsequent usage of the method group should look at extensions, because it came from a member access where we want to allow extension resolution. 
1. if we don't have a member access, we don't use this flag. For example, we don't use it when binding an identifier (see `ConstructBoundMemberGroupAndReportOmittedTypeArguments` call in `BindIdentifier`).
2. for a type receiver, we previously didn't set this flag
3. for the remaining cases, we generally set the flag for a member access with two exceptions: in `BindMemberAccessWithBoundLeftTypeOrInstance` if `!searchExtensionsIfNecessary` (only used ` for binding interpolation handlers, which purposefully exclude extension resolution) and `leftIsBaseReference` (no extension resolution on `base` access).

We want to set the flag for case 2 and have not yet identified any exceptions for that scenario.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
jcouv: I ended up reverting this. I still think the name isn't great, but it's not essential for our present purpose

jcouv:
>         comp.VerifyEmitDiagnostics(

:memo: Would like to discuss this scenario. We have a use-site diagnostic (error) from the constraint, but we're dropping it because `LookupExtensionMembersInSingleBinder` returns no candidate method.

---
Refers to: src/Compilers/CSharp/Test/Emit3/Semantics/ExtensionTests.cs:5951 in e98d317. [](commit_id = e98d31766bf534a916befdf4e8e6279af2c85182, deletion_comment = False)

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
```diff
@@ -176,6 +176,133 @@ protected void LookupMembersInternal(LookupResult result, NamespaceOrTypeSymbol
...
+                bool success = checkConstraints(extension, extension.TypeParameters, typeArguments, compilation, conversions, ref useSiteInfo);
+                if (!success)
+                {
+                    return null;
+                }
+
+                return extension.Construct(typeArguments);
+            }
+
+            static bool checkConstraints(Symbol symbol, ImmutableArray<TypeParameterSymbol> typeParams, ImmutableArray<TypeWithAnnotations> typeArgs,
```
AlekseyTs, line 279: >Symbol [](http://example.com/codeflow?start=41&length=6)

NamedTypeSymbol?

AlekseyTs, line 279: >ImmutableArray<TypeParameterSymbol> typeParams [](http://example.com/codeflow?start=56&length=46)

It looks like this parameter is redundant, we can get type parameters from extension type.

src/Compilers/CSharp/Portable/Symbols/NativeIntegerTypeSymbol.cs
```diff
@@ -184,6 +184,8 @@ internal override UseSiteInfo<AssemblySymbol> GetUseSiteInfo()
         internal sealed override bool IsRecordStruct => false;
         internal sealed override bool HasPossibleWellKnownCloneMethod() => false;
 
+        internal sealed override ParameterSymbol ExtensionParameter => throw ExceptionUtilities.Unreachable();
```
AlekseyTs, line 187: >throw ExceptionUtilities.Unreachable() [](http://example.com/codeflow?start=71&length=38)

It is probably safe to return `null` as we do in other similar cases.

AlekseyTs:
>             // (3,1): error CS0012: The type 'Missing' is defined in an assembly that is not referenced. You must add a reference to assembly 'missing, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.

I am not sure how this error is relevant to the compiled scenario.

---
Refers to: src/Compilers/CSharp/Test/Emit3/Semantics/ExtensionTests.cs:6039 in 8462cf7. [](commit_id = 8462cf795ca9eb239297077d87f2368295ce51cd, deletion_comment = False)

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -8389,40 +8469,179 @@ protected MethodGroupResolution BindExtensionMethod(
...
+
+            Debug.Assert((actualArguments == null) || !firstResult.IsEmpty);
+            actualArguments?.Free();
+            diagnostics.Free();
+
+            return firstResult;
+
+            static bool tryResolveExtensionMember(Binder binder, SyntaxNode expression, string memberName, AnalyzedArguments? analyzedArguments, BoundExpression left,
+                ImmutableArray<TypeWithAnnotations> typeArgumentsWithAnnotations, OverloadResolution.Options options, RefKind returnRefKind, in CallingConventionInfo callingConvention,
+                TypeSymbol? returnType, BindingDiagnosticBag diagnostics, ExtensionScope scope, out MethodGroupResolution extensionResult)
```
AlekseyTs, line 8543: >BindingDiagnosticBag diagnostics [](http://example.com/codeflow?start=40&length=32)

This approach doesn't feel right. Why would we want to attach an unrelated diagnostics to ```MethodGroupResolution``` that we create. The ```diagnostics``` probably contains errors from previous iterations of the loop above, i.e. a different extension.

AlekseyTs:

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -8389,40 +8469,179 @@ protected MethodGroupResolution BindExtensionMethod(
...
+                }
 
-            foreach (var scope in new ExtensionMethodScopes(this))
+                CompoundUseSiteInfo<AssemblySymbol> useSiteInfo = binder.GetNewCompoundUseSiteInfo(diagnostics);
+
+                scope.Binder.LookupExtensionMembersInSingleBinder(
+                    lookupResult, left.Type!, memberName, arity,
+                    basesBeingResolved: null, lookupOptions, originalBinder: binder, ref useSiteInfo);
+
+                diagnostics.Add(expression, useSiteInfo);
```
AlekseyTs, line 8560: >diagnostics.Add(expression, useSiteInfo); [](http://example.com/codeflow?start=16&length=41)

I think this is where the problem is. This is where we are dropping use-site in case diagnostics is dropped below. And the diagnostics is dropped for a good reason. I think use-site diagnostics should be collected separately on the ```ResolveExtension``` level. I.e. it should be passed in as a parameter and we shouldn't be merging it into diagnostic bags that might be dropped.

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -8368,18 +8428,38 @@ private BoundExpression BindMemberOfType(
...
-            AnalyzedArguments analyzedArguments,
+            string memberName,
+            AnalyzedArguments? analyzedArguments,
             BoundExpression left,
             ImmutableArray<TypeWithAnnotations> typeArgumentsWithAnnotations,
             OverloadResolution.Options options,
             RefKind returnRefKind,
-            TypeSymbol returnType,
+            TypeSymbol? returnType,
             bool withDependencies,
```
AlekseyTs, line 8459: >bool withDependencies, [](http://example.com/codeflow?start=12&length=22)

Here, I think that instead of passing ```withDependencies```, we should pass ```ref CompoundUseSiteInfo<AssemblySymbol> useSiteInfo```. And keep passing it down and collecting use-site diagnostics into it when lookup might be failing because of them, and in other similar cases.

AlekseyTs:
>         Assert.Equal("void N.E2.<>E__0<Container>.M()", model.GetSymbolInfo(memberAccess).Symbol.ToTestDisplayString());

This change is not expected

---
Refers to: src/Compilers/CSharp/Test/Emit3/Semantics/ExtensionTests.cs:6036 in bc0b7ed. [](commit_id = bc0b7ed2cf47921c27e9583a93625cb391da1e44, deletion_comment = True)

AlekseyTs:
>         extension<T>(T t)

Consider testing without this extension. We should still get use-site error.

---
Refers to: src/Compilers/CSharp/Test/Emit3/Semantics/ExtensionTests.cs:6030 in 8462cf7. [](commit_id = 8462cf795ca9eb239297077d87f2368295ce51cd, deletion_comment = False)

AlekseyTs:
>                 diagnostics.Add(expression, useSiteInfo); // PROTOTYPE test use-site info

It feels like this place shouldn't change. We are not dropping ```diagnostics``` and it is specific to the resolution that we return.

---
Refers to: src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs:8681 in 8462cf7. [](commit_id = 8462cf795ca9eb239297077d87f2368295ce51cd, deletion_comment = True)

jcouv:
>                 diagnostics.Add(expression, useSiteInfo); // PROTOTYPE test use-site info

Let's discuss next week. I don't understand why we would want some use-site diagnostics attached to the result, but not others.

I think we should treat all use-site diagnostics in `ResolveExtension` the same. 
Since the caller in an invocation (`ResolveMethodGroupInternal`) may choose to use the default method group result or the extension method group, and it's caller (`ResolveMethodGroup`) has further logic in case the method resolution has no errors, so I think we should collect all the use-site diagnostics from `ResolveExtension` and attach them all to the method resolution that we end up producing.

---
In reply to: [2660477061](https://github.com/dotnet/roslyn/pull/77172#issuecomment-2660477061) [](http://example.com/codeflow?ancestors=2660477061)

---
Refers to: src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs:8681 in 8462cf7. [](commit_id = 8462cf795ca9eb239297077d87f2368295ce51cd, deletion_comment = True)

AlekseyTs:
>                 diagnostics.Add(expression, useSiteInfo); // PROTOTYPE test use-site info

There is a good rational behind this argument. At the same time it looks like ```MethodInvocationOverloadResolution``` doesn't add use-site diagnostics for "bad" candidates into ```useSiteInfo``` parameter. Therefore, it appears, if we propagate it separately, we will collect only some of the info, not all of it. I guess, it is worth following up on this issue later. 

---
In reply to: [2661981228](https://github.com/dotnet/roslyn/pull/77172#issuecomment-2661981228) [](http://example.com/codeflow?ancestors=2661981228,2660477061)

---
Refers to: src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs:8681 in 8462cf7. [](commit_id = 8462cf795ca9eb239297077d87f2368295ce51cd, deletion_comment = True)

AlekseyTs (approved):
LGTM (commit 8)

jcouv:
@jjonescz for second review. Thanks

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
```diff
@@ -176,6 +176,134 @@ protected void LookupMembersInternal(LookupResult result, NamespaceOrTypeSymbol
             }
         }
 
+#nullable enable
+        protected void LookupExtensionMembersInSingleBinder(LookupResult result, TypeSymbol receiverType,
```
jjonescz, line 180:
```suggestion
        private void LookupExtensionMembersInSingleBinder(LookupResult result, TypeSymbol receiverType,
```

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
```diff
@@ -760,6 +888,19 @@ internal virtual void GetCandidateExtensionMethods(
...
 
+#nullable enable
+        /// <summary>
+        /// Return the extension containers from this specific binding scope
+        /// Since the lookup of extension members is iterative, proceeding one binding scope at a time,
+        /// GetExtensionContainers should not defer to the next binding scope. Instead, the caller is
+        /// responsible for walking the nested binding scopes from innermost to outermost. This method is overridden
+        /// to search the available members list in binding types that represent types, namespaces, and usings.
+        /// </summary>
+        internal virtual void GetExtensionContainers(ArrayBuilder<NamedTypeSymbol> extensions, Binder originalBinder)
```
jjonescz, line 899: Should the terminology here be "extension declaration" instead of "extension container"?

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
```diff
@@ -176,6 +176,134 @@ protected void LookupMembersInternal(LookupResult result, NamespaceOrTypeSymbol
...
+
+            var extensions = ArrayBuilder<NamedTypeSymbol>.GetInstance();
+            binder.GetExtensionContainers(extensions, originalBinder);
+
+            foreach (var extension in extensions)
+            {
+                addCompatibleExtension(binder, extension, receiverType, compatibleExtensions, ref useSiteInfo);
+            }
+
+            return;
```
jjonescz, line 224:
```suggestion
            extensions.Free();
            return;
```

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
```diff
@@ -176,6 +176,134 @@ protected void LookupMembersInternal(LookupResult result, NamespaceOrTypeSymbol
...
+
+                diagnosticsBuilder.Free();
+
+                if (useSiteDiagnosticsBuilder != null && useSiteDiagnosticsBuilder.Count > 0)
+                {
+                    foreach (TypeParameterDiagnosticInfo diagnostic in useSiteDiagnosticsBuilder)
+                    {
+                        useSiteInfo.Add(diagnostic.UseSiteInfo);
+                    }
+                }
```
jjonescz, line 300: Free `useSiteDiagnosticsBuilder`?

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
```diff
@@ -176,6 +176,134 @@ protected void LookupMembersInternal(LookupResult result, NamespaceOrTypeSymbol
...
+                // PROTOTYPE we should prefer extension members that apply to a more specific type
+                result.MergeEqual(tempResult);
+                tempResult.Clear();
+            }
+
+            tempResult.Free();
+            compatibleExtensions.Free();
+        }
+
+        private static void GetCompatibleExtensions(Binder binder, TypeSymbol receiverType, ArrayBuilder<NamedTypeSymbol> compatibleExtensions,
```
jjonescz, line 207: nit: looks like this could be a local function as well

src/Compilers/CSharp/Portable/Symbols/SubstitutedNamedTypeSymbol.cs
```diff
@@ -487,5 +492,35 @@ internal sealed override bool HasInlineArrayAttribute(out int length)
         }
 
         internal sealed override bool HasCompilerLoweringPreserveAttribute => _underlyingType.HasCompilerLoweringPreserveAttribute;
+
+#nullable enable
+        internal override ParameterSymbol? ExtensionParameter
```
jjonescz, line 497:
```suggestion
        internal sealed override ParameterSymbol? ExtensionParameter
```

src/Compilers/CSharp/Portable/Symbols/TypeSymbol.cs
```diff
@@ -555,9 +555,16 @@ public virtual bool IsAnonymousType
         /// </summary>
         internal virtual bool IsNativeIntegerWrapperType => false;
 
+#nullable enable
         public bool IsExtension
             => TypeKind == TypeKind.Extension;
 
+        /// <summary>
+        /// For the type representing an extension container, returns the receiver parameter symbol.
```
jjonescz, line 563:
```suggestion
        /// For the type representing an extension declaration, returns the receiver parameter symbol.
```

src/Compilers/CSharp/Test/Symbol/Symbols/TypeUnificationTests.cs
```diff
@@ -94,9 +94,6 @@ void M() { }
                     }
                 }
             }
-
-            AssertCanUnify(null, null);
-            AssertCannotUnify(classType, null);
         }
```
jjonescz, line 97: Why have these asserts been removed?

jjonescz (approved):

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -10550,10 +10772,10 @@ private MethodGroupResolution ResolveDefaultMethodGroup(
                 }
             }
 
-            if (node.SearchExtensionMethods)
+            if (node.ReceiverOpt is not BoundTypeExpression && node.SearchExtensions)
```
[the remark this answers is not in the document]
jcouv: It's not supported yet. The function type scenarios are out-of-scope for this PR. 
When given a type receiver, we will want to find new extension members but not classic extension methods. For now, the whole logic (related to classic extension methods) is skipped.

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
jcouv: Good catch. Thanks

src/Compilers/CSharp/Portable/Binder/Binder_Lookup.cs
jcouv: No, that one isn't pooled. See `ensureUseSiteDiagnosticsBuilder` which initializes this when needed.

src/Compilers/CSharp/Test/Symbol/Symbols/TypeUnificationTests.cs
jcouv: An earlier iteration of this PR included type unification changes. Will revert this test file too. Thanks

src/Compilers/CSharp/Portable/Binder/Binder_Expressions.cs
```diff
@@ -7952,6 +7969,46 @@ private BoundExpression MakeMemberAccessValue(BoundExpression expr, BindingDiagn
...
+        //   closer than any non-method extension member), then we return nothing and let the caller represent the failed member lookup with a BoundMethodGroup.
+        //   Note: Such method group will be resolved specially in scenarios that can handle method groups
+        //     (such as inferred local `var x = A.B;`, conversion to a delegate type `System.Action a = A.B;`).
+        //     It will be an error in other scenarios.
+        //
+        // - if the extension member lookup is ambiguous, then we'll use an error symbol as the result of the member access.
+        //
+        // - if the extension member lookup finds nothing, then we return nothing and let the caller represent the failed member lookup with a BoundMethodGroup.
+        internal BoundExpression? ResolveExtensionMemberAccessIfResultIsNonMethod(SyntaxNode syntax, BoundExpression receiver, string name,
+            SeparatedSyntaxList<TypeSyntax> typeArgumentsSyntax, ImmutableArray<TypeWithAnnotations> typeArgumentsOpt,
```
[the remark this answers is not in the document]
jcouv: That was a remnant from roles branch where we needed both for binding extension nested types. Simplified

PR #77173, opened 2025-02-12
Multi-target LSP tests to run against .NET

dibarbet:
Previously, all LSP tests ran against net472 only, even though the LSP code running in VSCode ran against .NET9.  So we've had a test gap here for a while.

But some refactorings Tomas did a couple weeks ago to TestWorkspace now make it relatively straightforward to multi-target these tests.  This gets us both CI coverage on .NET core as well as the ability to debug these tests using VSCode (it only ships with a .net core debugger).

The changes in the PR fit a few categories
1.  Removing EditorFeatures dependencies from the Protocol unit tests.  This generally resulted in removing usages of EditorTestWorkspace and text buffers.  Text buffers were previously mostly used as a proxy for opening a file in the workspace, and didn't actually require text buffers.
2.  Fixing new nullable warnings that came from bumping the tfm
3.  Refactoring the TestLspServer type a little bit to allow either the `LspTestWorkspace` or `EditorTestWorkspace` (for tests in EditorFeatures that require text buffers, like the document outline tests).

CI:
image


src/EditorFeatures/TestUtilities/Workspaces/EditorTestWorkspace.cs
```diff
@@ -31,44 +31,26 @@
 
 namespace Microsoft.CodeAnalysis.Test.Utilities;
 
-public partial class EditorTestWorkspace : TestWorkspace<EditorTestHostDocument, EditorTestHostProject, EditorTestHostSolution>, ILspWorkspace
+public partial class EditorTestWorkspace : TestWorkspace<EditorTestHostDocument, EditorTestHostProject, EditorTestHostSolution>
```
dibarbet, line 34: we're not using this anymore for any LSP tests that require mutation, so no need for it to implement the LSP mutation methods.  Only used by doc outline tests at the moment (which work with any workspace).

src/LanguageServer/Protocol.TestUtilities/LanguageServer/AbstractLanguageServerProtocolTests.cs
```diff
@@ -108,7 +100,7 @@ private protected class OrderLocations : Comparer<LSP.Location?>
             public override int Compare(LSP.Location? x, LSP.Location? y) => CompareLocations(x, y);
         }
 
-        protected virtual TestComposition Composition => EditorFeaturesLspComposition;
+        protected virtual TestComposition Composition => FeaturesLspComposition;
```
dibarbet, line 103: we generally don't need editorfeatures in the composition anymore.  Previously it was required to test LSP APIs that returned internal VS types - we would have a default service in the protocol layer, and override it in editor features (to return things like classified text runs).  

However now we have all our own types and the conversions are in the protocol layer too, so this is generally not necessary.

src/LanguageServer/Protocol.TestUtilities/LanguageServer/AbstractLanguageServerProtocolTests.cs
```diff
@@ -390,13 +382,19 @@ internal EditorTestWorkspace CreateWorkspace(
         /// Waits for the async operations on the workspace to complete.
         /// This ensures that events like workspace registration / workspace changes are processed by the time we exit this method.
         /// </summary>
-        protected static async Task WaitForWorkspaceOperationsAsync(EditorTestWorkspace workspace)
+        protected static async Task WaitForWorkspaceOperationsAsync<TDocument, TProject, TSolution>(TestWorkspace<TDocument, TProject, TSolution> workspace)
```
dibarbet, line 385: generics are required so this can be passed an EditorTestWorkspace (for a few tests that use that) or LspTestWorkspace (almost everything else).

src/LanguageServer/Protocol.TestUtilities/LanguageServer/AbstractLanguageServerProtocolTests.cs
```diff
@@ -531,44 +532,71 @@ private static LSP.DidCloseTextDocumentParams CreateDidCloseTextDocumentParams(U
...
+            public static async Task<TestLspServer> CreateAsync(LspTestWorkspace testWorkspace, InitializationOptions initializationOptions, AbstractLspLogger logger)
+            {
+                var locations = await GetAnnotatedLocationsAsync(testWorkspace, testWorkspace.CurrentSolution);
+                var server = new TestLspServer(testWorkspace, locations, initializationOptions, logger);
+                await server.InitializeAsync();
+                return server;
+            }
+        }
+
+        internal abstract class AbstractTestLspServer<TWorkspace, TDocument, TProject, TSolution> : IAsyncDisposable
```
dibarbet, line 555: refactored TestLspServer a bit so it can be passed any implementation of `TestWorkspace<TDocument, TProject, TSolution>`.  mostly this is just the LspTestWorkspace, but as above there are a few cases where we still use EditorTestWorkspace.

src/LanguageServer/Protocol.TestUtilities/LanguageServer/AbstractLanguageServerProtocolTests.cs
```diff
@@ -698,6 +709,33 @@ public async Task OpenDocumentAsync(Uri documentUri, string? text = null, string
                 await ExecuteRequestAsync<LSP.DidOpenTextDocumentParams, object>(LSP.Methods.TextDocumentDidOpenName, didOpenParams, CancellationToken.None);
             }
 
+            /// <summary>
+            /// Opens a document in the workspace only, and waits for workspace operations.
+            /// Use <see cref="OpenDocumentAsync(Uri, string, string)"/> if the document should be opened in LSP"/>
+            /// </summary>
+            public async Task OpenDocumentInWorkspaceAsync(DocumentId documentId, bool openAllLinkedDocuments, SourceText? text = null)
```
dibarbet, line 716: previously, some tests opened text buffers to simulate opening a file in the workspace (and setting things like active document context).

We don't actually need the buffers - instead we can just open the documents in the workspace directly.

src/LanguageServer/ProtocolUnitTests/Diagnostics/PullDiagnosticTests.cs
```diff
@@ -56,18 +56,15 @@ public async Task TestDocumentDiagnosticsForOpenFilesWithFSAOff(bool useVSDiagno
...
         var document = testLspServer.GetCurrentSolution().Projects.Single().Documents.Single();
 
         await OpenDocumentAsync(testLspServer, document);
 
         var results = await RunGetDocumentPullDiagnosticsAsync(
             testLspServer, document.GetURI(), useVSDiagnostics);
 
-        Assert.Equal("CS1513", results.Single().Diagnostics.Single().Code);
-        Assert.NotNull(results.Single().Diagnostics.Single().CodeDescription!.Href);
+        Assert.Equal("CS1513", results.Single().Diagnostics!.Single().Code);
```
dibarbet, line 66: everything in this file is removal of the text buffer calls and nullable updates

src/VisualStudio/CSharp/Test/DocumentOutline/DocumentOutlineTestsBase.cs
```diff
@@ -118,10 +119,25 @@ private async Task<TestLspServer> CreateTestLspServerAsync(EditorTestWorkspace w
             solution = solution.WithAnalyzerReferences([new TestAnalyzerReferenceByLanguage(DiagnosticExtensions.GetCompilerDiagnosticAnalyzersMap())]);
             await workspace.ChangeSolutionAsync(solution);
 
-            var server = await TestLspServer.CreateAsync(workspace, new InitializationOptions(), _logger);
+            var server = await EditorTestLspServer.CreateAsync(workspace, new InitializationOptions(), _logger);
             return server;
         }
 
+        internal class EditorTestLspServer : AbstractTestLspServer<EditorTestWorkspace, EditorTestHostDocument, EditorTestHostProject, EditorTestHostSolution>
```
dibarbet, line 126: these are the tests that still require the EditorTestWorkspace because they deal with text buffers in the tests.

src/EditorFeatures/Test/LanguageServer/VSTypeScriptHandlerTests.cs
```diff
@@ -3,33 +3,38 @@
...
-namespace Microsoft.CodeAnalysis.LanguageServer.UnitTests;
+namespace Microsoft.CodeAnalysis.Editor.UnitTests.LanguageServer;
 public class VSTypeScriptHandlerTests : AbstractLanguageServerProtocolTests
 {
     public VSTypeScriptHandlerTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
     {
     }
 
-    protected override TestComposition Composition => base.Composition.AddParts(typeof(TypeScriptHandlerFactory));
+    protected override TestComposition Composition => EditorTestCompositions.LanguageServerProtocolEditorFeatures
```
dibarbet, line 35: requires Editor features here as the VS typescript EA components live in editor features.

src/EditorFeatures/Test/LanguageServer/VSTypeScriptHandlerTests.cs
```diff
@@ -3,33 +3,38 @@
...
+using Microsoft.CommonLanguageServerProtocol.Framework;
 using Nerdbank.Streams;
+using Roslyn.LanguageServer.Protocol;
 using Roslyn.Test.Utilities;
 using StreamJsonRpc;
 using Xunit;
 using Xunit.Abstractions;
 
-namespace Microsoft.CodeAnalysis.LanguageServer.UnitTests;
+namespace Microsoft.CodeAnalysis.Editor.UnitTests.LanguageServer;
```
JoeRobich, line 28: nit:
```suggestion
namespace Microsoft.CodeAnalysis.Editor.UnitTests.LanguageServer;

```

JoeRobich (approved):

